PR #1736, opened 2013-11-10
Stop VanillaInThisDiscussion overrun by limiting to 20 with new config option & better sort

lungyiing:
*Also lists by last posted due to this change


Tama63:
@linc glad this got resolved, I think github is getting filtered into spam for me so I missed the notification - sorry about that!

PR #1807, opened 2014-03-21
Pass strings through T() for translations.

kemat1an:
I added T() calls everywhere in order to localize Vanilla Dashboard and
Interface as much as possible.


tburry:
I'm going to have to reject this PR as there are too many translates that aren't done properly here. I think we need to simplify some of the dashboard text if we want to accept a PR like this, but right now this is too many files changed to unwind.

PR #1887, opened 2014-04-29
There is an issue with parse_ini_file which for some reason change the t...

chanhong:
There is an issue with parse_ini_file which for some reason change the two \ into one \ in the cache ini file when Vanilla is running from a Windows network share.

For example: (two backslashes)
gdn = "\\networkshare\folder\Garden/library/core/class.gdn.php"

Will become: (one backslash)
gdn = "\networkshare\folder\Garden/library/core/class.gdn.php"

This fix will produce (forward slash)
gdn = "//networkshare/folder/Garden/library/core/class.gdn.php"

This fix will enable Vanilla to run on a Windows network share where without it, Vanilla will hit a "bonk" screen.

### DIFF
--- a/library/core/class.autoloader.php
+++ b/library/core/class.autoloader.php
@@ -893,6 +893,18 @@ class Gdn_Autoloader_Map {
    public function MapType() {
       return $this->MapInfo['maptype']; //GetValue('maptype', $this->MapInfo);
    }
+   
+    /**
+    * Try to fix path that has backslash in Windows OS
+    *
+    * @path string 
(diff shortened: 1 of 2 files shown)


chanhong:
I update my code to have all the changes that you commented previously.  Please let me know if there is anything else that you need.  Thanks

linc:
There were a great deal of issues with this PR still. Please review my changes here for future reference: e1f1e6bb182ea7922722eff013c071f4064bc376

PR #1954, opened 2014-05-28
Bugfix for CommentCount watch

ryacobi:
The watched comment calculation was wrong as Offset is not a page number but an actual offset.


ryacobi:
Hi

Thanks. Just signed it. (ryacobi)

PR #1955, opened 2014-05-29
Fix for offset in pagination

ryacobi:
For example, when asking for page=5 when ActualResponses=15 and Limit=20 - it will zero the offset.
But, if asking for page=15 when ActualResponses=21 and Limit=20 - it won't zero the offset and will return the last 6 comments. This is not a consistent behavior

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -112,8 +112,6 @@ class DiscussionController extends VanillaController {
             // (((67 comments / 10 perpage) = 6.7) rounded down = 6) * 10 perpage = offset 60;
             $this->Offset = floor($CountCommentWatch / $Limit) * $Limit;
          }
-         if ($ActualResponses <= $Limit)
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -112,8 +112,6 @@ public function Index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
             // (((67 comments / 10 perpage) = 6.7) rounded down = 6) * 10 perpage = offset 60;
             $this->Offset = floor($CountCommentWatch / $Limit) * $Limit;
          }
-         if ($ActualResponses <= $Limit)
-            $this->Offset = 0;
```
tburry: I think this is incorrect. If there are less comments than the page size then the offset is always zero when in auto-offset. Moving this removed statement to within the previous if statement is the way to go.

linc:
If you want to revisit this, let's do a new PR with enough information so it can be tested.

PR #2043, opened 2014-08-23
Added support for Twitch.tv and Hitbox.tv

adrianspeyer:
These are two popular video sites for video game streaming.


Basert:
Any new status on this pull request? Wanted to have sth like this, but before doing the same, a merge would be awesome.

linc:
@adrianspeyer Would you remove the back-link anchor tag at the end of each of those? I don't think there's any reason for those, and they just junk up discussions.

adrianspeyer:
@linc I thought it made sense to let members  visit the channel page, should they want. Personally, I don't think it's so distracting, because it's so small.

PR #2062, opened 2014-09-04
Permission defining and saving tweaks.

tburry:
* Do some additional schema checking when saving permissions.
* Set additional default permissions for administrators and moderators.

### DIFF
--- a/applications/dashboard/models/class.permissionmodel.php
+++ b/applications/dashboard/models/class.permissionmodel.php
@@ -90,6 +90,7 @@ class PermissionModel extends Gdn_Model {
       $this->ClearPermissions();
       if ($this->Schema) {
          // Redefine the schema if it has been defined to reflect the permissions that were just added.
+         $this->Schema = NULL;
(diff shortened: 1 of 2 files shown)


applications/dashboard/models/class.permissionmodel.php
```diff
@@ -88,8 +88,12 @@ public function Define($PermissionNames, $Type = 'tinyint', $JunctionTable = NUL
             ->Put('Permission', array(), $Where);
       }
       $this->ClearPermissions();
+      if ($this->Schema) {
+         // Redefine the schema if it has been defined to reflect the permissions that were just added.
+         $this->DefineSchema();
```
kaecyra: This is a no-op.

```
   public function DefineSchema() {
      if (!isset($this->Schema)) {
         SNIPPED
      }
      return $this->Schema;
   }
```

tburry:
I've added the schema redefining code and tested it with forum installs and plugin enables.

PR #2089, opened 2014-09-15
Let BeforeBodyBox change the BodyBox again

JasonBarnabe:
I'm creating a plugin that allows the user to choose the format (HTML or Markdown) of each of their individual posts.

Prior to [this commit](https://github.com/vanilla/vanilla/commit/e2b1f69393d0fe54ff7790b9957fe8402329e9bb#diff-bcdcd95ad070ab8428b1b3eb3e2353cbL150), `BeforeBodyBox`'s `BodyBox` parameter contained HTML for the textbox and the hidden `Format` input. My plugin would rewrite this string to remove the hidden input.

After that commit, `BeforeBodyBox` does not receive the hidden input in the `BodyBox` parameter - it's always `<div class="bodybox-wrap">`, which is pretty pointless.

This PR restores the old behaviour (the current behaviour in 2.1.x). It additionally makes it easier than in 2.1 for `BeforeBodyBox` to change `BodyBox`: it no longer needs to modify the passed string object; it can set a new one if it wants.

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -161,13 +161,12 @@ class Gdn_Form extends Gdn_Pluggable {
       $this->EventArguments['Column'] = $Column;
       $this->EventArguments['Attributes'] = $Attributes;
 
-      $Result = '<div class="bodybox-wrap">';
+      $Result = $this->TextBox($Column, $Attributes).$this->Hidden('Format');
+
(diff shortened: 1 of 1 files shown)


linc:
I see no issue with the latest versions of this. Same functionality, just slightly shifted to allow for hooks.

PR #2103, opened 2014-09-19
Fix pager bug in search page

R-J:
`PagerModule::Write` without any custom arguments has been erroneously used. Replaced it with `$this->Pager` (which gets initialized in the SearchController) and its method `ToString`. A hint have been that the wrong pager expressed "Next" while the controller sets `$this->Pager->MoreCode = 'More Results';`     
    
Tested successfully with search words that yield       
0 result,   
1 result,   
2 results and    
more than C('Garden.Search.PerPage') results


linc:
A next/previous pager is the desirable default pager.

Are you saying this PR creates a "More" pager?

linc:
Tested. Does create a "More" pager.

linc:
This probably just needs to get redone differently, so I'm gonna close this one.

PR #2104, opened 2014-09-19
Fix HTML special char bug

R-J:
`$this->Data('SearchTerm')` is already sanitized in controller with `$this->SetData('SearchTerm', Gdn_Format::Text($Search), TRUE);` so `htmlspecialchars()` mustn't be called in the output (again)


linc:
The hell? This wasn't merged.

PR #2116, opened 2014-09-23
Emoji api enhancements.

tburry:
Add enhancements to the emoji api so that emoji packs can do more.

- Add more options to the image format string for emoji so that packs can pass @2x retina images or sprite sheets.
- Add a seperate editor list for the default emoji so that more aliases can be added.
- Pass the emoji auto-complete data as a raw array since the new definition list format supports that.


tburry:
This PR has been rebased [here](https://github.com/vanilla/vanilla/pull/2425).

PR #2125, opened 2014-09-25
Fixed model validation field clearing

kaecyra:


linc:
Looks like this got changed directly on master, then reverted because it broke. *golf clap*

PR #2170, opened 2014-10-11
use HostAndPort in domain parsing

### LINKED ISSUE #2072: 2.2.16 On http://localhost:8888 server all pictures paths are without port
On http://localhost:8888 server pictures paths are without port.
http://localhost/uploads/favicon.png instead of http://localhost:8888/uploads/favicon.png

vanilla 2.2.16

imnotjames:
Since domain includes the scheme it's not that far fetched for it to have the port as well.  This fixes assets that are loaded with domain pointing at the wrong URL.

Fixes #2072

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -556,7 +556,7 @@ class Gdn_Request {
 
       $Domain = FALSE;
       if ($Domain === FALSE || $Domain == '')
-         $Domain = $this->Host();
+         $Domain = $this->HostAndPort();
 
       if ($Domain != '' && $Domain !== FALSE) {
          if (!stristr($Domain,'://'))
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -341,8 +341,14 @@ class Gdn_Request {
       $this->_EnvironmentElement('ConfigWebRoot', Gdn::Config('Garden.WebRoot'));
       $this->_EnvironmentElement('ConfigStripUrls', Gdn::Config('Garden.StripWebRoot', FALSE));
 
-      $this->RequestHost(     isset($_SERVER['HTTP_X_FORWARDED_HOST']) ? val('HTTP_X_FORWARDED_HOST',$_SERVER) : (isset($_SERVER['HTTP_HOST']) ? val('HTTP_HOST',$_SERVER) : val('SERVER_NAME',$_SERVER)));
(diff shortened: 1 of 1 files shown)


library/core/class.request.php
```diff
@@ -341,8 +341,14 @@ protected function _LoadEnvironment() {
       $this->_EnvironmentElement('ConfigWebRoot', Gdn::Config('Garden.WebRoot'));
       $this->_EnvironmentElement('ConfigStripUrls', Gdn::Config('Garden.StripWebRoot', FALSE));
 
-      $this->RequestHost(     isset($_SERVER['HTTP_X_FORWARDED_HOST']) ? val('HTTP_X_FORWARDED_HOST',$_SERVER) : (isset($_SERVER['HTTP_HOST']) ? val('HTTP_HOST',$_SERVER) : val('SERVER_NAME',$_SERVER)));
-      $this->RequestMethod(   isset($_SERVER['REQUEST_METHOD']) ? val('REQUEST_METHOD',$_SERVER) : 'CONSOLE');
+      $Host = isset($_SERVER['HTTP_X_FORWARDED_HOST']) ? val('HTTP_X_FORWARDED_HOST',$_SERVER) : (isset($_SERVER['HTTP_HOST']) ? val('HTTP_HOST',$_SERVER) : val('SERVER_NAME',$_SERVER));
```
kaecyra: I'm not a big fan of this inscrutable nested ternary if-statement. Would you mind expanding it?

Malet:
+1 this fixes an issue I was having when running on `localhost:3000` :+1:

library/core/class.request.php
Malet: Also `val()` doesn't actually do anything in this case, better off using `$_SERVER['SERVER_NAME']` etc.

kaecyra:
Bump @imnotjames, this needs cleanup per comment.

linc:
Note to self: This is confirmed clear, besides the syntax issue.

library/core/class.request.php
imnotjames: The `val()` function does very little in almost every case where it is used except make the code a little more indirect.

imnotjames:
I wasn't a fan of the ternary either but used it because it was in the original code.  Updated.

linc:
Thanks!

PR #2200, opened 2014-10-28
Add Recaptcha localization & standardize calls to Captcha

linc:
See https://github.com/vanilla/vanilla/pull/1796 for details & conversation. This simply re-targets against stage.


linc:
@beckyvb @MightyCoder Either of you interested in implementing Todd's feedback on this one to finish it? It looks like a straightforward chance to play in the Locale class and write some new code.

Supported languages list moved to here: https://developers.google.com/recaptcha/docs/language

linc:
They don't want RecaptchaOptions as a proper JSON-encoded array. The options are variables so they can't be quoted. Referencing here: https://developers.google.com/recaptcha/old/docs/customization

I've implemented `Gdn::Locale()->Language()` and the Google whitelist & `lang` parameter in my latest commit.

tburry:
Are you sure about that? I don't think javascript even works like that.

tburry:
@linc, pinging you on my last comment. Javascript objects are javascript objects and can accept quoted keys.

linc:
Yup, this is still on my todo list this month.

initvector:
Fixed a typo in Gdn_Locale::Language and switched up RecaptchaOptions to be stored as a PHP array, but sent as JSON.

Issue #2201, opened 2014-10-29
Mentions - auto complete for mouse click


John0x00:
A user has requested that mouse click work for autocomplete.

![screenshot](https://www.evernote.com/shard/s30/sh/02a7523f-bab6-40d6-a142-be2c91f6673e/1a52dae82a06a78414549488d1cde8c6/deep/0/New-Discussion---Vanilla-2.png)

PR #2219, opened 2014-11-01
Adds click-to-complete mentions

bleistivt:
- Options in the <code>@mention</code> flyout can now be selected with a mouseclick.
This was already the librarys default, but "overridden" by the
display_timeout setting of 0 in global.js.
(The inputs blur event is triggered before the click event, closing the flyout before anything
can be selected. This is the reason for the timeout being there.)

- The tab key no longer closes the flyout, but moves to the next item.
This required 2 minimal changes in the library.

Fixes https://github.com/vanilla/vanilla/issues/2201

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -1678,7 +1678,6 @@ jQuery(document).ready(function($) {
                   }
                }
             },
-            display_timeout: 0,
             cWindow: iframe_window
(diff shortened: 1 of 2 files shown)


linc:
Sweet!

PR #2228, opened 2014-11-04
Add Markdown support for single line breaks and strikeout

linc:
Reformulation of https://github.com/vanilla/vanilla/pull/2221 per notes on Spoilers. This version is much more conservative since it assumes future improved spoiler support will be core.

* Correct Markdown class markup & documentation per Vanilla standards
* Add single line breaks via `Gdn_Format`
* Add strikeout support


linc:
* Implemented soft breaks correctly.
* Fixed a hash problem with strikeouts.

I'm a much bigger fan of sub-classing to future-proof our extension than merging his future changes into ours. We're much more inclined to keep our changes precise & surgical if we extend it using the means provided in the parent class than if we just start slapping code in anywhere we feel like it.

tburry:
This really isn't surgical at this point and we'll have no way of knowing if we can catch bug fixes to methods we override. I want the fork. Subclassing isn't really for this purpose.

linc:
I don't have access to fork a repo under the Vanilla account.

PR #2230, opened 2014-11-04
Allow Emoji suggest template to be modified

linc:
Hard-coding stuff like emoji dimensions isn't good. Let this tpl get overwritten if it needs to be.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -1488,6 +1488,10 @@ jQuery(document).ready(function($) {
          ? $.parseJSON(emoji_array)
          : [];
 
+      // Emoji suggest template.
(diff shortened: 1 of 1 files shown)


tburry:
I'm a no on this. The suggest needs to work with the regular `Emoji->format` and not have a separate template. Also note there are other branch I have for adding more options to `Emoji->format` in [another PR](https://github.com/vanilla/vanilla/pull/2116).

tburry:
I didn't see your previous comment. You can pass formatted html to the editor and auto-lookup that has been pre-generated with `Emoji->img()`. Implementing that function in js is also not the worst, but only gains a bit of space in `gdn.meta`.

PR #2242, opened 2014-11-06
Use more secure error for invalid pagination ranges

linc:
* Disallow limit and offset clauses too large for MySQL.
* Make better use of the Offset method rather than allowing it to be touched directly elsewhere.


library/database/class.sqldriver.php
```diff
@@ -1283,10 +1283,16 @@ public function Like($Field, $Match = '', $Side = 'both', $Op = 'like') {
     * @return Gdn_SQLDriver $this
     */
    public function Limit($Limit, $Offset = FALSE) {
+      // SQL chokes on ints over 2^31
+      if ($Limit > 2147483648) {
```
tburry: Put this in a class constant please.

linc: Done, updated

library/database/class.sqldriver.php
```diff
@@ -1263,10 +1266,16 @@ public function Like($Field, $Match = '', $Side = 'both', $Op = 'like') {
     * @return Gdn_SQLDriver $this
     */
    public function Limit($Limit, $Offset = FALSE) {
+      // SQL chokes on ints over 2^31
+      if ($Limit > MAX_SIGNED_INT) {
+         throw new Exception(T('Invalid limit defined.'), 400);
```
tburry: 'Invalid limit.' instead please. The "defined" part seems too bleep bloop.

library/database/class.sqldriver.php
```diff
@@ -1367,6 +1376,11 @@ public function NoReset($NoReset = TRUE, $OneTime = TRUE) {
     * @return Gdn_SQLDriver $this
     */
    public function Offset($Offset) {
+      // SQL chokes on ints over 2^31
+      if ($Offset > MAX_SIGNED_INT) {
+         throw new Exception(T('Invalid offset defined.'), 400);
```
tburry: 'Invalid offset.'

kaecyra: image

linc:
Would appreciate it if, in general, we could catch more things on the first round of feedback rather than nickel-and-diming as we go, especially things like this that didn't change. Revisiting these is significant overhead.

PR #2262, opened 2014-11-14
Separate "embed theme" settings

daedeloth:
These changes add a new tab to the settings page "embed theme" (only visible when embedding is enabled).
Works very similar to the mobile theme.

A query parameter embed=1 is added to determine if the forum is embeded.
(As far as I understand from online documentation, this should not be a problem seo wise)

Working implementation:
http://catlab.be/quized-development-forum-holiday-release (embedded)
http://forum.catlab.eu/ (bootstrap theme)

More info:
http://vanillaforums.org/discussion/28467/separate-embed-theme-setting#latest


linc:
Howdy @daedeloth, I need you to sign our CLA before we can look at this. http://vanillaforums.org/contributors You'll need an account on vanilaforums.org to do that.

This would be better targeted at the stage branch since it's a feature addition that would need additional testing before going to master. Including more details about what issue(s) you're attempting to solve with this PR would be helpful as well.

Given all that overhead, I am going to close this for now, but please feel free to resubmit per my notes here.

daedeloth:
I have made a new pull request: #2579 

The issue I am trying to solve is that I want to have a forum that is available online (and looks good) + a forum that is embedded in my website (and looks good overthere as well). For me, this problem could be solved by having a seperate "embed" theme, that is displayed on the website, while the default theme is still used when accessing the forum directly.

These changes solve my problem, but I also believe this is something that could be used by anyone who's embedding their forum into their website.

PR #2284, opened 2014-11-22
Fix MyBB hash

linc:
Tested against MyBB 1.8 database.


tburry:
Okay I guess we can leave it. Wondering how it got that way though. That ':' concat has been there for a while so I'd expect there to be people with that hash too.

PR #2288, opened 2014-11-23
Give better feedback if ini files are busted

linc:
The goal here is to protect against a common problem: the .ini files in the `cache` folder get messed up for whatever reason during an upgrade and it completely baffles the user with an error like "Fatal error: Class 'Gdn' not found".

Examples: http://vanillaforums.org/search?adv=&search=%22Fatal+error%3A+Class+%27Gdn%27+not+found%22&title=&author=&cat=all&tags=&discussion_d=1&discussion_question=1&comment_c=1&comment_answer=1&page_p=1&within=1+day&date=

This is the most straightforward way I could see of making this an easy issue to sidestep for our open source users, but I'm not sure it's the best solution. Suggestions welcome.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -45,7 +45,8 @@ Gdn_Autoloader::Start();
 // Guard against broken cache files
 if (!class_exists('Gdn')) {
    // Throwing an exception here would result in a white screen for the user.
-   exit("Class GDN not found. This often indicates the .ini files in the folder 'cache' are out of date and should be deleted.");
+   // This error usually indicates the .ini files in /cache are out of date and should be deleted.
+   exit("Class Gdn not found.");
 }
 
 // Cache Layer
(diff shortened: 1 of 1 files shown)


linc:
Agreed; updated. I just added it as an inline comment for now.

linc:
Sending this to 2.1

oliveratgithub:
Okay, got it. Would the following be a compromise to suit all needs? ;)

```php
exit("Class Gdn not found. Try cleaning up your cache directory.");
```

linc:
I'm not sure vagueness serves either goal.

I've long wanted to eliminate the issue entirely by auto-wiping the cache folder's ini files when the Gdn class isn't found, but worry about a scenario where this could bog down the server if there was some other legitimate reason it couldn't find the class.

oliveratgithub:
> I've long wanted to eliminate the issue entirely by auto-wiping the cache folder's ini files when the Gdn class isn't found
> - @linc 

Sounds like an overall better and more professional solution.

PR #2295, opened 2014-11-27
Hotfix/sqldriver groups concat type

imnotjames:
this makes it possible to set an or/and explicitly in a beginwheregroup so that doing an `OrOp()` after starting a `BeginWhereGroup()` doesn't make the entire group an OR.  If this is all intentional then perhaps the documentation should be made clearer - and I'd be fine with updating the comments in a pull request, just not sure of what is exactly the correct outcome.

This is so that this code:
```php
Gdn::SQL()
    ->Select('*')
    ->From('Test t')
    ->Where('ID', 5)
    ->BeginWhereGroup()
    ->OrWhere('InsertIPAddress', 123456)
    ->EndWhereGroup();
```
would not generate:
```sql
select t.*
from Test t
where t.Id = :ID
  or (t.InsertIpAddress = :tInsertIpAddress)
```
instead of the correct output of:
```sql
select t.*
from Test t
where t.Id = :ID
  and (t.InsertIpAddress = :tInsertIpAddress)
```

This also adds `OrBeginWhereGroup`.

The scary thing is that while I've looked through the code for the few situations where this exists, and this is the "correct" behavior.. this could cause backwards compatibility issues - not just with the internals but with plugins as well.

### DIFF
--- a/applications/dashboard/models/class.messagemodel.php
+++ b/applications/dashboard/models/class.messagemodel.php
@@ -131,6 +131,7 @@ class MessageModel extends Gdn_Model {
          ->Where('Enabled', '1')
          ->BeginWhereGroup()
          ->WhereIn('Controller', $Exceptions)
+         ->OrOp()
          ->BeginWhereGroup()
          ->OrWhere('Application', $Application)
          ->Where('Controller', $Controller)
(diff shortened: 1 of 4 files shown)


tburry:
I think this is a good change and any regressions that are caused should be easily fixed.

PR #2297, opened 2014-11-27
Emotify: fix for new inline preview

bleistivt:


bleistivt:
I noticed that with Emotify the preview button in /post/discussion didn't work since it doesn't use the popup anymore:

    TypeError: frm.find is not a function

frm is not a jQuery object int the /post/discussion context, so I wrapped it in $() to fix that (it makes no difference in the discussion view where it already is a jquery object).

linc:
Awesome, thanks! Sorry if it sounds lazy, but I really need these to be easy to mentally parse when there's only 1 of me and dozens of folks submitting things I didn't even know were a problem.

PR #2346, opened 2014-12-13
[WIP] Login rate limit: no-cache implementation

bleistivt:
Currently login rate limiting is limited to installs with caching. This
tries to fix that.

WIP because I need feedback on how to implement IP rate limiting without
being too DB-expensive.

The actual fix is this part:
https://github.com/bleistivt/vanilla/blob/a2d656b136cd86b5a59a48f3249dc2896e5da7c1/applications/dashboard/models/class.usermodel.php#L3458-L3474

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -3454,75 +3454,94 @@ class UserModel extends Gdn_Model {
     * @param array $User
     * @param boolean $PasswordOK
     */
-   public static function RateLimit($User, $PasswordOK) {
-      if (!Gdn::Cache()->ActiveEnabled()) return FALSE;
-//      $CoolingDown = FALSE;
-//
-//      // 1. Check if we're in userid cooldown
-//      $UserCooldownKey = FormatString(self::LOGIN_COOLDOWN_KEY, array('Source' => $User['UserID']));
-//      if (!$CoolingDown) {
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -3454,75 +3454,113 @@ public function PasswordReset($UserID, $Password) {
     * @param array $User
     * @param boolean $PasswordOK
     */
-   public static function RateLimit($User, $PasswordOK) {
-      if (!Gdn::Cache()->ActiveEnabled()) return FALSE;
-//      $CoolingDown = FALSE;
+    public static function RateLimit($User, $PasswordOK) {
+        if (function_exists('apc_store')) {
```
tburry: The check to `Gdn::Cache()->ActiveEnabled()` must come first. On multiple application server environments you need a global check.

bleistivt: Ok, so I would rewrite it to be:

    if (Gdn::Cache()->ActiveEnabled())
        // Use cache
    elseif (C('Garden.Apc', false) && function_exists('apc_store'))
        // Use apc
    else
        // Use database (user attributes)

What do you think about  removing the large comment block of dead code?

tburry: Yes please with braces on top.

bleistivt:
new pull to current stage #

PR #2353, opened 2014-12-16
allow unicode chars in dropdown options, profile extender plugin

guillermofr:
I needed this to include options with spanish charcode, like "España".


linc:
Thanks!

PR #2359, opened 2014-12-19
Allow app-relative css file includes

kaecyra:


library/core/class.controller.php
```diff
@@ -1678,9 +1678,16 @@ public function RenderMaster() {
...
-                  $CssPaths = array(CombinePaths(array(PATH_ROOT, str_replace('/', DS, $CssFile))));
+               } elseif (strpos($CssFile, '/') !== FALSE) {
+                  $CssPaths = array();
+
+                  $AppFolder = $CssInfo['AppFolder'];
+                  if (empty($AppFolder)) {
+                     // A direct path to the file was given.
+                     $CssPaths[] = paths(PATH_ROOT, str_replace('/', DS, $CssFile));
+                  } else {
+                     $CssPaths[] = paths(PATH_APPLICATIONS, $AppFolder, 'design', $CssFile);
```
tburry: Handle the case where `$AppFolder` begins with "plugins/". See below on line 1706.

kaecyra: Specifying "plugins/" is not compatible with also specifying an app folder since plugins are global.

kaecyra: Oh I get what you're saying. Let me check.

PR #2368, opened 2014-12-28
Fix for GetUnread to include unread discussions without comments

R-J:
GetUnread only showed discussions with unread comments.This fix includes unread discussions without any comments.

Keep current branch up with changes in 2.1. See: https://github.com/vanilla/vanilla/pull/2367


linc:
I am merging this mostly for reference. This entire functionality needs to be refactored out of Vanilla to a plugin, but I don't want this fix to get lost in that shuffle.

PR #2370, opened 2014-12-29
Remove tab shortcut

bleistivt:
The tab shortcut interferes with the mentions autocomplete.

This also jumps the selection to the post button when the mention
autocomplete is active, see #2201

Commented out the line, because pressing tab already jumps to the post
button in my tests, making it redundant.


linc:
So, I think this might be a good one, but that branch is dead at the moment. I'm gonna have to delete it to try again with the latest version, probably early next week. Sorry for the false start.

bleistivt:
No problem. Noticed ButtonBar has the same shortcut which breaks the tab key on autocomplete.

I'd be interested to know if it was added for a reason here or if it was just copied over from ButtonBar. @danemacmillan

linc:
My guess? Copied over.

Issue #1666, opened 2013-08-25
Mobile: always use Modern view for Discussions


linc:
Table view is unwieldly. Replies & Views take 2/5 of the page.

PR #2371, opened 2014-12-29
Mobile: Always use modern layout for discussions

bleistivt:
Temporarily set the default (modern) view for discussions in the themehooks.

#1666


linc:
I like this change, but want @tburry to weigh in too.

linc:
For reference, this sort of change should be made to `stage` in the future. This isn't really a hotfix, it's a noticeable feature change.

bleistivt:
I can open this against stage right now, if you want

linc:
Yeah, let's do that, actually. That'd make me feel better because then we'll notice if a cloud client's site were unexpectedly altered by this before it made it into production. I also think, barring Todd reviewing this sooner, I'll hang off on merging this until next week after the current batch of staged features goes off to allow more review time.

PR #2372, opened 2014-12-30
Emotify: Surpress Emoji rendering and autocomplete

bleistivt:
Since Emoji are rendered through vanilla before Emotify comes in, a
number of smilies are broken.

This removes  the EmojiSet and the Emoji autocomplete list for the time
Emotify is enabled.

### DIFF
--- a/plugins/Emotify/class.emotify.plugin.php
+++ b/plugins/Emotify/class.emotify.plugin.php
@@ -28,6 +28,14 @@ class EmotifyPlugin implements Gdn_IPlugin {
    public function AssetModel_StyleCss_Handler($Sender) {
       $Sender->AddCssFile('emotify.css', 'plugins/Emotify');
    }
+   
+	/**
+	 * Surpress Emoji rendering and autocomplete.
+	 */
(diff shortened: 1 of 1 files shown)


linc:
This is a good compromise, I think.

Emotify should probably move to Addons for 2.next, too.

linc:
I had to revert this. It causes a fatal error.

"PHP Parse error:  syntax error, unexpected '[', expecting ')' in /var/www/frontend/plugins/Emotify/class.emotify.plugin.php on line 37"

bleistivt:
Sorry, I thought the PHP requirement for 2.2/2.3 was already 5.4?

linc:
Ah, so, looks like this is a fault in our legacy cluster still running 5.3. We'll need to resolve that and then get this back in.

bleistivt:
I could redo this with the array() syntax, but it might be worth investigating why setting Garden.EmojiSet to 'none' doesn't completely remove the emoji list, first (I put the call to setFromManifest for that reason.)

PR #2374, opened 2014-12-31
gzopen64 may work if gaopen doesn't exist

peregrine-web:
http://vanillaforums.org/discussion/comment/222668/#Comment_222668

some version of php have a bug and gzopen64 may work if gaopen doesn't exist.


tburry:
+1 to x00.

linc:
@peregrine-web This needs to be reworked per the comments above.

PR #2377, opened 2015-01-02
Fixes double-parsing of special Profile Extender fields

linc:
* Adds ability to use `Gdn_Format::Html()` without magic parsing
* Fixes Profile Extender's special Twitter field from linking to profile (mention) instead of Twitter profile
* Generally fix inadvertent magic parsing of profile fields and potentially getting weird stuff like embedded YouTube there

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -805,7 +805,7 @@ class Gdn_Format {
     * @param mixed $Mixed An object, array, or string to be formatted.
     * @return string
     */
-   public static function Html($Mixed) {
+   public static function Html($Mixed, $Options = array()) {
       if (!is_string($Mixed)) {
          return self::To($Mixed, 'Html');
       } else {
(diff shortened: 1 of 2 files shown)


tburry:
I'd rather you refactor our a different format method that does just the html stuff you want than have an `$options` parameter.

linc:
I considered that at first, but I thought the code-duplication was worse & potentially dangerous looking to the future. If we have 2 ways to make HTML safe, will we remember to update both of them consistently if we change it?

I'm fine with adding it if you think the `$options` is worse than the duplication. I can't think of a good name for it, tho. `SimpleHtml()`? `HtmlBasic()`? 

Maybe `Html()` should call the new one instead of duplicating the basic parsing?

tburry:
The idea is you refactor just the html stuff our of Gdn_Format::Html() and
call that method inside Gdn_Format::Html() too.

Issue #1779, opened 2014-02-04
Application Locale not found on enable


hgtonight:
In 2.1b2, upon enabling an application the locale is refreshed in an attempt to map the application's locale files. The locale mapping file is updated, but does not include the recently enabled application.

I have determined this is due to refreshing the locale mapping during the testing of the application rather than during the enabling of the application.

PR incoming

PR #2378, opened 2015-01-02
Update locale refreshing for applications. `$Locale` is not superfluous.

hgtonight:
Update locale refreshing on application manager to wait until the application is actually enabled.


hgtonight:
Closes #1779 on the master branch

PR #2381, opened 2015-01-03
Clear notifications immediately

linc:
When notifications are clicked in the MeModule, the `Alert` sticks around even tho they have been cleared on the backend. This adds an instant-clearing mechanism in the UI using generic class names we can use elsewhere to accomplish the same effect anywhere we access notifications or display its count.

### DIFF
--- a/applications/dashboard/views/home/updatemode.php
+++ b/applications/dashboard/views/home/updatemode.php
@@ -1,7 +1,7 @@
 <?php if (!defined('APPLICATION')) exit(); ?>
 
 <div class="SplashInfo">
-   <h1><?php echo T('wMaintenance Mode'); ?></h1>
+   <h1><?php echo T('Maintenance Mode'); ?></h1>
    <p><?php echo T('The site is currently undergoing maintenance.'); ?></p>
 </div>
 <!-- Domain: <?php echo Gdn::Config('Garden.Domain', ''); ?> -->
--- a/applications/dashboard/views/home/updatemode.php
+++ b/applications/dashboard/views/home/updatemode.php
@@ -1,7 +1,7 @@
 <?php if (!defined('APPLICATION')) exit(); ?>
 
 <div class="SplashInfo">
-   <h1><?php echo T('Maintenance Mode'); ?></h1>
+   <h1><?php echo T('wMaintenance Mode'); ?></h1>
    <p><?php echo T('The site is currently undergoing maintenance.'); ?></p>
 </div>
 <!-- Domain: <?php echo Gdn::Config('Garden.Domain', ''); ?> -->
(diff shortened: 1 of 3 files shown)


js/global.js
```diff
@@ -1086,6 +1086,11 @@ jQuery(document).ready(function($) {
       setInterval(pingForNotifications, 60000);
    }
 
+   // Clear notifications alerts when they are accessed anywhere.
+   $(document).delegate('.NotificationsClear', 'click', function() {
```
tburry: I think delegate is deprecated. `$(document).on()` is what you use.

linc: Thanks, updated

applications/dashboard/views/home/updatemode.php
```diff
@@ -1,7 +1,7 @@
 <?php if (!defined('APPLICATION')) exit(); ?>
 
 <div class="SplashInfo">
-   <h1><?php echo T('Maintenance Mode'); ?></h1>
+   <h1><?php echo T('wMaintenance Mode'); ?></h1>
```
tburry: Derp.

applications/dashboard/views/modules/me.php
```diff
@@ -31,10 +31,10 @@
       echo '<div class="MeMenu">';
          // Notifications
          $CountNotifications = $User->CountNotifications;
-         $CNotifications = is_numeric($CountNotifications) && $CountNotifications > 0 ? '<span class="Alert">'.$CountNotifications.'</span>' : '';
+         $CNotifications = is_numeric($CountNotifications) && $CountNotifications > 0 ? '<span class="Alert NotificationsAlert">'.$CountNotifications.'</span>' : '';
 
          echo '<span class="ToggleFlyout" rel="/profile/notificationspopin">';
-         echo Anchor(Sprite('SpNotifications', 'Sprite Sprite16').Wrap(T('Notifications'), 'em').$CNotifications, UserUrl($User), 'MeButton FlyoutButton', array('title' => T('Notifications')));
+         echo Anchor(Sprite('SpNotifications', 'Sprite Sprite16').Wrap(T('Notifications'), 'em').$CNotifications, UserUrl($User), 'MeButton FlyoutButton NotificationsClear', array('title' => T('Notifications')));
```
tburry: Sorry to nickle and dime you, but classes that are pure behaviour should be 'js-' prefixes. So something like `js-clear-notifications` is good. This is kind of a new thing, but will ensure that themers don't style on such classes.

tburry: Actually why not add a `.js-remove-class` and then a `data-class="NotificationsAlert"` to make this more generic?

linc:
I get what you're saying about the data-class & re-usability, but I don't really wanna use this PR to guess-and-check if I'm coding it right, so I took the easy way out. :) If you do an example of that at some point I'll be interested to take a peek.

PR #2387, opened 2015-01-08
Vanilla: Embed comments: Make signin (Comment as...) link url absolute.

beckyvb:
This PR makes the link to signin when commenting on an embedded site (the 'Comment as...' button) an absolute link rather than a relative link.

It fixes an IE bug where the relative link to the forum signin page pointed to the host of the website the forum is embedded in, rather than the forum.

### DIFF
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -481,11 +481,7 @@ function WriteEmbedCommentForm() {
          );
          echo $Controller->Form->Button('Post Comment', array('class' => 'Button CommentButton'));
       } else {
-         $root = '';
-         if (val('vanilla_forum_url', $Controller->Data('ForeignSource'))) {
-            $root = trim(val('vanilla_forum_url', $Controller->Data('ForeignSource'), ''), '/');
-         }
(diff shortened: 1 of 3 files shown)


applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -481,7 +481,11 @@ function WriteEmbedCommentForm() {
          );
          echo $Controller->Form->Button('Post Comment', array('class' => 'Button CommentButton'));
       } else {
-         $AuthenticationUrl = SignInUrl($ReturnUrl);
+         $root = '';
+         if (val('vanilla_forum_url', $Controller->Data('ForeignSource'))) {
+            $root = trim(val('vanilla_forum_url', $Controller->Data('ForeignSource'), ''), '/');
+         }
+         $AuthenticationUrl = $root.SignInUrl($ReturnUrl);
```
tburry: If you want to add the current domain then you want to use `Url($AuthenticationUrl, true)` and rollback the other file changes you made.

beckyvb: Would this work with an embedded comment form?

PR #2394, opened 2015-01-12
Fix url checks for profile photos.

tburry:
There are checks in some of the `ProfileController` methods for whether or not a user’s photo is a url or not. These needed to be updated to the now standardized `IsUrl()`. Without this the activity feed renders incorrect photo change activities and some user API calls will have incorrect photo urls.


tburry:
In order to test this PR you're going to need cloudfiles because it uses cf:// style urls.

PR #2396, opened 2015-01-13
Relax CSRF checks on api calls made with an SSO string.

tburry:
There is currently no way to make proper POST api calls because our CSRF checks kick in. This change will validate CSRF checks making a call against a .json endpoint with a valid SSO string (?sso=).


linc:
:+1:

PR #2404, opened 2015-01-15
Escalate the permissions required for mass banning.

tburry:
Make mass banning an admin permission and add some additional tracking information to the ban table. Note that the additional information on the ban table will take effect after a structure update, but will still work without it.


linc:
:+1: 

Kinda wanna backport this too.

PR #2405, opened 2015-01-15
Regenerate System user on utility update

linc:


tburry:
I feel like the code before happens to work if the system user is renamed, but this will break that.

linc:
This only invokes GetSystemUserID, which respects a renamed System user correctly. It does not force regeneration.

PR #2412, opened 2015-01-17
Render HTML date in conversations

bleistivt:
Dates in conversations were missing the <code>< time ></code> element.


linc:
Very nice.

PR #2413, opened 2015-01-19
Don’t pollute the Gdn_Validation object’s when validating.

tburry:
Currently, when calling Gdn_Validation->Validate() you end up applying rules
from the database schema which permanently alters the validation object. This can break subsequent calls to validate when for example calling validate on insert and then update. This change addresses the problem.


tburry:
Gunna have to close this. The validation object need more serious thought than this.

PR #2414, opened 2015-01-19
Don't Show Image When No Photo

shadowdare:
Don't output image HTML tag when there is no photo for the activity.

### DIFF
--- a/applications/dashboard/views/modules/recentactivity.php
+++ b/applications/dashboard/views/modules/recentactivity.php
@@ -5,12 +5,19 @@
       <?php
       $Data = $this->ActivityData;
       foreach ($Data->Result() as $Activity) {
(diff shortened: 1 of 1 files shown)


linc:
Thanks for the PR! I like the fix, but have some quibbles with the syntax. 

It'd simplify this nicely to always output `$PhotoAnchor`. Start with `$PhotoAnchor = '';`, then do your `if` statement to change it to `Anchor(stuff).' '` (note the space concat happens there), then you're clear to eliminate the second `if`.

Also, is this an issue in master branch? I'm more concerned with it getting changed there than 2.1 fo' sho'.

linc:
Gonna close this for now, but feel free to send it in again with those fixes. Cheers!

PR #2416, opened 2015-01-20
Change data name on activity controller

hgtonight:
This allows the `/activity/item` view to actually display a single activity.


linc:
Have you carefully checked that this isn't going to break anything?

linc:
I didn't notice it was just the `Item` method, was going a little too fast and changing a data variable gives me The Fear. Item is pretty benign tho so I'm fine with this.

PR #2418, opened 2015-01-20
DiscussionModel stopped respecting $DataSetType at some point.

kaecyra:


linc:
I can't think of a scenario where this would break anything short of plugin code that asked for an array but then coped with an object... which would be uberstupid.

PR #2421, opened 2015-01-21
Mobile: Always use modern layout for discussions

bleistivt:
Temporarily set the default (modern) view for discussions in the themehooks.

#2371 re-opened against stage


tburry:
We have a lot of mobile themes that are fine with the table view. Maybe the mobile css should just hide a few table columns by default.

Alternatively, I'd be okay with the mobile view pulling from a separate config setting for its various views.

PR #2425, opened 2015-01-21
Emoji api enhancements.

tburry:
Add enhancements to the emoji api so that emoji packs can do more. This is a rebase of #2116.

- Add more options to the image format string for emoji so that packs can pass @2x retina images or sprite sheets.
- Add a seperate editor list for the default emoji so that more aliases can be added.
- Pass the emoji auto-complete data as a raw array since the new definition list format supports that.
- Add an `Emoji->archive` array to enable the deprecating/renaming of emoji.
- [x] Make the autocomplete list obey `Emoji->format`
- [x] Clean up properties so that `Emoji->emojis` represents the list of emoji so we can use just emoji for a single emoji.

Note, there is a PR on the internal repository that goes along with this PR and can be used to test the various changes to the emoji API.


linc:
This is all delightful and good and reasonably bulletproof.

PR #2427, opened 2015-01-22
Allow non-magic HTML parsing & fix it in Profile Extender fields

linc:
Replaces #2377

* Adds `Gdn_Format::HtmlFilter()` as non-magic HTML parsing.
* Add `IsHtml()` convenience function.
* Fixes Profile Extender automatically magic-parsing its field values, which is bad on profiles.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -809,43 +809,6 @@ class Gdn_Format {
 
    /**
     * Takes a mixed variable, filters unsafe HTML and returns it.
-    * Use this instead of Gdn_Format::Html() when you do not want magic formatting.
-    *
-    * @param mixed $Mixed An object, array, or string to be formatted.
-    * @return string
-    */
-   public static function HtmlFilter($Mixed) {
-      if (!is_string($Mixed)) {
-         return self::To($Mixed, 'HtmlFilter');
-      } else {
-         if (IsHtml($Mixed)) {
-            // Purify HTML with our formatter.
-            $Formatter = Gdn::Factory('HtmlFormatter');
-            if (is_null($Formatter)) {
-               // If there is no HtmlFormatter then make sure that script injections won't work.
-               return self::Display($Mixed);
-            }
-
-            // Allow the code tag to keep all enclosed HTML encoded.
-            $Mixed = preg_replace(
-               array('/<code([^>]*)>(.+?)<\/code>/sei'),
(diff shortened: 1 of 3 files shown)


linc:
The diff is a bit weird on this since it makes it look like I renamed `Html()` when in fact I just created a similar method above it while editing `Html()` at the same time.

tburry:
You can move HtmlFilter() below Html() which is the correct alphabetical order and might help review the PR.

linc:
Good idea, that worked.

library/core/functions.general.php
```diff
@@ -1766,6 +1766,18 @@ function InSubArray($Needle, $Haystack) {
...
 
+if (!function_exists('IsHtml')) {
+   /**
+    * Determine whether text contains HTML.
+    *
+    * @param string $Text
+    * @return bool Whether HTML was detected.
+    */
+   function IsHtml($Text) {
+      return strpos($Text, '<') !== FALSE || (bool)preg_match('/&#?[a-z0-9]{1,10};/i', $Text);
```
tburry: This is kind of a bad thing to make a global function as it really doesn't  check for HTML. The idea behind the check is for code that looks like html as an optimization. We don't need to filter code that doesn't have any tags.

Making this a global function is going to be misleading to other programmers that might want to use it for a more general purpose. Adding more functionality will kind of kill the entire reason for it in the first place as it will lose its optimization.

library/core/class.format.php
```diff
@@ -874,6 +856,43 @@ public static function Html($Mixed) {
...
+            $Mixed = preg_replace(
+               array('/<code([^>]*)>(.+?)<\/code>/sei'),
+               array('\'<code\'.RemoveQuoteSlashes(\'\1\').\'>\'.htmlspecialchars(RemoveQuoteSlashes(\'\2\')).\'</code>\''),
+               $Mixed
+            );
+
+            // Do HTML filtering before our special changes.
+            $Result = $Formatter->Format($Mixed);
+         }
+         else {
```
tburry: Coding style.

library/core/functions.general.php
tburry: I recommend just making this a protected static on `Gdn_Format`.

tburry:
Other than my comments I think this looks fine.

library/core/class.format.php
linc: What about it? Are you asking for a same-line else? I didn't think we had a rule about that.

tburry: Fence the `} else {` pls.

linc:
Updated

PR #2428, opened 2015-01-22
Fix use of MAX_SIGNED_INT constant

bleistivt:
class constants need to be accessed with self:: inside of the class


bleistivt:
Also, shouldn't this be a Gdn_UserException?

linc:
Let us never speak of this again. ;) lol (Nice save).

linc:
To be honest, I'm a little murky on exception differentiation. This is a fairly critical and non-user-facing exception in all likelihood, so I went with a straight Exception.

bleistivt:
You can thank me for living dangerously on master :D
 I'm usually like "looks safe" and pull from master to my site. But once in a while -> dramatic trainwreck

PR #2429, opened 2015-01-23
Fix display bug where users without a title would show a bullet with nothing on the profile page.

tburry:
It looks like we added a line without adding braces a while back.


hgtonight:
+1

linc:
Coding standard: validated.

PR #2432, opened 2015-01-23
Delete unnecessary files, saving a full MB

linc:
The incredible shrinking Vanilla! Now 10% smaller! 58 fewer files! Amazing!


tburry:
Great stuff. The only one I disagree with is composer.lock. Once we start including libraries we'll want to keep this file so that all devs/deploys can be sure to be using the exact same libraries. Otherwise, it's a crap shoot.

linc:
TBH, I thought it was like an old `.lock` file that someone had forgotten to gitignore. My bad, shows what I know about Composer! Restored.

PR #2433, opened 2015-01-24
Popup: Process targets on confirm clicks

bleistivt:
Makes the confirm action consistent with the reveal.


linc:
I need some more notes one what we're doing here & why. Gotta give me a lazy-read when it's something that's not even on our radar.

linc:
Alright I dove in and figured it out. :P

Issue #2345, opened 2014-12-10
Handle EXIF orientation in UploadImage


bleistivt:
> Update: I got a couple of original images from a poster. According to exiftool, they have an "orientation" attribute defined, which doesn't appear to be handled by SaveImageAs in class.uploadimage.php.  The images are properly oriented when I view them on my desktop (Mac), but it looks like the SaveImageAs code simply reads the width/height attributes and ignores the request to rotate.

This Issue was generated from your [forums] 
[forums]: http://vanillaforums.org/discussion/comment/221216#Comment_221216

webstey:
I submitted a PR for this, but it created a new issue.  Is there a way I could have submitted my PR as a proposed fix for this issue instead?

hgtonight:
@webstey You just reference it as #number in your PR. Add a comment that says "Closes #2345" and github will do the linking for you.

webstey:
TYAH!  (Thanks Yet Again HgTonight :) )

linc:
And to cross-link on GitHub, just mention it.

Related:  #2438

linc:
![screen shot 2015-01-26 at 11 41 04 am](https://cloud.githubusercontent.com/assets/117672/5903541/45d37176-a550-11e4-906d-6f57cc3dc618.png)

webstey:
I think I've got it worked out. Awaiting confirmation from @linc and merge :)  FYI, works on 2.1.8.

Issue #182, opened 2009-11-30
Vanilla: Message "above main contet" appears below main content


linc:
Adding a message on All Pages "Above Main Content" instead puts it below main content.

Reported (fourth item): http://vanillaforums.org/discussion/10217/vanilla-2:-bugs-and-suggestions/#Item_11

PR #2438, opened 2015-01-26
Fix rotation issue with iPhone uploads

webstey:
As per http://vanillaforums.org/discussion/28683/images-are-rotated-when-resized , added rotation of image files if EXIF rotation tags are present.

### DIFF
--- a/library/core/class.uploadimage.php
+++ b/library/core/class.uploadimage.php
@@ -238,6 +238,25 @@ class Gdn_UploadImage extends Gdn_Upload {
          imagecopyresampled($TargetImage, $SourceImage, 0, 0, $XCoord, $YCoord, $Width, $Height, $WidthSource, $HeightSource);
          imagedestroy($SourceImage);
 
+         // Check for EXIF rotation tag, and rotate the image if present
+         if (function_exists('exif_read_data') && 
+               ( ($Type == IMAGETYPE_JPEG) || ($Type == IMAGETYPE_TIFF_II ) || ($Type == IMAGETYPE_TIFF_MM) ) ) { 
+            $ImageExif = exif_read_data($Source);
+            if (!empty($ImageExif['Orientation'])) {
+               switch ($ImageExif['Orientation']) {
+                  case 3:
+                     $TargetImage = imagerotate($TargetImage, 180, 0);
+                     break;
+                  case 6:
+                     $TargetImage = imagerotate($TargetImage, -90, 0);
+                     break;
+                  case 8:
+                     $TargetImage = imagerotate($TargetImage, 90, 0);
+                   break;
+                }
+            }
+         }
+
          // No need to check these, if we get here then whichever function we need will be available
          if ($OutputType == 'gif')
             imagegif($TargetImage, $TargetPath);
--- a/library/core/class.uploadimage.php
+++ b/library/core/class.uploadimage.php
@@ -249,9 +249,11 @@ class Gdn_UploadImage extends Gdn_Upload {
                      break;
                   case 6:
                      $TargetImage = imagerotate($TargetImage, -90, 0);
+                     list($Width, $Height) = array($Height, $Width);
                      break;
                   case 8:
                      $TargetImage = imagerotate($TargetImage, 90, 0);
+                     list($Width, $Height) = array($Height, $Width);
                    break;
                 }
             }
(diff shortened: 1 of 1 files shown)


linc:
YESSSSSSSSSSS

webstey:
I guess it worked for you too linc?  :)  FWIW, I've been using this on my forum since sometime in December with no complaints.

linc:
Haven't tested it yet, but this is a known grievance with core and a personal pet peeve. I've resorted to using Preview to rotate photos 90 degress and back again & save it to make sure it uploads correctly.

linc:
I tried this out and it didn't actually solve the problem for me. I attached my sample photos over here if you want to dig further. http://vanillaforums.org/discussion/comment/224734/#Comment_224734

bleistivt:
@webstey could you try checking for both $exif['Orientation'] and $exif['IFD0']['Orientation']?

webstey:
Sorry for flaking off here; I've been busy with family etc etc.  Going to take a look now.

linc:
8 days doesn't even rate for "flaking off" around here, lol. I know how it goes.

webstey:
Interesting.  I uploaded the 4 photos you posted at VF with the following results:

- using ImageUpload, all 4 photos are rotated properly.  
- using FileUpload, the thumbnails are rotated properly, but when I click on them to view the originals, they are *not* rotated.

Any ideas?

webstey:
This is the case for both your test images, and the test images from my user.  His were from an iPhone 5s, yours from iPhone 6 (according to the exif tags).  Both used the 'Orientation' exif tag.

bleistivt:
ImageUpload uses Gdn_UploadImage->SaveImageAs()
Fileupload uses Gdn_Upload->SaveAs()

oh, and I just stumbled upon a snippet using the IDF0 subarray, which is the reason i just posted this here. Take all the time you need.

bleistivt:
oh, and for the thumbnails FileUpload uses SaveImageAs, which seems to be the reason for this.

I'd argue that for a generic fileupload, not rotating the image is ok, since one may be interested in the original file, but as it's primarily used to attach images to comments, proper orientation would be expected by most users.

webstey:
Thanks @bleistivt !  Very helpful.  I am going to make the images get rotated by FileUpload.  2 questions for you and @linc : 
1. Should I add this to the fix, or should I leave it out as per @bleistivt comment?  I wasn't sure which you thought I should do :)
2. Presuming I am going to make FileUpload rotate images, should I put this in the FileUpload plugin code, or in the core SaveAs function?  My guess would be the former.

webstey:
Finally, @bleistivt do you have a sample photo that uses the IDF0 tag handy?

bleistivt:
lincs photos actually seemed to have that. I used an online exif tool:

http://exifdata.com/

webstey:
Ah, I was using exiftool - and I just looked up what IFD0 meant, so I've unconfused myself there.

I think the solution may lie in calling SaveImageAs in the FileUpload plugin if it's an image, instead of doing the straight up copy.  Weird that it doesn't actually use the SavaAs function already?  I'm not sure what the effect of the FireAs() function is, but it doesn't change anything if I change it.

bleistivt:
FireAs() just changes the Controller Name of the Hook that will be called.

The class is called "FileUploadPlugin" which would make the event <code>FileUploadPlugin_SaveAs_Handler</code>, but the intention seems to be for plugins that handle the <code>Gdn_Upload_SaveAs</code> Event to also hook into this.

Plugins can Handle this event as <code>Gdn_Upload_SaveAs_Handler()</code> just like with normal fileuploads.

That doesn't explain why the plugin is not using the SaveAs() function. Could be possible, that it uses it's own MediaModel in some way that Gdn_Upload would not allow.

webstey:
I've got it partly figured out, but think I may have noticed a bug in my edit to SaveImageAs - I'm not adjusting the image height/width after rotating. I haven't figured out if some other code is doing it after the fact yet.

webstey:
Update: yep, I was right.  ImageUpload doesn't seem to do anything with height/width after calling SaveImageAs, so it works just dandy.  However, FileUpload's thumb generator seems to add in CSS to size the image, so 90 or -90 rotated images were stretched.  I made SaveImageAs return the new height and width so the thumbnail generator (or any code using it) can get the proper dimensions.

I seem to have made it work, rotation, new dimensions, and all!  I had to edit class.fileupload.plugin and class.uploadimage.php.  Now I just have to figure out how to add my changes to the PR.

webstey:
OK, I've added my changes to class.uploadimage.php.  Can I somehow add my changes class.fileupload.php to this PR, even though it's in the addons repo?

bleistivt:
No, that will have to be a separate PR.

Great! Can't wait to have this in the app, so I don't have to manually rotate images of my users on the server anymore.

webstey:
I made the PR (#182) and referenced this issue.  Hopefully that works?

bleistivt:
Github will autolink that if you just paste the uri here.

webstey:
https://github.com/vanilla/addons/pull/182

webstey:
OK, thank @bleistivt !  I think I've got it now.

https://github.com/vanilla/vanilla/issues/2345

PR #2439, opened 2015-01-26

linc:
This replaces #1887 with syntax fixes.

### DIFF
--- a/library/core/class.autoloader.php
+++ b/library/core/class.autoloader.php
@@ -899,26 +899,29 @@ class Gdn_Autoloader_Map {
       return $this->MapInfo['maptype']; //GetValue('maptype', $this->MapInfo);
    }
    
-    /**
-    * Try to fix path that has backslash in Windows OS
+   /**
+    * Fix path with backslash for Windows OS.
     *
-    * @path string 
+    * @param string $path
     * @return string
     *
     * Check to see if the path is not empty and if it is not in a form of drive: then replace any "\" with "/" to avoid 
     * the parse_fn_file bug and it also further check to see if it has only one "/" and the OS is "WIN" then it will prepend "/" 
     * in front of the path so it has the correct file path.
-    *  
     */
-    function fixBackSlash($path) {
-        if (!empty($path) and !preg_match('/[A-Za-z]\:/',$path)) {
-            $path = str_replace("\\", "/", $path);  // convert to / to avoid parse_in_file create array with missing \
-            if (preg_match('/^\/{1}\w/',$path)==TRUE and (strtoupper(substr(PHP_OS, 0, 3)) == 'WIN')) { // for some reason there is only 1 / then add / to have a valid network path
-                $path = "/" . $path;
-            }
-        }
-        return $path;
-    }
+   function fixBackSlash($path) {
+      if (!empty($path) && !preg_match('/[A-Za-z]\:/', $path)) {
+         // Convert to slash to avoid parse_in_file create array with missing backslash.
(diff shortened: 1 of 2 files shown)


tburry:
Can I get a link to the specific Windows bug in php?

library/core/class.autoloader.php
```diff
@@ -898,6 +898,30 @@ protected function GetClassNameFromFile($File) {
...
+    *
+    * @param string $path
+    * @return string
+    *
+    * Check to see if the path is not empty and if it is not in a form of drive: then replace any "\" with "/" to avoid 
+    * the parse_fn_file bug and it also further check to see if it has only one "/" and the OS is "WIN" then it will prepend "/" 
+    * in front of the path so it has the correct file path.
+    */
+   function fixBackSlash($path) {
+      if (!empty($path) && !preg_match('/[A-Za-z]\:/', $path)) {
```
tburry: You need to match the beginning of the string in your regex.

~~~
!preg_match('`^[a-z]:`i', $path)
~~~

library/core/class.autoloader.php
```diff
@@ -898,6 +898,30 @@ protected function GetClassNameFromFile($File) {
...
+         // Convert to slash to avoid parse_in_file create array with missing backslash.
+         $path = str_replace("\\", "/", $path);
+
+         // If there is only 1 slash, add another to have a valid network path.
+         if (preg_match('/^\/{1}\w/', $path) == TRUE && (strtoupper(substr(PHP_OS, 0, 3)) == 'WIN')) {
```
tburry: Clean up the regex.

```
preg_match('`^/[^/]`', $path) && stripos(PHP_OS, 'win') === 0)
```

tburry:
I feel like if you are using a network drive to store the cached lookup files then you are doing something wrong or at least suboptimal. These files are supposed to be loaded every page request and should therefore be loaded from the local disk. This may be a misguided use case.

We run Vanilla with many, many application servers and all of these such caches are kept local just fine.

linc:
@chanhong Feedback is for you ^

You can submit a PR against this branch, or you can fork my changes here and do a whole new one.

library/core/class.autoloader.php
chanhong: I discover this bug when I was doing testing with Vanilla that reside on a network share and Vanilla just not working so I start a debug session and found out that the network share path was chop off by parse_ini_file so that is when I come up with a fix.

chanhong: Yes, your regex is much better.  Thanks

library/core/class.autoloader.php
chanhong: Nice!  Thanks

chanhong:
Yes, the use case is not optimal when reside on the network share but for development to have on a network share would simplify the need to do backup of the development work.

@linc, I am not sure what you are asking me to do?  Please clarify.

chanhong:
Are you asking me to create a new PR that incorporate all the feedback from you and from @tburry?

Thanks

chanhong:
@tburry, @linc 

Just want to clarify, I was attempting to configure Vanilla to run on multiple tiers layer. 

For example:
Web content layer: content serving
File layer: network share or SAN of cluster of servers
Cache layer: local hard drive, SSD or in memory
Database layer: one or cluster of servers

This way the up time is guarantee and scalable.
.
Another words with this configuration Vanilla would be ready for the "Cloud"

chanhong:
I just created a new PR to include @tburry feedback and the latest code from Vanilla github.  Thanks

PR #2442, opened 2015-01-27
Mark All Viewed: Fix child-category read bug & performance bug by not saving entire tree

linc:
Rescued from an abandoned branch. No idea what was going on here but it looks important so I'll have another look.

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -1579,7 +1579,7 @@ class CategoryModel extends Gdn_Model {
    }
 
    /**
-    * Grab the Category IDs of the tree.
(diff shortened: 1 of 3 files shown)


tburry:
The not saving the subtree might work for a terribly over-categoried forum, but it doesn't really make sense as a feature and will cause a lot of confusion.

linc:
I don't understand what the source of the confusion would be.

I believe the current bug is that reading all discussions in a parent category marks the children as read, even if you haven't read the discussions in the children.

tburry:
Right now I'm just not into this fix. We need to evaluate the UI to make sure it looks clear that an entire subtree or just a category is going to be marked as read. If there is a gear beside a heading category, say then it makes sense that it would mark all the subcategories read.

PR #2443, opened 2015-01-27
Markdown: reimplement as fork with Vanilla customizations.

linc:
Replaces #2228. Customizations are here: https://github.com/vanilla/php-markdown/commit/4c8db6115abf8b3c6a265f6bfb2bc53d939a97b4

* Adds: soft line breaks, spoilers, & strikeout.
* Code: Adds `pre` tags to multi-line code.
* Quotes: Adds class to blockquotes & fix next-line text getting pulled into them.


tburry:
Consider the following markdown:

```
This is a paragraph.

`echo 'one';
echo 'two';`
```

This will render an extra empty `<p>` tag in between the first paragraph and the source code which really shouldn't happen. This is a bug in current master too, but it's still a bug. I'm okay with filing a later ticket though.

tburry:
Here is another odd bug. Consider the following markdown:

<pre><code>```
&lt;http://example.com/&gt;
```</code></pre>

This will render a trailing sem-colon. Probably a bug in our autolinking more that markdown.

tburry:
I'm okay with merging this and adding the two issues I reported to our list.

linc:
Filed #2445

PR #2444, opened 2015-01-27
Fix typos in calls to DiscussionModel->GetID().

tburry:
Since `DiscussionModel->GetID()` can only return an object currently, specifying an array is wrong and just a land mine of a regression bug waiting to happen.


linc:
Agreed.

PR #2446, opened 2015-01-27
Rate limit logins on installs without a cache

bleistivt:
This chooses the best available store for rate limiting:
Gdn_Cache > APC store > Database (User attributes)

I have tested the APC and UserModel store variant. The Gdn_Cache implementation hasn't changed.

New pull for #2346 with a cleaner changeset


linc:
Kudos on getting this in!

PR #2447, opened 2015-01-28
Add support for default category for new discussions (Vanilla.Categories.Default)

michaeltyson:
This provides the facility to specify a default category for new discussions, via the Vanilla.Categories.Default option (which is a category URL code). We want this for our own forum in order to avoid fragmentation by users who don't care what category their new discussion goes into.

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -106,6 +106,10 @@ class PostController extends VanillaController {
       
       Gdn_Theme::Section('PostDiscussion');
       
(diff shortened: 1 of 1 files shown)


linc:
Howdy! Have you signed our CLA? http://vanillaforums.org/contributors/ We need that before we can review your code change.

Minor changes can be targeted against `master`; big changes can go to `stage`. Unfortunately, `develop` is a dead branch we are working on removing so we can't accept this there.

michaeltyson:
Ah, okay, no worries. It's not a particularly important change, and I'm not planning on any active development, so will probably just let this one go =)

PR #2452, opened 2015-01-28
Remove the markup breaking star emoji alias.

tburry:
Also added some other emoji fixes.


linc:
All of this is good and proper.

PR #2453, opened 2015-01-28
Log model: Add count increments to User model once pending discussions/c...

beckyvb:
...omments are approved.

Especially useful for harnessing Ranks to make a user verified after they have a set number of posts.


applications/dashboard/models/class.logmodel.php
```diff
@@ -452,14 +452,38 @@ public function Recalculate() {
...
+         $this->_RecalcIDs['Discussion'] = array();
+      }
+
+      $UserIDsComment = $this->_RecalcIDs['UserComment'];
+
+      $In = implode(',', array_keys($UserIDsComment));
+      if (!empty($In)) {
+
+         $Px = Gdn::Database()->DatabasePrefix;
+         $Sql = "update {$Px}User set CountComments = coalesce(CountComments, 0) + 1 where UserID in ($In)";
```
linc: While I understand the original sub-select using `Query()` directly (it's not immediately obvious to me if it's even possible otherwise in our SQL driver), I'm not sure these more straight-forward statements warrant it. Naked SQL is technical debt.

linc:
If this is tested & working I don't have any other feedback on this besides the SQL refactor; looks good otherwise.

PR #2455, opened 2015-01-30
Import: Force current user's password on new admin

initvector:
Removed UseCurrentPassword, Password form fields and references;
Forcing all conditional code related to UseCurrentPassword being true
to run on import


linc:
Let's add a note to the effect of "Your current password will become this user's password." to clarify what's about to happen.

tburry:
Just want to make sure you're not going to start a whitespace war with everyone else. Make sure you have your editorconfig plugin enabled and whatnot.

linc:
@tburry My assumption was that `importcontroller` was full of bad line endings and this was the first commit since you added that force option to the repo.

initvector:
EditorConfig IDE plug-in is installed and enabled, but I'm still seeing the same behavior.  Vim interprets the original file as DOS, so I'm also assuming the full change is related to line endings.

tburry:
Yeah I think it looks like you are ending the war. Not starting it.

PR #2460, opened 2015-01-31
First post

linc:
- [x] Throw event for a user's first comment / discussion / post.
- [x] Add attribute to comments & discussions that are a user's first.
- [ ] Add relevant CSS class to marked discussions / comments.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -446,7 +446,7 @@ class CommentModel extends VanillaModel {
 
             if ($CountWatch < $Discussion->CountCommentWatch)
                $CountWatch = $Discussion->CountCommentWatch;
-               
+
             if (isset($Discussion->DateLastViewed))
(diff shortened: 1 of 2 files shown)


linc:
Interested in some feedback before I push forward with the rest of this.

tburry:
This is not a feature that can be easily implemented without a lot of issues. So I'd recommend not pursuing.

The user count is not super accurate right now. Imports in particular can cause it to not be synchronized. Right now I think it's engineered to catch up every once in a while, but can't be relied upon for a feature like this.

linc:
I agree with bulletproofing post counts as a prerequisite. I also like attaching a badge to content as a feature, but I'd like to see the ability to CSS class & notify about these posts in core. I think this is helpful info in nearly _every_ forum scenario, regardless of whether you have gamification. Onboarding is super challenging and we should do a lot more to support it.

linc:
I do think this is an important feature, but I'm considering your comment about adding "1-off" features in core. There's a few strategic changes I'd like to see (#2514, for instance, and a way to hook a post count trigger once that's more solid) that would make this a nice plugin that I think would see a lot of use. So I'll approach the core changes needed from that perspective instead.

PR #2476, opened 2015-02-04
Profile editing: fallback to Preferences to match MeModule behavior

linc:
If a user doesn't have profile editing permission, the MeModule falls back to a "Preferences" link. The profile should do the same.


beckyvb:
:+1:

PR #2477, opened 2015-02-04
Fix PagerUrl to allow hooking & setting

linc:
* Setting `_PagerUrl` had no effect on `DiscussionsController->Index()`
* The Pager's `Url` option was ignored after page 1.


linc:
This is to address an open support ticket where the Q&A plugin's Unanswered section neither links to page 2 correctly, nor is capable of displaying a page 2 when navigated manually.

PR #2482, opened 2015-02-07
Fix malformed HTML in delete popup

bleistivt:
The div element around the buttons wasn't properly closed.


hgtonight:
+1 makes sense

PR #2484, opened 2015-02-09
Add dba/fixtree for categories after import/manipulation

linc:
We do imports that only set `Sort` & `ParentCategoryID`, and/or sometimes need to do manual manipulations of the category table. This lets us rebuild the tree on the fly.

### DIFF
--- a/applications/dashboard/controllers/class.dbacontroller.php
+++ b/applications/dashboard/controllers/class.dbacontroller.php
@@ -12,51 +12,51 @@
 
 class DbaController extends DashboardController {
    /// Properties ///
-   
+
    /**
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/class.dbacontroller.php
```diff
@@ -81,39 +81,57 @@ public function FixPermissions() {
       $this->AddSideMenu();
       $this->Render('Job');
    }
-   
+
+   /**
+    * Fix the category tree after an import that only gives a sort & parent.
+    */
+   public function FixTree() {
```
tburry: `RebuildCategoryTree()`

PR #2486, opened 2015-02-10
Refactor the validation object.

tburry:
The GDN_Validation object has some severe edge cases and this aims to deal with them.

- Remove a bunch of properties that can be calculated from other properties.
- Make `Gdn_Validation->Validate()` a bit more idempotent.
- Separate rules applied manually from those applied from the schema so that the schema does not pollute the validation object during `Validate()`.
- Protect methods that really shouldn’t be called externally.
- Remove some dangerous methods.
- Allow columns with default values to be left out during inserts.
- Add a ResetOnValidate property to Gdn_Validation to allow the validation results to reset whenever `Validate()` is called.
- Change the way auto-increment fields are validated.
- Allow empty strings to validate as required if the default field value is an empty string.

### DIFF
--- a/applications/dashboard/models/class.rolemodel.php
+++ b/applications/dashboard/models/class.rolemodel.php
@@ -10,35 +10,35 @@ Contact Vanilla Forums Inc. at support [at] vanillaforums [dot] com
 
 class RoleModel extends Gdn_Model {
    public static $Roles = NULL;
-   
+
    /**
     * Class constructor. Defines the related database table name.
     */
    public function __construct() {
       parent::__construct('Role');
    }
-   
+
    public function ClearCache() {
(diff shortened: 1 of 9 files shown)


tburry:
This is a big honking change. Once it's ready much testing will be needed. There may be some code that depends on bad behaviour within this class, but I'm hoping that there won't be much and those bugs can be fixed.

linc:
Getting a "DraftID required" error when a comment is auto-saved.

tburry:
I added an unset specifically to the draft model, but this has uncovered a bit of backwards incompatibility. Before, auto increment fields would not be validated at all which to me is wrong so I'd rather keep the validation object like it is and hunt the bugs down that rely on that bad behaviour.

PR #2487, opened 2015-02-10
Increment user model post counts for approved pending posts.

beckyvb:
Rebasing https://github.com/vanilla/vanilla/pull/2453 off master.


linc:
cleditor is being insane in every branch it's still in, demanding I commit formatting changes to the js and css files. I have no idea why it's showing in this diff, since it's not in master either. *rage*

PR #2493, opened 2015-02-12
Use as many recently used IP addresses in AllIPAddresses that will fit

### LINKED ISSUE #2005: Data too long for column 'AllIPAddresses'
A user has reported to me that when they try to sign in to my forum, they get

```
Fatal Error in Gdn_Database.Query();
Data too long for column 'AllIPAddresses' at row 1
update GDN_User `User` set DateLastActive = :DateLastActive, LastIPAddress = :LastIPAddress, CountVisits = :CountVisits, AllIPAddresses = :AllIPAddresses where UserID = :UserID
The error occurred on or near: /home/www/forum/library/database/class.database.php
```

I'm unsure what this field represents (all IP addresses used ever by this user?), but it seems like this should be truncated at the field size.

JasonBarnabe:
Fixes #2005.

- `AllIPAddresses` is now a list of the most recently used IP addresses rather than being "all".
- This list is now sorted from least recently used to most recently used rather than sorted alphanumerically.
- If a new IP address brings the list over the size of the column (currently 100 characters), least recently used IP address(es) will be dropped until it's back within the limit.

Note that 100 characters is enough room for between 6 and 14 IP addresses.

**Test 1 - a new IP address brings you over the limit**
- Update your user to have a nearly full IP address list: `update GDN_User SET AllIPAddresses = '123.123.123.109,123.123.123.107,123.123.123.105,123.123.123.103,123.123.123.101,123.123.123.100' where UserId = YOURUSERID;`
- Log out
- Log in

Expected result: no error, login works fine. `AllIPAddresses` is set to 
~~123.123.123.109,~~123.123.123.107,123.123.123.105,123.123.123.103,123.123.123.101,123.123.123.100**,127.0.0.1**
Oldest IP is removed, new IP is added.

**Test 2 - a new IP address was used in the past**
- Update your user to have a lot of IP addresses, including the IP address you're currently using: `update GDN_User SET AllIPAddresses = '127.0.0.1,123.123.123.107,123.123.123.105,123.123.123.103,123.123.123.101,123.123.123.100' where UserId = YOURUSERID;`
- Log out
- Log in

Expected result: no error, login works fine. `AllIPAddresses` is set to 
~~127.0.0.1,~~123.123.123.107,123.123.123.105,123.123.123.103,123.123.123.101,123.123.123.100**,127.0.0.1**. 
Previously used IP address is moved to the end of the list.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2455,8 +2455,8 @@ class UserModel extends Gdn_Model {
          $AllIPs[] = ForceIPv4($IP);
       if ($IP = GetValue('LastIPAddress', $User))
          $AllIPs[] = $IP;
-      $AllIPs = array_unique($AllIPs);
-      sort($AllIPs);
+      // This will be a unique list of IPs, most recently used last. array_unique keeps the first key found.
+      $AllIPs = array_reverse(array_unique(array_reverse($AllIPs)));
       $Fields['AllIPAddresses'] = $AllIPs;
 
       // Set the hour offset based on the client's clock.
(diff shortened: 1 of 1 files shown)


JasonBarnabe:
Tested on master and 2.1.7.

tburry:
Great idea. The one nitpick I'm going to make is why not use an `array_unshift()` and have the most recently used IP at the front of the list. Here are some reasons:

* Removes the heavier `array_reverse()` call.
* You can more easily see at a glance the more important of IP addresses.
* The most import IP addresses will never get truncated.
* A "move to front" is a pretty standard computer science algorithm.

JasonBarnabe:
IP list is now most recently used to least recently used.

PR #2501, opened 2015-02-17
Unicode support for mentions and usernames

linc:
* [x] Create `unicodeSupport()` to detect ability to use Unicode regex.
* [x] Allow Unicode (including connector punctuation) for username validation.
* [x] Allow Unicode (including connector punctuation) for mentions.
* [x] Expand allowed mention length to 64 characters because nerds.
* [x] Fix redundant non-Unicode regex fallback for username regex (`\d\w_` is the same as `\w`)


tburry:
This function name is misleading. You are actually testing for `unicodeRegexSupport()`.

linc:
Fixt

Issue #1729, opened 2013-11-02
Please remove eval modifier 'e' from preg_replace()


kiang:
I just found that when you using double quote in [code] tags and format using bbcode formater, it will add additional slash to each double quote. The problem caused by line # 265 in file library/core/class.format.php. It uses 'sie' as modifiers and the 'e' one executes eval() function to deal with the text. As mentioned in PHP manual, it also runs addslashes() function automatically.

As the eval() function will be removed in PHP 5.5 or later, maybe it's time to prevent using related functions.

PR #2502, opened 2015-02-17
Remove the /e modifier from Gdn_Format::HtmlFilter().

tburry:
The /e modifier is deprecated and will end up being removed before we know it. This change accomplishes what the /e modifier accomplished with `preg_replace_callback()`.


linc:
closes #1729

PR #2504, opened 2015-02-17

chanhong:
Per discussion with @linc and @tburry, Fix parse_ini_file for running from a Windows network share #2439


chanhong:
This PR is not very clean.  I will close this and make a clean PR.

PR #2506, opened 2015-02-18
Strip out unnecessary call to htmlspecialchars

R-J:
Function `Img()` uses function [`Attribute()`](https://github.com/vanilla/vanilla/blob/master/library/core/functions.render.php#L560) 
to transform the attributes array to html and [that function uses htmlspecialchars](https://github.com/vanilla/vanilla/blob/master/library/core/functions.general.php#L445)  for each value. So it is not needed to use htmlspecialchars already in the function call.


linc:
Thanks for the explanation! I love these little improvements but I definitely don't have the patience to look up the rationale for each one, so it's much appreciated.

PR #2507, opened 2015-02-18
Harmonize charset (where used) in htmlspecialchars

R-J:
htmlspecialchars uses a mixture of `C('Garden.Charset', '')`, `C('Garden.Charset', 'UTF-8')`, `UTF-8` and no charset info at all. I've harmonized all that uses any charset parameter to respect config setting and default to UTF-8 (`C('Garden.Charset', 'UTF-8')`)

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -640,7 +640,7 @@ class Gdn_Format {
       if (!is_string($Mixed))
          return self::To($Mixed, 'Display');
       else {
-         $Mixed = htmlspecialchars($Mixed, ENT_QUOTES, C('Garden.Charset', ''));
+         $Mixed = htmlspecialchars($Mixed, ENT_QUOTES, C('Garden.Charset', 'UTF-8'));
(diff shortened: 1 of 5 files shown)


bleistivt:
+1

@tburry For the future, what do you think about adding a wrapper function for htmlspecialchars like ``hsc()`` or just ``h()``?
I've seen this in a lot of frameworks, it saves typing and gives a chance to define a charset globally.

linc:
Looks good to me.

I am going to cautiously +1 the wrapper idea because we use `htmlspecialchars` in more than 150 places in core and clearly it's gotten out of whack.

linc:
@tburry thoughts?

tburry:
I'm up for the wrapper. List some options.

```
hsc()
h() // nope
hesc() // for html escape, html special chars is such a phpism.
```

linc:
I think I'd like `htmlEsc()` for clarity. I'm not a fan of abbreviating into obfuscation, and this isn't nearly at the level of `c()` or `t()` usage.

PR #2508, opened 2015-02-18

chanhong:
I found out where the actual fix were so I change only on that to implement the fix per @tburry feedback to get a clean PR.


linc:
ref #2439

PR #2510, opened 2015-02-20
Add the proper permission check for the me box's applicants link.

tburry:


linc:
Should get added to the check above it too, no?

PR #2512, opened 2015-02-21
Exploding HOST (localhost:9090 => [localhost, 9090] and cutting out the ...

linc:
...port results in incorrect paths when running Vanilla Forums locally on any post other than 80

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -96,9 +96,6 @@ class Gdn_Request {
                $Value = !is_null($Value) ? trim($Value, '/') : $Value;
                break;
             case 'HOST':
-               $HostParts = explode(':', $Value);
-               $Value = array_shift($HostParts);
-               break;
             case 'SCHEME':
             case 'METHOD':
             case 'FOLDER':


linc:
This was lost into the void of `develop` branch. See #1386

tburry:
This is the wrong fix. Host should return the host. Not the host and the port.

tburry:
We even have `Gdn_Request->HostAndPort()` which should be used instead of `Gdn_Request->Host()` where appropriate.

onurguzel:
I have encountered a malfunction related to this issue. I am running nginx, and in the `Gdn_Request->_LoadEnvirontment()` method obtains host information via `$_SERVER['HTTP_HOST']` which has a value of `host:port`. This causes invalid host information to be stored in environment (and bad URLs such as `http://host:port:port/index.php?p=/dashboard/setup`

I think the proper way to fix this is modifying the `_LoadEnvironment()` method. Should I make changes and create a new pull request?

linc:
@onurguzel Please start a new discussion on the forum with your version number for troubleshooting help. I suspect you're running an older version of Vanilla.

onurguzel:
@linc this problem still exists in 2.1.10, which is the latest version. but i saw that it has been fixed ( #2170  ) months ago in master.

linc:
Thanks for the heads up! We're going to be doing 2.2 beta shortly so hopefully it won't need to get backported.

Issue #2456, opened 2015-01-30
Dashboard organization: Move Max/Min Comment Length settings


linc:
"Flood Control" makes no sense. Should be under Advanced with the rest of them. Or, better, abolish Advanced and group them under a new page w/ Comments Per Page etc.

PR #2519, opened 2015-02-23
#2456 min/max comment length to advanced

russellwwest:
Moved min/max comment length to advanced section, instead of flood
control.


linc:
Howdy @russellwwest thanks for the PR! I'm curious why you added 'p' tags wrapping the Save button on the floodcontrol view.

russellwwest:
Hi, Only reason was to begin with the Save button was up against the above block without spacing. So i put those in temporarily and forgot to take them out.

PR #2520, opened 2015-02-24
Wrap inline CODE tags in FireFox

tophf:
Unlike Chrome, FireFox doesn't force wrapping of long lines inside code blocks with white-space:pre so such lines break the site layout.

### DIFF
--- a/applications/dashboard/design/style.css
+++ b/applications/dashboard/design/style.css
@@ -1998,18 +1998,22 @@ code, pre {
    -webkit-border-radius: 2px;
    background: #ff9;
    padding: 4px 8px;
-   white-space: pre;
    font-family: monospace;
    overflow: auto;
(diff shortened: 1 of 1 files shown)


tburry:
This looks like a change that will be successful in markdown, but not all of the input formats. I'm wondering if the following would work:

```css
code {
    whitespace: pre-wrap;
}

pre, pre code {
    whitespace: pre;
}
```

tophf:
Good catch.
I've updated the commit.

tburry:
Can you sign our contributor's agreement @tophf? http://vanillaforums.org/contributors

tophf:
Done.

P.S. wow such document much caution!

linc:
@tophf The world is run by lawyers! :)

PR #2525, opened 2015-02-26
Clear cache on config save for Zend OPcache

shadowdare:
A better solution than the previous pull request #2523. Tested and works.


linc:
Nice one @shadowdare!

PR #2526, opened 2015-02-26
Allow users without admin access to assign roles and edit permissions.

tburry:
- Define a set of ranking permissions. This means that users can only assign roles/permissions if they
have those ranking permissions.
- Administrators can assign any role.
- Community managers with the edit users permission can assign roles that they have ranking permission of.
- Other users with the edit users permission can assign roles without any ranking permission.
- Add a hidden Garden.Roles.Manage permission that can be defined in plugins that will allow non admins to add and edit roles.

### DIFF
--- a/applications/dashboard/controllers/class.rolecontroller.php
+++ b/applications/dashboard/controllers/class.rolecontroller.php
@@ -247,9 +247,24 @@ class RoleController extends DashboardController {
       $this->AddJsFile('jquery-ui.js');
       $this->Title(T('Roles & Permissions'));
 
-      if (!$RoleID)
-         $RoleData = $this->RoleModel->Get()->ResultArray();
-      else {
+      if (!$RoleID) {
+         $RoleData = $this->RoleModel->GetWithRankPermissions()->ResultArray();
+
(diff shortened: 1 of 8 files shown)


linc:
This changset is working.

My concern, after seeing it in action, is that we're empowering moderators to appoint new moderators AND it's an unlogged change. That means they could register a new account, make them a moderator, use that new account to ban the person they have an ax to grind with, and OOPS there's no record how that happened now.

At most, I'd let them appoint curator-level users (one "rank" under their own rank). Not only because of the abuse vector, but because I think joining the moderator team is a decision reserved to admins.

tburry:
I see your point. Your use-case needs to be taken as an example though and a more general solution needs to be made.

linc:
I don't follow. Are you saying moderators should usually be allowed to promote other moderators, and my scenario is an edge case? Because I think the inverse is true.

tburry:
Okay I've updated the roles a user is allowed to assign so that admins, community managers, and users with user/edit all have different levels of role assignability. Check out the updated description of this PR for details.

linc:
:+1:

Issue #2467, opened 2015-02-02
Smileys are rendered inside code tags


R-J:
Smiley shortcuts inside code tags will be rendered. See: http://vanillaforums.org/discussion/comment/225036/#Comment_225036   
Text is first parsed for smileys and afterwards handled by the js code renderer so that code already contains smileys.

JasonBarnabe:
More generally, nothing after the `code` start tag should be special except the `code` end tag.

valeryan:
This is also triggering on youtube/vimeo urls inside code blocks. Your formatting should probably skip over everything inside a code block

JasonBarnabe:
With this PR, YouTube links are also left alone when inside `<code>`.

PR #2531, opened 2015-02-27
Don't linkify @mentions and #hashtags in CODE blocks

tophf:
A straightforward/dumb implementation.
Arguably it should be generalized to protect code blocks from all kinds of content modification (see #2467), implemented via preg_replace_callback processing the inter-code spans (in one step, unlike the two-step replacement here), if the idea itself will be accepted.


initvector:
Addressed this issue with #2694

PR #2533, opened 2015-03-01
Don't show react button if social reactions are turned off

bleistivt:
The config option ``Plugins.GooglePlus.SocialReactions`` wasn't respected by the plugin


bleistivt:
fixes https://github.com/vanilla/addons/issues/140

The issues should be moved.

linc:
Going to look at moving them with a tool I found. Otherwise it'll have to be manual copy/paste.

PR #2535, opened 2015-03-02
Discussions controller: Add where param to Index method.

beckyvb:
Solves problem where pagination renders based on total number of discussions, rather than the view being rendered.


applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -73,8 +74,9 @@ public function Table($Page = '0') {
     * @access public
     *
     * @param int $Page Multiplied by PerPage option to determine offset.
+    * @param array $where Discussions query filter.
     */
-   public function Index($Page = FALSE) {
+   public function Index($Page = FALSE, $where = array()) {
```
tburry: We can't take a where clause in a controller method. A malicious user could pass something to this.

tburry:
This PR is insecure. I'm going to have to turn it down. We can talk about a better way.

PR #2540, opened 2015-03-03
Add events for TrollManagement fixes

linc:
Adds GetByUser events for filtering out troll content on their own profiles. Adds a 'UserLoaded' event to ProfileController to intercept userdata before logic is done on it.


linc:
I should mention that I am simply copying existing data/event naming for the DiscussionModel & CommentModel events so that the existing hooks in TrollManagement work in these cases as well. I have no idea why they are using completely different nomenclatures already and despair for our mortal souls, but feel I have chosen the lesser of two evils.

PR #2541, opened 2015-03-03
Fix array index assumption in profile activity array

linc:
We want the first activity, not necessarily index zero. Grouping activity means this could start with a higher number.


applications/dashboard/controllers/class.profilecontroller.php
```diff
@@ -135,7 +135,7 @@ public function Activity($UserReference = '', $Username = '', $UserID = '', $Pag
       $this->ActivityModel->JoinComments($Activities);
       $this->SetData('Activities', $Activities);
       if (count($Activities) > 0) {
-         $LastActivity = $Activities[0];
+         $LastActivity = array_shift(array_values($Activities));
```
tburry: `array_shift($Activities)` will suffice.

linc: `array_shift` would also remove it from `$Activities`, which isn't what this code was doing.

tburry: Okay then go with `reset()`.

PR #2546, opened 2015-03-04
CountString() ignored CssClass parameter for "normal" counts

R-J:
The function returned `" <span class=\"Count\">$Number</span>"` for `CountString(10, '', array('CssClass' => 'SomethingSpecial'))`.
Now it returns `"<span class=\"SomethingSpecial Count\">$Number</span>"`.

**Note that I've also stripped the preceding space from the output!** The function could also return an empty string or `"<span class=\"$CssClass\" rel=\"$Url\"></span>"` and they have no space as the first character so I think there is no problem stripping it.

I've changed "if is_string($Options) ... else" to "if is_array($Options) ... else" so that numerical parameters for $Options do not break the function. They will end as (useless) class="13 Count" though...

I made some changes to the rest of the function and this are the results in a speed comparison:

CountString(0): **80% faster**
CountString(Null, $Url, array('CssClass' => 'Something')): **25% faster**
CountString(Null, $Url, 'SomeCssClass'): **>60% faster**
CountString(10, '', array('CssClass' => 'Something')): *20% slower*, **but** only because CssClass is ignored in the old function. If that would be handled like it is done in the current function: 
~~~
$CssClass = ConcatSep(' ', $CssClass, 'Count');
$Result = " <span class=\"$CssClass\">$Number</span>";
~~~
the new function would be **30% faster** ;-)

Changes are mainly:
Set the easiest (`return '';`) right at the beginning
Return results as soon as possible and not storing them in a variable that is returned at the end of it all
If is_string($Options) $CssClass = $Options, don't stuff it in an array first to extract that same value afterwards
Use trim($CssClass.' Count/Popin', ' ') instead of ConcatSep()

Here's the code I used to do the speed comparison:

~~~
<?php

define('ITERATIONS', 100000);

	function val($key, $collection, $default = false) {
		if (is_array($collection) && array_key_exists($key, $collection)) {
			return $collection[$key];
		} elseif (is_object($collection) && property_exists($collection, $key)) {
			return $collection->$key;
      }
      return $default;
	}
    
    function ConcatSep($Sep, $Str1, $Str2) {
      if(is_array($Str2)) {
         $Strings = array_merge((array)$Str1, $Str2);
      } else {
         $Strings = func_get_args();
         array_shift($Strings);
      }

      $Result = '';
      foreach($Strings as $String) {
         if(!$String)
            continue;

         if($Result)
            $Result .= $Sep;
         $Result .= $String;
      }
      return $Result;
   }

   function CountString($Number, $Url = '', $Options = array()) {
      if (is_string($Options))
         $Options = array('cssclass' => $Options);
      $Options = array_change_key_case($Options);
      $CssClass = val('cssclass', $Options, '');

      if ($Number === NULL && $Url) {
         $CssClass = ConcatSep(' ', $CssClass, 'Popin TinyProgress');
         $Url = htmlspecialchars($Url);
         $Result = "<span class=\"$CssClass\" rel=\"$Url\"></span>";
      } elseif ($Number) {
//         $CssClass = ConcatSep(' ', $CssClass, 'Count');
//         $Result = " <span class=\"$CssClass\">$Number</span>";
           $Result = " <span class=\"Count\">$Number</span>";
      } else {
         $Result = '';
      }
      return $Result;
   }

   function CountStringNew($Number, $Url = '', $Options = array()) {
      if (!$Number && $Number !== Null) {
         return '';
      }
      
      if (is_array($Options)) {
         $Options = array_change_key_case($Options);
         $CssClass = val('cssclass', $Options, '');
      } else {
         $CssClass = $Options;
      }
      
      if ($Number) {
         $CssClass = trim($CssClass.' Count', ' ');
         return "<span class=\"$CssClass\">$Number</span>";
      } elseif ($Number === NULL && $Url) {
         $CssClass = trim($CssClass.' Popin TinyProgress', ' ');
         $Url = htmlspecialchars($Url);
         return "<span class=\"$CssClass\" rel=\"$Url\"></span>";
      } else {
         return '';
      }
   }

  
    
    
$time_start = microtime(true);
$time_end = microtime(true);
$time = $time_end - $time_start;
$test = '';


// test old
$time_start = microtime(true);
for ($k = 0; $k < ITERATIONS; $k++) {
    $test = CountString(10, '', array('CsSClAsS' => 'blubb'))."\n";
/*
    $test =  CountString(null, "vf.org", array('CsSClAsS' => 'blubb'))."\n";
    $test = CountString(10, '', array('CsSClAsS' => 'blubb'))."\n";
    $test = CountString(0)."\n";
*/
}
$time_end = microtime(true);
$time = $time_end - $time_start;
echo $test;
echo "$time seconds for old\n";

// test new
$time_start = microtime(true);
for ($k = 0; $k < ITERATIONS; $k++) {
    $test = CountStringNew(10, '', array('CsSClAsS' => 'blubb'))."\n";

/*
    $test =  CountStringNew(null, "vf.org", array('CsSClAsS' => 'blubb'))."\n";
    $test = CountStringNew(10, '', array('CsSClAsS' => 'blubb'))."\n";
    $test = CountStringNew(0)."\n";
*/
}
$time_end = microtime(true);
$time = $time_end - $time_start;
echo $test;
echo "$time seconds for new\n\n";
~~~

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -231,21 +231,27 @@ if (!function_exists('Condense')) {
 
 if (!function_exists('CountString')) {
    function CountString($Number, $Url = '', $Options = array()) {
-      if (is_string($Options))
-         $Options = array('cssclass' => $Options);
-      $Options = array_change_key_case($Options);
-      $CssClass = GetValue('cssclass', $Options, '');
-
-      if ($Number === NULL && $Url) {
-         $CssClass = ConcatSep(' ', $CssClass, 'Popin TinyProgress');
-         $Url = htmlspecialchars($Url);
-         $Result = "<span class=\"$CssClass\" rel=\"$Url\"></span>";
-      } elseif ($Number) {
-         $Result = " <span class=\"Count\">$Number</span>";
-      } else {
-         $Result = '';
+      if (!$Number && $Number !== Null) {
+         return '';
       }
-      return $Result;
+      
+      if (is_array($Options)) {
+         $Options = array_change_key_case($Options);
+         $CssClass = val('cssclass', $Options, '');
+      } else {
+         $CssClass = $Options;
+      }
+      
+      if ($Number) {
+         $CssClass = trim($CssClass.' Count', ' ');
+         return "<span class=\"$CssClass\">$Number</span>";
+      } elseif ($Number === NULL && $Url) {
+         $CssClass = trim($CssClass.' Popin TinyProgress', ' ');
+         $Url = htmlspecialchars($Url);
+         return "<span class=\"$CssClass\" rel=\"$Url\"></span>";
+      } else {
+         return '';
+      }
    }
 }
 
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -231,7 +231,7 @@ if (!function_exists('Condense')) {
 
 if (!function_exists('CountString')) {
    function CountString($Number, $Url = '', $Options = array()) {
-      if (!$Number && $Number !== Null) {
+      if (!$Number && $Number !== null) {
          return '';
       }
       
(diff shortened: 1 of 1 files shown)


linc:
My only quibble is you use both `Null` and `NULL` when we now use `null` :)

R-J:
Done!

PR #2549, opened 2015-03-04
Create an endpoint for accessing promoted content

linc:
* Add `Promoted` endpoint on `DiscussionsController` to access `PromotedContentModule`.
* Filter additional user info from API calls that should be PersonalInfo by default. It would otherwise be exposed by this endpoint.
* Refactor PromotedContentModule to allow controller access  via `Load` & `Validate` methods.
* Make the `SelectBy*` methods all work the same if only `selection` value is passed as `$Parameters`.
* Fix array index issue in `Prepare`.
* Fix passing ID array in `SelectByCategory`.
* Fix misc coding standards & docs throughout `PromotedContentModule`.
* Fix caching of multiple RoleIDs in selection (array to string conversion error).
* Allow Ranks & Categories to be selected by Name.
* Remove date keying of `Content` array.

### DIFF
--- a/applications/vanilla/modules/class.promotedcontentmodule.php
+++ b/applications/vanilla/modules/class.promotedcontentmodule.php
@@ -346,6 +346,14 @@ class PromotedContentModule extends Gdn_Module {
          $CategoryID = GetValue('CategoryID', $Parameters, NULL);
       }
 
+      // Allow category names, and validate category exists.
+      // Disallow multiple categories.
+      if (is_array($CategoryID)) {
+         $CategoryID = array_shift($CategoryID);
+      }
+      $Category = CategoryModel::Categories($CategoryID);
+      $CategoryID = val('CategoryID', $Category);
+
       if (!$CategoryID) {
          return false;
       }
--- a/applications/vanilla/modules/class.promotedcontentmodule.php
+++ b/applications/vanilla/modules/class.promotedcontentmodule.php
@@ -201,7 +201,9 @@ class PromotedContentModule extends Gdn_Module {
       }
 
       // Check cache
-      $SelectorRoleCacheKey = "modules.promotedcontent.role.{$RoleID}";
+      sort($RoleID);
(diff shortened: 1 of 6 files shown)


applications/vanilla/modules/class.promotedcontentmodule.php
```diff
@@ -89,10 +166,11 @@ public function GetData() {
    }
 
    /**
-    * Select content based on author RoleID
+    * Select content based on author RoleID.
     *
-    * @param array $Parameters
-    * @return boolean
+    * @param mixed $Parameters
```
tburry: FYI: You'd want to do `@param int|array $Parameters` instead of mixed.

applications/vanilla/views/discussions/promoted.php
```diff
@@ -0,0 +1 @@
+<?php if (!defined('APPLICATION')) exit();
```
tburry: I'd go ahead and write the module out in this view instead of nothing.

linc:
Implemented line comments above. I also discovered an event name conflict in the process - using `AfterBody`, which is the Debugger hook - so I swapped that too. Hoping that's not used in some obscure place but there's no way around it and it's really bad.

tburry:
This is really hard to test without any documentation or examples.

tburry:
Consider the following path: `/discussions/promoted?selector=role&selection=Administrator`

When I run this with debug on I get data just fine, but there are a lot of notices in the page. I'd rather those be cleaned up before merge.

tburry:
I'm also noticing that for the json endpoint the array is keyed by date which is very odd for non-php consumers. This should just be a regular javascript array.

linc:
I fixed the above issues, and added/fixed the ability to select ranks & categories by name. 

I've also blocked the ability for multiple ranks or categories to be passed for now. Role already allowed multiple (on purpose) and that could be in use, so I've left it be.

PR #2557, opened 2015-03-09
Flagging: validate incoming URLs

linc:


linc:
The URL being passed here is only ever a relative discussion/comment URL.

Would `HtmlFilter` be a more appropriate filter here?

tburry:
I'm pretty sure `htmlspecialchars()` is what you want. It will display the exact text of the url. That's what you want.

PR #2558, opened 2015-03-09
Smarty pocket: Add Sender param.

beckyvb:
Give pockets plugin more info to check page when rendering.


beckyvb:
Needed for https://github.com/vanilla/addons/pull/195

tburry:
This needs to be reverted. PR #2564.

Issue #1482, opened 2012-11-24
Issue with Social function (GetWebResource related ? )


cataldoc:
Hi, I am using the latest master baea4630b9 and I found an issue with the Social function. My forum is installed in the /forum directory. I changed the .htaccess file accordingly to have the correct rewritebase.

Problem is that if I try to use the Facebook or Twitter integration I get a "Page not Found" error due to Garden trying to locate the plugin in /forum/forum/plugins/Facebook (which ofc. does not exists).

This problem is not present if I install Garden in the apache root directory.

I found a previous (closed) issue regarding a similar problem (issue 712 here) related to the GetWebResource function not sure it's related but the effects looks exactly the same.

bleistivt:
I guess this is already fixed, the social plugins work for me (and a lot of other people), both in root and in a subfolder.

PR #2562, opened 2015-03-11
Workaround for HHVM #1482

bbs4us:
This is a workaround for HHVM by x00, as https://github.com/facebook/hhvm/issues/1482 is intentional.


linc:
Hi bbs4us, thanks for the contribution. Please sign our CLA here: http://vanillaforums.org/contributors

This code isn't following our coding standards for indenting or spaces around keywords.

PR #2563, opened 2015-03-11
Fix HHVM #1482

bbs4us:
CLA signed.


tburry:
Thanks for this. I'm hoping you can have a look at #2566 to see if that will work for you. It's a bit more efficient and easier to debug.

bbs4us:
Yeah, that worked. Thanks.

PR #2566, opened 2015-03-12
Fix for using iterator by reference in foreach

initvector:
Improves HHVM compatibility


linc:
Did you see #2563?

tburry:
If you look at the code here the if statement is moved outside of the tight loop and is thus more efficient and easier to follow in a debugger.

PR #2577, opened 2015-03-18
Apply filter to Title

linc:


tburry:
Won't this double encode if you add an html character to a discussion or category title or any of the wide variety of things that have already escaped the title?

Most escaping is actually done at the view level because views are constructed for not just html and thus only the view knows how its data should be escaped.

PR #2578, opened 2015-03-18
gitattributes: add files to ignore on archive

linc:


.gitattributes
```diff
@@ -6,3 +6,7 @@
 *.less  text eol=lf
 *.md    text eol=lf
 *.yml   text eol=lf
+
+# Skip files on archive
+.editorconfig export-ignore .gitattributes export-ignore .gitignore export-ignore
+.travis.yml export-ignore tests export-ignore
```
tburry: Why does this spacing look so odd?

PR #2579, opened 2015-03-18
Add embed theme options + advanced embedding options

daedeloth:
(This is a new version of https://github.com/vanilla/vanilla/pull/2262)

These changes add a new tab to the settings page "embed theme" (only visible when embedding is enabled). Works very similar to the mobile theme.

A query parameter embed=1 is added to determine if the forum is embeded.
(As far as I understand from online documentation, this should not be a problem seo wise)

Working implementation:
http://catlab.be/forum (embedded)
http://forum.catlab.eu/ (bootstrap theme)

More info:
http://vanillaforums.org/discussion/28467/separate-embed-theme-setting#latest


daedeloth:
I did some more cleanup in the embed.js file.
Forum is live at http://forum.catlab.eu/ and http://www.catlab.be/forum

tburry:
I like the idea of a separate embed theme, but this methodology is fundamentally incompatible with some of the work we're doing with assets (css and js). We'll be providing support for a separate embed context in the future, but it will have to be off of a virtual /embed subdirectory that Vanilla will look at to determine the embed theme.

daedeloth:
I don't really understand what you mean by that.
This implementation was basically just a copy of what was done on the mobile theme before...

linc:
Howdy @daedeloth, this was clearly a ton of work and we really appreciate the effort you put into this. There's a couple key things in this that conflict with some roadmap items we're working towards which is what Todd was alluding to.

1.  We're moving to all CSS being added via an asset model (because we're moving toward proper asset minification). This means it's not appropriate to pass state data via those URLs since it'll ultimately get lost / ignored.

2. State data in a URL should generally get done as a subdirectory being added before the route (rather than a parameter appended) so that it's preserved in all URLs; we can't guarantee 100% of links are passing thru `Anchor` so we don't want to rely on it for something like this. We're using a "fake" subdirectory for some internal features and that methodology has proven useful for this sort of problem, so we'd like to apply it here too, previous implementations notwithstanding.

3. We have some ideas about parent/child theme relationships we're going to be implementing soon that will probably help out with this.

There were also a few other miscellaneous concerns like some coding standards issues (especially brace positioning) and doing a couple bad things like accessing a superglobal directly. (This is just an FYI, not a direct reason we closed it).

We really do like your overall idea, but there's enough implementation conflicts with near-term plans that this pull request can't be easily reconciled. However, we want to circle back to this idea soon and would love to use some of the components from your work here when we do so (e.g. your embed settings page is great); we're just saying we can't merge this entire body of work in its current form because it's moving in a different direction than we want to go in some key ways. It would reduce overhead on both our sides to just table this for a few months while we get the missing pieces in place.

I've taken a look at your post history on .org and see you've mentioned this in a roundabout way recently and didn't get any feedback from core developers. Sorry about that. We're still doing work to improve our communication with the open source team & signalling our intentions for future planning like that. When you have architectural questions before embarking on projects like this, please feel free to mention me (`linc` on that site as well) to get some feedback. I may take a little to reply but I'll definitely make sure we get back to you.

PR #2581, opened 2015-03-18
YouTube embed fixes: start time, hyphens, data-youtube

linc:
* Fix hyphens being disallowed in YouTube URLs (regression).
* Fix support for start times via `#t=1` anchors (regression).
* Unify pattern matching in JS to match PHP.
* Use a `data-youtube` param instead of `id` to prevent conflicts & attribute rule breaking.
* Give a useful console log error by providing the bad videoid parsed.
* Adds more example cases for YouTube testing.
* Cleanup some comments & increment version to cache bust.

### DIFF
--- a/index.php
+++ b/index.php
@@ -13,7 +13,7 @@
  */
 
 define('APPLICATION', 'Vanilla');
-define('APPLICATION_VERSION', '2.2.17');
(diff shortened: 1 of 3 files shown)


beckyvb:
Good to go!

linc:
Nice catch, thanks!

PR #2583, opened 2015-03-18
Profile Extender: Add parentheses to white-listed characters for profile...

beckyvb:
... extender dropdown fields.
Country lists often use parentheses.


plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -271,7 +271,7 @@ public function SettingsController_ProfileFieldAddEdit_Create($Sender, $Args) {
 
          // Make Options an array
          if ($Options = GetValue('Options', $FormPostValues)) {
-            $Options = explode("\n", preg_replace('/[^\w\s-]/u', '', $Options));
+            $Options = explode("\n", preg_replace('/[^\w\s-()]/u', '', $Options));
```
tburry: Have you tested this? Usually the dash has to go last or it has a different meaning.

PR #2586, opened 2015-03-20
Locale refactor.

tburry:
Refactor the locales to add locale code canonicalization and change the default locale from "en-CA" to "en".

- Canonicalize locale codes so that different representations of the same locale will be compatible with each other (ex. en-us == en_US).
- Change the way locale translations are crawled and cached within the application.
- Switch the default locale from "en-CA" to "en".


linc:
Do not merge before 30 Mar

linc:
This is is ready to go. There are conflicts so I'm not merging.

The refactoring of `CanEdit` doesn't seem like it's appropriate in this pull request. Not sure how that ended up in here.

tburry:
Shoot. That's where that code went.

PR #2587, opened 2015-03-20
Allow the dirty cache to store null values.

tburry:
The user model stores null users when a user does not exist so that it isn’t queried again and again. The dirty cache does not properly handle this leading to some cases of the user getting queried too many times on a page.


tburry:
This is a small change with a lot of whitespace differences. To test you can try the following:

1. Turn on the debugger.
2. Turn off memcache.
3. Delete a user with a lot of comments.
4. Look at the queries without the PR.
5. Look at the queries with the PR.

PR #2637, opened 2015-03-21
Don't show the "Preferences" button to guests

bleistivt:
preferences


linc:
This should include moderators who can modify it as well.

bleistivt:
What permission do you have in mind?

If it's on of `Garden.Users.Edit` or `Moderation.Profiles.Edit`, hasEditProfile() will always return true for those:

https://github.com/vanilla/vanilla/blob/83ad6bbfa653b9747f393323ad76cfd9b9b26ee8/library/core/functions.render.php#L522

linc:
You're correct, I had the wrong assumption in my head (that `hasEditProfile` was also dependent on a config option).

PR #2641, opened 2015-03-23
Adding post-import data checks and fixes

initvector:
Adding checks for InsertUserID on imported discussions and comments; Adding check for valid role on imported users; Adding dba function to update missing InsertUserIDs on discussions and comments with SystemUserID; Adding dba function to apply specified role to users without a valid one


linc:
Tabs have snuck into this.

applications/dashboard/models/class.importmodel.php
```diff
@@ -1784,4 +1787,64 @@ public function UpdateCounts() {
 
       return TRUE;
 	}
+
+	/**
+	 * Verify imported data
+	 */
+	public function VerifyImport() {
+		// When was the latest discussion posted?
+		$LatestDiscussion = $this->SQL->Get('Discussion', 'DateInserted', 'desc', 1);
```
tburry: This is suboptimal on large tables. What you really want is:

```php
$MaxDate = $this->SQL->Select('DateInserted', 'max', 'MaxDate')->From('Discussion')->Get()->Value('MaxDate');
```

Check the query and you'll see:

```SQL
describe select max(DateInserted)
from GDN_Discussion;
```

linc:
There are still tabs in import.js.

PR #2642, opened 2015-03-23
Add master php view deletion reference for README upgrade process

hgtonight:
This file also needs to be deleted for upgrading from 2.0.x to 2.1.x


linc:
Awesome, thanks!

PR #2644, opened 2015-03-25
Very small performance enhancements for Anchor()

R-J:
- Move `if ($Destination == '' && $ForceAnchor === FALSE) return $Text;` to top in order to exit the function as early as possible in program flow.

- Avoid unneeded calls to function Attribute for empty parameters $CssClass and $Attributes

- Skipped testing for isset($Attributes[something]) if $Attributes is no array

- Only assign "static" values to $SSL and $WithDomain if they are not in $Attributes. Otherwise they will be assigned and might be changed two lines later

I have not made any performance tests this time :-/

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -370,33 +370,45 @@ if (!function_exists('Anchor')) {
    /**
     * Builds and returns an anchor tag.
     */
-   function Anchor($Text, $Destination = '', $CssClass = '', $Attributes = array(), $ForceAnchor = FALSE) {
-      if (!is_array($CssClass) && $CssClass != '')
-         $CssClass = array('class' => $CssClass);
-
-      if ($Destination == '' && $ForceAnchor === FALSE)
+   function Anchor($Text, $Destination = '', $CssClass = '', $Attributes = array(), $ForceAnchor = false) {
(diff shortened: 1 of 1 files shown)


linc:
What if we simplified that `$Attributes` logic block even more?


      $HtmlAttributes = '';
      $WithDomain = false;
      $SSL = null;
 
      if (is_array($Attributes)) {
         if (isset($Attributes['SSL'])) {
            $SSL = $Attributes['SSL'];
            unset($Attributes['SSL']);
         }
         if (isset($Attributes['WithDomain'])) {
            $WithDomain = $Attributes['WithDomain'];
            unset($Attributes['WithDomain']);
         }
         $HtmlAttributes = Attribute($Attributes);
      }

R-J:
I'd say it looks _much_ cleaner :-) 

That is how SSL and WithDomain have been defined before. In worst case, those three vars are assigned twice, which might be a very, very tiny bit slower, but I guess in this case it is to be preferred to chose the more simple code over that if/else spaghetti

But three weeks .ago I was of the opposite opinion. I'm happy that this is not my decision ;-)

linc:
Clarity over miniscule performance gains is a good rule.

PR #2645, opened 2015-03-27
Uses a mutex on Categories cache entry to prevent cache stampede.

kaecyra:


linc:
Given this has been smoke tested in the most extreme way possible...

PR #2646, opened 2015-03-27
Format BBCode: Update code tag regex

beckyvb:
Fixes bug where nested code tags would only render the text between the first opening tag and the first closing tag as code.


linc:
Consider this example:

     [code]code 1[/code]
     text
     [code]code 2[/code]
     text
     [code]code 3[/code]

Now renders as 1 blockl

linc:
My take is that the current functionality of the BBCode parser is fine. Not being allowed to include a closing code tag inside a code block makes perfect sense, and the penalty for doing so is simply a code block spanning the wrong text, not a trainwreck.

beckyvb:
Right now '[code]some text[code]more text[/code][/code]' is rendering 'some text[code]more text' inside the code div (with a renegade [/code] outside of the div). In this case, I think it makes more sense to have 'some text[code]more text[/code]' inside the code div.

PR #2648, opened 2015-03-29
Add div.Info to ConfigurationModule

R-J:
Adds an easy way of providing a description in the configuration module by simply adding `$configurationModule->setData('Description', t('Some Description'));`


bleistivt:
@R-J This is cool, but it should only add the div if there is a description.

PR #2649, opened 2015-03-29
Fix wall and other no-upload text areas losing image button

linc:
Revises #2543 to toggle `canUpload` & trade the 'uploads' button for the 'images' button.


linc:
Needs version increment now too.

PR #2652, opened 2015-03-30
Add support for mentions in brackets

R-J:
`(@someone)` wasn't recognized as a mention. Changed regex to allow that,


linc:
Looks good.

bleistivt:
Shouldn't this be `\(` ?

linc:
That's what he added. Can you elaborate?

bleistivt:
This adds a closing bracket which allows for )@mention but not (@mention which I think was the intention.

linc:
Looks like @R-J and I have complimentary cases of dyslexia.

PR #2655, opened 2015-03-31
Make Editor independent of the Vanilla application

shadowdare:
Plugin can remain enabled if Vanilla isn't enabled by not calling the PermissionCategory() method.


shadowdare:
Early pull request mistake. More commits to come

Issue #2572, opened 2015-03-17
Editor only highlights http:// the first time you post a link


linc:
When you hit the Url button to easily post a clickable link the Http:// in the popup is highlighted for easy pasting. When posting a second link, and each after, there is no highlight.

PR #2666, opened 2015-04-07
Highlight "http://" whenever link button is clicked

R-J:
Solves that issue: https://github.com/vanilla/vanilla/issues/2572


linc:
When you update static files (CSS or JS), increment the third point of the version number for the addon or core based on where it is. Currently anyone using Varnish (including cloud) needs this for the files to flush on update.

PR #2671, opened 2015-04-09
Don’t show misleading permissions when editing a role without any permissions.

tburry:
If there are no entries for a role in `GDN_Permission` then editing that role shows default permissions. This is misleading because that role really has no permissions. This fix makes the role show no permission so that admins know to fix the role.

This is an issue that usually won’t occur unless the data is incorrect from something like an import.


tburry:
The best way to test this fix would be to do the following:

1. Delete all everything in `GDN_Permission` for a given role ID.
2. Edit that role and make sure there are no permissions.
3. Make sure saving the role still works.
4. Do some basic regression testing on role editing.

initvector:
Everything seems to checkout

* Editing a role with no permission rows gives you an empty permission grid (no defaults)
* Editing a role *with* permission rows still properly populates the permission grid
* Adding a new role gives you default permissions pre-populated
* Deleting a role is unaffected
* Editing attributes of a role not related to permissions (name, description) is still functional

High-five on making that foreach loop a little less confusing.

PR #2676, opened 2015-04-14
Basic composer integration

tburry:
Add basic composer integration to Vanilla. This integration just adds composer autoloader support, but will be the basis for much deeper integration in the future.


tburry:
Let's hold off until 2.2 starts in earnest for this one.

tburry:
I'll redo this PR after the spacening.

bleistivt:
Are you planning to remove the core autoloader and (in the future) libraries from the repo, or are you going to update those locally and commit them to the repository?

tburry:
We can't use composer for addon classes because they shouldn't always be available.

I am planning on an autoloader rewrite in the future. We are going to remove the ability to have a context aware autoloader which will simplify and speed it up somewhat.

linc:
This is ready to go, but I'm unclear how we're dealing with this in our deploy process.

tburry:
Staged this to cl211.

tburry:
Rebased to current master.

PR #2677, opened 2015-04-15
Hotfix/display as category

beckyvb:
Fixes bug where site description displays on page for subcategory with 'Display as categories' option enabled.
Fixes bug where CategoryFilterToggle does not work with mixed view.
Fixes bug where subcategories with 'Display as categories' option enabled have no child categories appear due to max display depth not being updated.

### DIFF
--- a/applications/vanilla/controllers/class.categoriescontroller.php
+++ b/applications/vanilla/controllers/class.categoriescontroller.php
@@ -190,14 +190,14 @@ class CategoriesController extends VanillaController {
                switch($Layout) {
                   case 'mixed':
                      $this->View = 'discussions';
-                     $this->Discussions($CategoryIdentifier, FALSE);
+                     $this->Discussions($CategoryIdentifier);
                      break;
                   case 'table':
-                     $this->Table($CategoryIdentifier, FALSE);
+                     $this->Table($CategoryIdentifier);
                      break;
                   default:
                      $this->View = 'all';
-                     $this->All($CategoryIdentifier, FALSE);
+                     $this->All($CategoryIdentifier);
                      break;
                }
                return;
@@ -329,6 +329,7 @@ class CategoriesController extends VanillaController {
    /**
     * Show all (nested) categories.
     *
+    * @param string $Category The url code of the parent category.
     * @since 2.0.17
     * @access public
     */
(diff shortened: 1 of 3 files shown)


applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -109,12 +109,12 @@ public function Archives($Category, $Month, $Page = FALSE) {
    /**
     * "Table" layout for categories. Mimics more traditional forum category layout.
     */
-   public function Table($Category = '') {
+   public function Table($Category = '', $isRoot = TRUE) {
```
tburry: We can't add parameters to control methods like this. Remember that controller methods concert to url and an ?isRoot=true is not appropriate. Don't you know we are in the root if `!$Category`?

applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -190,14 +190,14 @@ public function Index($CategoryIdentifier = '', $Page = '0') {
                switch($Layout) {
                   case 'mixed':
                      $this->View = 'discussions';
-                     $this->Discussions();
+                     $this->Discussions($CategoryIdentifier, FALSE);
```
tburry: I think the second argument is invalid. Same below.

applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -384,29 +386,45 @@ public function All($Category = '') {
     * @since 2.0.0
     * @access public
     */
-   public function Discussions() {
+   public function Discussions($Category = '') {
```
tburry: Document this parameter.

applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -384,29 +386,45 @@ public function All($Category = '') {
...
       $this->Menu->HighlightRoute('/discussions');
       $this->AddJsFile('discussions.js');
-      $Title = C('Garden.HomepageTitle');
-      if ($Title)
-         $this->Title($Title, '');
-      else
-         $this->Title(T('All Categories'));
-      $this->Description(C('Garden.Description', NULL));
+
+      $this->Menu->HighlightRoute('/discussions');
```
tburry: Is this a duplicate of line 392?

applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -384,29 +386,45 @@ public function All($Category = '') {
...
+      if (!$this->Title()) {
+         $Title = C('Garden.HomepageTitle');
+         if ($Title)
+            $this->Title($Title, '');
+         else
+            $this->Title(T('All Categories'));
+      }
+
+      if (!$Category) {
+         $this->Description(C('Garden.Description', NULL));
```
tburry: If there is a category should the description be set?

PR #2680, opened 2015-04-16
Speed up the user page.

tburry:
Remove some costly features on the users page when there are too many users in the database.
To test this PR you can edit `UserModel->UserThreshold` to see the changes for the two cases of user counts.

### DIFF
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -89,8 +89,6 @@ class UserController extends DashboardController {
          $Filter = $Keywords;
       }
 
-      $this->SetData('RecordCount', $UserModel->SearchCount($Filter));
-
       // Sorting
       if (in_array($Order, array('DateInserted','DateFirstVisit', 'DateLastActive'))) {
          $Order = 'u.'.$Order;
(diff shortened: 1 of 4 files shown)


linc:
This looks good, but I'm wondering if we can raise that threshold significantly. It's millions of users that's hurting us, not a couple thousand.

linc:
The more I am playing with this the less I like the experience.

* A blank user table by default is bad. It should hide the headings or say something below it for the empty state.
* Not counting the user results of a search is actually a pretty severe restriction. Basic stuff like "how many of my users signed up with a Hotmail account" becomes impossible to find. This is a poor-man's analytics, but it's important.
* The regression of not being able to search an email without an '@' included is going to confuse folks without a note. It's already bad enough we don't note how wildcards work in the description.
* Only having 'Next' is making me realize that the pager is overlapped by my scrollbar in Firefox. Usually I clicked the number '2' so I didn't notice the lack of padding there before.

If I lost all this functionality on my site with 15K users, I'd be pissed. If we gotta do this, it's an inferior experience for sure, so we should crank that threshold as high as we can handle until we can make a better experience here.

tburry:
Raising the threshold is fine I guess. Improving the experience is really something that needs a search engine, but that search engine a significant undertaking.

PR #2681, opened 2015-04-17
Look up CSS from a single method and re-use.

kaecyra:
This refactor allows the Controller's RenderMaster to use AssetModel's lookup methods to find CSS files, thereby allowing us to centralize the rules for these and cause all CSS files to be included the same way.


kaecyra:
This was done because some of my hooks in the assetmodel_styleCss_handler were not working as expected, because AssetModel::CssPath() uses different logic than Controller::RenderMaster().

tburry:
Closed in favor of #2720.

PR #2683, opened 2015-04-20
Use PHP 5.2 function filter_var() for validations

R-J:
Vanilla requires PHP 5.2 and since this PHP version, there is a native validation function: http://php.net/manual/en/function.filter-var.php

Changes I've made:

ValidateRegex: about 20% faster
ValidateEmail: speed increase not measured, but current implementation calls external class which checks if filter_var() is valid function and uses it afterwards. So basically the previously done "external" check is now done inside of the function. The ValidateRequired check has not been passed the $Field parameter.
ValidateInteger: more than 35% faster for number in quotes, slightly slower (4%) for float, slightly faster for other values

Changes I have not made:
ValidateNoLinks: `return (filter_var($Value, FILTER_VALIDATE_URL) === false);` would have done the job but it is slower than the current implementation and it also works different on other protocols. `mail://bill@microsoft.com` is not a link for the current implementation, but it is a link for filter_var().
ValidateBoolean: FILTER_VALIDATE_BOOLEAN checks for "1", "true", "on", "yes" and "0", "false", "off", "no". This is far more than ValidateBoolean does.

### DIFF
--- a/library/core/functions.validation.php
+++ b/library/core/functions.validation.php
@@ -33,8 +33,7 @@ if (!function_exists('ValidateCaptcha')) {
 
 if (!function_exists('ValidateRegex')) {
    function ValidateRegex($Value, $Regex) {
-      preg_match($Regex, $Value, $Matches);
-      return is_array($Matches) && count($Matches) > 0 ? TRUE : FALSE;
+      return (!filter_var($Value, FILTER_VALIDATE_REGEXP, array('options' => array('regexp' => $Regex))) === false);
    }
 }
 
@@ -135,12 +134,11 @@ if (!function_exists('ValidateOldPassword')) {
 
 if (!function_exists('ValidateEmail')) {
    function ValidateEmail($Value, $Field = '') {
-      if (!ValidateRequired($Value))
-         return TRUE;
+      if (!ValidateRequired($Value, $Field)) {
+         return true;
+      }
 
-      $Result = PHPMailer::ValidateAddress($Value);
-      $Result = (bool)$Result;
-      return $Result;
+      return (!filter_var($Value, FILTER_VALIDATE_EMAIL) === false);
    }
 }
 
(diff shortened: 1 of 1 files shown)


linc:
Can you confirm that the email filter_var is better or at least equivalent to the PHPMailer check? Because the notes on the filter_var documentation make me wary of it. http://php.net/manual/en/function.filter-var.php

R-J:
PHPMailer is using filter_var() for validation: https://github.com/vanilla/vanilla/blob/master/library/vendors/phpmailer/class.phpmailer.php#L524-L534

linc:
Ha, lame. Good enough.

initvector:
For the most part, it looks good, @R-J!

Couple of things...

1. filter_var returns the value if the validation passes or false if it doesn't.  Why are you inverting the value returned and performing a strict equivalency check for false?  Why not just check if filter_var's return value !== false?
2. If you change it up to checking for a strict not-false value, you can probably ditch the check for 0 in ValidateInteger.

R-J:
@initvector: I've made the changes you've proposed and i do not know why I have made the comparison like !filter_var === false

There is one different result for the old and the new ValidateInteger, though (but this has already been different with my first try): if `$Value = '  9   ';` or `$Value = "9\r\n";` the value hasn't been validated as an integer by now, but filter_var does so.

initvector:
Instead of filter_var for validating integers, what do you think about using ctype_digit?  In my local tests, it ran about 60% faster than filter_var for checking integers and will not validate "integer" values with additional whitespace.

R-J:
ctype_digi() doesn''t recognize "+9" and "-9" as integers, as far as I know.

initvector:
Do you mind removing the updates to ValidateInteger?

There's some concern about keeping ValidateInteger fully compatible with its current implementation.  I think this PR going to be shipshape once you put those optimizations on hold.

R-J:
Done. I've only changed some brackets and upper/lower case of true/false/null

library/core/functions.validation.php
```diff
@@ -244,14 +242,18 @@ function ValidateMinimumAge($Value, $Field, $FormPostedValues) {
...
+        if (!$Value || (is_string($Value) && !trim($Value))) {
+            return true;
+        }
+        $Integer = intval($Value);
+        $String = strval($Integer);
+        if ($String == $Value) {
+            return true;
+        } else {
+            return false;
+        }
```
initvector: You can probably remove this whole `if ($String == $Value)...` conditional.  It seems kind of redundant, based on the return values.  I think you can get away with just: `return $String == $Value;`

R-J:
You're certainly right again!

initvector:
Thanks for your patience, @R-J.  Great work!

PR #2685, opened 2015-04-20
Remove /vanilla from non-ambiguous controller URLs

linc:
Use of `/vanilla` in these cases is inconsistent & redundant.


tburry:
Love.

PR #2687, opened 2015-04-20
Invites: allow resend option

linc:
Allow duplicate invites when declared by deleting the old one.

### DIFF
--- a/applications/dashboard/models/class.invitationmodel.php
+++ b/applications/dashboard/models/class.invitationmodel.php
@@ -92,14 +92,14 @@ class InvitationModel extends Gdn_Model {
 
          // Make sure that the email does not already belong to an invitation in the application.
          $TestData = $this->GetWhere(array('Email' => $Email));
+         $DeleteID = false;
          if ($TestData->NumRows() > 0) {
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.invitationmodel.php
```diff
@@ -79,21 +80,27 @@ public function Save($FormPostValues, $UserModel, $Options = array()) {
...
          if ($TestData->NumRows() > 0) {
-            $this->Validation->AddValidationResult('Email', 'An invitation has already been sent to the email you entered.');
-            return FALSE;
+            if (!$Resend) {
+               $this->Validation->AddValidationResult('Email', 'An invitation has already been sent to the email you entered.');
+               return false;
+            } else {
+               // Remove the old invitation.
+               $DeleteID = val('InvitationID', $TestData->FirstRow(DATASET_TYPE_ARRAY));
+               $this->Delete($DeleteID);
```
tburry: If there is another error then the invitation will not be saved, but the old invitation will still be deleted. You need to only delete if there are no errors.

linc:
Is this correct?

PR #2688, opened 2015-04-21
Editor: Revert #2089 and allow Format opt-out separately

linc:
* Allow BeforeBodyBox to once again change Format, `$Column`, and `$Attributes`.
* Allow for intent of #2089 by adding Format opt-out EventArg.

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -149,29 +149,29 @@ class Gdn_Form extends Gdn_Pluggable {
     */
    public function BodyBox($Column = 'Body', $Attributes = array()) {
       TouchValue('MultiLine', $Attributes, TRUE);
-      TouchValue('format', $Attributes, $this->GetValue('Format', Gdn_Format::DefaultFormat()));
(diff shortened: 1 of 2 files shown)


linc:
@JasonBarnabe fyi

library/core/class.form.php
```diff
@@ -157,15 +157,27 @@ public function BodyBox($Column = 'Body', $Attributes = array()) {
       $Attributes['format'] = htmlspecialchars($Attributes['format']);
       $this->SetValue('Format', $Attributes['format']);
```
tburry: Can you instead move this after the event and check for an empty string or something? The `IncludeFormat` just seems too kludgey.

linc: Sure, that'd be better.

tburry:
@linc, I made some changes. This PR is going to need some team testing though so I'm hoping that @beckyvb will have time this week too.

beckyvb:
I've tested this with different editors and configurations. Good to go!

PR #2690, opened 2015-04-21
Quotes: disable Emoji parse for FormatQuote

linc:
Fixes bug where Emoji get double-parsed when quoted post is a different input formatter.

To test, make a post with BBCode format containing Emoji. Switch to Html and quote the post. Without this patch, you'll get broken Emoji HTML in the post.

### DIFF
--- a/plugins/Quotes/class.quotes.plugin.php
+++ b/plugins/Quotes/class.quotes.plugin.php
@@ -302,11 +302,8 @@ BLOCKQUOTE;
 
    protected function FormatQuote($Type, $ID, &$QuoteData, $Format = FALSE) {
       // Temporarily disable Emoji parsing (prevent double-parsing to HTML)
-      $undoEmoji = false;
-      if (Emoji::instance()->enabled) {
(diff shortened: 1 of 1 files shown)


plugins/Quotes/class.quotes.plugin.php
```diff
@@ -302,6 +301,13 @@ public function PostController_BeforeCommentRender_Handler($Sender) {
    }
 
    protected function FormatQuote($Type, $ID, &$QuoteData, $Format = FALSE) {
+      // Temporarily disable Emoji parsing (prevent double-parsing to HTML)
+      $undoEmoji = false;
```
tburry: Usually you just do.

```php
$emojiEnabled = Emoji::instance()->enabled;
Emoji::instance()->enabled = false;

// ...

Emoji::instance()->enabled = $emojiEnabled;
```

linc: Y'know I even thought about that afterward and decided not to and now I have no idea why.

PR #2694, opened 2015-04-22
Don't do mentions, hashtag searches, emoji, or links inside <code>

### LINKED ISSUE #1495: Link parser shouldn't run between <code> tags
<code></code>

Produces:

     

Where 'http://site.com/example.gif&quot;' is a link.

### LINKED ISSUE #2467: Smileys are rendered inside code tags

JasonBarnabe:
This PR defines a new function - `Gdn_Format::ReplaceButProtectCodeBlocks`. This function works like `preg_replace` and `preg_replace_callback`, but avoids doing any replacements inside `<code>` tags. The function accepts four parameters: the first three are the same as `preg_replace` and `preg_replace_callback`, the fourth controls whether it acts like `preg_replace` (default/`FALSE`) or `preg_replace_callback` (`TRUE`).

Internally, `Gdn_Format::ReplaceButProtectCodeBlocks` replaces `<code>` tags and their contents with placeholders, does the requested replace, then puts the `<code>` tags back.

This PR changes the `@mention`, `#hashtagsearch`, and `/me does something` code to use this new function. I believe `/me` only runs if it's at the start of the comment, but I've included it for consistency and in case that behaviour changes in the future. This supercedes #2531. It also changes the emoji (#2467) and linkifying (#1495) code to use this new function.

Test code:

```
<code> @JasonBarnabe :) :smile: #search <img src="http://site.com/example.gif" /> </code>
```
Should appear as plain text with this change, as links/emoji without it.

Any regexp working on HTML content will have issues, but I don't think this has any more issues than what's there already. For example, the code before this change couldn't handle nested `<code>`s, and this one can't either. What's really needed is an HTML parser to handle these kinds of replacements.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1048,10 +1048,10 @@ class Gdn_Format {
 
          self::LinksCallback(NULL);
 
-         $Mixed = preg_replace_callback(
+         $Mixed = Gdn_Format::ReplaceButProtectCodeBlocks(
             $Regex,
          array('Gdn_Format', 'LinksCallback'),
-         $Mixed);
+         $Mixed, TRUE);
 
          Gdn::PluginManager()->FireAs('Format')->FireEvent('Links', array(
             'Mixed' => &$Mixed
@@ -1401,7 +1401,7 @@ EOT;
 
             // Unicode includes Numbers, Letters, Marks, & Connector punctuation.
             $Pattern = (unicodeRegexSupport()) ? '[\pN\pL\pM\pPc]' : '\w';
-            $Mixed = preg_replace(
+            $Mixed = Gdn_Format::ReplaceButProtectCodeBlocks(
                '/(^|[\s,\.>\)])@('.$Pattern.'{1,64})\b/i', //{3,20}
                '\1'.Anchor('@$2', $urlFormat),
                $Mixed
@@ -1410,7 +1410,7 @@ EOT;
 
          // Handle #hashtag searches
 			if(C('Garden.Format.Hashtags')) {
-				$Mixed = preg_replace(
+				$Mixed = Gdn_Format::ReplaceButProtectCodeBlocks(
 					'/(^|[\s,\.>])\#([\w\-]+)(?=[\s,\.!?<]|$)/i',
 					'\1'.Anchor('#\2', '/search?Search=%23\2&Mode=like').'\3',
 					$Mixed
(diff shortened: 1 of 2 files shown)


valeryan:
Hey Jason, have you looked at how video links for youtube or vimeo are handled? the are also getting formatted inside of code blocks.

JasonBarnabe:
They are also handled by this PR.

library/core/class.format.php
```diff
@@ -1430,6 +1430,52 @@ public static function Mentions($Mixed) {
...
+     // appear in the string, and keep track of what substring got
+     // replaced with what code.
+     $Replacements = array();
+     $Subject = preg_replace_callback(
+       '/<code>.*?<\/code>/i',
+       function($Matches) use (&$Replacements) {
+         // Random string and replacement index, surrounded by
+         // whitespace to try to prevent the characters from being
+         // picked up by $Pattern.
+         $ReplacementString = ' 5z9Ah9Y2sX5xnzUx8wSq'.count($Replacements).' ';
```
initvector: Instead of random-string-plus-running-total-of-replacements, why not use [sha1](http://php.net/sha1)?  You can use the hash as $ReplacementString.  You also wouldn't need to reverse your placeholder array at the end of your function.

library/core/class.format.php
```diff
@@ -1430,6 +1430,52 @@ public static function Mentions($Mixed) {
...
+     if ($IsCallback) {
+       $Subject = preg_replace_callback($Pattern, $Replacement, $Subject);
+     } else {
+       $Subject = preg_replace($Pattern, $Replacement, $Subject);
+     }
+
+     // Put back the code blocks. First in, last out so that e.g.
+     // random10 is not replaced by random1.
+     foreach (array_reverse($Replacements) as $Placeholder => $Original) {
+       $Subject = str_replace($Placeholder, $Original, $Subject);
```
initvector: [str_replace](http://php.net/str_replace) supports arrays in $search and $replace.  If you kept your placeholders in one array and your code block content in another, so long as they're synchronized, you should be able to ditch this foreach loop.

initvector:
This is looking pretty good @JasonBarnabe.  In addition to the comments I left in your function, please punch up the ReplaceButProtectCodeBlocks docblock). I'd like to see first line of comment text moved down beneath the opening comment line and some some description text for your parameters added in.

JasonBarnabe:
@initvector Updated per your comments.

initvector:
Nice work! Thanks @JasonBarnabe!

PR #2697, opened 2015-04-28
Checking if source file is readable before import

### LINKED ISSUE #2674: "ImportFile is required" after uploading it
I've exported a MyBB database with Vanilla Porter 2.1.9 and I'm trying to import it into a fresh checkout of the `master` branch ("Version 2.2.100"). The .tar.gz is about 1.8Mb and after I upload it i get is this:

image

[This thread](http://vanillaforums.org/discussion/27054/importing-from-phpbb-importfile-is-required-error-now-with-added-solution) suggested to check zlib; it's enabled:

image

After copying `export_qsforum_mybb_2015-04-10_045203.txt.gz` to the uploads folder and selecting it on the Import page, I get

> The import file is not in the correct format.

If gunzipped, the file starts with

```
Vanilla Export: 2.1.9, Source: MyBB

// Export Started: 2015-04-10 04:52:03
Table: User
UserID:int,Name:varchar(50),Email:varchar(200),Photo:varchar(255),DateInserted:datetime,DateFirstVisit:datetime,DateLastAct
ive:datetime,Password:varbinary(100),HashMethod:varchar(10)
```

and ends some 59,000 lines later with

```
// Exported Table: UserDiscussion (4295 rows, 00:00.19)

// export_qsforum_mybb_2015-04-10_045203.txt.gz
// Export Completed: 2015-04-10 04:52:05
// Elapsed Time: 00:02.39
```

### Solution

I had to set the r+go attribute. There could be a less misleading error message than the file not being in the correct format :-P

initvector:
When importing a file, ImportModel wasn't checking to see if the file was readable before it attempted to open it.  If the read permission wasn't granted to the user or group PHP was running under, it would throw an error informing the user that the file was in the wrong format, because attempting to read the file would fail when validating the file's header.  This fix implements a check before the import and alerts the user they need to verify permissions on the import file.

Fixes #2674

Also removing tabs from class.importmodel.php.

### DIFF
--- a/applications/dashboard/models/class.importmodel.php
+++ b/applications/dashboard/models/class.importmodel.php
@@ -12,75 +12,75 @@ Contact Vanilla Forums Inc. at support [at] vanillaforums [dot] com
  * Object for importing files created with VanillaPorter.
  */
 class ImportModel extends Gdn_Model {
-	const COMMENT = '//';
-	const DELIM = ',';
-	const ESCAPE = '\\';
-	const ID_PADDING = 1000; // padding to add to IDs that are incremented
-	const NEWLINE = "\n";
-	const NULL = '\N';
-	const TABLE_PREFIX = 'z';
-	const QUOTE = '"';
+   const COMMENT = '//';
+   const DELIM = ',';
+   const ESCAPE = '\\';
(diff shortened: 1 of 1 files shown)


initvector:
Translation updated!

PR #2700, opened 2015-04-29
ProfileExtender: Remove ability to add more than one DoB

### LINKED ISSUE #2675: Can't Save Profile without Setting Birthday
Since issue #2575 was merged, profile changes still can't be saved without setting the birthday field added via the Profile Extender plugin.

- Setting the birthday dropdown to "Month / Day / Year" will bring up the error "DateOfBirth is not a valid date"
- Also, on the dashboard page to add custom profile fields, you can add two birthday fields. If you add a second one, it will be have a label called "DateOfBirth" and a type of "DateOfBirth"

initvector:
The Profile Extender plug-in was allowing more than one DoB field to be set.  The problem being that this was a special type of field that mapped to the DoB field in the user table.  Multiple DoB fields didn't make sense and ultimately caused a fatal error when editing a profile.

This fix removes the ability to add a DoB field if one already exists.

Coupled with a previously applied fix, Fixes #2675


tburry:
If this is the case then shouldn't we use similar logic for fields like Title, Location, or any other fields in the user table?

initvector:
DoB was a special case in Profile Extender and actually classified as a "type".  If multiple instances of that type were added to a profile, it was throwing a fatal error.  This fix, in particular, just focuses on limiting usage of that DoB type by removing it as an option from the type dropdown if a DoB already exists.

Adding other fields that map directly to the User table require you to label them with the field name.  If you specify the same field more than once, it'll self-correct by appending an increment number (e.g. About, About1) to the field name.  This behavior avoids conflicting values in profiles.  You can have two "About" fields if you really, really want to, but only one of them is going to actually save to the User table.  The other one is going to save as About1 to UserMeta.

PR #2706, opened 2015-04-29
ButtonBar: Adding support for Wysiwyg input format

### LINKED ISSUE #2517: ButtonBar cannot handle the Wysiwyg input formatter
ButtonBar cannot handle the Wysiwyg input formatter. If you are using WYSIWYG Editor and switch to Button it just fails to load.

initvector:
If you happened to have InputFormatter configured to Wysiwyg, probably after using Advanced Editor, the Button Bar plug-in would not display.  This fix adds support for the Wysiwyg format by having ButtonBar treat it as Html.

Fixes #2517 

Also some IDE-assisted whitespace fixes to class.buttonbar.plugin.php and buttonbar.js


linc:
Increment the version so JS cache breaks and we're good to go here.

initvector:
Incremented ButtonBar version to 1.7, from 1.6

PR #2710, opened 2015-04-30
Dashboard Hooks: Wrap full site link in Url function.

beckyvb:


applications/dashboard/settings/class.hooks.php
```diff
@@ -263,7 +263,7 @@ public function SiteNavModule_all_handler($sender) {
       if (Gdn::Session()->IsValid()) {
          // Switch between the full site and mobile.
          if (IsMobile()) {
-            $sender->addLink('etc.nomobile', array('text' => t('Full Site'), 'url' => '/profile/nomobile', 'icon' => icon('resize-full'), 'sort' => 100));
+            $sender->addLink('etc.nomobile', array('text' => t('Full Site'), 'url' => Url('/profile/nomobile'), 'icon' => icon('resize-full'), 'sort' => 100));
```
tburry: This is actually not correct. The module wraps stuff appropriately. My guess is lithe-specific js is striking again if this is a bug.

tburry: I don't even see the bug on the site that there is supposed to be a bug. You should write about the bug you are fixing in the description of the PR.

PR #2711, opened 2015-04-30
Updating Editor upload handling to use Gdn_UploadImage::SaveImageAs

### LINKED ISSUE #2552: editor doesn't respect image exif orientation
Embedded images appear without respect to proper orientation.

Help me Obi @webstey you're my only hope.

initvector:
Image uploads made using Advanced Editor were not adjusting for orientation exif data.  Editor uploads now use the core function Gdn_UploadImage::SaveImageAs, which handles orientation adjustments.

Fixes #2552

### DIFF
--- a/plugins/editor/class.editor.plugin.php
+++ b/plugins/editor/class.editor.plugin.php
@@ -792,7 +792,7 @@ class EditorPlugin extends Gdn_Plugin {
 
          $fileExtension = strtolower($Upload->GetUploadedFileExtension());
          $fileName = $Upload->GetUploadedFileName();
-         list($tmpwidth, $tmpheight) = getimagesize($tmpFilePath);
+         list($tmpwidth, $tmpheight, $imageType) = getimagesize($tmpFilePath);
 
          // This will return the absolute destination path, including generated
          // filename based on md5_file, and the full path. It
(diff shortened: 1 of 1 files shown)


linc:
PHP Fatal error:  Allowed memory size of 67108864 bytes exhausted (tried to allocate 13056 bytes) in /.../library/core/class.uploadimage.php on line 229, referer: /discussion/99698/boop-ignore-our-bug-testing

linc:
The above error is triggered when uploading the ~1.7MB images I shared for testing via drag-and-drop with editor enabled when checked out on this branch. It does not trigger with my other standard image samples that are all under 300KB.

linc:
It's a significant possibility that there is something specific to my configuration that is doing this. However I'd feel a lot better if you have a closer look at how this could be possibly getting stuck in a loop or something that it's eating 67MB to upload a 1.7MB image.

linc:
FYI, this presented as a 500 error and an InformMessage "The file failed to upload" in the UI / console.

initvector:
I'm not seeing anything that would cause unnecessary memory allocation.  Based on your error, it's triggered by imagecreatetruecolor, but I don't see any out-of-the-ordinary looping anywhere near there.

Your error message says "67108864 bytes" is the allowed memory size.  That's only ~68 megs.  That seems unusually low for large image manipulation (the test images were 3264×2448).

I can't recreate the cited issue, but if @tburry could also confirm, that would be swell!

PR #2713, opened 2015-05-01
adding soundcloud

adrianspeyer:
All member needs to do is post share url and soundcloud track will be embeded.


linc:
See https://github.com/vanilla/vanilla/pull/2725 ; otherwise looks good

adrianspeyer:
Made the change as requested +1

adrianspeyer:
Added it back. I was too eager in making it less.

linc:
Just noticed this is against stage. :P We're not really using that branch unless it's a big change we wanna do more testing on.

PR #2715, opened 2015-05-01
Move from reliance on p parameter to using PATH_INFO

initvector:
Reliance on the p parameter to determine the requested path causes conflicts with vBulletin routes.  This fix allows usage of the PATH_INFO server variable, if available, to determine the requested path information.

Also some caching goodness added to .htaccess for the following extensions:

* css
* js
* woff
* ttf
* eot
* svg
* png
* gif
* jpeg
* jpg
* ico
* swf


tburry:
I think your branching has caused too many files to get pulled in to this PR. Do you need a re-base?

initvector:
Rebase here: #2718

PR #2716, opened 2015-05-01
Differentiate the ways users can be banned.

tburry:
Currently, there is no way to know the difference between a  user being banned manually and then by IP address. This can cause issue where users get blanked banned/unbanned when they aren't supposed to. This feature adds the following.

* Adds a bit mask to the banned flag to keep track of the various ways users get banned.
* Allows CMs to manually ban/unban users that have been caught by an IP ban, but keep that IP ban in effect.
* Adds some hooks for future bans by warnings and temporary banning.

### DIFF
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -333,16 +333,18 @@ class UserController extends DashboardController {
       if (!$User)
          throw NotFoundException($User);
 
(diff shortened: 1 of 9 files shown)


linc:
Only note from code review is that `explodeBans` is a bit of a php-ism which I've known you to actively avoid so I thought I'd point it out. Doing some testing now.

linc:
Slick.

PR #2717, opened 2015-05-01
Improve the view lookup logic against some edge cases.

tburry:
Also remove an old unused method from `Gdn_Controller`.


kaecyra:
This works for me.

PR #2718, opened 2015-05-01
Move from reliance on p parameter to using PATH_INFO

initvector:
Reliance on the p parameter to determine the requested path causes conflicts with vBulletin routes.  This fix allows usage of the PATH_INFO server variable, if available, to determine the requested path information.

Also some caching goodness added to .htaccess for the following extensions:

* css
* js
* woff
* ttf
* eot
* svg
* png
* gif
* jpeg
* jpg
* ico
* swf


tburry:
@linc, Can you see if anyone in the community is willing to test this? OSS is where this will see issues with the various environments.

library/core/class.request.php
```diff
@@ -435,7 +435,9 @@ protected function _LoadEnvironment() {
             SafeParseStr($Get, $Get, $Original);
          }
 
-         if (isset($Get['_p'])) {
+         if (isset($_SERVER['X_REWRITE']) && isset($_SERVER['PATH_INFO'])) {
```
tburry: Is `isset($_SERVER['PATH_INFO'])` always set? I'm thinking of the case where you are on the server root. Maybe we can just check for `isset($_SERVER['X_REWRITE'])`.

initvector:
It's part of the CGI spec, so I'm *pretty sure* it'll always be set.  I'll be more confident in that, once we're able to confirm functionality in less-common setups.

tburry:
One thing this PR will break is embed.js because it always sends index.php?p=. You will probably have to do the following:

1. Modify embed.js to use the proper path.
2. Make this work for sites that have X_REWRITE, but call the site with index.php?p=.

Let me know what you think here.

.htaccess
```diff
@@ -1,12 +1,29 @@
...
+
+    # 301 redirect urls that start with index.php
+    #RewriteCond %{REQUEST_METHOD} GET [NC]
+    #RewriteCond %{REQUEST_URI} ^(.*?)/index\.php(.*)$
+    #RewriteRule ^index\.php /%1%2 [QSA,R,L]
+</IfModule>
+
+<IfModule mod_headers.c>
+   <FilesMatch "\.(css|js|woff|ttf|eot|svg|png|gif|jpeg|jpg|ico|swf)$">
+      Header set Cache-Control "max-age=315360000"
```
tburry: Don't cache `embed.js` or `vanilla.embed.js`.

library/core/class.request.php
```diff
@@ -435,7 +435,9 @@ protected function _LoadEnvironment() {
             SafeParseStr($Get, $Get, $Original);
          }
 
-         if (isset($Get['_p'])) {
+         if (isset($_SERVER['X_REWRITE'])) {
+            $Path = $_SERVER['PATH_INFO'];
+         } elseif (isset($Get['_p'])) {
```
tburry: Remove support for `_p`.

tburry:
I think this PR is ready to go at this point, but we need some help with apache testing. Also @linc can maybe give it a smoke test locally and let us know that we are past OSS 2.2 before this is merged.

.htaccess
```diff
@@ -1,12 +1,32 @@
...
+    RewriteEngine On
+    # Certain hosts may require the following line.
+    # If vanilla is in a subfolder then you need to specify it after the /.
+    # (ex. You put Vanilla in /forum so change the next line to: RewriteBase /forum)
+    # RewriteBase /
+
+    RewriteCond %{QUERY_STRING} ^p=/?([^&]+)(&(.+))?$
+    RewriteRule ^index\.php index.php/%1?%3 [E=X_REWRITE:1,L]
+
+    # The basic rewrtie rule.
```
linc: rewrtie

PR #2719, opened 2015-05-01
Add support for internal controller methods.

tburry:
An internal method is a method that is public, but should not be dispatched to. It’s a method to be used by plugins and the framework, but not accessed from a url.


linc:
Would it make sense to use `isInternal()` to block all double-underscore methods rather than enumerating `__construct()`?

tburry:
Yeah good idea. I'll actually block anything that starts with an underscore.

PR #2723, opened 2015-05-04
Mark bans by system appropriately

### LINKED ISSUE #2709: Banning's InsertUserID is the user being banned
A user matching a Banning rule (e.g. IP address) gets an Activity record of type `Ban` that has an `InsertUserID` of their own UserID. We use this in other functionality as the origin of the ban which is nonsensical. Suggest we use the SystemUserID instead.

linc:
Closes #2709.


beckyvb:
Rebased here: https://github.com/vanilla/vanilla/pull/2736

PR #2726, opened 2015-05-05
global.js context bug: change `this` to `$elem`

xorith:
Change proposed here: http://vanillaforums.org/discussion/29836/possible-bug-in-core-global-js#latest
`this` is not in the proper context when the `complete` method is called. `$elem` should be used to reference the element that initiated the $.ajax request.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -676,7 +676,7 @@ jQuery(document).ready(function($) {
          data: { DeliveryType: 'VIEW', DeliveryMethod: 'JSON', TransientKey: gdn.definition('TransientKey') },
          dataType: 'json',
          complete: function() {
-            gdn.enable($elem);
+            gdn.enable($elem.get(0));
             $elem.removeClass(progressClass);
             $elem.attr('href', href);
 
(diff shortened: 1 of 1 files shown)


linc:
I verified this in the original discussion but would like another dev to review.

initvector:
[gdn.enable](https://github.com/xorith/vanilla/blob/patch-1/js/global.js#L167-L174) appears to be for making certain changes to an HTML element.  As such, it expects its parameter to be an HTML element.

What is going on in hijackClick to cause this error?

1. When hijackClick is called (elements with a class of "Hijack" clicked), it saves a jQuery object referencing `this` (the HTML element) in $elem.
2. gdn.disable is then called using `this`, which is an HTML element in this context.
3. The AJAX call is made and an anonymous function assigned to its "complete" event.
4. In the context of the "complete" event, `this` appears to be the AJAX call object (makes sense, right?). `$elem` is a jQuery object referencing the original HTML element.
5. gdn.enable needs to receive some kind of jQuery-compatible reference to an HTML element and it isn't getting that from an AJAX object (`this`).

Other uses of gdn.enable are passed an element.  I'd like to see this fix follow that practice, instead of passing it the jQuery object (which may still work). Changing `$elem` to `$elem.get(0)` in gdn.enable should do it.

tl;dr Seems to be a legitimate issue, but needs just a tiny tweak to maintain consistency with other gdn.enable calls.

xorith:
I updated per @initvector to use `$elem.get(0)` instead of `$elem`

initvector:
Thanks @xorith!

PR #2730, opened 2015-05-07
Don’t fire an event when an already disabled plugin is disabled again.

tburry:


initvector:
Seems like a sane change and doesn't appear to break anything.  Approved!

PR #2731, opened 2015-05-07
Replace ConsolidateArrayValuesByKey() with array_column().

tburry:
The new `array_column()` provides 99% of the functionality of `ConsolidateArrayValuesByKey()` and has the benifit of having a nicer name.


hgtonight:
This does bump the version requirement of PHP to 5.5 ([`array_column()`](http://php.net/manual/en/function.array-column.php))

tburry:
We have a compatibility version of it [here](https://github.com/vanilla/vanilla/blob/ec75dc79301c524ff4de2c619d7d95c9e2255298/library/core/functions.general.php#L97).

hgtonight:
+1 :)

tburry:
Also note that although this PR doesn't remove all ConsolidateArrayValuesByKey() uses I think it's good to get this merged before the code diverges.

PR #2734, opened 2015-05-07
Add a config setting to only sync roles on register and fix the sync roles config spelling mistake.

tburry:
The Garden.SSO.SyncRoleBehavior can be set to “register” so that roles will only synchronize when registering and not on subsequent sign ins.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -455,7 +455,7 @@ EOT;
       $UserID = GetValue('UserID', $Auth);
 
       // Check to synchronise roles upon connecting.
-      if (($this->Data('Trusted') || C('Garden.SSO.SynchRoles')) && $this->Form->GetFormValue('Roles', NULL) !== NULL) {
+      if (($this->Data('Trusted') || C('Garden.SSO.SyncRoles')) && $this->Form->GetFormValue('Roles', NULL) !== NULL) {
          $SaveRoles = TRUE;
 
          // Translate the role names to IDs.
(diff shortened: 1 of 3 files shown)


tburry:
Here are some testing scenarios. To test this you will need to set up an SSO connection that passes a "roles" property back to the SSO.

* [ ] Make sure a regular SSO connection works without any config settings to make sure roles don't sync.
* [ ] Turn on Garden.SSO.SyncRoles and make sure the roles sync when registering.
* [ ] Turn on Garden.SSO.SyncRoles and make sure the roles sync when signing in again.
* [ ] Also turn on Garden.SSO.SyncRolesBehavior = 'register' and make sure the roles sync when registering.
* [ ] Also turn on Garden.SSO.SyncRolesBehavior = 'register' and make sure the roles **don't** sync when registering.

applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -468,9 +468,15 @@ public function Connect($Method) {
             $RoleIDs = $this->UserModel->NewUserRoleIDs();
          }
 
+         if (C('Garden.SSO.SyncRolesBehavior') === 'register') {
+            $SaveRolesRegister = TRUE;
+            $SaveRoles = FALSE;
+         }
+
```
linc: It doesn't seem right that you can get out of this block without `$SaveRolesRegister` being set at all (if `Garden.SSO.SyncRoles` == `true` and `Garden.SSO.SyncRolesBehavior` == `false`).

tburry: `Garden.SSO.SyncRolesBehavior` not being set is the current behavior. If `Garden.SSO.SyncRoles` is set and the behavior isn't then it means to sync roles all the time which it currently does.

linc: In the current behavior, the call to `Register()` below would then have 'SaveRoles' set to `true` whereas now it would be `false`.

tburry: I see what you're saying. Fixed.

PR #2737, opened 2015-05-08
Fix bug where comments don’t update the latest post on categories.

tburry:
The issue is caused when memcached is enabled. The logic was looking at a property erroneously and that check was removed.


tburry:
Kind of had to merge this. It was a bug.

PR #2739, opened 2015-05-11
Change the priorization of formatters of To() function

R-J:
<b>At first: I'm <i>very</i> unsure about this one!</b>

As far as I can tell, by now it wouldn't be possible to override any standard formatter of class.format.php easily. The classes functions are always used first and if they do not implement an alternative way, a plugin author is lost.
Although such an alternative exists for function BBCode(), it leaves the output out of the control of the plugin author since there are three more transformations after the custom result is returned.
function Markdown() doesn't give an alternative at all.

So I think if someone wants to have a custom formatter, it should be used at first. If someone doesn't implement a formatter but simply installs a custom function formatSomething, that function should be used.

(By the way: I do not understand why there are those two possibilities: to me they look as if they would work exactly the same.)

And then - and only if there is no alternative! - the standard functions should be used. The formatting functions aren't even declared with an enclosing `if(!function_exists...` so I think that someone who wants to implement a custom Markdown formatter is lost.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -255,17 +255,7 @@ class Gdn_Format {
    public static function BBCode($Mixed) {
       if (!is_string($Mixed)) {
          return self::To($Mixed, 'BBCode');
-      } else {
-         // See if there is a custom BBCode formatter.
-         $BBCodeFormatter = Gdn::Factory('BBCodeFormatter');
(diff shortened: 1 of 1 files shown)


tburry:
As it stands this is going to break backwards compatibility as our built in formatters depend on mentions and such being baked in.

PR #2740, opened 2015-05-11
Only show Activity link to those with Garden.Activity.View permission

beckyvb:
Lithe uses the SiteNavModule to display its links. This fixes a bug where the Activity menu item shows up on Lithe whether a user has permission or not.


initvector:
Works for me!

PR #2741, opened 2015-05-11
Change class filter on user generated content to a whitelist

linc:
This does a global filter of the `class` attribute on all HTML elements, then re-allows a whiltelist of values of `class` for a whitelist of elements.

Whitelisted values for `class`:
Quote|Spoiler|CodeBlock|Spoiled|P|ImageResized|AuthorAction|MeAction, plus Align* and post-* (This last one is Editor's class prefix, which would be how plugin authors can sidestep this restriction going forward.

Elements allowed to have these `class` values: a,span,div,p,li,i,b,strong,em,code,blockquote,dd,dt,img,pre,h1,h2,h3,h4,h5,h6,s,del,sup,sub,q,small,cite,ins

### DIFF
--- a/plugins/HtmLawed/class.htmlawed.plugin.php
+++ b/plugins/HtmLawed/class.htmlawed.plugin.php
@@ -51,10 +51,10 @@ class HTMLawedPlugin extends Gdn_Plugin {
        'direct_list_nest' => 1,
        'balance' => 1
       );
-      
+
       // Turn embedded videos into simple links (legacy workaround)
       $Html = Gdn_Format::UnembedContent($Html);
-      
+
       // We check the flag within Gdn_Format to see
       // if htmLawed should place rel="nofollow" links
       // within output or not.
@@ -72,7 +72,7 @@ class HTMLawedPlugin extends Gdn_Plugin {
       if ($this->SafeStyles) {
          // Deny all class and style attributes.
          // A lot of damage can be done by hackers with these attributes.
-         $Config['deny_attribute'] .= ',style';
+         $Config['deny_attribute'] .= ',style,class';
 //      } else {
 //         $Config['hook_tag'] = 'HTMLawedHookTag';
       }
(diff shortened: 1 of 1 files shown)


tburry:
Does this whitelist only apply when safe styles is off?

linc:
No, this is for when it is on. There are classes you can currently add to content and signatures that abuse the layout of the forum even when filtering is enabled.

plugins/HtmLawed/class.htmlawed.plugin.php
```diff
@@ -97,7 +97,9 @@ public function Format($Html) {
          'Status' => 1,
       );
 
-      $Spec = 'object=-classid-type, -codebase; embed=type(oneof=application/x-shockwave-flash); a=class(noneof=Hijack|Dismiss|MorePager/nomatch=%pop[in|up|down]|flyout|ajax%i)';
+      $Spec = 'object=-classid-type, -codebase; embed=type(oneof=application/x-shockwave-flash); ';
+      //$Spec .= 'a=class(noneof=Hijack|Dismiss|MorePager/nomatch=%pop[in|up|down]|flyout|ajax%i); ';
+      $Spec .= 'a,span,div,p,li,i,b,strong,em,code,blockquote,dd,dt,img,pre,h1,h2,h3,h4,h5,h6,s,del,sup,sub,q,small,cite,ins=class(oneof=Quote|Spoiler|CodeBlock|Spoiled|P|ImageResized|AuthorAction|MeAction/match=%^(post-|Align)%i)';
```
tburry: I think we should remove: li,i,b,strong,em,dd,dt,h1,h2,h3,h4,h5,h6,s,del,sup,sub,q,small,cite,ins.

linc:
I revised the list to use the "normal" elements you'd find in forum posts (the definition lists I was on the fence about so I included 'em) and fixed my spec to work correctly; I don't think you can mix `oneof` and `match` without it using an `AND` operator which kinda defeats what I was shooting for, so I had to go full regex.

linc:
Pulled out the regex plus some classes that weren't needed.

beckyvb:
Looks good!

PR #2745, opened 2015-05-13
Fix issue with CategoryModel::MakeTree().

tburry:
Fix an issue where depths are not properly adjusted when calling `CategoryModel::MakeTree()` without specifying a root.


kaecyra:
Tested, works.

PR #2746, opened 2015-05-13
Add some more error checking to OpenID urls before discovery.

tburry:
Prevent users from using OpenID to examine non-standard urls.


initvector:
Looks good and held up against testing

PR #2750, opened 2015-05-14
Fixing Facebook scope

nervoustwit:


plugins/Facebook/class.facebook.plugin.php
```diff
@@ -522,13 +522,23 @@ public function GetProfile($AccessToken) {
 
    public function AuthorizeUri($Query = FALSE, $RedirectUri = FALSE) {
       $AppID = C('Plugins.Facebook.ApplicationID');
-      $FBScope = C('Plugins.Facebook.Scope', Array('email','public_profile'));
+      $FBScope = C('Plugins.Facebook.Scope', 'email,publish_actions');
+
+      if(is_array($FBScope))
+      {
```
tburry: Bracket style.

tburry: And a space after the if.

PR #2751, opened 2015-05-14
Fix an issue with generating urls when rewrite is off.

tburry:
Fix an issue when `Garden.RewriteUrls` is off where `index.php` gets added multiple times.


tburry:
This is likely a PR that needs to be back-ported to OSS.

PR #2752, opened 2015-05-14
Fix tag URLs when in a subdirectory.

tburry:
Fix an issue where tags are getting double url’d in the popular tags module.


beckyvb:
Looks good!

PR #2757, opened 2015-05-14
Allow assignment of permissions based on a Role's Type column

initvector:
Adds a Type column to the Role table.  Possible types are: Guest, Unconfirmed, Applicant, Member, Moderator, Administrator

Methods added to PermissionModel to allow assignment of permissions based on a Role's Type column.  Event also added to allow other applications and plug-ins to set default role type permissions.

Once the method is nailed down, the goal is to move permission assignment into post-install and post-import areas.  This is especially important for imports, because it has the potential of simplifying incoming permissions when migrating from a different platform.


applications/dashboard/settings/structure.php
```diff
@@ -28,6 +28,7 @@
    ->PrimaryKey('RoleID')
    ->Column('Name', 'varchar(100)')
    ->Column('Description', 'varchar(500)', TRUE)
+   ->Column('Type', array('Guest', 'Unconfirmed', 'Applicant', 'Member', 'Moderator', 'Administrator'), TRUE)
```
tburry: These enums also need to be constants on the `PermissionModel`. Use the constants in code.

tburry:
My impression is that the arrays that are assigned in the event are a bit too complex. What I'd do is add a convenience method to `PermissionModel` and then have the event handlers call that via `$Sender`. Make the permission model do all the logic work.

I'll be shortening your function and event names, but will go over that later.

initvector:
1. Added PermissionModel::TYPE_* class constants and use them instead of strings
2. Included AddDefault method to PermissionModel to allow other applications and plugins to pass along their defaults during the DefaultPermissions event
3. Shortened event and function names

applications/dashboard/models/class.permissionmodel.php
```diff
@@ -820,6 +965,47 @@ public function SaveAll($Permissions, $AllWhere = NULL) {
...
+    */
+   public function ResetRole($Role) {
+      // Grab the value of Type for this role.
+      $RoleType = $this->SQL->GetWhere('Role', array('Name' => $Role))->Value('Type');
+
+      if ($RoleType == '') {
+         $RoleType = self::TYPE_MEMBER;
+      }
+
+      $this->AssignDefaults();
```
tburry: You have an inefficiency here where calling `ResetAllRoles()` means that this is called multiple times even though one call gives enough information for all defaults.

I'd suggest making an accessor for `$_DefaultPermissions` and have that lazy populate the array. Also FYI: we don't use underscores on protected properties or methods anymore. That was an artifact of old PHP.

tburry:
Some testing scenarios.

- [ ] Install a new forum.
- [ ] Register with basic.
- [ ] Register with approval.
- [ ] Register with SSO.
- [ ] Register with an invite.
- [ ] Fix the role types from an old installation on utility/update.

PR #2747, opened 2015-05-13
Promoted Content api update

beckyvb:
Add Url field to post
Prune Author field 
Add Url, PhotoUrl, RankName and RoleNames fields to Author field, 
Add 'Re' to comment Name field

PR #2762, opened 2015-05-15
Promoted Content: Add and configure new promoted content view.

beckyvb:


beckyvb:
Relies on https://github.com/vanilla/vanilla/pull/2747/

applications/vanilla/views/modules/promoted.php
```diff
@@ -0,0 +1,112 @@
...
+   $categoryName = val('CategoryName', $row);
+   $date         = val('DateUpdated', $row) ?: val('DateInserted', $row);
+   $date         = Gdn_Format::Date($date, 'html');
+   $type         = val('RecordType', $row, 'post');
+   $id           = val('CommentID', $row, val('DiscussionID', $row, ''));
+   $author       = val('Author', $row);
+   $username     = val('Name', $author);
+   $userUrl      = val('Url', $author);
+   $userPhoto    = val('PhotoUrl', $author);
+   $cssClass     = val('_CssClass', $author);
```
linc: Typically we use these `val`s in context rather than calling a big stack of them like this.

linc:
I feel like these sorts of functions wanna be in a helper_functions type of file typically, no? Overriding this view would mean killing all these functions right now too.

applications/vanilla/views/modules/promoted.php
beckyvb: I'm writing my views with the idea that we may eventually decide to use templates for our html views. I'm trying to separate the data collection and logic from the html/view as much as possible to make that conversion as smooth as possible.

beckyvb:
Ready for review again. This is potentially breaking for any theme that has overridden the modules/helper_functions view.

PR #2764, opened 2015-05-16
Adds no follow to social logins

adrianspeyer:


plugins/Twitter/class.twitter.plugin.php
```diff
@@ -226,7 +226,7 @@ public function CommentModel_AfterSaveComment_Handler($Sender, $Args) {
 	private function _GetButton() {
       $Url = $this->_AuthorizeHref();
 
-      return SocialSigninButton('Twitter', $Url, 'icon', array('class' => 'js-extern'));
+      return SocialSigninButton('Twitter', $Url, 'icon', array('class' => 'js-extern'), 'rel' => 'nofollow'));
```
tburry: I think the `rel` needs to be before that bracket.

adrianspeyer:
Yep, fixed the errant parentheses.

Issue #1410, opened 2012-08-03
Moving discussion doesn't update "most recent post" for each category


linc:

linc:
Fixed by #2202

bleistivt:
Imo this part should be moved into its own method on the CategoryModel:

https://github.com/vanilla/vanilla/blob/fd6484796abd2af457ab0a62cbc463e1bd88886e/applications/vanilla/models/class.commentmodel.php#L906-L930

This could then be called when moving discussions.

linc:
From briefly glancing at that, I'd agree.

linc:
This is the dumbest bug ever.

PR #2766, opened 2015-05-20
Update recent post when moving a discussion

bleistivt:
#1410 This needs to be tested with caching enabled and multiple discussions.

### DIFF
--- a/applications/vanilla/controllers/class.moderationcontroller.php
+++ b/applications/vanilla/controllers/class.moderationcontroller.php
@@ -350,7 +350,7 @@ class ModerationController extends VanillaController {
 
       // Check for edit permissions on each discussion
       $AllowedDiscussions = array();
-      $DiscussionData = $DiscussionModel->SQL->Select('DiscussionID, Name, DateLastComment, CategoryID')->From('Discussion')->WhereIn('DiscussionID', $DiscussionIDs)->Get();
+      $DiscussionData = $DiscussionModel->SQL->Select('DiscussionID, Name, DateLastComment, CategoryID, CountComments')->From('Discussion')->WhereIn('DiscussionID', $DiscussionIDs)->Get();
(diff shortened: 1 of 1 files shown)


linc:
Can you make it also fix the counters? +1 / -1

linc:
Verified this does work with caching enabled.

bleistivt:
This became a little more verbose. Should I resubmit against stage?

linc:
Nah, we're avoiding stage nowadays.

linc:
My test went like this. I checked 2 discussions from Category A and 2 from Category B and moved them to Category C.

Expected counter changes:
* A: -2
* B: -2
* C: +4

Actual change:
* A: -4
* B: 0
* C: 0

I ran /dba/counts to fix my counters. I then repeated the test with D, E, & F, moving 4 from D and 1 from E all into F.

Actual changes:
* D: +2
* E: +1
* F: +7

Caching is disabled. Something is amiss here.

initvector:
It looks like this automatically assigns whatever discussion is being moved to be the most recent for a category, regardless of the date it was created.  If I move a discussion from an hour ago into a category with a discussion from five minutes ago, the hour-old one appears as the "Most Recent".  Running dba/counts corrects that to the five-minute-old discussion.

bleistivt:
This would mean that CategoryModel::SetRecentPost is not working as intended:
https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1815-L1826

I can't find anything though, it sorts by DateLastComment.

bleistivt:
@initvector I can't reproduce this. Did you do your test with caching enabled?

linc:
I couldn't reproduce either, and if that's a bug it has nothing to do with this PR, so we're all set here. :) Thanks!

PR #2793, opened 2015-06-02
Don't use the session user in ConversationModel

bleistivt:
When programatically sending messages to the current user (e.g. from a bot user), the InsertUserID gets a notification but the receiving user does not, because it excludes the session user (as opposed to the insert user) from the notifications.

This commit changes the behaviour to excluding the InsertUserID to make the Save() function session independent.


linc:
I like it.

PR #2794, opened 2015-06-02
Fix mention detector to allow longer usernames

linc:
Raise the max length from 20 to 64 to match `GdnFormat::Mentions()` which accounts for jsConnect & import names rather than the default {3,20} restriction.


initvector:
Solid work.

Issue #2134, opened 2014-09-26
typo in fireevent   $this->FireEvent('CustomNotificationPreferneces');  2.1 and Master.


peregrine-web:
https://github.com/vanilla/vanilla/blob/master/applications/dashboard/views/profile/preferences.php#L102

probably want Preferences not Preferneces

change from 

$this->FireEvent('CustomNotificationPreferneces');

to

$this->FireEvent('CustomNotificationPreferences');

PR #2797, opened 2015-06-03
Fix typo in CustomNotificationPreferences event name

linc:
Closes # #2134


initvector:
I agree that Preferences should not be spelled Preferneces.

PR #2801, opened 2015-06-04
Remove the "ImageResized" text under images & all related code

### LINKED ISSUE #2768: Remove “This image has been resized” text, or add a toggle option
Most users already understand they can try click an image to see a larger version, so it seems generally unnecessary.

linc:
Closes #2768


bleistivt:
Would you accept a second PR that adds a (silent) config option that also disables the autolinking?

linc:
Why would that be desirable?

bleistivt:
People in my community sometimes use hundreds of images in a post, so I added some javascript to lazyload all images which is a great benefit for mobile users. This requires me to change the src attribute to a placeholder image though. ImageResized then wraps the images in links to the placeholder image, which is really bad.

I know it's kind of a specific use case, but another one I could think of is when using images as design elements in a post.

And when a user puts an image between a tags, the browser will follow the innermost link which is the one added by ImageResized

linc:
Can you look at perhaps making it default to lazyload behavior when you have `n` or more images in a post or discussion page? Your hundreds-of-images use case makes a lot of sense, so let's optimize for that in core rather than making a crappy config toggle to address 10% of it.

tburry:
Is there a max-width: 100% to go somewhere with this PR?

linc:
I don't follow. This doesn't change any behavior or styling, it simply removes a line of needless text.

PR #2804, opened 2015-06-05
Fixes import steps page when no original query string is present

initvector:
Prior to this fix, the import "Go" page was displaying a popup window of the same page every time the page checked for an update to the current progress of steps.  This was due to the DeliveryType and DeliveryMethod JSON being lost during the status update requests, so the full markup of the page was returned, instead of just the relevant data.  The loss of those parameters occurred when no query string originally existed on the "Go" page.

The original JavaScript just tried to append `&DeliveryType=VIEW&DeliveryMethod=JSON` onto the end of a URL.  If no query string existed, these parameters obviously weren't being passed on.  This fix appends either a ? or a & to the target URL, depending on if a query string already exists, before attempting to pass the parameters.


linc:
The long national nightmare is over.

PR #2810, opened 2015-06-08
Change the minimum number of characters in a username from 3 to 1 when creating mentions.

nervoustwit:


linc:
This didn't shake out like I thought it would so I cherry-picked it to master.

PR #2812, opened 2015-06-08
Casening cleanup

linc:
Just what it says on the tin.


kaecyra:
This PR is cynamite. Merging...

PR #2814, opened 2015-06-09
Fix PHP 5.3 compatibility in MesageModel::getMessagesForLocation()

linc:
* Refactor away use of `$this` in closure by using a static method (which is probably shoulda been anyway).
* Cleanup docs & method casing found along the way.


nervoustwit:
Looks good, tested locally just fine.

PR #2816, opened 2015-06-09
Fix stack trace on invalid DeliveryType

linc:
If someone sets an invalid DeliveryType, just do nothing instead of gratifying them with a lurid stack trace.


nervoustwit:
Tested it locally, it looks good.

PR #2817, opened 2015-06-09
Check for Gdn_Role.Type before using in where clause

linc:
Dashboard's structure recently added Role.Type to the database, which needs to exist before attempting to use the Type column for additional queries.

This change has the unfortunate side effect of making the default role change not advertise itself on /utility/structure before it is run, but I think that's an acceptable tradeoff give the alternatives of A) Big red warning on that page instead or B) Refactoring how that page calculates changes to be done to not fatal when a column isn't present yet, which sounds like a nightmare.


beckyvb:
Looks good!

PR #2818, opened 2015-06-09
Improve handling of sign-in redirect and TrustedDomains

### LINKED ISSUE #2800: Open redirect in /entry endpoints & the TrustedDomains setting
1. `EntryController::Target()` has an insecure default. It is an open redirect unless the `Garden.TrustedDomains` config setting is manually set by the admin.
2. `Garden.TrustedDomains` is in the Dashboard under `/embed/advanced` and its description only references its functionality in relation to embedding. Embedding has a similarly permissive default of "unset TrustedDomains means embed anywhere".
3. Our documentation for SSO does not encourage users to set `TrustedDomains`.
4. Changing the behavior of `Target()` to no longer be an open redirect by default (without each site adding their trusted domains) would break hundreds of SSO implementations.

How can we close this open redirect without breaking existing implementations the moment they upgrade?

linc:
* [ ] Add a 'Leaving Vanilla' page that includes a timed redirect.
* [ ] Untrusted `Target` URLs should go thru Leaving page via `safeRedirect()` (see `EntryController::Target()`)
* [ ] Clarify `TrustedDomains` functionality with better copy & placement.
* [ ] Add `TrustedDomains` explanation to SSO & embed docs.

### Considerations:

* Do we want to be able to specify http vs https?
* Should jsConnect auto-add its URLs to TrustedDomains?
* Wildcard domains allowed?

Closes #2800

### DIFF
--- a/applications/dashboard/controllers/class.homecontroller.php
+++ b/applications/dashboard/controllers/class.homecontroller.php
@@ -154,4 +154,14 @@ class HomeController extends Gdn_Controller {
             $this->RenderException(permissionException());
         }
     }
+
+    /**
+     * Provide a safe landing page for external redirects.
+     *
+     * @since 2.2
+     * @access public
+     */
+     public function leaving() {
+        $this->render();
+     }
 }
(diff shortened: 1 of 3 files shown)


tburry:
I don't think we should have the timed redirect on the leaving page. This page is being put in for security so the user should make an explicit choice to leave.

PR #2819, opened 2015-06-09
Improve 404 handling & error views

### LINKED ISSUE #2774: Missing user profile sends error page "Whoops" instead of route "Default404"
Obstacle: Accessing an non existing user profile (like `/profile/xyz/`) results in a _Whoops!_ page (`dashboard/home/error`) with exception message "_User not found_". 

Expected: Make use of route Default404 (`dashboard/home/filenotfound`) like e.g. discussions (`/discussion/xyz`) do this

Reason: Send a user friendly *404* page instead of showing an exception message to visitors.

This happens also when accessing route `/profile` directly without any further URI.

Edith says: Affected version is 2.1.10

linc:
* Use `notFoundException()` for users & discussions.
* Add traces to `Default404` and `DefaultPermission` views.
* Change controller to prefer rendering those views.

Closes #2774

### DIFF
--- a/applications/dashboard/controllers/class.profilecontroller.php
+++ b/applications/dashboard/controllers/class.profilecontroller.php
@@ -480,9 +480,6 @@ class ProfileController extends Gdn_Controller {
             $this->Head->addTag('meta', array('name' => 'googlebot', 'content' => 'noindex'));
         }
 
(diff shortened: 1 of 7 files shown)


nervoustwit:
Tested locally. Worked fine for me.

PR #2826, opened 2015-06-11
Do not show google sign in button if google authentication not properly configured.

nervoustwit:
Google sign in button was showing if the plug in was turned on, even if there was no google clientID provided to the plugin. Added the isConfigured function to the condition for printing out the link, changed the functionality of the isConfigured function so that it pings google with the clientID and returns false if anything but a 200 or 301 is returned.

### DIFF
--- a/plugins/GooglePlus/class.googleplus.plugin.php
+++ b/plugins/GooglePlus/class.googleplus.plugin.php
@@ -132,17 +132,7 @@ class GooglePlusPlugin extends Gdn_Plugin {
      */
     public function isConfigured() {
         $Result = c('Plugins.GooglePlus.ClientID') && c('Plugins.GooglePlus.Secret');
-        if(!$Result) {
-            return $Result;
-        }
         return $Result;
-        $Url = $this->authorizeUri();
-        $Ping = $this->curl($Url, 'GET', true);
-        if($Ping === 302) {
-            return true;
-        } else {
-            return false;
-        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


plugins/GooglePlus/class.googleplus.plugin.php
```diff
@@ -198,9 +211,13 @@ public static function curl($Url, $Method = 'GET', $Data = array()) {
         }
 
         if ($HttpCode != 200) {
-            $Error = val('error', $Result, $Response);
+            if($Ping===false) {
```
linc: Spaces on either side of operator.

plugins/GooglePlus/class.googleplus.plugin.php
```diff
@@ -132,7 +132,20 @@ public function getAccessToken($Code) {
...
+        if(!$Result) {
+            return false;
+        }
+        /*
+         * If there is a GooglePlus client id check to see if it is valid
+         */
+        $Url = $this->authorizeUri();
+        $Ping = $this->curl($Url, 'GET', array(), true);
+
+        if($Ping === 302) {
```
linc: Space after `if`

linc:
I'd much rather see the plugin check this once when the configuration is updated than do it on every signin page load. Pinging Google that hard might not be something they care about, but our stack is gonna feel it, I think.

nervoustwit:
I was thinking about that. I will move it.

nervoustwit:
In the end I stripped out the the functionality of verifying the google clientID because:
* It is too much overhead on the page load, as you pointed out
* It is too difficult to integrate a custom validation function in the submitting of the config file
* The other social sign-ins don't validate either

plugins/GooglePlus/class.googleplus.plugin.php
```diff
@@ -198,9 +198,9 @@ public static function curl($Url, $Method = 'GET', $Data = array()) {
         }
 
         if ($HttpCode != 200) {
-            $Error = val('error', $Result, $Response);
+                $Error = val('error', $Result, $Response);
```
linc: Appears to have double indents here.

nervoustwit:
Thanks, I've been working in an environment where, if you indented at all you were one classy coder.

PR #2828, opened 2015-06-11
Editor: Enforce max-width of 100% on wysiwyg image uploads.

beckyvb:
Addresses issue where large uploaded images make editor difficult to use on mobile.


nervoustwit:
Looks good, tested on local.

PR #2830, opened 2015-06-12
Fixes typo in importmodel function names

initvector:
DeleteFiles and DeleteOverwriteTables were renamed to celeteFiles and celeteOverwriteTables during The Casening.  This just fixes those function names.


linc:
At least I'm consistent...?

PR #2831, opened 2015-06-12
Garden formatter: Add function to remove spoilers and invoke in plainText function.

beckyvb:
Fixes issue where spoilers appear in messages popin and messages/inbox view.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -964,24 +964,15 @@ class Gdn_Format {
      * @return string
      */
     protected static function replaceSpoilers($html) {
-        // Check if $html contains spoiler classes before we do the heavy lifting.
-        $found = false;
-        foreach(Gdn_Format::$spoilerClasses as $spoiler) {
-            if (strpos($html, $spoiler)) {
-                $found = true;
+        if (preg_match('/class="(User)?Spoiler"/i', $html)) {
+            // Transform $html into a dom object and replace the spoiler block.
+            if (!defined('HDOM_TYPE_ELEMENT')) {
+                require_once(PATH_LIBRARY.'/vendors/simplehtmldom/simple_html_dom.php');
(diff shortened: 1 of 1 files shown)


library/core/class.format.php
```diff
@@ -955,6 +958,35 @@ protected static function isHtml($Text) {
...
+            return $html;
+        }
+
+        // Transform $html into a dom object and replace the spoiler block.
+        if (!defined('HDOM_TYPE_ELEMENT')) {
+            require_once(PATH_LIBRARY . '/vendors/simplehtmldom/simple_html_dom.php');
+        }
+        $html = str_get_html($html);
+        foreach (Gdn_Format::$spoilerClasses as $spoiler) {
+            $html->find('.' . $spoiler, 0)->outertext = t('Spoiler Replacement', t('Spoiler'));
```
tburry: Check to see if `$html->find('.', implode(',.'.self::$spoilerClasses))` works instead of the loop.

Also try and make the translation code work on its own. So something like `t('(Spoiler)')`.

library/core/class.format.php
```diff
@@ -955,6 +958,35 @@ protected static function isHtml($Text) {
...
     /**
+     * Checks to see if $html has a spoiler element and replaces the element.
+     *
+     * @param string $html A html-formatted string
+     * @return string
+     */
+    protected static function replaceSpoilers($html) {
+        // Check if $html contains spoiler classes before we do the heavy lifting.
+        $found = false;
+        foreach(Gdn_Format::$spoilerClasses as $spoiler) {
```
tburry: Not a fan of this loop. I'd maybe just hardcode the spoiler regex for speed: `#class="(User)?Spoiler"#i`

beckyvb:
Looks good!

PR #2840, opened 2015-06-16
Add dropdown component

beckyvb:
Adds support for quick configuration of a sortable dropdown menu.


applications/dashboard/modules/class.dropdownmodule.php
```diff
@@ -0,0 +1,149 @@
+<?php if (!defined('APPLICATION')) exit();
```
linc: We no longer add this to declaraction-only files.

applications/dashboard/modules/class.sortablemodule.php
```diff
@@ -0,0 +1,467 @@
+<?php if (!defined('APPLICATION')) exit();
```
linc: same here

linc:
This looks fine to me otherwise.

When adding this much code to core, I think it deserves more than a two-sentence description. ;) I was a bit put off from digging into this with no context or overview.

linc:
`dropdown-twbs.php` - a typo of 'tabs' or an acronym I don't know?

beckyvb:
Twitter Bootstrap. The view follows the CSS class naming conventions of Twitter Bootstrap. Thinking about it now, this doesn't need to be in core. An alternate dropdown view is nice to offer, but should live elsewhere.

PR #2851, opened 2015-06-19
Fix mentions in brackets

R-J:
PR for 2.2
Fixes https://github.com/vanilla/vanilla/pull/2652
Regex was done wrong and allowed `)@someone` instead of  `(@someone`

@bleistivt told me I did it wrong, but I missed to correct it... :(


linc:
Backported for 2.2.

PR #2852, opened 2015-06-19
Applying Vanilla new coding standard.

chanhong:
Just some left over minor code changes based on the new coding standard.


linc:
* Multi-line checks for APPLICATION defined / exit() is not desireable.
* There are many incorrect spacing and lines being added / deleted.

I recommend against trying to automate coding standards changes at this point. We've gone as far as the tools can take us effectively.

chanhong:
ok, thanks

PR #2856, opened 2015-06-22
Fix issue with saving roles when creating a new user via embedded SSO

tburry:
This issue manifested itself when Garden.SSO.SyncRoles was turned on and then a new user was registered via embedded SSO. In this case the new user would have no roles rather than those passed via the SSO.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -668,7 +668,7 @@ class UserModel extends Gdn_Model {
             $CurrentUser = $this->getID($CurrentUser, DATASET_TYPE_ARRAY);
         }
 
-        // Don't synch the user photo if they've uploaded one already.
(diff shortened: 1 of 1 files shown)


tburry:
I'm hoping that @linc can help with some testing on this one when he gets back in, but for now this will have to merge.

linc:
Oh man cheatin' with the PRs the minute I go on vacation, I see how it is. :ghost:

applications/dashboard/models/class.usermodel.php
```diff
@@ -4354,4 +4345,28 @@ public function getPermissionsIncrement() {
         return $PermissionsKeyValue;
         ;
     }
+
+    /**
+     * @param $Roles
```
linc: Not adding a summary to the docblock. For shaaaaaaame.

PR #2859, opened 2015-06-24
Change order of values in Year select input.

beckyvb:
Descending years is usually a better user experience than ascending.


linc:
The tweak namespace! :tada:

PR #2860, opened 2015-06-25
Fix for plugins controllers on windows

bleistivt:
Controllers in plugins don't work on windows, because the plugin path that the dispatcher gets from the autoloader can contain backslashes.

Further explanation:

This is how application controllers are saved in the controller maps:

    discussionscontroller = "Z:\web\vanilla-master/applications/vanilla/controllers/class.discussionscontroller.php"

This is how plugin controllers are saved in the core library map:

    testcontroller = "Z:\web\vanilla-master\plugins\test/controllers/class.testcontroller.php"

The path is not being properly split and the dispatcher doesn't recognize it as a plugin controller and throws a 404.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -630,7 +630,7 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
                 if ($ControllerPath) {
                     $InterimPath = explode('/controllers/', $ControllerPath);
                     array_pop($InterimPath); // Get rid of the end. Useless;
-                    $InterimPath = explode('/', trim(array_pop($InterimPath)));
+                    $InterimPath = explode('/', str_replace('\\', '/', trim(array_pop($InterimPath))));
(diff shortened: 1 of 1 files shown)


tburry:
I'd rather this be fixed in the map.

bleistivt:
Reopened for discussion:

This would fix it:
https://github.com/vanilla/vanilla/compare/vanilla:master...bleistivt:patch-4

But it should definitely go into the fixBackSlash function which already does that:
https://github.com/vanilla/vanilla/blob/master/library/core/class.autoloader.php#L1083

@chanhong What was the reason for **not** replacing backslashes when there is a drive at the start like `A:\...` ?

Could this condition just be omitted so it always replaces backslashes?

Afaik Windows takes paths with slashes without a problem.

chanhong:
If you have a use case that require it then why not.
I only change what was need in my use case.

PR #2867, opened 2015-07-01
Fix mentions formatting for usernames with special characters

tburry:
Fixes an issue where a username like todd#123 would not format the URLs correctly.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1471,14 +1471,20 @@ EOT;
 
             // Handle @mentions.
             if (C('Garden.Format.Mentions')) {
-                $urlFormat = str_replace('{name}', '$2', self::$MentionsUrlFormat);
-
                 // Unicode includes Numbers, Letters, Marks, & Connector punctuation.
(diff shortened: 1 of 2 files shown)


library/core/class.format.php
```diff
@@ -1471,14 +1471,20 @@ public static function mentions($Mixed) {
...
             if (C('Garden.Format.Mentions')) {
-                $urlFormat = str_replace('{name}', '$2', self::$MentionsUrlFormat);
-
                 // Unicode includes Numbers, Letters, Marks, & Connector punctuation.
                 $Pattern = (unicodeRegexSupport()) ? '[\pN\pL\pM\pPc]' : '\w';
                 $Mixed = Gdn_Format::replaceButProtectCodeBlocks(
-                    '/(^|[\s,\.>\(])@('.$Pattern.'{1,64})\b/i', //{3,20}
-                    '\1'.anchor('@$2', $urlFormat),
-                    $Mixed
+                    '/(^|[\s,\.>\(])@('.$Pattern.'{1,100})\b/i', //{3,20}
```
linc: This length change needs to be reflected in `getMentions()` as well.

PR #2869, opened 2015-07-02
Normalize paths in autoloader maps

bleistivt:
This fixes #2860 so that plugin controllers are recognized as such on windows.

### DIFF
--- a/library/core/class.autoloader.php
+++ b/library/core/class.autoloader.php
@@ -1068,24 +1068,20 @@ class Gdn_Autoloader_Map {
     }
 
     /**
-     * Fix path with backslash for Windows OS.
+     * Normalize paths
      *
-     * Check to see if the path is not empty and if it is not in a form of drive: then replace any "\" with "/" to avoid
-     * the parse_fn_file bug and it also further check to see if it has only one "/" and the OS is "WIN" then it will prepend "/"
-     * in front of the path so it has the correct file path.
+     * Replaces any "\" with "/" and prepends another slash to a single leading slash. 
      *
      * @param string $path
      * @return string
      */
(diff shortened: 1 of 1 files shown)


bleistivt:
Remind me to backport this if it gets accepted.

linc:
This looks fine to me but I don't have a Windows environment setup to test this so I'd appreciate a gander by @tburry too before we merge it.

library/core/class.autoloader.php
```diff
@@ -1068,24 +1068,20 @@ public function mapType() {
...
-     * Check to see if the path is not empty and if it is not in a form of drive: then replace any "\" with "/" to avoid
-     * the parse_fn_file bug and it also further check to see if it has only one "/" and the OS is "WIN" then it will prepend "/"
-     * in front of the path so it has the correct file path.
+     * Replaces any "\" with "/" and prepends another slash to a single leading slash. 
      *
      * @param string $path
      * @return string
      */
     function fixBackSlash($path) {
-        if (!empty($path) && !preg_match('`^[a-z]:`i', $path)) {
```
initvector: Why was the if-not-empty-and-doesn't-start-with-a-drive-letter-and-colon condition removed?

bleistivt: This function was originally written to accommodate for windows network paths, but if the files are on a local disc, it will start with a drive letter. See the issue for this #2860 for an example.

And if it is empty, nothing will be replaced anyway, so I dropped the condition.

initvector:
Just to recap: This PR broadens the scope to allow GDN_Autholoader:: fixBackSlash to process local paths.  Before this update, only network paths were fixed.

Looks good to me!

PR #2870, opened 2015-07-02
Fix initial permissions during site spawning

linc:
* Add second permission setup call in structure so it is triggered when called remotely.
* Minor cleanup in code I read while evaluating this issue.

### DIFF
--- a/applications/dashboard/controllers/class.setupcontroller.php
+++ b/applications/dashboard/controllers/class.setupcontroller.php
@@ -85,9 +85,8 @@ class SetupController extends DashboardController {
                     // Now that the application is installed, select a more user friendly error page.
                     $Config = array('Garden.Installed' => true);
                     saveToConfig($Config);
+                    $this->fireAs('UpdateModel')->fireEvent('AfterStructure');
                     $this->fireEvent('Installed');
-
-                    PermissionModel::ResetAllRoles();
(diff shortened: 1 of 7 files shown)


tburry:
I don't like that call in vanilla. What about conversations and everything else in the future?

tburry:
Try putting the check in `UpdateModel->runStructure()`.

tburry:
Bonus points for conditionally registering a handler using `Gdn::pluginManager()->registerHandler()` in dashboard's structure. We've never done this before, but it's the future.

linc:
Honestly putting it in Vanilla's structure instead of Dashboard's was a mistake; I lost track after testing several variations on how to do this. I'll have a look at those other solutions.

tburry:
I appreciate your honesty. The house of lies must be burned down.

linc:
:fire: :house_with_garden: :fire: :vertical_traffic_light: :fire_engine:

linc:
Can you elaborate on this `Gdn::pluginManager()->registerHandler()`? I'm not following what you're laying down. It seems like I'd already have to know what things I'm looking for installed (Vanilla, Conversations) to do that rather than things actually hooking.

kaecyra:
Tested on cl111, works as expected.

PR #2872, opened 2015-07-02
Remove double translate on registration forms

tburry:
Gdn_Form->CheckBox() already translates its string so this code was being translated twice.


beckyvb:
Looks good!

PR #2875, opened 2015-07-03
Throw permission exception when attempting to merge discussion redirects

### LINKED ISSUE #2782: Merging into a redirected discussion
The issue is that if you merge an existing thread with a moved thread, making the moved thread the main discussion, the secondary thread vanishes entirely.

**Steps to Recreate**
* Create a thread and name it, "Test Thread 1". 
* Create a thread and name it, "Test Thread 2". 
* Move the thread named "Test Thread 1" to another category and leave a redirect link. 
* In the original category, select "Test Thread 2" and the redirect link of the moved "Test Thread 1", then select "Merge". 
* Elect to Merge the two selected threads. Choose to make the redirect link of the moved "Test Thread 1" as the primary (or master) thread to merge the "Test Thread 2" (secondary) link into. 

**The result:** the secondary thread, "Test Thread 2" vanishes with any posts that it contained. 
**Expected result:** the secondary thread would be merged into the primary master thread. 
**Alternative expected result:** you can merge a redirect links and only actual threads can be merged.

linc:
Disallow using a discussion redirect as part of a merge. Currently, merging into a redirect is a backdoor delete (sort of). Fixes #2782.

### DIFF
--- a/plugins/SplitMerge/class.splitmerge.plugin.php
+++ b/plugins/SplitMerge/class.splitmerge.plugin.php
@@ -149,7 +149,7 @@ class SplitMergePlugin extends Gdn_Plugin {
         // Make sure none of the selected discussions are ghost redirects.
         $discussionTypes = array_column($Discussions, 'Type');
         if (in_array('redirect', $discussionTypes)) {
-            throw permissionException('@You cannot merge redirects.');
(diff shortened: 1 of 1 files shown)


tburry:
I don't think a permission exception is the right exception here. I think a `Gdn_UserException('...', 400)` would be better.

PR #2879, opened 2015-07-03
Read directory exception to rewrite rules in htaccess

linc:
Removing this rule destroyed access to any `/folder` that isn't Vanilla. Broke statistics for my sites and whole slew of other things I didn't notice until now.


initvector:
Works for me!

Issue #2864, opened 2015-06-29
PM button missing on iOS (default mobile)


linc:
http://vanillaforums.org/discussion/29924/cannot-delete-pms-on-mobile-devices
> When the confirmation popup appears 'Are you sure you want to do that', the Cancel button works but the Okay button does not.

@bleistivt [suggests](http://vanillaforums.org/discussion/comment/230714/#Comment_230714):
> `.Button` class needs: `-webkit-appearance: none;`

PR #2882, opened 2015-07-05
Remove webkit-appearance from .Button

bleistivt:
#2864 I have no idea how that could have fixed the users problem with deleting PMs, but removing the default style can't hurt.


tburry:
Wha? This looks like pure applesauce.

PR #2883, opened 2015-07-08
Comments view: Add event before rendering comment.

beckyvb:
Allow plugins to choose whether to render a comment.


tburry:
Is this the right thing to do? Usually you want the plugin to remove the actual comment data on the before render event.

tburry:
I'll just finalize and say this is not the right thing to do. Views need less logic, not more.

PR #2890, opened 2015-07-13
Always remove redundant columns from permission value updates.

tburry:
There are some cases where JunctionID goes through to a permission model save twice and causes an SQL error.


linc:
Should this get backported?

PR #2895, opened 2015-07-16
Move Js search code to AssetModel and refactor in line with CSS lookup

kaecyra:


linc:
A couple bullet points on the how & why make PRs like this a lot less painful to review. Coming at these things cold is a non-trivial task for us mere mortals. :ghost:  I assume this is for future minification but we shouldn't be leaving it up to reviewers to infer what's happening and why.

This appears to work as far as I can tell, including in a subdirectory.

PR #2896, opened 2015-07-20
Recognize Nintendo 3DS and New 3DS browser as mobile

bleistivt:
Only tested with the 3DS, but the new 3DS should work too, since its user agent string is `New Nintendo 3DS`.


linc:
Sweet!!!

PR #2900, opened 2015-07-21
Let userPhoto() fallback to PhotoUrl

linc:
* Fix coding standards in `userPhoto()`.
* Fix `userPhoto()` to accept 'PhotoUrl' when no 'Photo' is provided. Fixes interaction with PromotedContent module + Reactions, which only sets PhotoUrl, which seems reasonable.


beckyvb:
Looks good!

PR #2901, opened 2015-07-21
Support v2 Vanillicon in Gravatar

beckyvb:
Fixes bug where Gravatar's default user image url defaults to Vanillicon v1, no matter which Vanillicon version is selected.

Also adds caching to the user's Photo field, so the Gravatar api is only being invoked once per user.

### DIFF
--- a/plugins/Gravatar/default.php
+++ b/plugins/Gravatar/default.php
@@ -108,7 +108,8 @@ class GravatarPlugin extends Gdn_Plugin {
                     $url .= '&d='.$default;
                 }
             } else {
-                $default = urlencode(Asset(c('Plugins.Gravatar.DefaultAvatar', 'plugins/Gravatar/default.png'), true));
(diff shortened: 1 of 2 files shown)


initvector:
Just a couple minor things:

1. Gravatar + Plugins.Gravatar.UseVanillicon + Vanillicon v1 gives me pixelated avatars on profile pages.    It looks like it's attempting to stretch a 50px image to fill a 200px space.  Vanillicon v1 works A-OK when not using Gravatar.  I'm assuming this just needs a valid size passed to VanilliconPlugin::vanilliconUrl around [plugins/Gravatar/default.php:111](https://github.com/vanilla/vanilla/blob/gravatar-vanillicon/plugins/Gravatar/default.php#L111)
2. Asset should be lowercase at [plugins/Gravatar/default.php:116](https://github.com/vanilla/vanilla/blob/gravatar-vanillicon/plugins/Gravatar/default.php#L116)

PR #2902, opened 2015-07-22
Fix underlying sharding issue (sort order) and add auto sharding + randomization

kaecyra:
Cache sharding was breaking when the number of servers was > 9, because ksort() sorts numerically by default and was putting "10" just after "1", followed by "2" and "3" etc. Added SORT_NATURAL to fix.

Also added some sharding boundaries / limits to make it more efficient. Don't create shards that are too small (< 10000 bytes), and automatically shard keys that are too large (> 100000 bytes).


linc:
Looks good. Didn't test it but couldn't find any logical issues with it.

PR #2908, opened 2015-07-27
Editor: Load discussion media for answers, as well as comments.

beckyvb:
Fixes bug where file upload attachments do not render for answers.


plugins/editor/class.editor.plugin.php
```diff
@@ -1096,6 +1096,13 @@ protected function cacheAttachedMedia($Sender) {
 
         $DiscussionID = null;
         $Comments = $Sender->data('Comments');
+        if ($answers = $Sender->data('Answers')) {
+            $commentsArray = $Comments->resultObject();
```
linc: For the record, a method named `resultObject()` that returns an array is the work of the devil.

PR #2910, opened 2015-07-29
Fix quote plugin's stray bullet item

mtschirs:
A bullet item is now only shown if the quote button is visible. Guests no longer see a 'stray' bullet item. Fixes http://vanillaforums.org/discussion/30497/issue-quote-plugin-shows-stray-bullet-item-for-guests


linc:
Great tweak.

PR #2913, opened 2015-07-30
Fixed descending ordering of years in Gdn_Form::Date()

mtschirs:
While days and month where sorted in ascending order, the years where
unexpectedly sorted in descending order. This commit fixed this strange
behavior.

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -834,7 +834,7 @@ class Gdn_Form extends Gdn_Pluggable {
 
         $Years = array();
         $Years[0] = T('Year');
-        for ($i = $EndYear; $i >= $StartYear; --$i) {
+        for ($i = $StartYear; $i <= $EndYear; ++$i) {
             $Years[$i] = $i;
         }
 


hgtonight:
This was recently changed for usability: https://github.com/vanilla/vanilla/commit/e3d34ba375b6d257a3fc5d97e08a3df4c2ec18b9

mtschirs:
Ah, thanks, I didn't notice. Under which conditions does this ordering provide better usability?

hgtonight:
The PR was #2859.

I would assume most dates input would be closer to today versus 1900.

linc:
Yeah, UX > logical consistency.

linc:
I'd be fine with adding an attribute for sort order as long as we leave the current situation as the default.

mtschirs:
I guess that in a specific use case (year of birth?), a descending ordering provides better usability. However, in other use cases such as selecting future dates, this is confusing and the traditional ascending ordering provides a better UX experience. See the references pull request.

PR #2914, opened 2015-07-30
Made ordering of years in Gdn_Form::Date() dependend on supplied year range

mtschirs:
A year range of e.g. 2016-2018 procudes an ascending ordering, while a year range of e.g. 2018-2016 procudes a descending ordering.

Without supplied year range, the behavior is as before (descending ordering).

See also https://github.com/vanilla/vanilla/pull/2913


linc:
@beckyvb Would you make sure this is fine for whatever prompted your other PR?

beckyvb:
I like it. Looks good!

Issue #2871, opened 2015-07-02
Advanced Notification sends wrong message when post includes a mention


linc:
When someone is mentioned in an OP, the advanced notification of new post in a category sends out mention notifications instead of new post notifications.

Possibly related to closed issue #1292 ?

PR #2915, opened 2015-07-30
Prevent $Activity['HeadlineFormat'] from being overwriten

R-J:
Solves issue: https://github.com/vanilla/vanilla/issues/2871

Notifications for new discussions with advanced notifications were done wrong: at first the HeadlineFormat.Discussion was set to $Activity['HeadlineFormat'], then it was checked if users were mentioned and were to be informed. If yes,  $Activity['HeadlineFormat'] was set to HeadlineFormat.Mention. At last the users with advanced notification were notified but with the last HeadlineFormat.

Simply changing the order to first inform the users with advaanced notification before changing the HeadlineFormat solved the problem.


linc:
Nice work!

linc:
Picked this to 2.2 also.

PR #2916, opened 2015-07-30
Fix bugs caused by legacy columns in the discussion model.

mtschirs:
- Removed legacy columns from the discussion model
- Replaced references to those legacy columns from discussion views and
the Tagging plugin as well as the IndexPhotos plugin

This fixes http://vanillaforums.org/discussion/30496/issue-recent-discussions-module-doesnt-show-author#latest

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -144,7 +144,7 @@ class DiscussionModel extends VanillaModel {
      *
      * @param array $AdditionalFields Allows selection of additional fields as Alias=>Table.Fieldname.
      */
-    public function discussionSummaryQuery($AdditionalFields = array(), $Join = true) {
+    public function discussionSummaryQuery($AdditionalFields = array()) {
         // Verify permissions (restricting by category if necessary)
         if ($this->Watching) {
             $Perms = CategoryModel::CategoryWatch();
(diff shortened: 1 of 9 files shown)


linc:
Some serious heavy lifting in here, thanks. At first blush, I like this plan. I'm going to need to take a deeper dive on it and do some internal code vetting on our side to make sure this isn't going to mess up cloud, and I'm gonna want @tburry to take a look to make sure these query changes aren't gonna undo some sort of sidehatch performance gain I don't know about. This might take a bit.

mtschirs:
Thank you for looking into it! Since I only removed superfluous joins from the existing queries, the performance *should* improve a bit. But who knows...

linc:
This will absolutely clobber our VIP themes as it stands. Need to put a hold on this until we can get that resolved. Shouldn't be a huge deal to clean up, but it's definitely some time to invest and we'll need to wait for a release cycle after it's fixed everywhere.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -654,14 +631,6 @@ public function calculate(&$Discussion) {
...
         $Discussion->PermissionCategoryID = $Category['PermissionCategoryID'];
 
-        // Add some legacy calculated columns.
-        if (!property_exists($Discussion, 'FirstUserID')) {
-            $Discussion->FirstUserID = $Discussion->InsertUserID;
-            $Discussion->FirstDate = $Discussion->DateInserted;
-            $Discussion->LastUserID = $Discussion->LastCommentUserID;
-            $Discussion->LastDate = $Discussion->DateLastComment;
-        }
-
```
tburry: I think leaving something like this in for a period of time would help with the deprecation process.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -716,11 +685,6 @@ public function calculate(&$Discussion) {
 
         $Discussion->CountCommentWatch = is_numeric($Discussion->CountCommentWatch) ? $Discussion->CountCommentWatch : null;
 
-        if ($Discussion->LastUserID == null) {
-            $Discussion->LastUserID = $Discussion->InsertUserID;
-            $Discussion->LastDate = $Discussion->DateInserted;
-        }
-
```
tburry: You can see here that we set a user in the case of a discussion with no comments. Removing this kind of logic could lead to theme breakage.

tburry:
I think the original intention was to remove that join, but we had forgot about it. I like these changes, but really want some measure of backwards compatibility for a transition period as I've noted in my line comments.

applications/vanilla/models/class.discussionmodel.php
mtschirs: I would assume those are already legacy (=deprecated) columns.

applications/vanilla/models/class.discussionmodel.php
mtschirs: This code is flawed and causes actual theme breakage right now (e.g. see vanillaforums.org/discussion/30496/issue-recent-discussions-module-doesnt-show-author).

That's what this PR is all about - remove these incomplete and actually already theme-breaking legacy calculated columns and adapt old themes to no longer rely on their presence.

tburry: Ah OK. Fair enough.

Issue #2862, opened 2015-06-29
Setup failing in 2.2


j-norwood-young:
Hi there

When I try run a setup on a clean database, it fails without a specific error. The red error bar shows up, but with content. 

I've isolated the problem to this command in `dashboard/models/class.permissionmodel.php`:

```
$this->SQL
->set($this->_Backtick($DefaultPermissions), '', false)
->replace('Permission', array(), array('RoleID' => 0, 'JunctionTable' => $JunctionTable, 'JunctionColumn' => $JunctionColumn), true);
```

This is called by `dashboard/settings/structure.php` at:

```
// Define the set of permissions that Garden uses.
$PermissionModel->define(array(
    'Garden.Email.View' => 'Garden.SignIn.Allow',
    'Garden.Settings.Manage',
    'Garden.Settings.View',
    'Garden.SignIn.Allow' => 1,
    'Garden.Users.Add',
    'Garden.Users.Edit',
    'Garden.Users.Delete',
    'Garden.Users.Approve',
    'Garden.Activity.Delete',
    'Garden.Activity.View' => 1,
    'Garden.Profiles.View' => 1,
    'Garden.Profiles.Edit' => 'Garden.SignIn.Allow',
    'Garden.Curation.Manage' => 'Garden.Moderation.Manage',
    'Garden.Moderation.Manage',
    'Garden.PersonalInfo.View' => 'Garden.Moderation.Manage',
    'Garden.AdvancedNotifications.Allow',
    'Garden.Community.Manage' => 'Garden.Settings.Manage'
));
```

Unfortunately I don't know enough about the app to know where to start fixing this.

I'm running on PHP 5.4.39 on OSX. Vanilla version 2.2.100.4.

linc:
What is you MySQL version and do you have "strict mode" enabled?

j-norwood-young:
```
mysql> SELECT @@GLOBAL.sql_mode;
+--------------------------------------------+
| @@GLOBAL.sql_mode                          |
+--------------------------------------------+
| STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION |
+--------------------------------------------+
```

```
mysql> select @@VERSION;
+-----------+
| @@VERSION |
+-----------+
| 5.6.24    |
+-----------+
```

linc:
Thanks!

chanhong:
This sound like the same problem that I am having.  Thanks

geoclicks:
[...]

PR #2919, opened 2015-07-31
Replace permission names that are inserted into a tinyint column

R-J:
Refers to that issue: https://github.com/vanilla/vanilla/issues/2862

Inserting strings into permission table results into an error when database is in strict mode, since permission columns are of type TinyInt. That's why installation would fail if sql_model = strict.

Default value for those columns is '0', which is the value that is inserted if you try to insert a string into an integer column. So I simply commented out the wrong values. I haven't deleted them because I do not know which purpose they should have had...

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -305,7 +305,7 @@ if ($PermissionModel instanceof PermissionModel) {
 
 // Define the set of permissions that Garden uses.
 $PermissionModel->define(array(
-    'Garden.Email.View' => 'Garden.SignIn.Allow',
+    'Garden.Email.View', // => 'Garden.SignIn.Allow',
     'Garden.Settings.Manage',
     'Garden.Settings.View',
     'Garden.SignIn.Allow' => 1,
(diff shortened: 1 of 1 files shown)


bleistivt:
I think this is an error in PermissionModel::define(), because it shouldn't even try to insert string values.

see https://github.com/vanilla/vanilla/blob/master/applications/dashboard/models/class.permissionmodel.php#L229-L233

It should use the value of the referenced field.

R-J:
Although I cannot understand the code without running some data through it, I'm convinced you are right. But fixing the permission model would result in some other starting permissions than there are by now:

'Garden.SignIn.Allow' => 1,
'Garden.Email.View' => 'Garden.SignIn.Allow',
'Garden.Profiles.Edit' => 'Garden.SignIn.Allow',

Because of the error in the model, the two permissions above are initially set to 0 but with a working function define, they would be initially set to 1, is that correct?

bleistivt:
Yes, but it would be set to 3, because all permissions in the first row (RoleID 0) have their values incremented by 2. It looks like this is used for checking if a permission is global.

linc:
The expected behavior is that the permissions are added with a default of the existing permission named. Typically this is done for new permissions not in existing installs so we don't accidentally lock people out of parts of the applications. This PR is not a correct solution to the tinyint problem.

PR #2922, opened 2015-07-31
Fixed misspelled function_exists('DiscussionSorter')

mtschirs:
Replaced misspelled function_exists('DiscussionSorter') with function_exists('WriteDiscussionSorter') in applications/vanilla/views/discussions/helper_functions.php so function WriteDiscussionSorter() can be overwritten.


linc:
Thanks!

PR #2929, opened 2015-08-03
Fix stashing of non-embedded comments written by non-logged-in users.

mtschirs:
Fixes these issues:

The DiscussionController::index() method looks for stashed comments, but fails to correctly put these stashed values into the comment form.

The $('a.Stash').click() event handler in global.js detects a stashed comment, but fails to differ between non-embedded and embedded stashed comments. Also, it redirects automatically to the login form even if c('Garden.SignIn.Popup') is set to true. 

The view vanilla/view/post/comment.php incorrectly asks for $this->data('ForeignUrl') even though this view is for non-embedded comments where the concept of a ForeignUrl doesn't make sense.

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -209,6 +209,7 @@ class DiscussionController extends VanillaController {
         // Look in the session stash for a comment
         $StashComment = $Session->Stash('CommentForDiscussionID_'.$this->Discussion->DiscussionID, '', false);
         if ($StashComment) {
+            $this->Form->setValue('Body', $StashComment);
             $this->Form->setFormValue('Body', $StashComment);
         }
 
--- a/js/global.js
+++ b/js/global.js
@@ -1136,19 +1136,24 @@ jQuery(document).ready(function($) {
 
     // When a stash anchor is clicked, look for inputs with values to stash
     $('a.Stash').click(function(e) {
-        // Embedded comments
         var comment = $('#Form_Comment textarea').val(),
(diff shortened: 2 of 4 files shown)


applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -209,6 +209,7 @@ public function index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
         // Look in the session stash for a comment
         $StashComment = $Session->Stash('CommentForDiscussionID_'.$this->Discussion->DiscussionID, '', false);
         if ($StashComment) {
+        	$this->Form->setValue('Body', $StashComment);
```
linc: Needs another space.

js/global.js
```diff
@@ -1136,19 +1136,25 @@ jQuery(document).ready(function($) {
...
         var comment = $('#Form_Comment textarea').val(),
-            placeholder = $('#Form_Comment textarea').attr('placeholder'),
-            vanilla_identifier = gdn.definition('vanilla_identifier');
-
-        if (vanilla_identifier && comment != '' && comment != placeholder) {
+            placeholder = $('#Form_Comment textarea').attr('placeholder');
+        
+        // Stash a comment:
+        if (comment != '' && comment != placeholder) {
+        	var vanilla_identifier = gdn.definition('vanilla_identifier', false);
```
linc: Some funky spacing all in here, too.

linc:
Tested and working.

It's helpful to increment the fourth number in the version number in index.php file when making CSS or JS updates. Currently our asset caching is tied to that (I know, it pains us too) so it's needed to break caches reliably.

PR #2933, opened 2015-08-04
Added check for dependend plugins when disabling an application.

mtschirs:
If plugins are found that require a to-be-deactivated application to be enabled, an error is thrown with the names of the dependend plugins in the error message.

This fixes possibly fatal errors caused by enabled plugins that expect certain application classes to be loaded even though the application has been deactivated.

Fixes http://vanillaforums.org/discussion/30540/fatal-error-after-disabling-vanilla-application

### DIFF
--- a/library/core/class.applicationmanager.php
+++ b/library/core/class.applicationmanager.php
@@ -294,6 +294,24 @@ class Gdn_ApplicationManager {
             }
         }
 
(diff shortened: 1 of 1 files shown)


linc:
Great enhancement! 

In related news, I with there was a way to gracefully kill actions on that page without presenting the whole stack trace. So icky. :/

PR #2935, opened 2015-08-04
Fix issue involving posting new announcements

tburry:
This issue manifests itself when memcache is enabled and a new discussion is posted as an announcement. The discussion does not show up in in the announcements list and is thus removed from the recent discussions list.


linc:
copypasta casening

tburry:
I was keeping it consistent since it's just a small bug fix. I didn't want to redecorate the guts of the method.

PR #2936, opened 2015-08-04
Default Avatar

beckyvb:
Adds support for uploading and editting the thumbnails for a default avatar.
Also adds method for resolving the default avatar for a user.


linc:
Some issues:

* Standards: 2 `CopyLocal`, 1 `RemoveFromConfig`.
* Default Avatar page does not highlight when you're there.
* Might want more agnostic language than "Your forum's" on that page because Dashboard isn't a forum. It's for users in all contexts. A nitpick but something to keep in mind so we don't accidentally kill the independence of our apps over time.
* If we're creating a whole new page in the Dashboard for this, I think we should just call it "Avatars" so its ready for more related functionality.

This is probably outside the scope of this pull request, but here's what else I'd like to see on this page eventually:

* Reveal the other avatar configs as options: `Garden.Thumbnail.Size`, `Garden.Profile.MaxHeight`, `Garden.Profile.MaxWidth` with a note that changing the thumbnail size is not retroactive.
* Roll Gravatar & Vanillicon plugins into core and have them available on this page.

This could be a great one-stop shop for a fundamental piece of forum functionality we're not addressing in a coherent way right now.

Awesome feature, and the page works great so far.

applications/dashboard/design/cropimage.css
```diff
@@ -0,0 +1,68 @@
...
+
+.custom .jcrop-vline,
+.custom .jcrop-hline {
+    background: yellow;
+}
+
+.custom .jcrop-handle {
+    border-color: black;
+    background-color: #C7BB00;
+    border-radius: 3px;
```
linc: Declare vendor-specific rules before the standard.

linc:
Javascript needs docs, minor CSS thing noted inline, and there's a merge conflict.

beckyvb:
Updated docs, changed css and resolved conflicts!

linc:
Looks great.

I noticed on the cropping page that the headers were padded off-center to the right and were pretty small as an `h4`. The top heading on the page is an `h1`. It looks like updating those subheadings to `h2` made them line up and larger, and are the correct next tier after `h1`, so I updated them to that. Let me know if I missed rationale for it being an `h4`.

applications/dashboard/models/class.usermodel.php
```diff
@@ -941,7 +941,6 @@ public function joinUsers(&$Data, $Columns, $Options = array()) {
 
         // Join the user data using prefixes (ex: 'Name' for 'InsertUserID' becomes 'InsertName')
         $Join = val('Join', $Options, array('Name', 'Email', 'Photo'));
-        $UserPhotoDefaultUrl = function_exists('UserPhotoDefaultUrl');
```
bleistivt: With this removed, activity items like "x users joined" are missing the profile picture if the first user has only a vanillicon. (ActivityModel::getWhere gets the avatars from UserModel::joinUsers)

beckyvb: This should be resolved with this PR: https://github.com/vanilla/vanilla/pull/3068

PR #2937, opened 2015-08-05
Fix settings link for social plugins without settings URL

mtschirs:
Fix dashboard/view/social to not show non-functional settings link for social plugins without settings URL. The settings icon is now only shown if the plugin has a settings URL / settings page.


linc:
Nice enhancement.

PR #2944, opened 2015-08-05
Extract CanDelete out of managecategories view and set per category

linc:
* Pull the `CanDelete` logic out of the managecategories view up to the controller.
* Set the `CanDelete` flag per-category instead of globally so it can be overridden by plugins.


initvector:
Works for me!

PR #2945, opened 2015-08-06
Add hook to private messages for civiltongue plugin

nervoustwit:


linc:
Isn't `$this->MessageData` an object, so specifying by-reference would be redundant?

I'd just go with "beforeMessages" - it's firing before the standard list of messages (in a conversation), not a single one.

nervoustwit:
Good points. Thanks.

PR #2947, opened 2015-08-07
Allow disabling commenting ajax so that page will refresh on submission

nervoustwit:
In order to test, add 
$Configuration['Vanilla']['Discussion']['DisableAjax'] = true; 
to config.

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -57,6 +57,10 @@ class DiscussionController extends VanillaController {
         $this->addJsFile('discussion.js');
         Gdn_Theme::section('Discussion');
 
+        if (c("Vanilla.Discussion.DisableAjax")) {
+            $this->addDefinition("disableAjax", "true");
+        }
+
         // Load the discussion record
         $DiscussionID = (is_numeric($DiscussionID) && $DiscussionID > 0) ? $DiscussionID : 0;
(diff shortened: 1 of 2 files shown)


mtschirs:
All relevant page content should already refresh on submission of a new comment... see vanilla/js/discussion.js in line 182: // If adding a new comment, show all new comments since the page last loaded, including the new one.

linc:
@mtschirs I don't believe that comment reflects current functionality. We no longer return partial comment payloads because it causes irregular comment queries which effects performance.

tburry:
I'm really not in favor of this as a core change. Sounds like a better place for a theme hack to me. What's the reasoning behind this?

linc:
It's a common problem on very busy forums to end up having an out-of-sequence comment you just made on the page. Since it's a widespread issue, it seemed more useful as a core modification so we can apply it to other sites having the same issue rather than hacking all their themes.

linc:
I'm not against doing it as a theme hack if you really wanna do it that way, but I think it's time to admit our standard behavior here doesn't scale very well.

tburry:
Here is what I'd prefer:

1. Look into how to remove the event handler on the comment form.
2. Create a js recipe that can be done in a theme or pocket.
3. Hack into sites that want it.

bleistivt:
$('.CommentButton').on('click', function (e) {
        e.stopPropagation();
    });

PR #2948, opened 2015-08-07
Fix client-side form caching via removal of "no-store" from "Cache-Control" header

mtschirs:
By default, Gdn_Controller sends the following Cache-Control header to prevent caching:
"Cache-Control: private, no-cache, no-store, max-age=0, must-revalidate"

The "no-store" attribute however causes the client browser not to cache form input fields. As a result, users may experience data loss when writing e.g. comments and navigating back and forth in history. Instead of the last cached form content, a blank form field is shown.

See also http://blogs.atlassian.com/2007/12/cache_control_no_store_considered_harmful/

Side-effect: Without no-store, navigating through the browser history is much faster (at least on mozilla based browsers). This is because without no-store, history really means history - pages are not re-loaded but taken from the browser's history cache.

Fixes http://vanillaforums.org/discussion/30584/auto-drafts-vs-usability#latest

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -247,7 +247,7 @@ class Gdn_Controller extends Gdn_Pluggable {
 
         if (Gdn::session()->isValid()) {
             $this->_Headers = array_merge($this->_Headers, array(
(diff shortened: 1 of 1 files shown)


linc:
Thanks for the thorough explanation!

PR #2949, opened 2015-08-07
Fix UTF8 decode/encode in global.js to support multibyte unicode strings

mtschirs:
The replacements finally handle full 3 or 4 byte utf8 unicode representations well.

See http://ecmanaut.blogspot.de/2006/07/encoding-decoding-utf8-in-javascript.html

This fixes http://vanillaforums.org/discussion/30590/issue-unicode-emoji-mangled-by-vanilla-js

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -43,46 +43,14 @@ jQuery(document).ready(function($) {
 
     var keyString = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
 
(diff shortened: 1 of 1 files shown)


mtschirs:
Alternative and perhaps better solution: Remove base64-encoding of JSON data. 
Can anyone tell me what the reason was to start base64-encoding certain JSON strings?

linc:
>  Can anyone tell me what the reason was to start base64-encoding certain JSON strings?

Predates even me. I'll poke @tburry if he wants to chime in. My guess is that it was to prevent special characters in the markup from gumming up the JSON reply.

PR #2956, opened 2015-08-09
Remove error prone magic_quotes support

mtschirs:
Removed or made deprecate all code related to handling PHP's deprecated magic quotes.
Exception: Vendor libraries. I didn't touch them.

* Made TrueStripSlashes() and removeQuoteSlashes() from functions.general.php deprecated and removedreferences to it.
* Removed all "if ($MagicQuotes) ..." or similar code.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3148,7 +3148,11 @@ if (!function_exists('removeKeysFromNestedArray')) {
 }
 
 if (!function_exists('removeQuoteSlashes')) {
+    /**
+     * @deprecated
+     */
     function removeQuoteSlashes($String) {
+        deprecated('removeQuoteSlashes()');
         return str_replace("\\\"", '"', $String);
     }
 }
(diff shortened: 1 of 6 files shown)


library/core/functions.general.php
```diff
@@ -3147,12 +3147,6 @@ function removeKeysFromNestedArray($Array, $Matches) {
     }
 }
 
-if (!function_exists('removeQuoteSlashes')) {
```
tburry: You don't want to remove this function at this point. Just make a call to `deprecated()`.

library/core/functions.general.php
```diff
@@ -3616,18 +3610,6 @@ function trace($Value = null, $Type = TRACE_INFO) {
     }
 }
 
-if (!function_exists('trueStripSlashes')) {
```
tburry: You don't want to remove this function at this point. Just make a call to `deprecated()`.

mtschirs:
Updated; removed functions are back with call to deprecated().

PR #2958, opened 2015-08-10
Add CORS allow-origin header for trusted domain-originating requests

initvector:
If delivering JSON and receiving an Origin header from the client, attempt to verify the host of Origin against Garden.TrustedDomains.

Origin values should match if they are a legitimate part of a domain in Garden.TrustedDomains.  For example: If google.com is a trusted domain, we should be able to use it to match mail.google.com, www.google.com or google.com, but not fakegoogle.com.

Adds isTrustedDomain function to determine if a supplied domain is a match for an entry in TrustedDomains.  If no domain is provided, fallback to the request's host.

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -1227,6 +1227,23 @@ class Gdn_Controller extends Gdn_Pluggable {
             }
             $this->contentType('application/json; charset='.c('Garden.Charset', 'utf-8'));
             $this->setHeader('X-Content-Type-Options', 'nosniff');
+
+            // Cross-Origin Resource Sharing (CORS)
+
+            /**
+             * Access-Control-Allow-Origin
+             * If a Origin header is sent by the client, attempt to verify it against the list of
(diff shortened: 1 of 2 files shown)


initvector:
I'm on the fence about allowing isTrustedDomain to fallback to the referring host in the event no domain is provided.  Any feedback would be appreciated!

linc:
Funny enough, that was exactly the same thing I was on the fence about after reading over this.

Here's my take: If you can think of a solid, plausible use case for wanting that fallback in there right now, then leave it. If you can't, strike it. We shouldn't add code on spec.

Otherwise, this looks right and good. Particularly deft on the regex; I wouldn't have thought to preg_quote and opened a big ol' security hole.

initvector:
My initial thought was that it would be convenient for verifying where traffic was coming from in the absence of a valid Origin header, but now I leaning towards it being more of an obscure default.  As such, I've yanked it out.  It's an explicitly-defined domain or nothing.

Issue #2905, opened 2015-07-25
Gdn_Plugin->Dispatch not working as expected


mtschirs:
The method `dispatch` in **vanilla/library/core/class.plugin.php** takes two arguments, `$Sender` and `$RequestArgs`. However, apart from checking whether or not it is an array, the `$RequestArgs` argument is competely ignored.

I would expect the dispatch method to offer me the possibility to hand over some arguments to the dispatched method, e.g. by the `$RequestArgs` argument.

Also, what is the reason for having `$Sender->Form = new Gdn_Form();` in the `dispatch()`-method?

bleistivt:
Noticed this too. Ideally it would splat the requestArguments, so that plugin controller methods could be used like real controller methods.

I would guess that the form is for convenience, because most real controllers initalize a form automatically using the $Uses property.

PR #2959, opened 2015-08-11
Add $RequestArgs as parameter for dispatcher function

R-J:
Assuming you want to use the dispatcher in a plugin and access the request arguments, you had to always access them with $sender->RequestArguments.
Following code, which I expected to work flawlessly did fail since the $args parameter is not passed to the custom function:
~~~
public function vanillaController_something_create($sender, $args) {
    $this->dispatch($sender, $args);
}

// For example: /vanilla/something/edit/1
public function Controller_Edit($sender, $args) {
    // will fail because $args are not passed, although this is a common way to access request arguments
    $somethingId = $args[0];

   // by now you have to do it like that:
    $somethingId = $sender->RequestArguments[1];
}
~~~

### DIFF
--- a/library/core/class.plugin.php
+++ b/library/core/class.plugin.php
@@ -327,12 +327,13 @@ abstract class Gdn_Plugin extends Gdn_Pluggable implements Gdn_IPlugin {
             $TestControllerMethod = 'Controller_'.$MethodName;
             if (method_exists($this, $TestControllerMethod)) {
                 $ControllerMethod = $TestControllerMethod;
+                array_shift($RequestArgs);
             }
         }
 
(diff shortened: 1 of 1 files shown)


mtschirs:
Good, this might resolve the following issue: https://github.com/vanilla/vanilla/issues/2905

For the future, making $sender the $this object and "splat[ting] the requestArguments, so that plugin controller methods could be used like real controller methods" (@Bleistivt) would be a cleaner solution.

But for now, this PR is a very welcome fix!

R-J:
I'm happy there haven't been a fix by now! I simply forgot to search if that problem has already been addressed...

linc:
Nice catch and clean fix.

PR #2960, opened 2015-08-11
Drop unneccessary if condition

R-J:
$CategoryID is already set: false = zero for new categories and with the correct value for existing categories. So the where clause can be used either way. Makes the code shorter => more readable...


mtschirs:
Actually, when it comes to simplifying DB queries, I am ok with accepting a little bit more code complexity... also, your proposal implicitly relies on CategoryID = 0 never being used. For me, explicitly handling the special case of CategoryID = false helps to understand and verify the correctness of the code.

PR #2965, opened 2015-08-12
Tweak profile picture removal

tburry:
* Add a link to remove the profile picture in the edit profile menu no matter what. Previously it was just when editing someone else’s profile.
* Change the way the transient key is added to the remove picture URL. This fixes an issue when user URLs have no user IDs.

Note that this tweak breaks index.php?p= style configs, but we are deprecating that behavior.


linc:
I'd really like to get #3002 fixed before we start blowing up our fallback routing.

PR #2971, opened 2015-08-14
Add category options to event arguments

bleistivt:
The CategoryOptions event is useless, because the function builds a string that is not exposed.
This adds the string to the event arguments.


linc:
@bleistivt See https://github.com/vanilla/vanilla/pull/3611/files#r56415564 if this is something you're using.

bleistivt:
Thanks for the notification!

And @beckyvb good job the dropdown module! This makes everything so much easier, adding to dropdowns used to be a pain.

beckyvb:
@bleistivt Hey thanks!

PR #2972, opened 2015-08-14
Adds support for inline Imgur GifV embeds

kaecyra:
Tested, works locally.


bleistivt:
Awesome! I just had a user asking about it today.

PR #2973, opened 2015-08-15
Add intent checks & sanitize misc outputs

linc:
* Increase coverage of postback authentication.
* Sanitize JSON output from `user/tagsearch`.
* Add sane upper bound to `limit` param for `/modules` endpoint.
* Sanitize some template variables.
* Sanitize modulecontroller's applicationfolder handling.
* Add visibility management to `activity/item`.

### DIFF
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -1071,14 +1071,18 @@ class UserController extends DashboardController {
     }
 
     /**
+     * JSON output of a username search.
      *
-     *
-     * @param $q
+     * @param string $query
      * @param int $limit
      */
-    public function tagSearch($q, $limit = 10) {
-        $Data = Gdn::userModel()->tagSearch($q, $limit);
-        die(json_encode($Data));
+    public function tagSearch($query, $limit = 10) {
+        $data = Gdn::userModel()->tagSearch($query, $limit);
+        foreach ($data as &$user) {
+            // Sanitize usernames.
+            $user['name'] = Gdn_Format::text($user['name']);
+        }
+        die(json_encode($data));
     }
 
     /**
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -1079,10 +1079,7 @@ class UserController extends DashboardController {
      */
     public function tagSearch($query, $limit = 10) {
         $data = Gdn::userModel()->tagSearch($query, $limit);
-        foreach ($data as &$user) {
-            // Sanitize usernames.
-            $user['name'] = htmlspecialchars($user['name']);
-        }
+        header('Content-type: application/json');
         die(json_encode($data));
     }
 
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -919,7 +919,7 @@ class UserController extends DashboardController {
      */
     public function save() {
         $this->permission('Garden.Users.Edit');
-        if (!Gdn::request()->isPostBack()) {
+        if (!Gdn::request()->isAuthenticatedPostBack()) {
             throw new Exception('Requires POST', 405);
         }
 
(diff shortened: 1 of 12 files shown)


applications/dashboard/controllers/class.usercontroller.php
```diff
@@ -1071,14 +1071,18 @@ public function sso($UserID = false) {
...
      * @param int $limit
      */
-    public function tagSearch($q, $limit = 10) {
-        $Data = Gdn::userModel()->tagSearch($q, $limit);
-        die(json_encode($Data));
+    public function tagSearch($query, $limit = 10) {
+        $data = Gdn::userModel()->tagSearch($query, $limit);
+        foreach ($data as &$user) {
+            // Sanitize usernames.
+            $user['name'] = Gdn_Format::text($user['name']);
```
tburry: Should most likely be htmlspecialchars, but if this can be fixed on js side then it should be.

applications/dashboard/views/admin.master.php
```diff
@@ -1,28 +1,28 @@
 <?php echo '<?xml version="1.0" encoding="utf-8"?>'; ?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="<?php echo Gdn::Locale()->Locale; ?>">
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="<?php echo Gdn_Format::text(Gdn::locale()->Locale); ?>">
```
tburry: htmlspecialchars

applications/dashboard/views/admin.master.php
```diff
@@ -1,28 +1,28 @@
...
 <head>
-    <?php $this->RenderAsset('Head'); ?>
+    <?php $this->renderAsset('Head'); ?>
     <!-- Robots should not see the dashboard, but tell them not to index it just in case. -->
     <meta name="robots" content="noindex,nofollow"/>
 </head>
-<body id="<?php echo $BodyIdentifier; ?>" class="<?php echo $this->CssClass; ?>">
+<body id="<?php echo Gdn_Format::text($BodyIdentifier); ?>" class="<?php echo Gdn_Format::text($this->CssClass); ?>">
```
tburry: htmlspecialchars

library/core/class.smarty.php
```diff
@@ -34,7 +34,7 @@ public function init($Path, $Controller) {
 
         // Get an ID for the body.
         $BodyIdentifier = strtolower($Controller->ApplicationFolder.'_'.$ControllerName.'_'.Gdn_Format::alphaNumeric(strtolower($Controller->RequestMethod)));
-        $Smarty->assign('BodyID', $BodyIdentifier);
+        $Smarty->assign('BodyID', Gdn_Format::text($BodyIdentifier));
```
tburry: Aren't you double formatting this?

linc: I don't think so; I only format in the PHP-based admin.master.php which doesn't go thru smarty.

applications/dashboard/controllers/class.usercontroller.php
linc: Problem with that is you can still hit `user/tagsearch` endpoint directly and end up executing Javascript directly.

applications/dashboard/controllers/class.usercontroller.php
```diff
@@ -1071,14 +1072,18 @@ public function sso($UserID = false) {
...
+     * @param string $query
      * @param int $limit
      */
-    public function tagSearch($q, $limit = 10) {
-        $Data = Gdn::userModel()->tagSearch($q, $limit);
-        die(json_encode($Data));
+    public function tagSearch($query, $limit = 10) {
+        $data = Gdn::userModel()->tagSearch($query, $limit);
+        foreach ($data as &$user) {
+            // Sanitize usernames.
```
mtschirs: header('Content-type: application/json');  should be the right way to do this. Htmlescape is not needed

linc: Good call, thanks.

PR #2974, opened 2015-08-15
Rework cookie expiration check

linc:
* `array_slice` was used wrong in `getCookiePayload` (it had an inverted offset so it returned only last element instead of all remaining payload).
* `checkCookie` call in `getCookiePayload` is redundant and would then create loop if used from `checkCookie`.
* `checkCookie` should call `getCookiePayload` to get secure expiration.

### DIFF
--- a/library/core/class.cookieidentity.php
+++ b/library/core/class.cookieidentity.php
@@ -103,7 +103,7 @@ class Gdn_CookieIdentity {
             return 0;
         }
 
-        list($UserID, $Expiration) = $this->getCookiePayload($this->CookieName);
+        list($UserID) = self::getCookiePayload($this->CookieName);
 
         if (!is_numeric($UserID) || $UserID < -2) { // allow for handshake special id
             return 0;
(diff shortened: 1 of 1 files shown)


tburry:
Your analysis of array_slice seems incorrect. The cookie payload is everything *from* index 4, not the 4 last elements.

linc:
Index 4 is the timestamp of the expiration date. I don't follow why payload = a timestamp and discard the rest of the cookie data.

linc:
It occurs to me that maybe we should timeout if the key expiration is passed OR the cookie expiration is passed.

tburry:
The intent of the cookie is to have indexes 0-3 be used for security stuff and then the rest to be an optional payload. As it turns out we don't use that payload, but using -4 is still incorrect.

The key expiration is the correct one to use as it's the one that is secured and unspoofable.

mtschirs:
A suggestion: Is it really necessary to offer optional spoofable payload? Computing a HMAC for all the payload/cookie data instead of just a small part of it has no performance penalty.

tburry:
We have another cookie algorithm that I want to get integrated at some point and it will do what you suggest @mtschirs. Right now I just want this PR to just deal with the issue of expiry. Great suggestion.

linc:
Reverted the slice change.

PR #2980, opened 2015-08-17
Application manager: Add correct version to application assets.

beckyvb:
Fixes bug where application assets were being appended with the default APPLICATION_VERSION in the head.

### DIFF
--- a/library/core/class.applicationmanager.php
+++ b/library/core/class.applicationmanager.php
@@ -311,7 +311,7 @@ class Gdn_ApplicationManager {
                 )
             );
         }
-
+        
         // 2. Disable it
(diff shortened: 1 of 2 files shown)


library/core/class.applicationmanager.php
```diff
@@ -118,6 +118,7 @@ public function checkApplication($applicationName) {
      * @return bool|mixed Returns the application's info, a specific value, or false if the application cannot be found.
      */
     public function getApplicationInfo($applicationName, $key = null) {
+        $applicationName = ucfirst($applicationName);
```
tburry: I think this is too risky. Maybe look where the version is being added and add the ucfirst before this call instead.

PR #2984, opened 2015-08-18
Rework Gdn_Format::bbcode()

tburry:
* Remove the deprecated \e modifier from preg_replace().
* Fix newline parsing.
* Remove the .PreContainer from the pre tags.
* Do a bit of trimming around pre tags to remove extra newlines.


bleistivt:
This finally fixes the extra newlines that appear when you post HTML tables, awesome!

PR #2988, opened 2015-08-19
Speed up user deletes on sites with a lot of categories

tburry:
Deleting user content on sites with many categories is prohibitively slow.
It turns out that this is because `CategoryModel->setRecentPost()` rebuilds the entire category cache each time because of issues in `CategoryModel->setField()`.

This fix doesn’t address that concern because the code will take some work to untangle. Rather, this code instead makes sure that only the categories that the user actually has the most recent post for are recalculated. Before all categories with no posts were also being included.

There is also an issue with the DateLastComment being set improperly for discussions with no comments that was also fixed.

### DIFF
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -39,109 +39,108 @@ class VanillaHooks implements Gdn_IPlugin {
      *
      * @since 2.1
      *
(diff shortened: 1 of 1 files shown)


linc:
Heartless banisher of our glorious `consolidateArrayValuesByKey()`.

linc:
(Confirmed working with caching enabled)

Issue #2707, opened 2015-04-29
SSO disables /profile/edit which blocks access to still-necessary fields


LauraMakesArt:
When SSO is enabled, its not quite giving users the functionality they need to give their users access to the right things. Turning on sso gets rid of any links to "/profile/edit" Which is fine, but that page also contains location/website/twitter information, plus any other fields that maybe added with profile extender.  

The current workaround for this is to set Garden.Profile.Locations(or Titles) = TRUE,  what if they don't want "Locations" specifically as a field? What if they want to edit email/gender without adding an extra field?

PR #2998, opened 2015-08-22
Method to allow Edit Profile with SSO enabled

valeryan:
Enabling SSO and setting registration method to connect will disable the edit profile link as mentioned in #2707. Although enabling locations or titles will enable the edit profile link, I don't think the edit link should get disabled because of things like the profile extension plugin, setting gender, etc. 

This PR serves as an possible example of what I think might work. 

1. Commented out the section of the core render function that removes the edit link.
2. Change the can edit username and email logic to be disable if registration method is "connect".
3. Set a message that explains why both are disabled.
4. Added logic to display message in the edit profile view.

### DIFF
--- a/applications/dashboard/views/profile/edit.php
+++ b/applications/dashboard/views/profile/edit.php
@@ -6,6 +6,13 @@
     echo $this->Form->errors();
     ?>
     <ul>
+        <?php
+        if ($this->data('_WarnConnect')) {
+            echo '<li><div class="Gloss">',
+                t('Editing Username and Email is disabled.', 'Single Sign On is managing users. Editing your Username and Email is disabled.'),
+                '</div></li>';
+        }
+        ?>
         <li class="User-Name">
             <?php
             echo $this->Form->label('Username', 'Name');
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -559,16 +559,7 @@ if (!function_exists('hasEditProfile')) {
         if ($userID != Gdn::session()->UserID) {
             return false;
         }
-
-        $result = checkPermission('Garden.Profiles.Edit') && c('Garden.UserAccount.AllowEdit');
-
-        // $result &= (
-        //     C('Garden.Profile.Titles') ||
-        //     C('Garden.Profile.Locations', false) ||
-        //     C('Garden.Registration.Method') != 'Connect'
-        // );
-
-        return $result;
+        return checkPermission('Garden.Profiles.Edit') && c('Garden.UserAccount.AllowEdit');
     }
 }
 
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -562,11 +562,11 @@ if (!function_exists('hasEditProfile')) {
 
         $result = checkPermission('Garden.Profiles.Edit') && c('Garden.UserAccount.AllowEdit');
 
-        $result &= (
-            C('Garden.Profile.Titles') ||
-            C('Garden.Profile.Locations', false) ||
-            C('Garden.Registration.Method') != 'Connect'
-        );
+        // $result &= (
+        //     C('Garden.Profile.Titles') ||
+        //     C('Garden.Profile.Locations', false) ||
+        //     C('Garden.Registration.Method') != 'Connect'
+        // );
 
         return $result;
     }
--- a/applications/dashboard/controllers/class.profilecontroller.php
+++ b/applications/dashboard/controllers/class.profilecontroller.php
@@ -357,15 +357,21 @@ class ProfileController extends Gdn_Controller {
         $User = Gdn::userModel()->getID($UserID, DATASET_TYPE_ARRAY);
         $this->Form->setModel(Gdn::userModel());
         $this->Form->setData($User);
-        $this->setData('User', $User);
 
-        // Decide if they have ability to edit the username
-        $CanEditUsername = (bool)c("Garden.Profile.EditUsernames") || Gdn::session()->checkPermission('Garden.Users.Edit');
+        // only allow editing if registration method is not connect
+        if (c('Garden.Registration.Method') != 'Connect') {
+            // Decide if they have ability to edit the username
+            $CanEditUsername = (bool)c("Garden.Profile.EditUsernames") ||Gdn::session()->checkPermission('Garden.Users.Edit');
+            // Decide if they have ability to edit the email
+            $EmailEnabled = (bool)c('Garden.Profile.EditEmails', true) && !UserModel::noEmail();
+            $CanEditEmail = ($EmailEnabled && $UserID == Gdn::session()->UserID) || checkPermission('Garden.Users.Edit');
+        } else {
+            $CanEditUsername = false;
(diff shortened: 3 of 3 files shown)


valeryan:
Hey Lincoln, 

Does the awaiting feedback label reference internal feedback or are you wanting some extra feedback from me?

linc:
For a PR, it means it's waiting on someone to review and provide feedback. If there's an action waiting on the author, we tag it 'Revisit'.

valeryan:
Thanks for the clarification.

library/core/functions.render.php
```diff
@@ -560,15 +560,7 @@ function hasEditProfile($userID) {
...
-        $result = checkPermission('Garden.Profiles.Edit') && C('Garden.UserAccount.AllowEdit');
-
-        $result &= (
-            C('Garden.Profile.Titles') ||
-            C('Garden.Profile.Locations', false) ||
-            C('Garden.Registration.Method') != 'Connect'
-        );
-
-        return $result;
+        return checkPermission('Garden.Profiles.Edit') && C('Garden.UserAccount.AllowEdit');
```
valeryan: My thinking here is that this check is only in place to allow a admin to turn profile editing back on using Locations or Titles. This check is not needed you are not disabling the edit profile functionality but rather only disabling the fields that will be managed by the single sign on.

applications/dashboard/views/profile/edit.php
```diff
@@ -6,6 +6,13 @@
     echo $this->Form->errors();
     ?>
     <ul>
+        <?php
```
valeryan: This message will be seen by everyone including Admins. No one will be able to edit the username or email field when SSO is enabled. However, Admins can still edit users through dashboard. I think this is a lot less confusing to admins and users.

valeryan:
I added a few line notes to this to clarify my thoughts. I did not really do this PR to be merged per say but rather to illustrate another possible method to handle using SSO for registration.  I am sure this can be clean up a bit.

linc:
I don't see why auto-disabling the ability to edit emails and names is better than allowing it to be done using a config setting. There are many scenarios where you want users to be able to change those things after SSO, and this blocks that ability.

I agree the conditions in `hasEditProfile()` are hacky at best, but it needs some more careful thinking to refactor that kludge rather than just deleting it. That's going to cause expected behavior for folks currently using that logic.

valeryan:
@linc The problem I face, is that most of the information on this edit profile page is very useful for my users. However, we use SSO to manage the username and email exclusively on our forums. The users get confused because they change their username on the forum but it gets overwritten by SSO. My moderators get overrun by support request asking why usernames keep getting overwritten and we have to direct the users to our post that explains that any username changes have to happen on the main site, and that they will then sync to the forum on their next login. 

So the best case scenario for our usage, is to have the edit profile page available, but have the fields that are managed through SSO disabled. 

I am willing to help develop a solution if you have any ideas of what would work for your purposes and mine...

An idea I have is maybe to add a config item like `$Configuration['Garden']['ProfileEditableWithSSO'] = true; ` and have that override the current behavior and apply my behavior. When this flag is set the edit profile link is shown, the full edit profile page is available, but the settings that are managed by SSO are disabled. When it is not present or set to false the current expected behavior is applied. 

Would you be interested in looking at this solution? Should I do it under this PR, or create a new one?

linc:
Why don't you just disable editing usernames and emails?

     Garden.Profile.EditUsernames = false
     Garden.Profile.EditEmails = false

valeryan:
I believe with the registration method set to connect, the edit profile page disappears. Then you have to set some location setting to  enabled to get it back and then disable these two fields. This will accomplish most of the behavior through series of settings, but still gives no context to the user why these fields are being disable. 

My Thought is to make a single setting that would setup a set of behaviors that works well with SSO and explains to the users what is going on. 

To accomplish what I am trying to do currently, I would have to take these steps I think:

Set registration to "Connect" in admin dashboard or in config. (disables edit profile page all together)
Set `Garden.Profile.Locations = true` (adds back edit profile page)
Set `Garden.Profile.EditUsernames = false` and `Garden.Profile.EditEmails = false`
Then override the edit profile view in my theme to set a custom message that explains to users that SSO is managing the username and email. 

It seems to me that the behavior I am trying to create is achieved here through this multistep process, but I think it could be narrowed to just one config item to make it easier on admins. 

I would think that in fact if this type of logic is added, that the config item could also be set through the jsconnet plugin. 

I think that putting the hooks to do this in place in the vanilla app would allow any SSO plugin to set that setting have enable this SSO specific behavior.

valeryan:
I guess because this PR is closed it doesn't show updates to my branch. @linc please take a moment and look at https://github.com/valeryan/vanilla/commit/ba6cba01aa8226ea8f8002f27e061e53a0f675fe I retooled this to be less intrusive as I discussed, but I don't want to clutter up you repo with another PR unless you guys are interested in exploring this...

In my opinion this will create a behavior that is what I would expect when using an SSO system and allow plugins like jsconnect to set `Garden.Profile.SSOManaged` as part of its settings and get this alternate behavior with only one config change. 

If you do want your team to evaluate it, let me know and I will do a new PR if needed but I assume that if you reopen this PR it would sync with my repo.

500knots:
I am trying to do the same thing. I am using SSO but need to allow users to edit their own profile information (specifically fields from ProfileExtender). Is there an update to this issue and a fix?

valeryan:
No, this PR was closed and the vanilla team probably doesn't view this as an important issue. You can look at my code. This is what I use on our forum and it works. valeryan@ba6cba0 Its cleaner than what was done in this PR and you can toggle the behavior with a config item.

500knots:
Thanks for the info. What settings do I need to edit in admin for the user permissions and/or in theme?

valeryan:
If you have applied my changes to your core vanilla file just remember anytime you update this will need to be redone. To active it go into your config and add `
$Configuration['Garden']['Profile']['SSOManaged'] = true;` and you will see the new behavior.

500knots:
Thanks. It works but only when I have the user setting as Profiles > Edit and Users > Edit. The problem now is that all users can edit each other's profiles. The edit button shows up on all profiles, not just the user...

Any ideas?

valeryan:
Yeah, I don't know I will have to look an see if it does that on mine and
get back with you.

500knots:
I hope not for the sake of your users. You wouldn't want them to be able to edit each other's profiles.

valeryan:
An admin will always see the edit button. I logged in as a normal user and the button is not there. You scared me...

500knots:
It is the opposite for me. I am logged in as a 'member' and can see edit button in all user profiles.

valeryan:
Was the user ever an admin? I find that its really hard to un admin a user...

500knots:
I checked with a couple different users who were never admins. 

What is the logic to block edit profile once SSO is being used and/or registration method is changed to Connect? Is this done in the SSO plugin or in the core code?

500knots:
Apparently it is a theme issue. I switched to basic theme and it works fine.

PR #2999, opened 2015-08-23
Add text alignment classes to Dashboard

austins:


linc:
Please format PR titles like a user-facing changelog sentence. It's also helpful if you elaborate on why you're making a change. In this case I'm just gonna take it on faith this is making something easier for you in your app development but no one else on the team has enough background to guess that.

PR #3004, opened 2015-08-26
Path typo which broke the oauth redirect.

btafoya:
The path was preceded by a forward slash, breaking the route on redirect.


btafoya:
Same applies for the twitter oauth path. :)

btafoya:
Adding Twitter also...

Issue #2906, opened 2015-07-25
resources/fonts/vanillicon/index.html -> update stylesheet path


mtschirs:
In **resources/fonts/vanillicon/index.html** the stylesheet path changed. It schould now be `<link type="text/css" rel="stylesheet" href="../../design/vanillicon.css">`.

linc:
This was fixed.

PR #3007, opened 2015-08-26
Fix font path in vanillicon font preview page.

beckyvb:


mtschirs:
https://github.com/vanilla/vanilla/issues/2906

tburry:
This all gets built in the [font-vanillicon](https://github.com/vanilla/font-vanillicon) repo and copied here so you'll need to make the fix there.

PR #3008, opened 2015-08-26
Add X_PATH_INFO fallback to deal with naughty hosts

### LINKED ISSUE #3002: .htaccess issue in 2.2
Running on Apache/2.4.12, and root works, but all sub-directories do not.

Leads to "no input file specified" issue on discussions and comments.

Please see discussion:
https://github.com/vanilla/vanilla/commit/3b3d5d4ba0863a0431d8b102b3fd71543ab6ae17#commitcomment-12598537

linc:
Fixes #3002 probably. @adrianspeyer can you test this branch?

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -13,7 +13,7 @@
     # The basic rewrite rule.
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteCond %{REQUEST_FILENAME} !-d
-    RewriteRule ^(.*)$ index.php/$1 [QSA,E=X_REWRITE:1,L]
+    RewriteRule ^(.*)$ index.php/$1 [QSA,E=X_REWRITE:1,E=X_PATH_INFO:/$1,L]
 
     # Add the proper X_REWRITE server variable for rewritten requests.
     RewriteCond %{ENV:REDIRECT_X_REWRITE} .+
--- a/.htaccess
+++ b/.htaccess
@@ -14,6 +14,7 @@
 
     # The basic rewrite rule.
     RewriteCond %{REQUEST_FILENAME} !-f
+    RewriteCond %{REQUEST_FILENAME} !-d
     RewriteRule ^(.*)$ index.php [QSA,E=X_REWRITE:1,E=X_PATH_INFO:/$1,L]
 
     # Add the proper X_REWRITE server variable for rewritten requests.
--- a/.htaccess
+++ b/.htaccess
@@ -7,7 +7,7 @@
     # Certain hosts may require the following line.
     # If vanilla is in a subfolder then you need to specify it after the /.
     # (ex. You put Vanilla in /forum so change the next line to: RewriteBase /forum)
-    RewriteBase /
+    # RewriteBase /
 
     RewriteCond %{QUERY_STRING} ^p=/?([^&]+)(&([^?]+))?$
     RewriteRule ^index\.php %1?%3 [E=X_REWRITE:1,L]
--- a/.htaccess
+++ b/.htaccess
@@ -10,16 +10,16 @@
     RewriteBase /
 
     RewriteCond %{QUERY_STRING} ^p=/?([^&]+)(&([^?]+))?$
-    RewriteRule ^index\.php %1?%3 [E=X_REWRITE:2,L]
+    RewriteRule ^index\.php %1?%3 [E=X_REWRITE:1,L]
 
     # The basic rewrite rule.
     RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteRule ^(.*)$ index.php [QSA,E=X_REWRITE:2,E=X_PATH_INFO:/$1,L]
+    RewriteRule ^(.*)$ index.php [QSA,E=X_REWRITE:1,E=X_PATH_INFO:/$1,L]
 
     # Add the proper X_REWRITE server variable for rewritten requests.
     RewriteCond %{ENV:REDIRECT_X_REWRITE} .+
     RewriteCond %{ENV:REDIRECT_X_PATH_INFO} (.+)
-    RewriteRule ^index\.php - [QSA,E=X_REWRITE:2,E=!REDIRECT_X_REWRITE,E=X_PATH_INFO:%1,E=!REDIRECT_X_PATH_INFO,L]
+    RewriteRule ^index\.php - [QSA,E=X_REWRITE:1,E=!REDIRECT_X_REWRITE,E=X_PATH_INFO:%1,E=!REDIRECT_X_PATH_INFO,L]
 
     # 301 redirect urls that start with index.php
     #RewriteCond %{REQUEST_METHOD} GET [NC]
--- a/.htaccess
+++ b/.htaccess
@@ -1,23 +1,25 @@
 # Original
 # If you modify this file then change the above line to: # Modified
+
 <IfModule mod_rewrite.c>
     RewriteEngine On
+
     # Certain hosts may require the following line.
     # If vanilla is in a subfolder then you need to specify it after the /.
     # (ex. You put Vanilla in /forum so change the next line to: RewriteBase /forum)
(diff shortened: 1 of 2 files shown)


linc:
This is falling short because GoDaddy cannot handle the URL form of `index.php/anything` because it looks for a folder named `index.php` instead of the file. The workaround is to delete `/$1` from the end of the RewriteRule so that it can see index.php and use the fallback server variable we set. That however will break everyone doing the correct CGI `PATH_INFO` behavior. 

So, in order of preference, we need one of:

* A way to force naughty hosts into the correct behavior.
* A way to detect & give different rules to naughty hosts.
* Add instructions for dealing with naughty hosts to the setup instructions.

bleistivt:
I encountered a similar problem with the X_REWRITE variable not being passed correctly:
http://vanillaforums.org/discussion/comment/232821/#Comment_232821

linc:
Very interested to know more about that, and how widespread it might be.

bleistivt:
I just tried setting an environment variable with SetEnv (and nothing else), which also got the prefix.

I opened a ticket there, because this is definitely a bug on their end.

library/core/class.request.php
```diff
@@ -495,6 +495,12 @@ protected function _loadEnvironment() {
 
             if (!empty($_SERVER['X_REWRITE'])) {
```
tburry: I might have to acquiesce and just accept:

```php
if (!empty($_SERVER['X_REWRITE']) || !empty($_SERVER['REDIRECT_X_REWRITE']) ) {
```

Daazku:
I ended up pushing Todd fix because when you use apache with mod_fastcgi && php-fpm you will always endup with REDIRECT_X_REWRITE even if we have the following code in the .htaccess
```
    # Add the proper X_REWRITE server variable for rewritten requests.
    RewriteCond %{ENV:REDIRECT_X_REWRITE} .+
    RewriteRule ^index\.php - [QSA,E=X_REWRITE:1,E=!REDIRECT_X_REWRITE,L]
```

It is because apache modifies the header just before sending it to php-fpm. Not much we can do about it.

adrianspeyer:
This is the most up-to-date pull on this: https://github.com/vanilla/vanilla/pull/3338

tburry:
I rebased this off of master and am looking to get a yay or nay from @Daazku and @initvector.

Daazku:
Works for me!

adrianspeyer:
Still getting "No input file specified".

adrianspeyer:
I should add @initvector fix for Htaccess here: https://github.com/vanilla/vanilla/blob/2c4b2065f30496485752f8c40dafa877d871441d/.htaccess 

works fine.

Daazku:
Wouldn't it be simpler to just parse $_SERVER['REQUEST_URI'] ???

library/core/class.request.php
```diff
@@ -514,8 +514,18 @@ protected function _loadEnvironment() {
                 safeParseStr($get, $get, $original);
             }
 
-            if (!empty($_SERVER['X_REWRITE'])) {
+            if (!empty($_SERVER['X_REWRITE']) || !empty($_SERVER['REDIRECT_X_REWRITE'])) {
                 $path = $_SERVER['PATH_INFO'];
```
initvector: I acknowledge this isn't caused by this PR, *BUT*...I recommend this be changed to something like
```php
$path = $this->getValueFrom(self::INPUT_SERVER, 'PATH_INFO')
```
`PATH_INFO` isn't guaranteed to be set.  A notice is thrown from here under my GoDaddy account, because of an invalid array index.

tburry: Good idea. Add a commit to this PR.

.htaccess
```diff
@@ -1,23 +1,25 @@
...
     # RewriteBase /
 
-    RewriteCond %{QUERY_STRING} ^p=/?([^&]+)(&(.+))?$
-    RewriteRule ^index\.php index.php/%1?%3 [E=X_REWRITE:1,L]
+    RewriteCond %{QUERY_STRING} ^p=/?([^&]+)(&([^?]+))?$
+    RewriteRule ^index\.php %1?%3 [E=X_REWRITE:1,L]
 
     # The basic rewrite rule.
     RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteCond %{REQUEST_FILENAME} !-d
```
tburry: I want to make sure this line was knowingly removed and why it shouldn't be here. Can you let me know.

PS: I know it's my commit, but I copied it from another PR by I think @initvector.

Daazku: My guess is that it was removed because if a directory named categories is present on disk we it would prevent us from seeing http://www.example.com/categories

linc: @Daazku Which is incorrect, because then any other app in a subdirectory is blocked from reaching its index.php file while invoking its folder name.

Daazku: @Linc /categories would be parsed by vanilla but /categories/index.php would still be handled correctly. It's not that dramatic IMO.

Daazku: @Linc I did some more tests and here are the behaviour with that change:

http://www.example.com/categories will be handled by vanilla unless there is a file in that directory that match the ``DirectoryIndex`` directive.
In the end that change won't break anything unless mixed with other rewrite rules.

linc: `/something/index.php` is not how you'd access that file; you'd type `/something` because webservers should automatically look for the index.php file in the directory when you do this. However, removing this rule blocks that behavior and therefore breaks normal access to the app.

I know for a fact it breaks things because it did so on my webserver the _last_ time that change snuck into master branch.

Daazku:
```    
    # The basic rewrite rule.
    RewriteCond %{REQUEST_FILENAME} !-f
    RewriteRule ^(.*)$ index.php/$1 [QSA,E=X_REWRITE:1,L]
```

I have a folder named categories.

Empty folder:
http://vanilla.dev:8080/categories/ VANILLA

Folder with index.php
http://vanilla.dev:8080/categories/ INDEX.PHP

Folder with index2.php
http://vanilla.dev:8080/categories/ VANILLA

initvector: I believe that particular line was removed when we based our .htaccess off [this gist](https://gist.github.com/tburry/3aef5128b8a25b304f9e).  I'm not sure what the original logic behind removing it was, but I'm in favor of restoring it.

initvector:
While the .htaccess now seems to be working fine and dandy on my GoDaddy account, it might be killing the unit testing.  Travis-CI is reporting four of the following exception messages:

```
VanillaTests\APIv0\StandardTest
exception 'Exception' with message 'Maximum (10) redirects followed (302)' in /home/travis/build/vanilla/vanilla/tests/APIv0/APIv0.php:134
```

tburry:
@initvector remember to check the build. You should be fixing or reverting a commit that breaks the build ASAP.

initvector:
The error I cited yesterday causing the build to fail (`Maximum (10) redirects followed (302)`) seems to have somehow been related to me updating the $path assignment from

```php
$path = $_SERVER['PATH_INFO'];
```

to

```php
$path = $this->getValueFrom(self::INPUT_SERVER, 'PATH_INFO');
```

I updated the assignment to `$path = val('PATH_INFO', $_SERVER, '');` and now it seems to be fine.  This "too many redirects" behavior only seemed to occur on Travis/nginx.  I did not experience any issues with Apache in my local environment or on GoDaddy.

linc:
@initvector Is this ready to go, then?

initvector:
This is working A-OK for me in my GoDaddy environment.  However, as of yesterday, @adrianspeyer reported he might be having some issues with it and possibly some incompatibility with the Q&A plug-in.  He's going to knock it around a little more and let us know for sure.  If he gives us the all-clear, I think it's ready for primetime.

adrianspeyer:
Seems to be working now like a charm :+1:

PR #3013, opened 2015-08-28
Add event before notFoundException in /discussion/index

initvector:
If a discussion was not found with /discussion/(discussion ID), Vanilla would immediately move to throw notFoundException.  This update adds an event just prior to this, so that custom handling of "discussion not found" behavior can be made with plug-ins.


linc:
`DiscussionNotFound` would be better. We already have a generic `NotFound` event, so `BeforeNotFound` is a bit of a non-sequitur.

initvector:
Event name has been updated

PR #3019, opened 2015-08-30
Quotes: JS refactor

bleistivt:
* Remove livequery dependency.
* Remove the editor stack concept and insert quotes in the last TextBox that had focus instead.
* Pull the function definitions out of the constructor.
* Make quote folding strings translateable.
* Remove commented out code.
* Remove dead code.
* Added comments.
* Lord Brackos

Tested with: ButtonBar, editor.

### DIFF
--- a/plugins/Quotes/js/quotes.js
+++ b/plugins/Quotes/js/quotes.js
@@ -119,7 +119,15 @@ Gdn_Quotes.prototype.Quote = function(ObjectID, QuoteLink) {
         return;
     }
 
-    var ScrollY = this.GetEditor().offset().top - 100;
+    var ScrollY;
+
+    // DEPRECATED: cleditor support
+    if ($('div.cleditorMain').length) {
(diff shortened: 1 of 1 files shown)


linc:
In general, try to do code format refactors as a separate PR before changing all the functionality. It gets really hard to read the diff when you do both together. We'll muddle thru for this one. 

I need to check internally here to see if any clients are still stuck on cleditor + quotes before we merge this. Hopefully we can expedite them getting off of it.

bleistivt:
Yes, this commit became larger and larger as i moved on :laughing: 

I removed cleditor because it was in the archive section of vanilla/addons.
I could add it back in for the time it is still being used if you want.

linc:
If you want to add it back, that'd definitely clear us to merge this quicker. Just clearly document what we can rip back out once the coast is clear. I bet there are open source folks still using somewhere too.

Issue #1509, opened 2013-01-07
API keys should trim


linc:
Specifically, the Twitter Consumer Key does not trim spaces.

PR #3023, opened 2015-08-30
Trim user-entered API keys and config strings

linc:
* Trim incoming values to ConfigurationModule generally - I cannot think of a use case where you'd ever purposefully enter a space at the start or end of a config value, but it can seriously mess up folks when it's a copy/paste error.
* Trim Twitter & Facebook API keys entered by user (since they don't use ConfigurationModule yet).
* Fix some misc casening in config module.

See #1509


initvector:
It may be possible for certain passwords (MySQL?) to start or end with a space.  If all config values are being trimmed, this might cause those password values to be rendered invalid, even though they were entered correctly.

I admit this is an extremely rare case, but it's up to you if you want to support it. :shit:

tburry:
I like this, but will feel happier if you add an option in a ConfigurationModule row to opt-out of the trimming.

Issue #2957, opened 2015-08-10
Livequery removal


bleistivt:
I would like to get rid of livequery in vanilla.

It is an antipattern and encourages lazy code (subscribing to DOM mutations instead of handling the proper events).

Based on #1942 this seems to be in tune with what you want (stop me if it is not).

I'll be using this to track my progress:

- [x] Remove livequery uses from /js: #2954
- [x] Remove livequery uses from /plugins: #2955, #3019, #3021
- [x] Remove livequery uses from vanilla/addons https://github.com/vanilla/addons/pull/251, https://github.com/vanilla/addons/pull/252, https://github.com/vanilla/addons/pull/254, https://github.com/vanilla/addons/pull/255
- [x] Remove jquery.livequery.js from all controllers (and remove the library). #3025

linc:
It is indeed, thanks!

x00:
Have you considered all used cases?

Obviously there is delegation, which which would replace many uses.

There are some use case that require detection of inserted elements. You have DOMNodeInserted and Mutation Observer, but these are not universally applied across all browsers.

DOMNodeInserted is not that fast.

bleistivt:
DOMNodeInserted / Mutation Observers shouldn't really be needed either.

> There are some use case that require detection of inserted elements.

Imo, that means, vanilla should fire a js event there instead.

bleistivt:
Saving application state in the DOM is an outdated pratice.

Maybe at some point, you want to make use of pjax loading in vanilla or use a VDOM implementation that updates the DOM by diffing.

The entire concept of "this element was inserted means that..." breaks down there while an event based approach will still work.

The last PR didn't remove the library btw. so you could still include it if you need it.

x00:
Ok perfect. That makes sense.

PR #3025, opened 2015-08-30
Remove jquery.livequery.js from all controllers

bleistivt:
Blocked by all issues referenced here: #2957
I wouldn't recommend merging this anytime soon as third party plugin might rely on livequery. But this can be used to test if removing it breaks something.


linc:
Last blocker is #3019

linc:
Still have livequery calls (1 each) in cleditor (deprecated),  Google Prettify (core), Badges (internal), and Groups (internal).

bleistivt:
I can fix prettify and cleditor later when I'm home.

The addons app also uses it:
https://github.com/vanilla/addons/search?utf8=%E2%9C%93&q=livequery

A quick temporary fix for all of those would be to just add an addJsFile call in the plugins that need it.

PR #3033, opened 2015-08-31
addjsfile fails to find discussion.js on profile page

valeryan:
class.hooks is being called from dashboard and no default application is specified for the call to
disscussion.js so the application tries to find it in dashboard. adding vanilla as default location fixes
the issues.


linc:
Which now makes me wonder how long that's been broken...

PR #3035, opened 2015-08-31
fix call to undefined object in the Category View

valeryan:
Adds an if statment that checks if Category->Depth is set. This will prevent the undefined object notice.


linc:
I believe you simply want `$this->data('Category.Depth', 0)` instead.

valeryan:
Thanks, I guess I should have looked up the data() function and checked if it had a way to specify a default. All fixed up now.

linc:
Thanks!

PR #3038, opened 2015-09-01
Remove base64-encoding of JSON data

mtschirs:
JSON encoding fails for binary data (data that isn't an UTF-8 string or another primitive, e.g. raw JPEG data). Such data would have to be e.g. base64-encoded before it can be encoded as JSON.

Vanilla allows access to possibly binary values stored in Controller->Data via DeliveryType=DATA. However, in this case, Vanilla currently does not perform base64-encoding:
* Controller::renderData():
```
case DELIVERY_METHOD_JSON:
default:
    ...
    safeHeader('Content-Type: application/json; charset='.c('Garden.Charset', 'utf-8'), true);
    // This is a regular json request.
    echo json_encode($Data);
    return true;
```

Only if DeliveryType != DATA and DeliveryMethod = JSON, then Vanilla performs base64-encoding. However, in this case, base64-encoding is IMHO unneccessary since the to-be-encoded data is already a (UTF-8) string:
* Controller::xRender():
```
if ($this->_DeliveryMethod == DELIVERY_METHOD_JSON) {
    ...
    $this->setJson('Data', base64_encode(($View instanceof Gdn_IModule) ? $View->toString() : $View));
```

Now, the solution to this inconsistency could be to add base64-encoding to the first case and remove it from the second one. Alternatively (since everything seems to perform well currently with no base64-encoding in case of DeliveryType=DATA), never base64-encode anything.

This PR makes Vanilla faster (CPU), reduces bandwith (base64-bloat) and unifies & simplifies data handling.

* Remove base64-encoding / decoding of informMessageStack
* Remove base64-encoding / decoding of json-encoded Data

* Remove UTF8-encoding / decoding functions
* Remove base64-encoding / decoding functions

* Deprecate $.postParseJson()

@see https://github.com/vanilla/vanilla/pull/2949#issuecomment-136207828

Testing required; this PR is work in progress. Your input is appreciated.

### DIFF
--- a/applications/conversations/js/conversations.js
+++ b/applications/conversations/js/conversations.js
@@ -33,8 +33,6 @@ jQuery(document).ready(function($) {
                gdn.informError(xhr);
             },
             success: function(json) {
-               json = $.postParseJson(json);
-
                // Remove any old errors from the form
                $(frm).find('div.Errors').remove();
 
--- a/applications/dashboard/js/activity.js
+++ b/applications/dashboard/js/activity.js
@@ -85,8 +85,6 @@ jQuery(document).ready(function($) {
                 gdn.informError(xhr);
             },
             success: function(json) {
-                json = $.postParseJson(json);
-
                 gdn.inform(json);
 
                 // Remove any old errors from the form
@@ -133,8 +131,6 @@ jQuery(document).ready(function($) {
                     gdn.informError(xhr);
                 },
                 success: function(json) {
-                    json = $.postParseJson(json);
-
                     gdn.inform(json);
 
                     if (json['FormSaved'] == true) {
--- a/applications/dashboard/js/applications.js
+++ b/applications/dashboard/js/applications.js
@@ -15,8 +15,6 @@ jQuery(document).ready(function($) {
                 $(btn).after('<span class="Progress">&#160;</span>');
             },
             success: function(json) {
-                json = $.postParseJson(json);
-
                 if (json.FormSaved == true) {
                     gdn.inform(json);
                     if (json.RedirectUrl) {
--- a/applications/dashboard/js/import.js
+++ b/applications/dashboard/js/import.js
@@ -9,8 +9,6 @@ jQuery(document).ready(function($) {
             url: url,
             dataType: 'json',
             success: function(json) {
-                json = $.postParseJson(json);
-
                 // Refresh the view.
                 $('#Content').html(json.Data);
                 bindAjaxForm();
@@ -35,8 +33,6 @@ jQuery(document).ready(function($) {
         $('form').ajaxForm({
             dataType: 'json',
             success: function(json) {
-                json = $.postParseJson(json);
-
                 $('#Content').html(json.Data);
                 bindAjaxForm();
 
--- a/applications/dashboard/js/profile.js
+++ b/applications/dashboard/js/profile.js
@@ -29,7 +29,6 @@ jQuery(document).ready(function($) {
                 $.popup({}, XMLHttpRequest.responseText);
             },
             success: function(json) {
-                json = $.postParseJson(json);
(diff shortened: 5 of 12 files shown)


linc:
@tburry

tburry:
Just for a bit of context. We originally had to add that base64 encoding stuff because we were getting a lot of issues reported regarding encoding issues. I think this had to do with the fact that some of the endpoints actually send back an entire view and not just data and that view can have screwy encodings. Basically, json_encode() is very unforgiving and can completely fail with not much to go on.

All that being said, this may be the right time for a change like this. It may be the case that more current versions of PHP don't have these issues. To be honest, finding a crappy string is kind of tough though. That will be needed to make sure this is okay.

My guess is that this kind of PR will take a bit of thinking on, but I'd love to see if we can make this happen. Thanks so much for scalpel work.

mtschirs:
Making Vanilla UTF-8-only would already reduce the testing complexity: https://github.com/vanilla/vanilla/issues/3047

tburry:
The odd character didn't come from other encodings, but rather weird characters that were part of imports or submitted by browsers. Again, this might be something that modern PHP handles, but we had the issues crop up quite frequently 4+ years ago.

I'm all for removing Garden.Charset tho.

mtschirs:
As far as I know, incoming data is - apart from an incoming SSOString in user/save - never base64 encoded/decoded. The only decoding/encoding this PR removes is base64 encoding on the server side and base64 decoding in the client browser.

So we can already exclude odd characters submitted by browsers.

If there are non-utf-8 characters send as a part of a view from the server to the client via external imports / whatever, then this is clearly a bug in itself that needs to be fixed serverside (since all output to the client should be UTF-8 / Garden.Charset encoded).

linc:
It seems very likely to me that we were missing proper encoding headers in some places, and/or we had databases that were imported incorrectly from, say, vBulletin that uses Latin-9. There was all sorts of ugly like that in the early days.

linc:
My suggestion is we send this to staging for a bit when it's ready and let it chill there for a couple weeks so we have time to collect support issues from clients and do some more QA.

tburry:
Sometimes data is just bad, usually before we even see it. We've had cases where the encoding was botched before us. We've had cases where old international browsers submit garbage data. But it doesn't matter because that data makes it into the database. It displays fine and works with a bit of wonkiness in views and whatnot, but it won't JSON encode because json_encode() is more picky than other functions.

Consider the following:

```php
$x = "\xF0\xA4\xAD"."\xF0\xA4\xAD\xA2"."\xF0\xA4\xAD\xA2";
$json = json_encode($x);
```

In this case $json will be false. The solution could be just specifying JSON_PARTIAL_OUTPUT_ON_ERROR in our flags, but I'm just not sure. And the bug reports that come in if we make a mistake here are really tedious.

What we really need is a string that has invalid UTF-8 in it that otherwise displays okay. The invalid UTF-8 should be bookended with valid ASCII. If this string works fine before this PR, but borks after then there is more work to do.

tburry:
There is a great [post](http://stackoverflow.com/questions/3715264/how-to-handle-user-input-of-invalid-utf-8-characters) about some strategies in dealing with invalid UTF-8. As you can see though, the road ahead is not a straightforward one and requires a very comprehensive solution before we can remove the base64 encoding.

mtschirs:
I just want to point out that there is already no base64-encoding in case of DeliveryType=DATA. So do you experience the aforementioned problems with your Vanilla API?

tburry:
We probably would, but the people that use the API are a small subset of everyone that uses our software. I do want to get this to work though. I may be overstating the problem based on some PSTD over having to debug these issues several years ago.

Mark my word though. We will get this in.

tburry:
Closing in favor of #3455.

PR #3040, opened 2015-09-01
Escape Smarty version of username for output

linc:


tburry:
There is an escape filter in smarty.

PR #3045, opened 2015-09-01
Remove calls to viewLocation() from Gdn_Controller

mtschirs:
Gdn_Controller::fetchViewLocation() makes a call to viewLocation() without using its return value (only for creating a trace() in case its return value diverges from the view location it determines on its own).

Each call to viewLocation() results in one or more calls to file_exist(). On a test installation, a normal request resulted in about 75 calls to file_exist() with unique path argument which translated into a delay of about max. 10 milliseconds.

This PR makes Vanilla faster and cleans-up the codebase. You need to check for references to viewLocation() in your proprietary code.

* Removed calls to viewLocation() from Gdn_Controller::fetchViewLocation()
and Gdn_Controller::renderMaster().

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -772,8 +772,6 @@ class Gdn_Controller extends Gdn_Pluggable {
             $View .= '_rss';
         }
 
-        $ViewPath2 = viewLocation($View, $ControllerName, $ApplicationFolder);
-
         $LocationName = concatSep('/', strtolower($ApplicationFolder), $ControllerName, $View);
         $ViewPath = arrayValue($LocationName, $this->_ViewLocations, false);
         if ($ViewPath === false) {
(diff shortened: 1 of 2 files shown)


library/core/functions.general.php
```diff
@@ -3771,8 +3771,10 @@ function url($Path = '', $WithDomain = false, $RemoveSyndication = false) {
      * @param string $Controller The name of the controller invoking the view or blank.
      * @param string $Folder The application folder or plugins/plugin folder.
      * @return string|false The path to the view or false if it wasn't found.
+     * @deprecated
      */
     function viewLocation($View, $Controller, $Folder) {
+    	deprecated('viewLocation()');
```
linc: spacing

library/core/class.controller.php
```diff
@@ -851,10 +849,6 @@ public function fetchViewLocation($View = '', $ControllerName = false, $Applicat
 //         trigger_error(ErrorMessage("Could not find a '$View' view for the '$ControllerName' controller in the '$ApplicationFolder' application.", $this->ClassName, 'FetchViewLocation'), E_USER_ERROR);
         }
 
-        if ($ViewPath2 != $ViewPath) {
-            Trace("View paths do not match: $ViewPath != $ViewPath2", TRACE_WARNING);
```
linc: casing

linc:
This code was added as part of a refactor to verify we hadn't altered functionality when moving away from `viewLocation()`. It's probably safe to remove now.

mtschirs:
Fixed spacing

PR #3049, opened 2015-09-02
Remove Minify

mtschirs:
* Remove UtilityController::css()
* Remove AssetModel::serveCss()
* Remove library/vendor/Minify

In 2012, Minify has been removed as a plugin and added to the core via https://github.com/vanilla/vanilla/commit/a68c975de5b74e20d10e2ea6fd50648c76e31549

Minify is outdated and has some issues, e.g. wrong paths in Minify/groupsConfig.php:
```
// Look for plugins in plugins/Minify/min/index.php to figure out where the web root is.
$Key = array_search('plugins', $WebRoot);
```
As far as I can tell, the dashboard/utility/css endpoint is unused. Since this is the only reference to Minify, I suggest to remove it. This PR does exactly that.


linc:
Would like @kaecyra to sign off on this since I'm unsure where our asset consolidation goals are headed.

linc:
I followed up with @tburry who says we have plans to pursue the use of Minify in this capacity again soon, including the use of `utility/css`.

mtschirs:
Even if you plan on re-activating Minify in the future, its endpoint should probably not be in the dashboard application while Minify is part of the core library.

linc:
@mtschirs Agreed & understood.

When #2676 is merged, this would be a prime candidate for refactoring as a Composer package since it's code that's not currently in production.

PR #3052, opened 2015-09-03
Fix jquery.popup.js; make $.popup respect targetUrl setting.

mtschirs:
Fixes http://vanillaforums.org/discussion/25089/opening-popup-programatically


linc:
Nice refactoring and enhancement, thanks

PR #3059, opened 2015-09-08
Clear up JSHint warnings in global.js

bleistivt:
Checked with JSHint. There are still a few warnings.

This fixes mostly missing semicolons and undefined variables, strict condition checks where possible and a leaky global smiliar to #3051.

One functional change that comes with that:
`null` can now be used as a default fallback in `gdn.definiton`.
This is actually fixes a check (line 810) that would always return true:

    if (gdn.definition('TickExtra', null) != null)


linc:
Wow! Thanks for the Javascript love.

PR #3062, opened 2015-09-08
Remove bad utility methods that had been commented out

tburry:
These methods are security risks and there is no reason to keep them.


mtschirs:
I already thought of sending a PR doing exactly the same; good idea. Leaving them in the codebase might inspire other developers to copy parts or the method as a whole which is bad.

PR #3063, opened 2015-09-08
Clean up the translations on the getting started plugin

tburry:
The getting started plugin had a lot of sloppy calls to t().
This PR addresses them.

- Give long translations a pithy code.
- Combine concatenated translations into one string using formatString() for links.
- Remove references to “left” or “right” so that RTL locales won’t be confusing.


linc:
Looks good.

PR #3064, opened 2015-09-08
Fix the path to the default avatar.

beckyvb:
Saves the path to the uploaded cloud file avatar in config.


linc:
Can I ask about the disappearing `trim()`? Not sure what's going on there.

beckyvb:
The str_replace is doing the same thing as trim was. The second param of trim applies a character mask to the subject, which was probably never the right solution for this. Running trim($basename, '/uploads') would return something like 'efaultavatar/filename.jpg'

PR #3072, opened 2015-09-10
Remove json encoding from Gdn_Controller->_Definitions['ResolvedArgs']

mtschirs:
The definition list is already json encoded. This double encoding (and
decoding) serves no purpose.


tburry:
Give me a bit to make a couple of plugins compatible with this. Then I'll merge.

Issue #3057, opened 2015-09-07
What is jQuery UI still used for?


bleistivt:
I'm asking because with ~300kB is size it makes up for the largest part in vanillas JS load and an additional HTTP request on the first load.

I know that it is used for the yellow fade/highlight effect in various places, but other than that, there seem to be no dependencies.

The effect could easily be replaced with a CSS animation which would also improve themeability.
see https://github.com/bleistivt/InfiniteScroll/blob/master/design/infinitescroll.css#L120-L128

I've actually been running my forum without jquery-ui for a month now without problems (using a dummy $.fn.effect which just applies a CSS class).

mtschirs:
There are references to the datepicker defined by jquery-ui in **dashboard/design/style.css** as well as **js/global.js**:
```
// Convert date fields to datepickers
if ($.fn.datepicker) {
		$('input.DatePicker').datepicker({
				showOn: "focus",
				dateFormat: 'mm/dd/yy'
		});
}
```
However, e.g. Gdn_Form::calendar() simply prints a text input with class="DateBox" and 
Gdn_Form::date() prints some unrelated dropdowns. So at least the datepicker seems not to be used by any core component or application.

I would be happy if it could be removed from the core.

bleistivt:
This is the "replacement" I am using for $.fn.effect
```javascript
$.fn.effect = function() {
    var that = this;
    return this
        .addClass('highlight')
        .on('animationend webkitAnimationEnd', function () {
            that.removeClass('highlight');
        });
};
```

I some part of jQuery UI is really needed, a special build which just includes that could also be used.
http://jqueryui.com/download/

linc:
nestedSortable requires "jQuery UI Sortable 1.10+". 

I'm all for finding a reasonable way to minimize how much of it we're using, and generally obsoleting as much of our jQuery use as possible.

mtschirs:
The following applications include nestedSortable:
* Articles - class.hooks.php (superfluous import, it doesn't seem to need it).
```
        $Sender->AddJsFile('js/library/nestedSortable.1.3.4/jquery-ui-1.8.11.custom.min.js');
        $Sender->AddJsFile('js/library/nestedSortable.1.3.4/jquery.ui.nestedSortable.js');
```
* BasicPages - class.pagesettingscontroller.php
```
        $this->AddJsFile('nestedSortable.1.3.4/jquery-ui-1.8.11.custom.min.js');
[...]

PR #3079, opened 2015-09-13
Remove jQuery UI from all controllers

bleistivt:
see #3057

This replaces jQuery UIs .effect() with a shim using CSS animations and
removes the library from all controllers.


kaecyra:
You're not seeing issues after making this change yourself? Testing reveals no weird console errors?

bleistivt:
Yes, I've been running my forum for more than a month with this now.

Apart from the highlight `.effect`, everything that needs jQuery UI imports its own custom build anyway, like nestedSortable or the editors upload bar widget.

bleistivt:
Definitely requires testing in case you might use it internally somewhere, but I believe the bandwith (and on low-end devices likely parsing speed) savings are worth it.
ss 2015-09-18 at 06 04 54

linc:
Cleaned up the merge with #3315

Issue #2363, opened 2014-12-24
System User email address raised by Anonymoose.


peregrine-web:
see discussion

http://vanillaforums.org/discussion/28826/system-user-account

anonymoose comment: 
System user account has email of system@domain.com which is a valid email, since domain.com is a valid domain. Potentially bad for resetting the password of the "System" account.

Anonymoosed:
invalid.com is not a reserved domain, see http://invalid.com

SchlesserClaude:
example.org and example.com at domains especially reserverd for that purpose by the Internet Assigned Numbers Authority: http://www.iana.org/domains/reserved

hgtonight:
@SchlesserClaude from your link: http://tools.ietf.org/html/rfc6761 section 6.4 recognizes `invalid.` domains as special.

@Anonymoosed people not following standards is not a reason to not follow them.

Anonymoosed:
@hgtonight That's not someone not following standards, that's a valid domain. 

You were thinking of this: http://domain.invalid/ is invalid. http://invalid.com/ is a valid domain.

PR #3080, opened 2015-09-13
Use system@example.com for system user

bleistivt:
see #2363 and http://www.iana.org/domains/reserved

domain.com is a registered domain, while example.com can not be registered.


tburry:
Nice one!

PR #3081, opened 2015-09-13
Make HTML in categories description consistent

### LINKED ISSUE #2435: HTML in category description - inconsistent behaviour
Category descriptions allow raw HTML, but tags are stripped in the /category view while in the /categories view they are not.

This Issue was generated from your [forums] 
[forums]: http://vanillaforums.org/discussion/comment/224605#Comment_224605

bleistivt:
Since HTML is already allowed in categories/all, descriptions should not be rendered as plain text inside of a category.

fixes #2435


tburry:
Your fix would make the description in the <head> tag also contain HTML which is not valid.

PR #3087, opened 2015-09-15
Make HTML in categories description consistent

bleistivt:
new PR for #2435


linc:
@beckyvb What's the theme implications here for clients? Is this a relatively safe change?

beckyvb:
This seems like a safe change, and it's completely sane to want to standardize the way we output the category description between the two views. While we're at it though, I'd recommend calling Gdn_Format::htmlFilter on the category description data before outputting.

linc:
@bleistivt up for adding a quick `htmlFilter()`? Thanks!

bleistivt:
Done!

PR #3089, opened 2015-09-15
Change the default format to Markdown

tburry:
This changes the default input formatter to Markdown. Any forums installed before this change will keep their old default so this only affects new communities moving forward.


linc:
I was under the impressions that `Garden.InputFormatter` would only be saved to the `config.php` IF you were using Editor AND you'd saved your settings for it (changed the default).

tburry:
There was a change implemented a long time ago to make way for this. Look [here](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/settings/structure.php#L759).

PR #3091, opened 2015-09-15
Show only the categories for which the user has “Add” permissions in drop down on “Add Discussion”

nervoustwit:


nervoustwit:
In Form::categoryDropDown() extract the permissions from the options passed to determine which categories to show. In /post/discussion we pass the "Add" permission, elsewhere we use the default "View" permission.

linc:
What's the behavior now if you cannot Add in the parent category but can in the child? Does that break the tree in the dropdown?

The current behavior should be that categories you lack the Add permission for should appear as part of said tree but `disabled` and unselectable.

nervoustwit:
I assumed that if you couldn't add into the parent, you couldn't add into the child. I will disable the link.

nervoustwit:
In the categoryDropDown function class.form.php in we loop through all the categories and create a variable to disable a category if the user does not have permissions. When this value is passed by reference it changes the first iteration to disabled and all the other iterations are not disabled. This is fixed by removing the reference and assigning the value directly.

library/core/class.form.php
```diff
@@ -368,7 +368,7 @@ public function categoryDropDown($FieldName = 'CategoryID', $Options = false) {
                 }
 
                 if ($Category['AllowDiscussions']) {
-                    $Disabled &= $Permission == 'add' && !$Category['PermsDiscussionsAdd'];
+                    $Disabled = $Permission == 'add' && !$Category['PermsDiscussionsAdd'];
```
tburry: This should probably be |=, but at this point we should just simplify this with another if statement.

```php
if ($Permission == 'add' && !$Category['PermsDiscussionsAdd']) {
    $Disabled = true;
}
```

You want the row that is disabled for some other reason to stay disabled.

tburry:
> I don't see why we are passing disabled by reference. When you do it this way you disable any links to categories where there is not permission to add. Maybe I am not seeing something.

Just so you know @nervoustwit, `=&` is a reference while `&=` means AND equals.

library/core/class.form.php
```diff
@@ -367,8 +367,11 @@ public function categoryDropDown($FieldName = 'CategoryID', $Options = false) {
                     $ForceCleanSelection = false;
                 }
 
+                $Disabled = false;
```
linc: This is not correct - you are overriding any previous setting of disabled. Look just 7 lines above this.

nervoustwit:
Didn't see that.

PR #3092, opened 2015-09-15
Google Prettify: Render in messages.

beckyvb:
Lets prettifier style code blocks in private messages.


tburry:
The term would be "conversations" not "messages" :stuck_out_tongue_closed_eyes:

Issue #2419, opened 2015-01-20
PHP 5.2 / 5.4 check


linc:
* index.php must be validated as PHP 5.2 compatible so it does not fatal on 5.3+.
* Need to throw an exception or exit on PHP <  5.4.

linc:
Also add checks for PDO & Unicode support for regex.

tburry:
I wonder if a standalone checker would be nice. That way we can also increase requirements over time and people can just run a single page to see if they meet them.

linc:
What I'd really like to see is a something in the Dashboard per #2018 that also incorporated a warning if the coming upgrade was going to fail on their current PHP version.

PR #3095, opened 2015-09-16
Check the PHP version before executing.

tburry:
A basic version check will help cut down on support requests.

### DIFF
--- a/index.php
+++ b/index.php
@@ -8,6 +8,10 @@
  * @since 2.0
  */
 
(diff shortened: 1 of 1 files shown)


tburry:
I'd be curious to know what people think about this.

tburry:
See also #2419. Checking for the correct extensions should be done in the setup routine. Checking for the basic PHP version early is necessary or else users will get parse errors.

PR #3103, opened 2015-09-17
Add SPAM check when editing discussions or comments

initvector:
When comments or discussions are edited, they aren't passed through the SPAM filter.  This means there is a very real possibility a spammer could create a legitimate discussion, have it pass the SPAM filter, then go back and edit it with something that wouldn't have been able to pass in the first place.

The SpamModel::IsSpam blocks for discussions and comments have been moved above the insert- and update-specific routines, so it should be able to hit them both.

The $this->CheckForSpam condition is not used in this fix, because it is not pluggable. SpamModel::IsSpam uses a handy CheckSpam event, so plug-ins like Akismet can do their thing.

### DIFF
--- a/applications/dashboard/models/class.spammodel.php
+++ b/applications/dashboard/models/class.spammodel.php
@@ -113,14 +113,7 @@ class SpamModel extends Gdn_Pluggable {
             // If this is a discussion or a comment, it needs some special handling.
             if ($RecordType == 'Comment' || $RecordType == 'Discussion') {
                 // Grab the record ID, if available.
-                switch ($RecordType) {
-                    case 'Comment':
-                        $recordID = intval(val('CommentID', $Data, false));
-                        break;
-                    case 'Discussion':
-                        $recordID = intval(val('DiscussionID', $Data, false));
-                        break;
-                }
+                $recordID = intval(val("{$RecordType}ID", $Data));
 
                 /**
                  * If we have a valid record ID, run it through flagForReview.  This will allow us to purge existing
@@ -164,7 +157,7 @@ class SpamModel extends Gdn_Pluggable {
                 break;
             case 'Discussion':
                 $model = new DiscussionModel();
-                $row = (array)$model->getID($id);
+                $row = $model->getID($id, DATASET_TYPE_ARRAY);
 
                 /**
                  * If our discussion has more than three comments, it might be worth saving.  Hold off on deleting and
@@ -173,7 +166,7 @@ class SpamModel extends Gdn_Pluggable {
                 if ($row['CountComments'] > 3) {
                     $deleteRow = false;
                 } elseif ($row['CountComments'] > 0) {
-                    $comments = Gdn::database()->SQL()->getWhere(
+                    $comments = Gdn::database()->sql()->getWhere(
                         'Comment',
                         array('DiscussionID' => $id)
                     )->resultArray();
@@ -186,7 +179,7 @@ class SpamModel extends Gdn_Pluggable {
                 }
                 break;
             default:
-                throw NotFoundException($recordType);
+                throw notFoundException($recordType);
         }
 
         $overrideFields = array('Name', 'Body');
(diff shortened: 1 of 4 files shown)


linc:
Need to carefully test what happens when spam is intercepted at this point; I'm concerned it will fork the comment into two places.

initvector:
I've updated the checking behavior to remove the original discussion/comment when it pulls an edited row into the SPAM moderation queue

applications/dashboard/models/class.spammodel.php
```diff
@@ -110,9 +110,99 @@ public static function isSpam($RecordType, $Data, $Options = array()) {
                     break;
             }
 
-            LogModel::insert('Spam', $RecordType, $Data, $LogOptions);
+            // If this is a discussion or a comment, it needs some special handling.
+            if ($RecordType == 'Comment' || $RecordType == 'Discussion') {
+                // Grab the record ID, if available.
```
linc: What do you think of collapsing this entire `switch` into just `$recordID = intval(val($RecordType.'ID', $Data, 0));` ?

applications/dashboard/models/class.spammodel.php
```diff
@@ -110,9 +110,99 @@ public static function isSpam($RecordType, $Data, $Options = array()) {
...
+         * Error out if we're not dealing with a discussion or comment.
+         */
+        switch ($recordType) {
+            case 'Comment':
+                $model = new CommentModel();
+                $row = $model->getID($id, DATASET_TYPE_ARRAY);
+                break;
+            case 'Discussion':
+                $model = new DiscussionModel();
+                $row = (array)$model->getID($id);
```
linc: Can we pick either array-casting or using `DATASET_TYPE_ARRAY` and stick with it?

linc:
For this workflow:

* Create discussion not detected as spam.
* 4 comments are made.
* OP is edited with something detected as SPAM.
* Discussion record goes to SPAM log.
* Moderator approves in SPAM log.

Won't the end effect still be to have a second, duplicate discussion?

applications/dashboard/models/class.spammodel.php
```diff
@@ -110,9 +110,99 @@ public static function isSpam($RecordType, $Data, $Options = array()) {
...
+
+                    if (!array_key_exists('_Data', $row)) {
+                        $row['_Data'] = array();
+                    }
+
+                    $row['_Data']['Comment'] = $comments;
+                }
+                break;
+            default:
+                throw NotFoundException($recordType);
```
linc: casing

applications/dashboard/models/class.spammodel.php
```diff
@@ -110,9 +110,99 @@ public static function isSpam($RecordType, $Data, $Options = array()) {
...
+                $row = (array)$model->getID($id);
+
+                /**
+                 * If our discussion has more than three comments, it might be worth saving.  Hold off on deleting and
+                 * just flag it.  If we have between 0 and 3 comments, save them along with the discussion.
+                 */
+                if ($row['CountComments'] > 3) {
+                    $deleteRow = false;
+                } elseif ($row['CountComments'] > 0) {
+                    $comments = Gdn::database()->SQL()->getWhere(
```
linc: `sql()`

linc:
That should cover everything; the rest looks good.

initvector:
>For this workflow:
>1. Create discussion not detected as spam.
>2. 4 comments are made.
>3. OP is edited with something detected as SPAM.
>4. Discussion record goes to SPAM log.
>5. Moderator approves in SPAM log.
>6. Won't the end effect still be to have a second, duplicate discussion?

The updated discussion body content should be quarantined in the SPAM queue when flagged.  If the moderator flags the entry as "Not SPAM", the body of the original discussion should be updated with the new content.

initvector:
This PR now relies on https://github.com/vanilla/vanilla/pull/3256

PR #3104, opened 2015-09-18
Fixes the dashboard's "edit user" so it populates

bryanjamesmiller:
The dashboard's "edit user" wasn't prepopulating with the information saved from additional custom fields when using ProfileExtender to add that additional profile information on the profile page.


linc:
Howdy Bryan, thanks for contributing. Would you sign our Contributors' Agreement? http://vanillaforums.org/contributors

bryanjamesmiller:
This fix only works for older versions of Vanilla.  I didn't realize it at the time I created it.  After upgrading to the newest version of Vanilla, this fix breaks the newest version.  Thus, please disregard this pull request, but perhaps some code could be created with better backwards compatibility.  I was working with the version of Vanilla that comes pre-installed on GoDaddy servers, which is a sadly outdated at version 2.0.17.9

PR #3106, opened 2015-09-18
Make title(), description(), error- and informMessage() translatable by default.

R-J:
Add the t() function inside the controller methods so that the methods texts are translatable by default. Changing this will lead to some double translations throughout the code. I do not think that this will effect performance too badly, since the methods are used sparsely and not in loops or anything like that.

Also fixed casening (completely?) in class.controller.php - *even in doc blocks!*

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -18,7 +18,7 @@
  * A base class that all controllers can inherit for common controller
  * properties and methods.
  *
- * @method void Render($View = '', $ControllerName = false, $ApplicationFolder = false, $AssetName = 'Content') Render the controller's view.
+ * @method void render($View = '', $ControllerName = false, $ApplicationFolder = false, $AssetName = 'Content') Render the controller's view.
  */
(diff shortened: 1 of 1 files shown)


tburry:
I'm not in to stuff like this because it doesn't allow for the default argument to t() which is all the more important in description type messages. Also, the page description is usually user-generated data and we don't want that going through translation capture at all.

R-J:
While I see your point (especially for the description), I think there are some inconsistencies. In this class (class.controller.php) there already is the addBreadcrumb function which always translates a breadcrumbs name. In the class form everything is passed through the t() function.

I think there should be one consistent way of handling that. Either you have a framework that is built for being translatable from the ground on up and that would mean that translation functions must be implemented in the most basic functions.
Or you use the translation function only when text is returned to the user (like the form class which generates markup). 

But mixing both options forces a developer to always look up how strings are handled in one special case.
Since I'm lazy, **I** would tend to always use the t() function. It's typed down fast, but the downside is, that sometimes strings are translated twice.

PR #3108, opened 2015-09-18
Move Markdown customizations out of vendor

kaecyra:
This PR moves most of the Vanilla Markdown customizations to a new class that extends the library. It also adds a flag to Gdn_Format::Markdown() to allow call-time toggling of these rules.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1478,9 +1478,10 @@ EOT;
      * Format a string using Markdown syntax. Also purifies the output html.
      *
      * @param mixed $Mixed An object, array, or string to be formatted.
+     * @param boolean $Flavored Optional. Parse with Vanilla-flavored settings? Default true
      * @return string
      */
-    public static function markdown($Mixed) {
+    public static function markdown($Mixed, $Flavored = true) {
         if (!is_string($Mixed)) {
             return self::to($Mixed, 'Markdown');
         } else {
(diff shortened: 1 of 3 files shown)


library/core/class.format.php
```diff
@@ -1478,9 +1478,10 @@ public static function getEmbedSize() {
      * Format a string using Markdown syntax. Also purifies the output html.
      *
      * @param mixed $Mixed An object, array, or string to be formatted.
+     * @param boolean $Flavored Optional. Parse with Vanilla-flavored settings? Default true
      * @return string
      */
-    public static function markdown($Mixed) {
+    public static function markdown($Mixed, $Flavored = true) {
```
hgtonight: This implementation would require you to change every call of this function if you want to remove the flavor. I would think this should default to a configuration option instead of true.

    public static function markdown($mixed, $flavored = null) {
      if(is_null($flavored)) {
        $flavored = c('Garden.InputFormatter.Flavored', true);
      }
    ...
    }

kaecyra:
My initial drive to do this is to support call-time invocation for special cases (such as on our marketing site). Allowing it to be globally configurable is a fine goal, but can be done later. This change does not remove functionality or configurability, simply refactors to improve design and opens the door for later configuration.

tburry:
We need another comparison of the Markdown base class to the original project. I know there were other changes in there at some point and blindly updating the library overwrote them. Now would be a good time to know what everything is.

kaecyra:
I've added a warning to not update the file, which will stand until we've had a change to perform the aforementioned diff in advance of moving to a clean version of the vendor library.

Issue #3105, opened 2015-09-18
Custom ProfileExtender fields in registration form not saving 


bryanjamesmiller:
After adding customized ProfileExtender fields that display during registration, those fields do not get saved and the user is forced to re-enter the same profile information on their profile page after logging in.

linc:
You're using the Applicant registration method, right?

bryanjamesmiller:
I just realized this issue is for 2.0.17.9 and everything works fine on 2.2.100.8.  Now I just have to figure out how to upgrade from 2.0.17.9 to 2.2.100.8

bryanjamesmiller:
The reason why we're using this version of Vanilla is because it comes pre-installed on GoDaddy's servers...

linc:
Wow 2.0.17 really? That's so bad. :-/

bryanjamesmiller:
@linc Maybe you could re-open this issue.  You might want to check out this thread: http://vanillaforums.org/discussion/comment/234016#Comment_234016 (you probably have seen it already).  Apparently there are some server configurations that prevent ProfileExtender from saving its custom fields that were filled out during registration.  One of those server configurations is GoDaddy's default settings.  I've successfully upgraded to 2.1.11 and I'm still having this issue.  However, I believe I got it working on Digital Ocean, but I also need to use GoDaddy.  I think there may be some 755 settings on GoDaddy that need to be turned to 777, but I'm not sure which.

bryanjamesmiller:
@linc Here's a fix that worked for me https://github.com/vanilla/vanilla/pull/3120 thanks to @peregrine

PR #3120, opened 2015-09-21
Update class.profileextender.plugin.php

bryanjamesmiller:
This successfully fixed the following issue: https://github.com/vanilla/vanilla/issues/3105 on my GoDaddy server

### DIFF
--- a/plugins/ProfileExtender/class.profileextender.plugin.php
+++ b/plugins/ProfileExtender/class.profileextender.plugin.php
@@ -1,21 +1,25 @@
-<?php if (!defined('APPLICATION')) exit();
+<?php
 /**
- * @copyright 2013 Vanilla Forums Inc.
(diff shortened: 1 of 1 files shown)


linc:
Can you re-target this to `master` branch? You appear to be attempting to modify the master version but are backporting to 2.1 but we've had too many formatting alterations since then.

bryanjamesmiller:
The 2.2 master branch doesn't have this bug and doesn't need any changes here as far as I know.  This fix is only for 2.1.  Forgive my ignorance - is 2.1 no longer getting any updates?

linc:
We're on the cusp of releasing 2.2 so we've only been releasing security fixes for 2.1. Can you isolate what the fix was rather than copy/pasting the entire file? This is nearly all formatting stuff.

PR #3123, opened 2015-09-22
Cleanup the API access for role/get

tburry:
Add some additional information when getting a single role as data.

- Return the role as a single “Role” key in the data array.
- Add the role’s permissions to the result.


tburry:
This PR does change the API from the docs slightly, but does not otherwise affect the behavior of the app.

PR #3133, opened 2015-09-25
Pretty print JSON responses

tburry:
This makes debugging a bit easier when looking at the API without a JSON view type plugin.


initvector:
I am a big fan of this update.

linc:
Likewise.

mtschirs:
Why not tie pretty printing to Garden.Debug = true?

kaecyra:
Enjoy your 4000 lines of Trace() output then

mtschirs:
My bad, I thought traces aren't visible in JSON responses.

PR #3136, opened 2015-09-25
Rename “Delete Conversation” to “Leave Conversation”

tburry:
Users don’t really delete conversations so the wording was confusing. This change also moves the confirmation into a proper postback rather than a GET request that validates the CSRF token.


linc:
YAS.

PR #3137, opened 2015-09-25
Make utility/structure require an authenticated postback

tburry:
- utility/structure now presents a start page on GET.
- The user must postback to scan and run the database structure.


tburry:
This PR might be a bit easer to look over if you go through each commit separately.

PR #3142, opened 2015-09-25
Add some basic sanitization to the HomeController

tburry:
Also, pass safer errors from /categories/archives.


mtschirs:
I favor explicit sanitizing, either in error.php or at the source, before user input is fed into the constructor of Gdn_Exception. This implicit automatism is hard to trace and debug.

tburry:
Good idea.

PR #3143, opened 2015-09-25
Refactor model form filtering

tburry:
- Add the property Gdn_Model->filterFields to control which fields get filtered on Gdn_Model->save().
- Add convenience methods to manipulate Gdn_Model->filterFields.

The idea behind this functionality is allowing more control on form filtering.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -60,6 +60,10 @@ class UserModel extends Gdn_Model {
             'LastIPAddress', 'AllIPAddresses', 'DateFirstVisit', 'DateLastActive', 'CountDiscussions', 'CountComments',
             'Score'
         ));
+
+        if (!Gdn::session()->checkPermission('Garden.Moderation.Manage')) {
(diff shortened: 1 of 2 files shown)


mtschirs:
What do you think about switching to explicit whitelisting (was a suggestion from @bleistivt)?

tburry:
Explicit whitelisting is something we are going to do for v2 of our API, but aren't currently planning for the rest of the app.

mtschirs:
You should really consider whitelisting POST fields that are written back to the database. Whitelisting could e.g. have prevented the admin registration vulnerability earlier this year.

tburry:
It has been considered and will be implemented in v2 of our API. However, this PR is not the place for it.

mtschirs:
Sorry if I was not clear enough. I was not talking about the API, but about model form filtering. This PR refactors model form filtering by giving subclasses of Gdn_Model the ability to modify the blacklist. However, in my opinion, a whitelisting approach offers advantages over this blacklisting approach, i.e. instead of offering subclasses the ability to modify the blacklist I recommend offering subclasses the ability to modify a whitelist.

If that is what you already understood from reading my comment, then all is well. I was just confused by you bringing up the API here.

PR #3146, opened 2015-09-26
IPv6 Compatibility

FerusGrim:
The goal of this PR is to enable the use of IPv6 addresses, rather than forcing '0.0.0.1/2' on users.

This PR is still being tested, but is being made so that progress can be reported, and tips/comments can be made.

TODO:
- [x] Database changes
- [x] Remove references to ForceIPv4
- [ ] Prettify _*_
- [ ] Fix the search function

_* Would like some comments on how people think this should be done._


FerusGrim:
Currently looking around for strategies on how ban/moderation rules could be applied to users using an IPv6 address. Give me a few hours/days until I can come up with a decent solution.

tburry:
IPv6 compatibility will probably have to span multiple PRs. It's a big job as I'm sure you've found. We are actually going to start working through it with the core team.

I've added a PR #3159 to consolidate the IP address fields to one type which will allow us to change the storage type more easily (and optionally) when we add IPv6 support.

PR #3150, opened 2015-09-28
Use a different method to generate random strings

tburry:
Change to a better random string method for transient keys. Also, expire the transient key cookie on sign out.


linc:
I don't think the second commit is supposed to be here.

tburry:
Yeah the second commit was an accident. I did a force reset to the first commit, but if you have a local copy of this branch you'll need to delete it first.

PR #3157, opened 2015-09-30
Correct 'You' links in activity

linc:
Our default userAnchor change from `/{id}/{name}` to just `/{name}` has broken they previously-passable links to `/{id}/You`. This corrects the formulation to quarantine the derived username in case it's "You".


tburry:
Isn't this now going to be missing the proper css class information for the user anchor?

PR #3159, opened 2015-10-01
Add support for an “ipaddress” database type

tburry:
This type acts as an alias for all columns that store a single IP address.
This helps lays the groundwork for changing IP address storage in support
of ipv6.


tburry:
There was an IP address in GDN_Attachment that was varchar(64) which will get downsized after this PR. That table should be small enough to not cause undue DB load.

PR #3165, opened 2015-10-05
Promoted Content: Change excerpt formatting.

beckyvb:
Slicing before stripping tags results in some tags being sliced and therefore, not stripped. This fixes a bug where tags were showing up in promoted content excerpts.

### DIFF
--- a/applications/vanilla/views/modules/helper_functions.php
+++ b/applications/vanilla/views/modules/helper_functions.php
@@ -87,7 +87,7 @@ if (!function_exists('WritePromotedContent')):
                 class="Title"><?php echo anchor(Gdn_Format::text(sliceString($Content['Name'], $Sender->TitleLimit), false), $ContentURL, 'DiscussionLink'); ?></div>
             <div class="Body">
                 <?php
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/modules/helper_functions.php
```diff
@@ -87,7 +87,7 @@ function writePromotedContent($Content, $Sender) {
                 class="Title"><?php echo anchor(Gdn_Format::text(sliceString($Content['Name'], $Sender->TitleLimit), false), $ContentURL, 'DiscussionLink'); ?></div>
             <div class="Body">
                 <?php
-                echo anchor(strip_tags(Gdn_Format::to(sliceString($Content['Body'], $Sender->BodyLimit), $Content['Format'])), $ContentURL, 'BodyLink');
+                echo anchor(sliceString(Gdn_Format::PlainText($Content['Body']), $Sender->BodyLimit), $ContentURL, 'BodyLink');
```
tburry: We just had another XSS merge where this actually needs an htmlspecialchars() around that sliceString().

PR #3166, opened 2015-10-05
Refactor CORS into its own method

linc:


library/core/class.controller.php
```diff
@@ -1343,6 +1329,24 @@ public function xRender($View = '', $ControllerName = false, $ApplicationFolder
...
 
     /**
+     * Set Access-Control-Allow-Origin header.
+     *
+     * If a Origin header is sent by the client, attempt to verify it against the list of
+     * trusted domains in Garden.TrustedDomains.  If the value of Origin is verified as
+     * being part of a trusted domain, add the Access-Control-Allow-Origin header to the
+     * response using the client's Origin header value.
+     */
+    public function setAccessControl() {
```
tburry: Can this be protected? Otherwise, but it in the internal array.

linc: Good catch, thanks.

initvector:
Works for me!

PR #3176, opened 2015-10-07
Order user tag search by comment count

bleistivt:
I found that ordering the tag search results by comment count generally gives better results, as these are the users you are more likely to mention / message.


linc:
I agree this would be a desirable enhancement but cannot remember if ordering on a non-indexed column is a concern. @tburry ?

tburry:
In this case it won't make a difference. Let's give it a try.

PR #3178, opened 2015-10-07
Restore user count to default user list

linc:
* Add a new "megasite" threshold for user total (in addition to existing lower "userthreshold").
* Always show a user total when not searching for users. For megasites, we approximate based on table status. Everyone else gets an accurate count. Previously, only sites below userthreshold got any total for this.
* Preserve "next/last" pager with no count for searches once we cross the userthreshold.
* Fix plural case for "X users found." to not hedge with (s).

### DIFF
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -112,25 +112,27 @@ class UserController extends DashboardController {
         // Get user list
         $this->UserData = $UserModel->search($Filter, $Order, $OrderDir, $Limit, $Offset);
         $this->setData('Users', $this->UserData);
-        $userCount = $this->UserData->count();
 
-        // If & how we display a user count depends on how huge our site is & whether we're searching.
-        // On sites past UserThreshold, zero users will be displayed by default.
-        if ($this->pastUserMegaThreshold()) {
-            // Dang, this site is mega-huge yo.
-
-
-        } elseif ($this->pastUserThreshold()) {
-            // Still big enough to choke an unoptimized query so tread lightly.
-            if ($userCount) {
-                $this->setData('_CurrentRecords', $userCount);
-            } else {
-                // No users have been searched for, so give the total users overall.
-                $this->setData('RecordCount', $UserModel->getCount());
-            }
-        } else {
+        // Figure out our number of results and users.
+        $showUserCount = $this->UserData->count();
+        if (!$this->pastUserThreshold()) {
             // Pfft, query that sucker however you want.
             $this->setData('RecordCount', $UserModel->searchCount($Filter));
+        } else {
+            // We have a user search, so at least set enough data for the Next pager.
+            if ($showUserCount) {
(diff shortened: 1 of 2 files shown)


applications/dashboard/controllers/class.usercontroller.php
```diff
@@ -106,15 +106,20 @@ public function index($Keywords = '', $Page = '', $Order = '') {
...
+        $this->UserData = $UserModel->search($Filter, $Order, $OrderDir, $Limit, $Offset);
         $this->setData('Users', $this->UserData);
-        if ($this->PastUserThreshold()) {
-            $this->setData('_CurrentRecords', $this->UserData->count());
+        if ($this->pastUserThreshold()) {
+            if ($count = $this->UserData->count()) {
+                $this->setData('_CurrentRecords', $count);
+            } else {
+                // No users have been searched for, so give the total users overall.
+                $this->setData('RecordCount', $UserModel->searchCount());
```
tburry: Is this optimized for the other end?

applications/dashboard/views/user/index.php
```diff
@@ -23,7 +23,8 @@
         echo $this->Form->textBox('Keywords');
         echo ' ', $this->Form->button(t('Go'));
         if ($this->data('RecordCount', null) !== null) {
-            echo ' ', sprintf(t('%s user(s) found.'), $this->data('RecordCount'));
+            $count = $this->data('RecordCount');
+            echo ' ', sprintf(plural($count, t('%s user found.'), t('%s users found.')), $count);
```
tburry: plural() translates so this is a double translate.

applications/dashboard/controllers/class.usercontroller.php
linc: Other end of what?

tburry: Looks like it isn't. This isn't okay then.

applications/dashboard/controllers/class.usercontroller.php
```diff
@@ -106,15 +106,20 @@ public function index($Keywords = '', $Page = '', $Order = '') {
...
+        $this->UserData = $UserModel->search($Filter, $Order, $OrderDir, $Limit, $Offset);
         $this->setData('Users', $this->UserData);
-        if ($this->PastUserThreshold()) {
-            $this->setData('_CurrentRecords', $this->UserData->count());
+        if ($this->pastUserThreshold()) {
+            if ($count = $this->UserData->count()) {
+                $this->setData('_CurrentRecords', $count);
+            } else {
+                // No users have been searched for, so give the total users overall.
+                $this->setData('RecordCount', $UserModel->getCount());
```
tburry: This should really be shown if there is no count AND no filter. Otherwise it will look a bit odd.

linc:
@tburry Updated per our conversation and I've edited the description accordingly.

tburry:
Just rebase this and we're good to go.

linc:
Why rebase a PR that will merge cleanly? Do we not like merges now?

PR #3180, opened 2015-10-07
Rewrite CDN links to always use SSL version

kaecyra:


kaecyra:
This is working on cl412, can it be merged?

PR #3183, opened 2015-10-08
Add ManageCaptcha config to optionally block key editing

linc:
Adds config `Garden.Registration.ManageCaptcha` that, when set to `false`, blocks access & editing of the Captcha API key.

### DIFF
--- a/applications/dashboard/views/settings/registration.php
+++ b/applications/dashboard/views/settings/registration.php
@@ -55,6 +55,7 @@ echo Gdn::slice('/dashboard/role/defaultroleswarning');
                 </tbody>
             </table>
         </li>
+        <?php if (c('Garden.Registration.ManageCaptcha', true)) : ?>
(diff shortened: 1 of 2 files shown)


applications/dashboard/views/settings/registration.php
```diff
@@ -55,6 +55,7 @@
                 </tbody>
             </table>
         </li>
+        <?php if (c('Garden.Registration.ManageCaptcha', true)) : ?>
```
kaecyra: Since you're checking `c('Garden.Registration.ManageCaptcha', true)` in the controller, its value should be assigned to the data array and checked there, instead of a) double checking the config in the view and b) making the view depend on the config

linc: Something was nagging me about this and I couldn't put my finger on what it was. Thanks.

PR #3186, opened 2015-10-08
Add phing support

initvector:
[Phing](https://www.phing.info) support has been added.  The build script will currently install dependencies via composer and package up files in the Vanilla directory, per package-contents.txt.  The result is a build.zip containing a read-to-deploy copy of Vanilla.

1. To use this build project, you'll first need to install Phing.  There are several different methods of installation outlined here: [Installing Phing](https://www.phing.info/trac/wiki/Users/Installation)
2. Once Phing is installed, navigate to the build directory and execute the phing command (no parameters necessary).  The build.xml project file should automatically be recognized and the steps within carried out.  The progress and status of each step of the build process will be displayed.  If the build completes successfully, a build.zip should now be present in the build directory.
* Done.


kaecyra:
I don't really know what it should be, but package-contents doesn't seem like it should have a txt extension.

initvector:
A few other options I had in mind:

* .dat
* .inc
* No extension

Easy-to-edit was the motivator behind my choice of TXT.

PR #3187, opened 2015-10-08
Add X-Path-Info for improved host compatibility

initvector:
We're trying to move away from using the p URL parameter.  Path info was our primary means to do this, but it isn't compatible with a few big hosts.  This update introduces X-Path-Info to Gdn_Request and .htaccess, which should improve that compatibility.

This should wait for merging until post-2.2.

Tested on GoDaddy.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,23 +1,25 @@
 # Original
 # If you modify this file then change the above line to: # Modified
+
 <IfModule mod_rewrite.c>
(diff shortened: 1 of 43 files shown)


linc:
2.2 is already forked, on the `release/2.2` branch. Nothing merged to core would effect it unless a second PR was created for that branch.

linc:
Merge conflict :-/

PR #3189, opened 2015-10-09
Fix ProfileExtender dropping data when email confirmation is used with applicant registration

linc:
Rename the event arg 'Users' in the `register()` method to `RegisteringUser` (and pass by reference for greater utility) to prevent naming conflict with the event arg in `setCalculatedFields()` which will overwrite this arg when sending a confirmation email for applicant users.


linc:
This breaks backward compatibility for events listening on events in this workflow.

PR #3190, opened 2015-10-09
Address invalid array_key_exists calls in UserModel::sso

initvector:
There were some invalid array_key_exists calls in UserModel::sso that were causing a couple SSO data fields to be erroneously removed.  This update addresses those function calls.


linc:
brutal.

PR #3197, opened 2015-10-14
Don't complain about non-writable folders unless being asked to create them.

kaecyra:


linc:
Isn't this just using bitwise operators as a harder-to-read boolean check?

linc:
Nevermind, figured it out.

PR #3199, opened 2015-10-14
Update README.md to include Phing blurb

initvector:


linc:
Let's move this down a ways to below Contributing. This is useful info only to maintainers, not the general public.

initvector:
Blurb moved!

PR #3201, opened 2015-10-15
Removes PHP 5.5 requirement from utility controller.

### LINKED ISSUE #3200: Vanilla 2.1.12 now requires PHP 5.5
From commit cb8f05af02cb7e913a74d7212ff7572793b95e8e

In `applications/dashboard/controllers/class.utilitycontroller.php` line 195. `empty()` prior to 5.5 requires a variable. http://php.net/manual/en/function.empty.php.

hgtonight:
This fixes #3200.

Also expanded the logic to be more readable.

### DIFF
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -292,7 +292,15 @@ class UtilityController extends DashboardController {
             // The form requires a postback to do anything.
             $step = 'start';
         } else {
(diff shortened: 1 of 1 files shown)


kaecyra:
I appreciate this commit, thank you. For core commits, we require strict adherence to the coding style. 

There needs to be a space after `if` and before the `(expression)`, and the `else` block should be on the same line as the closing `}` of the `if`.

So:

```
if (!empty($scan)) {
    $step = 'scan';
} else if (!empty($run)) {
    $step = 'run';
}
```

hgtonight:
Let me know if you need anything else.

linc:
Can you send this to master as well? We're not on 5.5 yet.

PR #3203, opened 2015-10-15
Hotfix/php 5.3 compatibility

hgtonight:
This adds a userland implementation of array_column for PHP < 5.5. It is recommended per the PHP documentation: http://php.net/manual/en/function.array-column.php

I also removed some more short arrays.


hgtonight:
This fixes #3200

linc:
fyi, the same array_column substitute is in `master` in `core/functions.general.php` :)

hgtonight:
@linc I was looking for it, but didn't find it.

Good eye :)

linc:
I cleaned this up to match that implementation so it doesn't trip us up with extra file loading at odd spots.

hgtonight:
:+1:

PR #3205, opened 2015-10-15
Remove PHP 5.5 requirement from utility controller

hgtonight:
Submitted against master per request from @linc at #3201


linc:
Sure was handy having this sitting here when my 5.4 server crapped out on /utility/structure! :+1:

PR #3208, opened 2015-10-15
Add option to not include usernames in "to" field of outgoing e-mails

initvector:
Sometimes you don't want to include a forum user's username in the "to" field of your e-mails.  This update adds a new configuration option (`Garden.Email.OmitToName `) and a corresponding new form control on the "Outgoing Email" page of the dashboard.  When enabled, this option stops usernames from being included in the "to" field of outgoing e-mails.  Only the e-mail address will be passed to that field.  By default, the value is false, so Vanilla's current behavior of including usernames is not affected unless explicitly changed.

### DIFF
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -607,7 +607,7 @@ class SettingsController extends DashboardController {
             'Garden.Email.SmtpPassword',
             'Garden.Email.SmtpPort',
             'Garden.Email.SmtpSecurity',
-            'Garden.Email.ToNoUsername'
+            'Garden.Email.OmitToName'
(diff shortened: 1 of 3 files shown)


linc:
`ToNoUsername` is making my head hurt. Can we just go with `NoUsername`? Or `OmitToName`? _Something something putting direct object first is weird_...

linc:
I think it's that, when reading a list of config variables, `Garden.Email.ToNoUsername` doesn't mean anything. "To No" is just confusing. I'd blank-stare at that and never be able to guess what it's for.

initvector:
I can appreciate that. I've updated this to use `Garden.Email.OmitToName` instead of `Garden.Email.ToNoUsername`.

PR #3211, opened 2015-10-19
Make activity headline a dynamic translation

x00:
This allows people to translate activity `HeadlineFormat` once an for all. Without having to do database changes. 

You can use the core standard. `HeadlineFormat.{ActivityType}` or the actual string (tested) or the actual activity. 

Where there is another definition format e.g. `Plugin.HeadlineFormat.{ActivityType}` they could translate that then `$Definition['`Plugin.HeadlineFormat.{ActivityType}'] = $Definition['HeadlineFormat.{ActivityType}']`. But good to start a convention. 

I hope this get in to 2.2 it would be a bonus if it could make it into any 2.1 updates.

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -144,7 +144,7 @@ class ActivityModel extends Gdn_Model {
         $Row['Url'] = ExternalUrl($Row['Route']);
 
         if ($Row['HeadlineFormat']) {
-            $Row['Headline'] = formatString($Row['HeadlineFormat'], $Row);
+            $Row['Headline'] = formatString(t("HeadlineFormat.{$Row['ActivityType']}", $Row['HeadlineFormat']), $Row);
         } else {
             $Row['Headline'] = Gdn_Format::activityHeadline($Row);
         }
@@ -895,7 +895,7 @@ class ActivityModel extends Gdn_Model {
         if (val('HeadlineFormat', $Activity)) {
             $SessionUserID = Gdn::session()->UserID;
             Gdn::session()->UserID = $User['UserID'];
-            $Activity['Headline'] = formatString($Activity['HeadlineFormat'], $Activity);
+            $Activity['Headline'] = formatString(t("HeadlineFormat.{$Activity['ActivityType']}", $Activity['HeadlineFormat']), $Activity);
             Gdn::session()->UserID = $SessionUserID;
         } else {
             if (!isset($Activity['ActivityGender'])) {


x00:
in fact best just set

`$Definition['HeadlineFormat.{ActivityType}']` there is no need to change what goes in the database at all. 

This is a good convention as `ActivityType` has to be unique.

tburry:
This will break backwards compatibility. We want to be able to have different headlines for the same activity type. A change will have to involve storing the translation code in the database rather than the end translation.

That also means that every language will need to be translated when a new code goes in or else the translation code will come back to the user.

x00:
@tburry 

`t("HeadlineFormat.{$Activity['ActivityType']}", $Activity['HeadlineFormat'])`

means it always defaults to what is in the database. Just it can be overridden. So there is no need translation for every language. 

Besides what to you mean different headlines, do you have an example? If you mean something like subjects, then it is pointless translating these anyway becuase they are totally unique. 

The sort of thing I'm talking about is "x commented on y", I very much doubt there different headlines in that case. If so, why not use a different activity type?

You could just have the `$Activity['HeadlineFormat']` as the reference, that way it is isn't linked to an activity type. I just thought a simpler reference would be nice. You could have a combination of the two as well. 

No one is forced to override the headlines, it is just a long term issue of translation.

x00:
Also having a translation code in the database, isn't backwards compatible, unless you can do backwards/forwards migration on the database, you will have record that are mixed literal and code.

You wouldn't translate the old record unless you provide translations in the two formats.

x00:
@tburry 

at the moment problem is being exasperated by offering translation on input. 

just `grep -r "HeadlineFormat\." .` and you will see what I mean. You get lines of 

` 'HeadlineFormat' => T('HeadlineFormat.`

What if someone translated them?

That is why it opted for a neutral reference on output, becuase otherwise you can end up chasing you tail. Entering a new reference in the database isn't going to change those records. there would be a continuity problem unless you updated those row, but them if there were imputed translated you have a problem. 

if it was to be translated on output but without reference, this is not backward compatable. 

it is not perfectly backward comparable whatever you do. 

I'd go with the least invasive solution.

`t($Activity['HeadlineFormat'])` might be reasonable. It not perfect, but covers most situations.

You could do a fix where you reverse translate `$Activity['HeadlineFormat']` to get the original text or reference the translate that. that could work if the reliant locale is loaded. 

 the reference idea should be abandoned unless you did something like this:

`t("HeadlineFormat.{$Activity['ActivityType']}", t($Activity['HeadlineFormat']))`

That give the option to override by text or by reference, which could help ease issue with dodgy records.

x00:
This is a long term issue. Can you read the previous comment about backwards compatibility. 

Would you accept that solution, or a solution like it with an optional config for overriding?

I would like to sort this issue as clients are requesting locale of activity, and at possible it is not made easy.

PR #3213, opened 2015-10-20
HTML email template

beckyvb:
Add support for sending html emails. Adds class for compiling and sanitizing content and basic view.


linc:
Needs to be tested with VanillaPop.

PR #3216, opened 2015-10-21
Fix coding standards around categories

linc:
* Casening
* Doc blocks
* Deprecated call replacements


nervoustwit:
Looks like Captain Casening is up to his old hijinx again.

PR #3224, opened 2015-10-26
Allow creation of separate values and labels in dropdowns in profile extender plugin.

nervoustwit:
This will allow us to manually include an array in the config file called OptionsLabel which is merged with the Options array to create dropdowns which have different values and labels. Presently when you create a dropdown we take the array of options provided by the user and merge it to itself to make something like:
&lt;option value='Canada'&gt;Canada&lt;/option&gt;

This PR provides the opportunity for someone to manually include an array in the config that would be merged with the options to provide user friendly labels:

&lt;option value='ca'&gt;Canada&lt;/option&gt;

If no OptionsLabel array is present, it continues to work as it always has.

### DIFF
--- a/plugins/ProfileExtender/views/profilefields.php
+++ b/plugins/ProfileExtender/views/profilefields.php
@@ -4,10 +4,16 @@ if (is_array($this->ProfileFields)) {
     foreach ($this->ProfileFields as $Name => $Field) {
         $Options = array();
         if ($Field['FormType'] == 'Dropdown') {
-            if(is_array($Field['OptionsLabels'])) {
-                $Options = array_combine($Field['Options'], $Field['OptionsLabels']);
-            } else {
-                $Options = array_combine($Field['Options'], $Field['Options']);
+            $values = $Field['Options'];
+            $labels = (array_key_exists('OptionsLabels', $Field)) ? $Field['OptionsLabels'] : false;
+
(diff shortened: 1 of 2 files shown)


linc:
This is a confusing way to code this - two nearly-identical lines without an obvious rationale for the 2 code paths. Consider assigning one or the other value to a clearly-named variable instead, so that it becomes clear (and self-documenting) what that extra option is doing.

nervoustwit:
I also copied the logic over to the view called "registrationfields".

plugins/ProfileExtender/views/profilefields.php
```diff
@@ -3,8 +3,19 @@
 if (is_array($this->ProfileFields)) {
     foreach ($this->ProfileFields as $Name => $Field) {
         $Options = array();
-        if ($Field['FormType'] == 'Dropdown')
-            $Options = array_combine($Field['Options'], $Field['Options']);
+        if ($Field['FormType'] == 'Dropdown') {
+            $values = $Field['Options'];
+            $labels = (array_key_exists('OptionsLabels', $Field)) ? $Field['OptionsLabels'] : false;
```
linc: Simplify all this:
     $Values = $Field['Options'];
     $Labels = val('OptionsLabels', $Field, $Field['Options']);
     $Options = array_combine($Values, $Labels);

PR #3225, opened 2015-10-27
Fix decho() behavior to show info only to admins

R-J:
Fighting zombies: https://github.com/vanilla/vanilla/pull/1867 :-D

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -797,9 +797,9 @@ if (!function_exists('decho')) {
      * Echo's debug variables if user is root admin.
      */
     function decho($Mixed, $Prefix = 'DEBUG', $Permission = false) {
-        $Prefix = StringEndsWith($Prefix, ': ', true, true).': ';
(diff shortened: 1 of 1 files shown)


tburry:
Since the $Permission parameter is a bit odd, would you mind adding @param comments to decho() in this PR?

R-J:
I've tried my best, but I'd say I'm not the best for that :-/

library/core/functions.general.php
```diff
@@ -795,11 +795,16 @@ function consolidateArrayValuesByKey($Array, $Key, $ValueKey = '', $DefaultValue
 if (!function_exists('decho')) {
     /**
      * Echo's debug variables if user is root admin.
+     * 
+     * @param mixed $Mixed The variable to echo.
+     * @param string $Prefix The text to be used as a prefix for the output.
+     * @param bool $Permission Whether or not output is visible for everyone.
+     * @return string The variable in human-readable format.
```
hgtonight: This function doesn't return anything.

R-J:
That's true... :-/

Issue #3227, opened 2015-10-28
Old roles not removed during saveRoles()


kaecyra:
When `saveRoles()` is called, if the user in question has UserRole entries for roles that no longer exist, these roles are not removed. They probably should be.

PR #3229, opened 2015-10-28
Fix bug when saving a non-existent role to a user that already has that non-existent role

tburry:
To reproduce this bug:

1. Assign a user to a dummy role ID that doesn’t exist in GDN_Role.
2. Try and save that role to the user. This isn’t that easy to do. Currently it gets done by adding a bad role ID in the config as the default member role and then confirming a user’s email address. You can also do this by editing the source code of UserController->edit() and just specifying a bad role ID there.


tburry:
Fixes #3227.

applications/dashboard/models/class.usermodel.php
```diff
@@ -2212,8 +2212,8 @@ public function saveRoles($UserID, $RoleIDs, $RecordEvent) {
         $OldRoleIDs = array();
         $OldRoleData = $this->SQL
             ->select('ur.RoleID, r.Name')
-            ->from('Role r')
-            ->join('UserRole ur', 'r.RoleID = ur.RoleID')
+            ->from('UserRole ur')
+            ->join('Role r', 'r.RoleID = ur.RoleID', 'left')
```
tburry: This is the main fix. Always bring in all entries from GDN_UserRole regardless of them existing in GDN_Role.

applications/dashboard/models/class.usermodel.php
```diff
@@ -2223,9 +2223,16 @@ public function saveRoles($UserID, $RoleIDs, $RecordEvent) {
         }
 
         // 1a) Figure out which roles to delete.
-        $DeleteRoleIDs = array_diff($OldRoleIDs, $RoleIDs);
+        $DeleteRoleIDs = [];
+        foreach ($OldRoleData as $row) {
+            // The role should be deleted if it is an orphan or the user has not been assigned the role.
```
tburry: This comment explains the rest of the change.

linc:
Got blocked by https://github.com/vanilla/vanilla/pull/3326

initvector:
Works A-OK for me.

Steps taken to reproduce:

1. Add a new entry to `UserRole` with a bogus role ID.
2. Hack `UserController::edit`, setting `$RequestedRoles` to a static value, including the invalid role ID.
3. Edit and save the user from step 1.

The invalid role ID was purged from the user's role associations.

PR #3245, opened 2015-11-02
Add discussion-specific view permission check.

beckyvb:
Add new method to test whether a user has permission to view a specific discussion by first checking category-specific permissions and then firing an event so it can be overridden.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -993,7 +993,7 @@ class CommentModel extends VanillaModel {
             $BookmarkData = $DiscussionModel->GetBookmarkUsers($DiscussionID);
             foreach ($BookmarkData->result() as $Bookmark) {
                 // Check user can still see the discussion.
-                if (!$discussionModel->canViewDiscussion($Discussion, $Bookmark->UserID)) {
+                if (!$discussionModel->canView($Discussion, $Bookmark->UserID)) {
                     continue;
                 }
 
@@ -1005,7 +1005,7 @@ class CommentModel extends VanillaModel {
             // Notify users who have participated in the discussion.
             $ParticipatedData = $DiscussionModel->GetParticipatedUsers($DiscussionID);
             foreach ($ParticipatedData->result() as $UserRow) {
-                if (!$discussionModel->canViewDiscussion($Discussion, $UserRow->UserID)) {
+                if (!$discussionModel->canView($Discussion, $UserRow->UserID)) {
                     continue;
                 }
 
@@ -1018,7 +1018,7 @@ class CommentModel extends VanillaModel {
             if ($Discussion != false) {
                 $InsertUserID = val('InsertUserID', $Discussion);
                 // Check user can still see the discussion.
-                if ($discussionModel->canViewDiscussion($Discussion, $InsertUserID)) {
+                if ($discussionModel->canView($Discussion, $InsertUserID)) {
                     $Activity['NotifyUserID'] = $InsertUserID;
                     $Activity['Data']['Reason'] = 'mine';
                     $ActivityModel->Queue($Activity, 'DiscussionComment');
@@ -1042,7 +1042,7 @@ class CommentModel extends VanillaModel {
                 }
 
                 // Check user can still see the discussion.
-                if (!$discussionModel->canViewDiscussion($Discussion, $User->UserID)) {
+                if (!$discussionModel->canView($Discussion, $User->UserID)) {
                     continue;
                 }
 
@@ -1112,7 +1112,7 @@ class CommentModel extends VanillaModel {
             $UserID = $Row['UserID'];
             // Check user can still see the discussion.
             $discussionModel = new DiscussionModel();
-            if (!$discussionModel->canViewDiscussion($Discussion, $UserID)) {
+            if (!$discussionModel->canView($Discussion, $UserID)) {
                 continue;
             }
 
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -896,7 +896,7 @@ class CommentModel extends VanillaModel {
      */
     public function save2($CommentID, $Insert, $CheckExisting = true, $IncUser = false) {
         $Session = Gdn::session();
-        $UserModel = Gdn::userModel();
+        $discussionModel = new DiscussionModel();
 
         // Load comment data
         $Fields = $this->getID($CommentID, DATASET_TYPE_ARRAY);
@@ -993,7 +993,7 @@ class CommentModel extends VanillaModel {
             $BookmarkData = $DiscussionModel->GetBookmarkUsers($DiscussionID);
             foreach ($BookmarkData->result() as $Bookmark) {
                 // Check user can still see the discussion.
-                if (!$UserModel->GetCategoryViewPermission($Bookmark->UserID, $Discussion->CategoryID)) {
+                if (!$discussionModel->canViewDiscussion($Discussion, $Bookmark->UserID)) {
                     continue;
                 }
 
@@ -1005,7 +1005,7 @@ class CommentModel extends VanillaModel {
             // Notify users who have participated in the discussion.
             $ParticipatedData = $DiscussionModel->GetParticipatedUsers($DiscussionID);
             foreach ($ParticipatedData->result() as $UserRow) {
-                if (!$UserModel->GetCategoryViewPermission($UserRow->UserID, $Discussion->CategoryID)) {
+                if (!$discussionModel->canViewDiscussion($Discussion, $UserRow->UserID)) {
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
...
+     * Tests whether a user has permission to view a specific discussion by checking category-specific permissions.
+     * Fires an event that can override the category view permission.
+     *
+     * @param object|integer $discussion The discussion ID or the discussion to test.
+     * @param string|integer $userId The ID of the user to test permission for. If empty, it defaults to Session user.
+     * @param string $categoryPermission The category permission to test against the user.
+     * @return bool Whether the user can view the discussion.
+     * @throws Exception
+     */
+    public function canViewDiscussion($discussion, $userId = '', $categoryPermission = 'Vanilla.Discussions.View') {
```
tburry: There is already a `DiscussionModel->canEdit()` so maybe this should just be called `DiscussionModel->canView()`.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
 
         return self::$AllowedSortFields;
     }
+
+
+    /**
+     * Tests whether a user has permission to view a specific discussion by checking category-specific permissions.
+     * Fires an event that can override the category view permission.
+     *
+     * @param object|integer $discussion The discussion ID or the discussion to test.
```
tburry: object|array|integer

applications/vanilla/models/class.discussionmodel.php
tburry: I'd put the $userID parameter last.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
...
+        if (!$userId) {
+            $userId = val('UserID', Gdn::session(), false);
+        }
+        // Fetch discussion.
+        if (is_numeric($discussion)) {
+            $discussion = $this->getID($discussion);
+        }
+        $userModel = new UserModel();
+        // Get category permission.
+        $canView = $userId && $userModel->getCategoryViewPermission($userId, val('CategoryID', $discussion), $categoryPermission);
```
tburry: This is wrong. You need to grab the category from CategoryID and then check against its PermissionCategoryID.

tburry: Maybe check to see if $discussion already has PermissionCategoryID too.

tburry: Oh never-mind. I see that userModel->getCategoryViewPermission does all this. Sorry.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
...
+
+        // Fire event to override permission ruling.
+        $this->EventArguments['discussion'] = $discussion;
+        $this->EventArguments['userId'] = $userId;
+        $this->EventArguments['categoryPermission'] = $categoryPermission;
+        $this->EventArguments['canView'] = &$canView;
+        $this->fireEvent('viewPermission');
```
tburry: I'd rather the event arguments stay uppercase to be consistent with the rest of the framework. This isn't something we can easily change.

Also, We use ID, not Id and that won't change.

applications/vanilla/models/class.discussionmodel.php
beckyvb: The userID argument is used often here, and the category permission is rarely changed. If we change the order, then dev will have to call canView($discussion, 'Vanilla.Discussions.View', $userID); rather than simply canView($discussion, $userID);

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
...
+        $userModel = new UserModel();
+        // Get category permission.
+        $canView = $userID && $userModel->getCategoryViewPermission($userID, val('CategoryID', $discussion), $categoryPermission);
+
+        // Fire event to override permission ruling.
+        $this->EventArguments['Discussion'] = $discussion;
+        $this->EventArguments['UserID'] = $userID;
+        $this->EventArguments['CategoryPermission'] = $categoryPermission;
+        $this->EventArguments['CanView'] = &$canView;
+        $this->fireEvent('viewPermission');
```
tburry: Wouldn't it be better to name this event "canView" to match the method name?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
...
+    public function canView($discussion, $userID = '', $categoryPermission = 'Vanilla.Discussions.View') {
+        // Default to session user.
+        if (!$userID) {
+            $userID = val('UserID', Gdn::session(), false);
+        }
+        // Fetch discussion.
+        if (is_numeric($discussion)) {
+            $discussion = $this->getID($discussion);
+        }
+        $userModel = new UserModel();
```
tburry: Save an object instantiation:

```php
$userModel = Gdn::userModel();
```

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2464,4 +2464,38 @@ public static function allowedSortFields($Allowed = null) {
...
+     * Tests whether a user has permission to view a specific discussion by checking category-specific permissions.
+     * Fires an event that can override the category view permission.
+     *
+     * @param object|array|integer $discussion The discussion ID or the discussion to test.
+     * @param string|integer $userID The ID of the user to test permission for. If empty, it defaults to Session user.
+     * @param string $categoryPermission The category permission to test against the user.
+     * @return bool Whether the user can view the discussion.
+     * @throws Exception
+     */
+    public function canView($discussion, $userID = '', $categoryPermission = 'Vanilla.Discussions.View') {
```
tburry: Using `$userID = 0` will allow this method to have the userID be just an integer which is a move we want to make to help hint to future compilers.

applications/vanilla/models/class.commentmodel.php
```diff
@@ -1033,15 +1033,16 @@ public function save2($CommentID, $Insert, $CheckExisting = true, $IncUser = fal
 
             // Notify any users who were mentioned in the comment.
             $Usernames = GetMentions($Fields['Body']);
+            $userModel = new UserModel();
```
tburry:
```php
$userModel = Gdn::userModel();
```

applications/vanilla/models/class.discussionmodel.php
beckyvb: The argument is referencing the variable, not the method. Also, I agree with your feedback from a couple days ago:
>I'd rather the event arguments stay uppercase to be consistent with the rest of the framework. This isn't something we can easily change.
I think this should stay uppercase.

R-J:
I think it is a great descision to add such a method to the **Discussion**Model! That's where I would expect it and I've already missed it many times without really knowing what I've missed ;-)

But instead of having separate functions canEdit and canView wouldn't it be wise to choose a more DRY way? What about the Sink, Announce, Close etc. permissions? If there is a function to check Edit and View permission, a developer would expect having a canWhateverpermission, too. If they are all implemented like canEdit and canView, there would be a lot of unneccessary code repetition.

I think you, @beckyvb, have already had the same idea and that's why the permission is an argument to the new canView method and it might also be the reason for not being sure about the best name for the event.

Wouldn't it be more intuitive and DRY to have a `getDiscussionPermissions($discussion, $user)` which returns an array of all "Vanilla.Discussions.\*" permissions the user has? canEdit, canView, canClose, canSink, canAnnounce and all the like could be based on that.
I do not know which is more performant: getting all permissions or only a single one. So instead of getting **all** permissions a `getDiscussionPermission($discussion, $user, $permission = 'View')` might be better.

As I've said before: if I have canEdit and canView, I would expect to have canEverypermission.
By implementing a general method, you could easily implement e.g. `canClose($discussion, $user)` with just 3 lines of code!
But you can also say "we implement only complex canSomethings because a simple 'canSink' is only one line of code: `$canSink = $discussionModel->getDiscussionPermission($discussion, $user, 'Sink');` and that's why we make our model fatter than needed"

<hr>

The above is what I think should really be done if you like to keep your codebase DRY. What's following is a "whishlist" if you do so:

**1. References to the permission**
I would really appreciate it, if the permissions are only referenced with their last part in a new method. So if I have to pass a permission as an argument to `getDiscussionPermission()`, I would like to use "view" and not "Vanilla.Discussions.View". This could be easily achieved with a `$permission = 'Vanilla.Discussions.'.ucfirst(strtolower($permission));` in the method. Since we are in the disccussion model, I wouldn't expect to check any other permission than "Vanilla.Discussions.\*" anyway.
As a result of `getDiscussionPermissions()`, I would like to see it like that `array('view', 'edit', 'sink')`. The same reason here: if I  want to get the permissions for a discsussion, I would expect to see that I have **view** or **edit** permission, but not "Vanilla.Discussions.View" or "Vanilla.Discussions.Edit".

Just imagine how it would look if you write code for it:
`$canAnnounce = $discussionModel->getDiscussionPermission($discussion, $user, 'Vanilla.Discussions.Announce');`
4 x "discussion" in there...

That leads directly to point 2...

**2. Method name**
That's why my first thought has been to name it `checkPermission` - which I would still favor - but I would also understand if there are concerns regarding the already existing functions. But maybe `getPermissions` (as it is already called in class.session.php) would be a better or even the best name.

beckyvb:
Thanks for the feedback @R-J! I've gone in and refactored a few things to make the method more universal. What do you think? (also @tburry)

PR #3260, opened 2015-11-06
Add button enter key handler to wysiwyg-format editor.

beckyvb:
If wysihtml5 has a custom key handler, it's failing here. This fixes a bug where clicking enter from the fileupload dropdown submits the post.


linc:
Needs version increment to accompany this.

PR #3261, opened 2015-11-06
Use explicitly defined SelfUrl in inform messages, if it exists.

beckyvb:
Use SelfUrl rather than the document url.
Fixes bug where cancelling the moderation inform message in Vanilla comments brings the user to /index.php?p=/discussion/embed/ rather than staying on the current page.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -924,9 +924,13 @@ jQuery(document).ready(function($) {
                 message = '<div class="InformMessage">' + message + '</div>';
                 // Insert any transient keys into the message (prevents csrf attacks in follow-on action urls).
                 message = message.replace(/{TransientKey}/g, gdn.definition('TransientKey'));
-                // Insert the current url as a target for inform anchors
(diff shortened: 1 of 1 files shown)


js/global.js
```diff
@@ -924,9 +924,13 @@ jQuery(document).ready(function($) {
                 message = '<div class="InformMessage">' + message + '</div>';
                 // Insert any transient keys into the message (prevents csrf attacks in follow-on action urls).
                 message = message.replace(/{TransientKey}/g, gdn.definition('TransientKey'));
-                // Insert the current url as a target for inform anchors
-                message = message.replace(/{SelfUrl}/g, document.URL);
-
+                if (gdn.definition('SelfUrl')) {
```
bleistivt: This will always be true, because gdn.definition returns the key if the value can't be found (as opposed to gdn.getMeta).

beckyvb: So it does. Thank you!

initvector:
Works for me!

PR #3266, opened 2015-11-09
Detect initial setup in permissionmodel->define() and force mappings

linc:
See #2862

**Summary**: MySQL 5.6+ defaults to strict mode, so we need to support it. Our method of mapping permissions to defaults causes a strict mode failure on initial setup. This works around this problem by detecting this situation and force-mapping the permissions to off.

**Only functional change is the new block starting at 250 ($DefaultRow)**. Everything else is to improve the clarity (docs, casening, line length).


linc:
Needs to be ported to `release/2.2` once finalized.

tburry:
This looks fine.

PR #3268, opened 2015-11-10
Add BeforeCaptcha event to easily enable custom captchas

R-J:
Added a new event to form->captcha to allow plugins to easily output custom captchas instead of the default reCAPTCHA. 

Also changed spelling of vars to lowercase and eliminated one unneeded if clause. For an explanation: the original code would have been extended to:
~~~
if (!in_array($Language, $Whitelist)) {
    if (in_array(Gdn::locale()->Locale, $Whitelist)) {
        $Language = Gdn::locale()->Locale;
    } else {
        $Language = false;
    }
}
if ($Language) {
    $Options['lang'] = $Language;
}
~~~

I've shortened
~~~
if (!in_array($Language, $Whitelist)) {
...
}
if ($Language) {
    $Options['lang'] = $Language;
}
~~~

to
~~~
if (in_array($Language, $Whitelist)) {
    $Options['lang'] = $Language;
} else {
...
}
~~~
and kicked out the now unneed
~~~
    } else {
        $Language = false;
~~~

Thanks @peregrine for [making me think](http://vanillaforums.org/discussion/comment/235400/#Comment_235400) about that :-)

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -243,14 +243,23 @@ class Gdn_Form extends Gdn_Pluggable {
     /**
      * Returns Captcha HTML & adds translations to document head.
      *
+     * Events: BeforeCaptcha
+     * 
      * @return string
      */
     public function captcha() {
+        $handled = false;
+        $this->EventArguments['Handled'] =& $handled;
+        $this->fireEvent('BeforeCaptcha');
+        if ($handled) {
+            // A plugin handled the captcha so don't display anything more.
+            return;
+        }
         // Google whitelist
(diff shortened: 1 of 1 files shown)


library/core/class.form.php
```diff
@@ -265,16 +274,17 @@ public function captcha() {
...
-        if (!in_array($Language, $Whitelist)) {
-            $Language = (in_array(Gdn::locale()->Locale, $Whitelist)) ? Gdn::locale()->Locale : false;
-        }
-        if ($Language) {
-            $Options['lang'] = $Language;
+        $language = Gdn::locale()->language();
+        if (in_array($language, $whitelist)) {
+            $options['lang'] = $language;
+        } else {
+            if (in_array(Gdn::locale()->Locale, $whitelist)) {
```
linc: `elseif` would make this neater.

library/core/class.form.php
```diff
@@ -243,14 +243,23 @@ public function calendar($FieldName, $Attributes = false) {
...
      * Returns Captcha HTML & adds translations to document head.
      *
+     * Events: BeforeCaptcha
+     * 
      * @return string
      */
     public function captcha() {
+        $handled = false;
+        $this->EventArguments['Handled'] =& $handled;
+        $this->fireEvent('BeforeCaptcha');
```
linc: Because the primary purpose of this event is to replace rather than prep, I think it would be appropriate to simply call this event `'Captcha'`.

linc:
I like this a lot. A nice little refactor + an event that moves us closer to something I've been wanting to do for a long time: make registration much more flexible.

linc:
Thanks!

PR #3270, opened 2015-11-10
Make it possible to skip captcha if it has not been set up correct

R-J:
When captcha is not configured, new users are seeing the misleading "To use reCAPTCHA you must get an API key from https://www.google.com/recaptcha/admin" and are not able to register. Bad.

Adding a new config setting 'Garden.Registration.CaptchaPassByDefault' which defaults to true would skip the captcha if it isn't set up. That way admins could decide to do without captchas by not configuring it or they can use them as they do by now, but no new users will be rejected before the captcha is configured.

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -247,7 +247,7 @@ class Gdn_Form extends Gdn_Pluggable {
      */
     public function captcha() {
         // Stop if captcha isn't configured and CaptchaPassByDefault is true.
-        if (!c('Garden.Registration.CaptchaPrivateKey') && c('Garden.Registration.CaptchaPassByDefault', false)) {
+        if (!c('Garden.Registration.CaptchaPrivateKey') && c('Garden.Registration.CaptchaPassByDefault', true)) {
             return;
         }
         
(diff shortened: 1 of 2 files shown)


linc:
There's a danger here that the admin not even realize they're _supposed_ to set up CAPTCHA ("everything works, sweet!"), and a week later their board is flooded with spam. This silent-off functionality with no opt-in prompt seems hostile to new forum owners.

R-J:
At first I had set the default to false so that the current behavior remains. But the way it is right now is user unfriendly while the way it is done in my PR is admin unfriendly. Forums are made for users and that's why I think it is better this way.

Looking at the risks of both possible ways, I'd say that losing potential users because they are not able to register is more "severe" than having to clean spammers.
But adding a hint to the dashboard that is displayed as long as captcha is needed but not configured could be a way to handle that.

linc:
There's a lot we could do better for this:

* Show a notice in the Dashboard when Captcha is active but no key is set up.
* Prompt the admin to get one as part of setup, or immediately after.
* Shut down the register page entirely when Captcha is active but no key is set up.

I see bad things happening by enabling folks to set up forums with zero spam protection by default and letting them run free. We're gonna get a bad rap as spam traps.

PR #3271, opened 2015-11-10
Let activity page length be configurable

hgtonight:
Used the naming scheme for Vanilla.Discussions.PerPage.

PR #3272, opened 2015-11-10
Activity page length configuration

hgtonight:
It is currently not possible to change the length of an activity page.

I don't think this warrants a GUI, but I do think it is important.


hgtonight:
This is identical to #3271.

PR #3275, opened 2015-11-11
Force wysiwyg embedded comments to Text format.

beckyvb:
Fixes bug where newlines are not being rendered in Vanilla comments if the editor is set to WYSIWYG.

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -397,6 +397,7 @@ class PostController extends VanillaController {
         $vanilla_category_id = $this->Form->getFormValue('vanilla_category_id', '');
         $Attributes = array('ForeignUrl' => $vanilla_url);
         $vanilla_identifier = $this->Form->getFormValue('vanilla_identifier', '');
+        $isEmbeddedComments = $vanilla_url != '' && $vanilla_identifier != '';
 
         // Only allow vanilla identifiers of 32 chars or less - md5 if larger
         if (strlen($vanilla_identifier) > 32) {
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -397,6 +397,7 @@ public function comment($DiscussionID = '') {
         $vanilla_category_id = $this->Form->getFormValue('vanilla_category_id', '');
         $Attributes = array('ForeignUrl' => $vanilla_url);
         $vanilla_identifier = $this->Form->getFormValue('vanilla_identifier', '');
+        $isEmbeddedComments = $vanilla_url != '' && $vanilla_identifier != '';
```
linc: Shouldn't it be equivalent to just check `$vanilla_url && $vanilla_identifier`?

applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -529,6 +530,12 @@ public function comment($DiscussionID = '') {
             $this->Form->addError(t('Failed to find discussion for commenting.'));
         }
 
+        // Vanilla Comments save as Text, no matter the format.
+        // Kludge to set format to Text if we're Wysiwyg to preserve newlines.
+        if ($isEmbeddedComments && ($this->Form->getFormValue('Format', c('Garden.InputFormatter')) === 'Wysiwyg')) {
```
linc: You say it saves as next no matter what, but only override if it's `Wysiwyg` - so which is it?

applications/vanilla/controllers/class.postcontroller.php
linc: I see now you just duplicated the old logic, which is safe and fine.

applications/vanilla/controllers/class.postcontroller.php
beckyvb: I was hoping to preserve the ability for a user to navigate to the forum and edit their post in the format specified by the editor. So, if we set every comment as Text and the editor format is BBCode, and if the user edits their comment on the forum and adds BBCode then BBCode won't render. WYSIWYG is the only format with the newline issue and there's a setting to reinterpret existing posts as WYSIWYG, so I think it's best to limit this to WYSIWYG.

PR #3276, opened 2015-11-12
Make alternate() use parameters correctly

R-J:
I first thought of changing alternate() when I saw it uses a modulo calculation when handling a bool value would also be enough, but looking closer at it showed that the output of `function alternate($Odd = 'Alt', $Even = '', $AttributeName = 'class')` is in almost any combination of parameters not what you expect.

Called without any parameters, it echos the following:
Odd:
Even: ` class="Alt"`
...
Already here it is faulty: it uses the standard $Odd parameter for the even lines.

If I pass `alternate('odd', 'even')`, I get
Odd:  `even`
Even: `odd` 
That they are mixed up was to be expected after point 1, but I would expect those to be enclosed in class="..."

Even if I pass `alternate('odd', 'even', 'arbitraryName')`, I only get
Odd:  `even`
Even: `odd` 

So it's obviously not as useful as the parameters makes us think it is. That's why I have *totally* rewritten it. Good news is, that the function is neither used in Vanilla nor in any of the many, many plugins and themes that I have downloaded (I did a grep through more than 100 plugins)

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -8,14 +8,31 @@
  * @since 2.0
  */
 
+/**
+ * Write alternating strings on each call.
+ *
+ * Useful for adding different classes to alternating lines in a list
+ * or table to enhance their readability.
+ *
+ * @param string $odd The text for the first and every further "odd" call.
+ * @param string $even The text for the second and every further "even" call.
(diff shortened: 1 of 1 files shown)


linc:
Looks good. That `if ($b = !$b)` makes me twitch a little but I'll allow it here.

PR #3283, opened 2015-11-13
Fix invalid closing span in activities

bleistivt:
See: https://validator.w3.org/nu/?doc=http%3A%2F%2Fvanillaforums.org%2Factivity


linc:
Thanks!

PR #3285, opened 2015-11-13
Move ability to set avatar to URL into /profile/picture

initvector:
Move the ability to set a user's photo to a URL to `/profile/picture` and tighten the security involved in such an action.


linc:
* PR title pls
* I want to make sure this `realpath()` check will not break Windows installs by returning a different file path reference than `getcwd()` does when we're establishing `PATH_ROOT` (and therefore `PATH_UPLOAD`), i.e. do they both give a path starting with a drive letter?
* I think add a `filter_var()` here in addition to the `isUrl()` check. Still imperfect but another step in the right direction.

initvector:
Testing PHP 5.5 on Windows 10, `getcwd` and `realpath` returned equivalent values, including drive letter.

PR #3286, opened 2015-11-13
Add Garden HTTP library

initvector:
Add Garden HTTP as a library for use in Vanilla

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
 
 	],
 	"require": {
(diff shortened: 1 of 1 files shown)


linc:
We're pulling a self-authored library into core here, which means the whole thing should get evaluated as if it was new code in this PR. No one but @tburry has ever gone thru it.

tburry:
@linc, you're going to have to assign this PR to someone then or else it's never going to get merged.

PR #3287, opened 2015-11-13
Allow config values to be permission-based

initvector:
This update is intended to allow some configuration values to be specified as a permission slug.  That slug will then be used to check against the signed-in user's permissions.  Fetching a permission-based config value will return the value of the permission check.  The check to resolve a permission to a configuration value is based on a third parameter to config functions: `$resolvePermission`

### Example

The `Garden.Profile.EditPhotos` configuration value might be set to `Garden.Settings.Manage`.  When the `Garden.Profile.EditPhotos` is checked using `Gdn::config('Garden.Profile.EditPhotos', true, true)`, the current user's permissions will be evaluated and if they have the `Garden.Settings.Manage` permission, the config value for `Garden.Profile.EditPhotos` will be true.  Otherwise, it will be false.

### DIFF
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -58,13 +58,12 @@ class Gdn_Session {
     /**
      * Check the given permission, but also return true if the user has a higher permission.
      *
-     * @param mixed $permission The permission (or array of permissions) to check.
-     * @param bool $fullMatch If $Permission is an array, $FullMatch indicates if all permissions specified are required. If false, the user only needs one of the specified permissions.
+     * @param bool|string $permission The permission to check.  Bool to force true/false.
      * @param string $junctionTable The name of the junction table for a junction permission.
-     * @param in $junctionID The ID of the junction permission.
-     * * @return boolean True on valid authorization, false on failure to authorize
+     * @param int $junctionID The ID of the junction permission.
+     * @return boolean True on valid authorization, false on failure to authorize
      */
-    public function checkRankedPermission($permission, $fullMatch = true, $junctionTable = '', $junctionID = '') {
+    public function checkRankedPermission($permission, $junctionTable = '', $junctionID = '') {
         $permissionsRanked = array(
             'Garden.Settings.Manage',
             'Garden.Community.Manage',
@@ -89,7 +88,7 @@ class Gdn_Session {
              */
             if ($currentPermissionRank !== false) {
                 for ($i = 0; $i <= $currentPermissionRank; $i++) {
-                    if ($this->checkPermission($permissionsRanked[$i], $fullMatch, $junctionTable, $junctionID)) {
+                    if ($this->checkPermission($permissionsRanked[$i], false, $junctionTable, $junctionID)) {
                         return true;
                     }
                 }
@@ -97,7 +96,7 @@ class Gdn_Session {
         }
 
         // Check to see if the user has at least the given permission.
-        return $this->checkPermission($permission, $fullMatch, $junctionTable, $junctionID);
+        return $this->checkPermission($permission, false, $junctionTable, $junctionID);
     }
 
     /**
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -59,11 +59,9 @@ class Gdn_Session {
      * Check the given permission, but also return true if the user has a higher permission.
      *
      * @param bool|string $permission The permission to check.  Bool to force true/false.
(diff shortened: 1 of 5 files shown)


tburry:
This shouldn't be on the configuration object. I was thinking more something like `Gdn_Session`.

```php
/**
 * Check the given permission, but also return true if the user has a higher permission.
 */
public function checkRankedPermission($name, ...) {
    if ($name === true) {
        return true;
    } elseif ($name === false) {
        return false;
    } elseif (in_array($name, []) {
        // Check to see if the user has at least the given permission.
    } else {
        return $this->checkPermission($name);
    }
}

// This can be combined with config settings to turn features on/off depending on a permission.
if (Gdn::session()->checkRankedPermission(c('Garden.Profile.EditPhotos')) {
    // photo editing is allowed.
}
```

initvector:
Ranked permission checking has been yanked out of `Gdn_Config` and pulled it into `Gdn_Session`.

library/core/class.session.php
```diff
@@ -56,6 +56,51 @@ public function addPermissions($perms) {
...
+     *
+     * @param mixed $permission The permission (or array of permissions) to check.
+     * @param bool $fullMatch If $Permission is an array, $FullMatch indicates if all permissions specified are required. If false, the user only needs one of the specified permissions.
+     * @param string $junctionTable The name of the junction table for a junction permission.
+     * @param in $junctionID The ID of the junction permission.
+     * * @return boolean True on valid authorization, false on failure to authorize
+     */
+    public function checkRankedPermission($permission, $fullMatch = true, $junctionTable = '', $junctionID = '') {
+        $permissionsRanked = array(
+            'Garden.Settings.Manage',
```
tburry: Isn't this list backwards?

library/core/class.session.php
```diff
@@ -56,6 +56,51 @@ public function addPermissions($perms) {
...
+             * user.  This is done so that a user with a permission like Garden.Settings.Manage can still validate
+             * permissions against a Garden.Moderation.Manage permission check, without explicitly having it
+             * assigned to their role.
+             */
+            if ($currentPermissionRank !== false) {
+                for ($i = 0; $i <= $currentPermissionRank; $i++) {
+                    if ($this->checkPermission($permissionsRanked[$i], $fullMatch, $junctionTable, $junctionID)) {
+                        return true;
+                    }
+                }
```
tburry: Add `return false` after the loop to avoid a redundant permission check.

library/core/class.session.php
```diff
@@ -56,6 +56,51 @@ public function addPermissions($perms) {
...
     /**
+     * Check the given permission, but also return true if the user has a higher permission.
+     *
+     * @param mixed $permission The permission (or array of permissions) to check.
+     * @param bool $fullMatch If $Permission is an array, $FullMatch indicates if all permissions specified are required. If false, the user only needs one of the specified permissions.
+     * @param string $junctionTable The name of the junction table for a junction permission.
+     * @param in $junctionID The ID of the junction permission.
+     * * @return boolean True on valid authorization, false on failure to authorize
+     */
+    public function checkRankedPermission($permission, $fullMatch = true, $junctionTable = '', $junctionID = '') {
```
tburry: Ditch the `$fullMatch = true, $junctionTable = '', $junctionID = ''` parameters for now. I really don't like those and would rather they stay out until we can refactor into a nicer representation.

initvector:
`$fullMatch`, `$junctionTable` and `$junctionID` have been ditched.

library/core/class.session.php
```diff
@@ -56,6 +56,49 @@ public function addPermissions($perms) {
...
+            $currentPermissionRank = array_search($permission, $permissionsRanked);
+
+            /**
+             * If the current permission is in our ranked list, iterate through the list, starting from the highest
+             * ranked permission down to our target permission, and determine if any are applicable to the current
+             * user.  This is done so that a user with a permission like Garden.Settings.Manage can still validate
+             * permissions against a Garden.Moderation.Manage permission check, without explicitly having it
+             * assigned to their role.
+             */
+            if ($currentPermissionRank !== false) {
```
tburry: This will never be false because of the above `in_array()` call, right?

tburry:
This looks good. I'm worried now that RoleModel->RankedPermissions will be confused with this. I think this is a better use of the name tho. I'll have to think of a different term for the one in rolemodel.

PR #3292, opened 2015-11-17
Respect format coming from params on comment preview

JasonBarnabe:
I have a [plugin](https://github.com/JasonBarnabe/ChooseYourOwnFormat) that lets users choose whether to post in Markdown or HTML.

In 2.2, Vanilla is not reading the `Format` parameter on an AJAX comment preview. It reads it correctly when posting the comment, as well as posting or previewing discussions, but just not previewing comments. This PR fixes that.


JasonBarnabe:
The fix is in 2.2.1.

PR #3296, opened 2015-11-17
Fix class constructors for forwards compatibility

tburry:
PHP 7 deprecates class constructors named with the class name. Renaming them to __constuct() ensures that we will be compatible with future versions of PHP.


tburry:
Many of these classes are never used. The main exception is Phpass which is fundamental to signing in.

PR #3301, opened 2015-11-19
Fix strict mode error in role saving by forcing PersonalInfo to 0 or 1

linc:


applications/dashboard/models/class.permissionmodel.php
```diff
@@ -702,6 +702,7 @@ public function getPermissions($RoleID, $LimitToSuffix = '') {
         $Result = array();
 
         $GlobalPermissions = $this->GetGlobalPermissions($RoleID, $LimitToSuffix);
+        print_r($GlobalPermissions);
```
beckyvb: Should this be here?

linc: oof

PR #3316, opened 2015-11-21
Eliminate Garden.Charset

linc:
Hardcode all charset values to UTF-8.

https://github.com/vanilla/vanilla/pull/3038#issuecomment-137195756


initvector:
Solid.

PR #3334, opened 2015-11-24
Replace getUserInfo() with setUserInfo()

R-J:
"getSomething" is a wrong name for a method that only sets internal values, but returns always true. So I have changed "getUserInfo" to "setUserInfo".

Moreover I've eliminated the extra "$this->editMode" calls and implemented it as a "setUserInfo" parameter.

### DIFF
--- a/applications/dashboard/controllers/class.profilecontroller.php
+++ b/applications/dashboard/controllers/class.profilecontroller.php
@@ -1,1610 +1,1732 @@
-<?php if (!defined('APPLICATION')) exit();
-
+<?php
 /**
  * Manages individual user profiles.
  *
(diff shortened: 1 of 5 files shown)


linc:
Source was deleted.

R-J:
Too bad I did something stupid like that. I do not know what I have changed exactly, but my point that the naming for that function is misleading is still there. What is your opinion about renaming ProfileControllers getUserInfo() to setUserInfo()?

The method sets up some of the controllers data and fires the event "UserLoaded". It is always used in the way `$this->getUserInfo();` so that the returned result of the method isn't of any interest at all (it always returns true). To me that is a strong indicator that it doesn't **get**s anything, but **set**s something

PR #3338, opened 2015-11-24
Add X-Path-Info for improved host compatibility

initvector:
We're trying to move away from using the p URL parameter.  Path info was our primary means to do this, but it isn't compatible with a few big hosts.  This update introduces X-Path-Info to Gdn_Request and .htaccess, which should improve that compatibility.

This should wait for merging until post-2.2.

Tested on GoDaddy.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,23 +1,25 @@
 # Original
 # If you modify this file then change the above line to: # Modified
+
 <IfModule mod_rewrite.c>
     RewriteEngine On
+
     # Certain hosts may require the following line.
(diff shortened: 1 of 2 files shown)


Daazku:
That won't work with a setup using mod_fastgi && php-fpm.

You will end up with apache rewriting X_REDIRECT as REDIRECT_X_REDIRECT before sending the header to php-fpm.
Tested on my machine.

initvector:
This PR is redundant since fd68f4ddfbf33fe239d3ee8553e00ba839acef53

Closing.

PR #3345, opened 2015-11-26
Feature/decho enhancement

Daazku:
Improve the decho function.

1. Now print the line and file from which the function was called. (Way easier to track down where you called those multiple dechos while working)
2. Now print "falsy" values wrapped in safePrint{VALUE}. This is a big one because [0,"",'',false,null] are returning nothing on screen when printed using print_r. (Also true return 1 so it's wrapped as well)

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -792,6 +792,36 @@ if (!function_exists('ConsolidateArrayValuesByKey')) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


francisbo:
Good improvement to decho.  There were a few empty fields it didn't quite catch though.

Daazku:
This pull request has been recreated correctly here: #3346

PR #3346, opened 2015-11-26
decho function enhancement

Daazku:
Improve the decho function

1. Prepend the file and line from which the function was called from to the output.
2. Replace "falsy" values by their string representation wrapped by "safePrint{}". This is done to circumvent the fact that print_r (and echo) print nothing for [false,null,0,(empty string)] and we cannot easily know what is the value when nothing is printed. (Also true become 1 so it's replaced as well)

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -804,19 +804,29 @@ if (!function_exists('safePrint')) {
 
         $functionName = __FUNCTION__;
 
-        $replaceCastedValues = function(&$value) use ($functionName) {
-            if ($value === true) { $value = $functionName.'{true}'; }
-            elseif ($value === false) { $value = $functionName.'{false}'; }
-            elseif ($value === null) { $value = $functionName.'{null}'; }
-            elseif ($value === '') { $value = $functionName.'{empty string}'; }
-            elseif ($value === 0) { $value = $functionName.'{0}'; }
+        $replaceCastedValues = function(&$value) use (&$replaceCastedValues, $functionName) {
+            if (is_array($value) || is_object($value)) {
+                foreach($value as &$content) {
+                    $replaceCastedValues($content);
+                }
+                unset($content);
+                return;
+            }
+
+            if ($value === '') {
+                $value = $functionName.'{empty string}';
+            } elseif ($value === true) {
+                $value = $functionName.'{true}';
+            } elseif ($value === false) {
+                $value = $functionName.'{false}';
+            } elseif ($value === null) {
+                $value = $functionName.'{null}';
+            } elseif ($value === 0) {
+                $value = $functionName.'{0}';
+            }
         };
 
-        if (is_string($mixed)) {
-            $replaceCastedValues($mixed);
-        } else {
-            array_walk_recursive($mixed, $replaceCastedValues);
-        }
+        $replaceCastedValues($mixed);
 
         return print_r($mixed, $returnData);
     }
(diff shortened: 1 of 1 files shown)


Daazku:
@initvector said something worthy of mention. Instead of creating a safePrint function we could use var_export. I personally prefer the print_r version but that's something we could debate.
On the plus side for var_export, it can read protected and private attributes from objects.

var_export:
```php
decho called from /var/repos/vanilla/applications/vanilla/controllers/class.categoriescontroller.php line: 124
DEBUG: CategoriesController::__set_state(array(
   'Uses' => 
  array (
    0 => 'Database',
    1 => 'Form',
    2 => 'CategoryModel',
  ),
   'CategoryModel' => 
  CategoryModel::__set_state(array(
     'Watching' => false,
     'JoinRecentUsers' => true,
     'Data' => NULL,
     'Database' => 
    Gdn_Database::__set_state(array(
       'ClassName' => 'Gdn_Database',
       '_CurrentResultSet' => 
      Gdn_DataSet::__set_state(array(
         'Connection' => 
        PDO::__set_state(array(
        )),
         '_Cursor' => -1,
         '_DatasetType' => 'array',
         '_EOF' => false,
         '_PDOStatement' => NULL,
```

Current version:
```php
decho called from /var/repos/vanilla/applications/vanilla/controllers/class.categoriescontroller.php line: 124
DEBUG: CategoriesController Object
(
    [Uses] => Array
        (
            [0] => Database
            [1] => Form
            [2] => CategoryModel
        )

    [CategoryModel] => CategoryModel Object
        (
            [Watching] => safePrint{false}
            [JoinRecentUsers] => safePrint{true}
            [Data] => safePrint{null}
            [Database] => Gdn_Database Object
                (
                    [ClassName] => Gdn_Database
                    [_CurrentResultSet:Gdn_Database:private] => Gdn_DataSet Object
                        (
                            [Connection] => PDO Object
                                (
                                )

                            [_Cursor:Gdn_DataSet:private] => -1
                            [_DatasetType:protected] => array
                            [_EOF:protected] => 
                            [_PDOStatement:Gdn_DataSet:private] => 
```

linc:
I'd rather simply use `var_export` for 2 reasons:

* Less code.
* Using PHP's baked-in functionality is always a better default. I don't want to build new tools when we have an equivalent one already.

I like the file & line number addition, but wonder if this would be more efficient for reading, since I think we can infer this output is indeed from `decho()` and that we're in our `PATH_ROOT`:

```php
$backtrace = 'Line '.$stack[0]['line'].' in '.$stack[0]['file']."\n";
if (defined('PATH_ROOT')) {
    $backtrace = str_replace(PATH_ROOT, '', $backtrace);
}
```

Daazku:
The shortening of the file path is a must do IMO.

For the use of another function than print_r I'm still not convinced.
The reason I created the safePrint function is to preserve the current decho function as much as possible. I summed up some pros and cons of the 3 functions we can use:

print_r():
* pros
 * Designed to be human readable
 * It has the prettiest output and easy to read output
 * It is what decho already use
* cons
 * I think the function use echo internally (falsy values are just blank and true is converted to 1) <-- that is what the safePrint function was for.

var_dump:
* pros
 * You cannot misinterpret the output
 * You have some nice formatting and color with xDebug (pretty widely used)
 * The basic formatting is well done.
* cons
 * Formatting without xDebug is actually not really nice...
 * Noisy.... like.. really noisy output. No seriously!

var_export():
* pros
 * You cannot misinterpret the output
 * The output is actually valid PHP code
* cons
 * Generate notices about circular references
 * http://php.net/manual/en/function.var-export.php#refsect1-function.var-export-notes
 * The output formatting could use some love (test it on big objects)!

linc:
Great. Get the line fix in and we can merge.

PR #3365, opened 2015-12-02
Set plaintext email styles default.

beckyvb:
Set config setting in preparation for new HTML email feature.


linc:
Let's drop 'Styles.' from this so we don't create our first 4-level config.

Issue #3298, opened 2015-11-18
Reset Link in error message now escaped.


hgtonight:
237c4339b0181f5c4d838a8ead351d35b4c47c43

This breaks the reset password link in the error message when signing in with hash type 'reset'.

Also present on 2.2.

PR #3367, opened 2015-12-03
Filter form errors rather than htmlspecialchars

linc:
Revises 237c4339b0181f5c4d838a8ead351d35b4c47c43 to use purification instead of escaping. As @hgtonight points out, this disabled links in error messages even internally, which are useful and good UI. Needs to be reevaluated against the security condition that prompted the change in the first place.


linc:
Reference https://github.com/vanilla/vanilla-patches/pull/22

My opinion is we should be blocking user-generated strings (the filename) from the error message rather than modifying this core component.

linc:
Closes #3298

PR #3375, opened 2015-12-04
Remove variable variable references in the managers

tburry:
These throw notices in PHP 7.


hgtonight:
Looks cleaner too. +1

PR #3380, opened 2015-12-07
Refactor methods to match base class signatures

tburry:
This will get rid of PHP 7 notices and allow us to use our models in a more OOP way.

- [x] Merge #3375.
- [x] Merge #3378.
- [x] Rebase this PR to master.
- [x] Clean up remaining notices that are easily found.


tburry:
I refactored this against master and fixed the merge conflict.

PR #3381, opened 2015-12-07
Validate the utility/update Target before redirecting.

beckyvb:
Makes sure the Target param is safe before redirecting in utility/update


linc:
I think it would be better to use `safeRedirect()` so that trusted domains are allowed as well.

Issue #3369, opened 2015-12-03
Bans unable to be deleted in Dashboard


JulieVanilla:
Bans are currently not being properly deleted.

1. Go to the dashboard as an administrator
2. Go to Moderation -> Banning
3. Click “Add Item” to ban an IP address
4. After the IP ban is in place, click the Delete button to remove it.
5. The IP ban shows as removed properly.
6. Refresh the page.
 
BUG: The IP ban is visible again, as though it was undeleted, or as though it were never actually deleted.

This actually is happening for names and email addresses as well. Reproduced on my test site.

**Expected functionality**

Bans should be deleted when the administrator deletes them.

PR #3383, opened 2015-12-09
Fix CSRF check when deleting a ban

initvector:
How the "delete ban" workflow currently happens:

1. User clicks "Delete" on a record row on /settings/bans.
2. User clicks "Okay" in the popup to confirm the deletion.
3. `SettingsController::bans` checks to see if `isAuthenticatedPostback` and fails, because the request is not sent with POST, nor does it contain the session's `TransientKey`.

This update adds the transient key to the delete link by way of the `Gdn_Theme::link` function and has `SettingsController::bans` check for a valid transient key in the URL parameters before committing the deletion.

### DIFF
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -499,7 +499,7 @@ class SettingsController extends DashboardController {
                 $this->View = 'Ban';
                 break;
             case 'delete':
-                if ($this->Form->authenticatedPostBack()) {
+                if (Gdn::session()->validateTransientKey(Gdn::request()->get('TransientKey'))) {
                     $BanModel->delete(array('BanID' => $ID));
                     $this->View = 'BanDelete';
                 }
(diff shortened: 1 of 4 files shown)


linc:
We're trying to avoid passing the TransientKey during additional `GET` requests. Is it possible to remedy this via `POST`?

initvector:
It's a little trickier, since the request goes through a confirmation popup.  If we want to keep things the way they are, but have the request made over POST, it might be necessary to edit our popup JavaScript to allow post request as part of its configuration on a link/button.

linc:
I'm OK with some tinkering necessary; part of our ongoing security effort has been to get this out of the URL entirely, so I'm loath to take a step backwards here and add it in. That said, if it's a many-hour modification then let's table it for now and just use this.

initvector:
No big deal.  I've updated our `jquery.popup` capabilities to allow specification of request method.  It defaults to `GET`, so everything else relying on these popups should continue to function without any updates.  If the method is `POST`, the user's transient key is added to the request data.

Daazku:
Related: #3369

PR #3384, opened 2015-12-10
Allow forced embed on mobile

initvector:
Vanilla explicitly avoids forced embedding of a forum when viewing on a mobile device.  This update removes that restriction, allowing mobile browsers to also be forced into embedding.

I'm not sure if this was put in place 3+ years ago for UE or compatibility.  I didn't run into any issues with compatibility and I figure the UE for this kind of advanced implementation should be the responsibility of the admin/developer/designer, if they want to pursue it.

### DIFF
--- a/applications/dashboard/controllers/class.embedcontroller.php
+++ b/applications/dashboard/controllers/class.embedcontroller.php
@@ -114,7 +114,16 @@ class EmbedController extends DashboardController {
 
         $Validation = new Gdn_Validation();
         $ConfigurationModel = new Gdn_ConfigurationModel($Validation);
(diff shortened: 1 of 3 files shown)


linc:
This is going to work. Right now the expected behavior would be that mobile exists with a separate theme that may not work via embed (see: Lithe). Lots of folk have "Force embed" enabled under this assumption, and suddenly removing this condition is going to cause chaos. To accomplish this, you'd need a secondary option in the Dashboard to explicitly force embed for mobile that this then respects as an override.

initvector:
Advanced configuration area for embedding in the admin area has been updated to include a separate option for forcing mobile devices.

PR #3393, opened 2015-12-15
Update Advanced Editor to show Html as Wysiwyg

initvector:
Advanced Editor currently renders posts with `Html` set as their format as markup.  This update moves to make these posts render in the same way as posts with `Wysiwyg` as their format: in WYSIWYG with a toolbar button to toggle the HTML code view.

### DIFF
--- a/plugins/editor/class.editor.plugin.php
+++ b/plugins/editor/class.editor.plugin.php
@@ -11,7 +11,7 @@
 $PluginInfo['editor'] = array(
    'Name' => 'Advanced Editor',
    'Description' => 'Enables advanced editing of posts in several formats, including WYSIWYG, simple HTML, Markdown, and BBCode.',
-   'Version' => '1.7.6',
+   'Version' => '1.7.7',
    'Author' => "Dane MacMillan",
    'AuthorUrl' => 'http://www.vanillaforums.org/profile/dane',
    'RequiredApplications' => array('Vanilla' => '>=2.2'),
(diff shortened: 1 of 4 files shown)


linc:
Does this effect all use of HTML or strictly the "reinterpret" use case you were trying to solve? This seems like it's going to effectively end the use of HTML as a separate rendering.

initvector:
This update only makes Advanced Editor initially display posts with the `Html` format in the WYSIWYG view.  The code toggle button is now present for `Html` formatted posts too, so the raw markup (how `Html` was previously displayed) can still be viewed.

linc:
That's not how the `Html` option is intended to work - that's how `Wysiwyg` works.

The issue was that "reinterpretting" an `Html` post as `Wysiwyg` via the editor settings would cause an issue. This pull request seems to remove the distinction between how Html & Wysiwyg function as independent input formatters rather than targeting that reinterpretting issue specifically.

initvector:
The avoidance of reinterpreting `Html` as `Wysiwyg` was explicitly implemented in the Advanced Editor plug-in.  Just removing that exception allows the post to be "reinterpreted" as `Wysiwyg`.  If that's all we're aiming to do, I have that in a separate PR, here: https://github.com/vanilla/vanilla/pull/3401

initvector:
Closing in favor of https://github.com/vanilla/vanilla/pull/3401

PR #3398, opened 2015-12-17
Split Travis CI into a lint and test matrix

tburry:
Make the linting and unit testing run seperately so they can be concurrent.


linc:
Is this still a WIP?

tburry:
I think I'm going to close this. I'm just not sure about it it being necessary if we can get more unit tests and less API calls.

PR #3401, opened 2015-12-18
Remove Html exception for Wysiwyg reinterpretation

initvector:
This update removes the exception for `Html` for being reinterpreted as `Wysiwyg` in Advanced Editor.

Re: https://github.com/vanilla/vanilla/pull/3393#issuecomment-165848168


linc:
This will give it a trip thru `Gdn_Format::to()` it doesn't need, but on balance, easier to let it do that than muddy the code further.

PR #3409, opened 2015-12-29
Editor: Use the actual file name for attachment downloads

bleistivt:
When downloading an attachment, this will retain the actual filename (as opposed to a random string).

Note, that this will also show the download prompt for image files.

[More info on the download attribute](https://davidwalsh.name/download-attribute)


linc:
Thanks!

PR #3410, opened 2015-12-30
Pass license info when analyzing an addon

linc:
Need to add a License field check to analyzed addons. To do this would duplicate code in 4 places, so we abstract it (along with the Description and Version checks) to new method `checkRequiredFields()` and use its results to detect failed validation.


linc:
Required for https://github.com/vanilla/community/issues/17

hgtonight:
Give it to him!

+1

PR #3415, opened 2016-01-02
Fix draft saving

### LINKED ISSUE #3411: New discussion draft save requires category & name
This creates a form error out of nowhere when you're just typing a discussion. Also blocks you from actually saving the draft, and isn't saving incomplete data the point of a draft?

bleistivt:
Spaces in field names need to be replaced with underscores because of https://github.com/vanilla/vanilla/commit/4435940f8c486ab1c473e8245e8e34be26f8c373

 fixes #3411


linc:
I just sat down to debug this because it's blocking a deployment. Many thanks!

PR #3417, opened 2016-01-03
reCAPTCHA 2

x00:
This is the reCAPTCHA 2 PR as promised. 

Possibly an addition could be a more explicit warning if they have not set up the keys. At moment it will say on the form. 

You don't really have anything like that, and the stable release has a translation bug which either would be superseded by this or the fix need to be backpacked.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2717,7 +2717,7 @@ class UserModel extends Gdn_Model {
             if (val('CheckCaptcha', $Options, true)) {
                 $CaptchaValid = ValidateCaptcha();
                 if ($CaptchaValid !== true) {
-                    $this->Validation->addValidationResult('Garden.Registration.CaptchaPublicKey', 'The reCAPTCHA value was not entered correctly. Please try again.');
+                    $this->Validation->addValidationResult('Garden.Registration.CaptchaPublicKey', 'The Security Check was not completed correctly. Please try again.');
                     return false;
                 }
             }
@@ -2788,7 +2788,7 @@ class UserModel extends Gdn_Model {
                 $CaptchaPublicKey = val('Garden.Registration.CaptchaPublicKey', $FormPostValues, '');
                 $CaptchaValid = validateCaptcha($CaptchaPublicKey);
                 if ($CaptchaValid !== true) {
-                    $this->Validation->addValidationResult('Garden.Registration.CaptchaPublicKey', 'The reCAPTCHA value was not entered correctly. Please try again.');
+                    $this->Validation->addValidationResult('Garden.Registration.CaptchaPublicKey', 'The Security Check was not completed correctly. Please try again.');
                     return false;
                 }
             }
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -267,6 +267,7 @@ class Gdn_Form extends Gdn_Pluggable {
      * @return string
      */
     public function captcha() {
+
         $handled = false;
         $this->EventArguments['Handled'] =& $handled;
         $this->fireEvent('Captcha');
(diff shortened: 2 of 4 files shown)


sebastianbond:
:-1: This has some issues, it makes a post live when the a draft is saved.

Would be great to change the capatcha to this one.

x00:
hmm doesn't sound very likely. 

Can you provide steps to repeat.

sebastianbond:
Could be me, I just pasted class.form.php and functions.validation.php the files here on top of the ones in my installation for 2.2. 

I didn't do the other ones as the changes were just copy.

sebastianbond:
Just tried it again - the right way by only copying the capatcha functions from the two files above and its all fine. I guess this wasn't branched off the same code as 2.2?

x00:
copy and pasting is not how you do it. The is a PR with a commit in a specif time and place. You would need to merge commits and resole any conflicts. 

There is no point creating a fork unless you are goign to support it, or you know this PR will be adopted.

library/core/functions.validation.php
```diff
@@ -27,16 +27,45 @@
...
-        require_once PATH_LIBRARY.'/vendors/recaptcha/functions.recaptchalib.php';
-
-        $CaptchaPrivateKey = c('Garden.Registration.CaptchaPrivateKey', '');
-        $Response = recaptcha_check_answer(
-            $CaptchaPrivateKey,
-            Gdn::Request()->IpAddress(),
-            Gdn::Request()->Post('recaptcha_challenge_field', ''),
-            Gdn::Request()->Post('recaptcha_response_field', '')
+        
+        $response = arrayValue('g-recaptcha-response', $_POST, '');
```
linc: Accessing $_POST directly is not correct. Gdn::request() has a method for this.

applications/dashboard/models/class.usermodel.php
```diff
@@ -2788,7 +2788,7 @@ public function insertForBasic($FormPostValues, $CheckCaptcha = true, $Options =
                 $CaptchaPublicKey = val('Garden.Registration.CaptchaPublicKey', $FormPostValues, '');
                 $CaptchaValid = validateCaptcha($CaptchaPublicKey);
                 if ($CaptchaValid !== true) {
-                    $this->Validation->addValidationResult('Garden.Registration.CaptchaPublicKey', 'The reCAPTCHA value was not entered correctly. Please try again.');
+                    $this->Validation->addValidationResult('Garden.Registration.CaptchaPublicKey', 'The Security Check was not completed correctly. Please try again.');
```
linc: I agree with rephrasing this a little, but we can still call it "reCAPTCHA" instead of obfuscating it as a "security check". Also we'll wanna get this into the locales repo ASAP.

library/core/class.form.php
```diff
@@ -267,45 +267,37 @@ public function calendar($FieldName, $Attributes = false) {
      * @return string
      */
     public function captcha() {
+
```
linc: We don't add blank lines after function definitions.

library/core/class.form.php
```diff
@@ -267,45 +267,37 @@ public function calendar($FieldName, $Attributes = false) {
...
         $this->fireEvent('Captcha');
         if ($handled) {
             // A plugin handled the captcha so don't display anything more.
             return;
         }
-        // Google whitelist
-        $whitelist = array('ar', 'bg', 'ca', 'zh-CN', 'zh-TW', 'hr', 'cs', 'da', 'nl', 'en-GB', 'en', 'fil', 'fi', 'fr', 'fr-CA', 'de', 'de-AT', 'de-CH', 'el', 'iw', 'hi', 'hu', 'id', 'it', 'ja', 'ko', 'lv', 'lt', 'no', 'fa', 'pl', 'pt', 'pt-BR', 'pt-PT', 'ro', 'ru', 'sr', 'sk', 'sl', 'es', 'es-419', 'sv', 'th', 'tr', 'uk', 'vi');
+        
+        if (!c('Garden.Registration.CaptchaPublicKey') || !c('Garden.Registration.CaptchaPrivateKey')) {
+            return '<div class="Warning">' . t('Security Check has not been set up by the site administrator in registration settings. This is required to register.') .  '</div>';
```
linc: Good enhancement. Still should just call it "reCAPTCHA" here too, tho.

library/core/functions.validation.php
```diff
@@ -27,16 +27,45 @@
...
+        }
+
+        $url = 'https://www.google.com/recaptcha/api/siteverify';
+        $data = array(
+            'secret' => c('Garden.Registration.CaptchaPrivateKey'),
+            'response' => $response
         );
-        return $Response->is_valid ? true : 'The reCAPTCHA value was not entered correctly. Please try again.';
+
+        $handler = curl_init();
```
linc: Did you look at using ProxyRequest for this instead of opening curl directly?

library/core/functions.validation.php
```diff
@@ -27,16 +27,45 @@
...
+
+        $response = curl_exec($handler);
+
+        if ($response) {
+            $result = json_decode($response);
+            $errorCodes = val('error_codes', $result);
+            if ($result && val('success', $result)) {
+                return true;
+            } else if (!empty($errorCodes) && $errorCodes != array('invalid-input-response')) {
+                throw new Exception(formatString(t('Could not get check if human! Error codes: {ErrorCodes}'), array('ErrorCodes' => join(', ', $errorCodes))));
```
linc: "Could not get reCAPTCHA response" would be a better string to use both here and below. We can just append the error codes without wrapping it into the translation or adding "Error codes: " as another label to translate.

library/core/functions.validation.php
```diff
@@ -27,16 +27,45 @@
...
         );
-        return $Response->is_valid ? true : 'The reCAPTCHA value was not entered correctly. Please try again.';
+
+        $handler = curl_init();
+        curl_setopt($handler, CURLOPT_URL, $url);
+        curl_setopt($handler, CURLOPT_PORT, '443');
+        curl_setopt($handler, CURLOPT_SSL_VERIFYPEER, false);
+        curl_setopt($handler, CURLOPT_HEADER, false);
+        curl_setopt($handler, CURLOPT_HTTPHEADER, array("Content-Type: application/x-www-form-urlencoded"));
+        curl_setopt($handler, CURLOPT_USERAGENT, arrayValue('HTTP_USER_AGENT', $_SERVER, 'NoCaptchaReCaptcha Vanilla'));
```
linc: $_SERVER access

PR #3423, opened 2016-01-05
Extend drafts to store types and attributes and add events for save

austins:
I've noticed that the DraftModel doesn't store the discussion type, which is set by addons, such as QnA when asking a question, while creating/editing a new discussion via the PostController.

When a draft is saved and the user wants to edit it, the type and additional related form data has been lost and the draft has reverted to a standard discussion, so the post view for the custom type won't show up when the draft is opened for editing.

Relevant vanillaforums.org discussion: http://vanillaforums.org/discussion/31426/storing-discussion-type-with-drafts

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -162,6 +162,10 @@ class PostController extends VanillaController {
             $this->title(t('New Discussion'));
         }
 
+        if (isset($this->Draft) && $this->Draft->Type) {
+            $this->setData('Type', $this->Draft->Type);
+        }
+
         touchValue('Type', $this->Data, 'Discussion');
 
         // See if we should hide the category dropdown.
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -162,11 +162,7 @@ class PostController extends VanillaController {
             $this->title(t('New Discussion'));
         }
 
-        if (isset($this->Draft) && $this->Draft->Type) {
-            $this->setData('Type', $this->Draft->Type);
-        }
-
-        touchValue('Type', $this->Data, 'Discussion');
+        touchValue('Type', $this->Data, val('Type', $this->Draft, 'Discussion'));
 
         // See if we should hide the category dropdown.
         $AllowedCategories = CategoryModel::getByPermission('Discussions.Add', $this->Form->getValue('CategoryID', $this->CategoryID), array('Archived' => 0, 'AllowDiscussions' => 1), array('AllowedDiscussionTypes' => $this->Data['Type']));
--- a/applications/vanilla/models/class.draftmodel.php
+++ b/applications/vanilla/models/class.draftmodel.php
@@ -167,6 +167,11 @@ class DraftModel extends VanillaModel {
             unset($FormPostValues['Sink']);
         }
 
+        // Prep and fire event
+        $this->EventArguments['FormPostValues'] = &$FormPostValues;
+        $this->EventArguments['DraftID'] = $DraftID;
+        $this->fireEvent('BeforeSaveDraft');
+
         // Validate the form posted values
         if ($this->validate($FormPostValues, $Insert)) {
             $Fields = $this->Validation->SchemaValidationFields(); // All fields on the form that relate to the schema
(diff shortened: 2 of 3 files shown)


applications/vanilla/models/class.draftmodel.php
```diff
@@ -183,6 +188,12 @@ public function save($FormPostValues) {
                 $DraftID = $this->SQL->insert($this->Name, $Fields);
                 $this->UpdateUser($Session->UserID);
             }
+
+            // Fire an event that the draft was saved
+            $this->EventArguments['FormPostValues'] = $FormPostValues;
```
linc: This is redundant since you passed this variable by reference earlier.

applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -162,6 +162,10 @@ public function discussion($CategoryUrlCode = '') {
             $this->title(t('New Discussion'));
         }
 
+        if (isset($this->Draft) && $this->Draft->Type) {
```
linc: `if (val('Type', $this->Draft))` would handle this.

linc: Actually I'd just ditch this `if` entirely and just go with:

`$this->setData('Type', val('Type', $this->Draft, null));`

linc:
Since I have background it doesn't matter so much, but in general you should always include an adequate description with rationale for the change, including any links to relevant .org discussions. You and I know what's going on here, but no one else does, and this is a public discussion so it's just rude to be obscure. :)

austins:
@linc, thank you for those great suggestions. I just copied some lines over from the DiscussionModel and didn't look things through in detail. I've updated the first post with more info about the changes. I'll submit a cleaner pull request soon. Is there a way to add new commits to the same pull request or should I make a new one?

linc:
You can keep using this one and I generally encourage you to do so unless a PR gets truly messed up. Just make a new commit on the same branch and push to the same repo, and it'll show up here.

applications/vanilla/controllers/class.postcontroller.php
austins: Good suggestion. `$this->setData('Type', val('Type', $this->Draft, null));` could go in the else closure of the `if (isset($this->Discussion))` block.

linc:
In the future, I recommend using a `feature/x` branch on your repo instead of `master` so that your `master` doesn't get blocked by needing to wait on a merge before you can do anything else on it (otherwise it'll all show up here too).

applications/vanilla/controllers/class.postcontroller.php
austins: Actually, what about modifying the touchValue line under that to: `touchValue('Type', $this->Data, val('Type', $this->Draft, 'Discussion'));`?

austins:
@linc, I've been trying to figure out how to make the branches work since I've recently installed the GitHub desktop client. There was a button on the website before that easily created another branch for pull requests automatically, but this time it didn't create one for some reason. I'll make sure that happens next time.

applications/vanilla/controllers/class.postcontroller.php
austins: That would make the `touchValue('Type', $this->Data, 'Discussion');` line not set the type to `Discussion` if it's set to null since the array key exists.

applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -162,7 +162,7 @@ public function discussion($CategoryUrlCode = '') {
             $this->title(t('New Discussion'));
         }
 
-        touchValue('Type', $this->Data, 'Discussion');
+        touchValue('Type', $this->Data, val('Type', $this->Draft, 'Discussion'));
```
linc: This is not safe because `$this->Draft` may not exist.

linc:
@austins See my note above re: `$this->Draft`.

austins:
If the changes discussed in pull request #3438 are merged, I believe we can settle on one of the single-line of codes of code above without having to worry about if and how the literal `Discussion` type is set by default. @linc, if you agree with this, I'll try to make another pull request with proper branching in case this one causes conflicts with the master.

linc:
Yeah, I'm onboard with Becky's suggestions.

PR #3427, opened 2016-01-05
Add support for a composer-local.json to add additional repositories and requirements to Vanilla.

tburry:
Drop a separate composer-local.json file in the same directory as the composer.json file and run `composer update`. The format of the file is a subset of the composer.local where you can specify the following:

- repositories
- require
- require-dev

Here is a sample composer-local for testing:

```json
{
	"require": {
		"vanilla/htmlawed": "*",
		"symfony/polyfill-mbstring": "~1.0"
	}
}
```


evach-vf:
Also `composer.lock` is not up to date with `composer.json` which gives a warning on `composer install`

evach-vf:
Overall, I really like this feature.  It's easy and clean to add dependencies dev or not, or to test different versions during development.

The only issue I see is if we commit our `composer.lock` with changes included from a dev `composer-local.json` which we shouldn't!

It could work as expected when a user first checkout the project and run `composer install`.
But it would likely break something when the user later run `composer update`.

I guess we should just be extra careful when committing `composer.lock`.

tburry:
The composer.lock is definitely a gotcha we'll have to monitor. I'll look into the problems with the repositories merge. Great catch.

tburry:
I've update the support for composer-local repositories. Here is the composer-local.json I tested with:

```json
{
	"repositories": [
		{
			"type": "vcs",
			"url": "git@github.com:vanilla/nbbc.git"
		}
	],
	"require": {
		"vanilla/nbbc": "*"
	}
}
```

PR #3442, opened 2016-01-11
Update default footer logo

linc:
* Delete unused logo image (probably from last iteration, safe to delete by now).
* Update footer logo to sexy black.
* Now with retina image quality.
* Retain last-generation image for now in case themes still use it.


tburry:
What's the browser support looking like for svg?

linc:
I don't know that any better than you do. Are you saying we can only update our grainy 2009 logo if it's directly to SVG? Do we even have an SVG version of the logo?

hgtonight:
> Do we even have an SVG version of the logo?

I would hope you have a vector version by whomever created this image.

hgtonight:
Also, http://caniuse.com/#feat=svg looks promising

PR #3458, opened 2016-01-12
Use a cache-busting hash from the config instead of addon version numbers

tburry:
This allows us to make use of the code-pusher’s cluster-wide etagb and remove the race condition when pushing code to multiple servers.


tburry:
This PR requires support from infrastructure. Do not merge.

tburry:
I'm closing this in favor of #3795. I'll revisit this solution later, once the addon manager is implemented.

PR #3463, opened 2016-01-13
Upgrade to PHP’s password_* functions for passwords if possible

tburry:
Attempt to use PHP 5.5’s password_* functions for password hashing and verification unless portable passwords are required.

- [x] Test on PHP 5.4 and make sure the passwords still verify.
- [x] Generate a password hash on PHP 5.5+ and then make sure PHP 5.4 can still verify it.
- [ ] Test on infrastructure.

### DIFF
--- a/library/core/class.passwordhash.php
+++ b/library/core/class.passwordhash.php
@@ -24,7 +24,7 @@ class Gdn_PasswordHash extends PasswordHash {
     /**
      * Constructor.
      */
-    function __construct() {
+    public function __construct() {
         // 8 iteration to create a Portable hash
         parent::__construct(8, false);
     }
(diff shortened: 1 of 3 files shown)


tburry:
Here is a hash I generated using `password_hash`.

Password: letmein
Hash: $2y$10$rGZuSnSep20h2wu20N1me.ExybolmiTuqfF14AJEPlBT4B.46rNSO

That hash should be able to be pasted into the user table and then verify against the plaintext password.

initvector:
Before this, I could reset someone's password in the DB by updating the `Password` field to the plain-text value.  When the person signed in, their password would be hashed and their user record would be updated.  However, I can't do this anymore.  It looks like this block may be killing it: [class.passwordhash.php:206](https://github.com/vanilla/vanilla/pull/3463/files#diff-ef24707abb9201c5d25a5144072405daR206)

tburry:
I added support for the plaintext use case and also some unit tests.

Issue #3430, opened 2016-01-06
Wysiwyg mode in Vanilla Comments breaks quoting and other formatting


linc:
See PostController:

        // Vanilla Comments save as Text, no matter the format.
        // Kludge to set format to Text if we're Wysiwyg to preserve newlines.
        if ($isEmbeddedComments && ($this->Form->getFormValue('Format', c('Garden.InputFormatter')) === 'Wysiwyg')) {
            $this->Form->setFormValue('Format', 'Text');
        }

The first comment there is a damned lie. If you're using Markdown on the forum, Markdown works in Vanilla Comments. So what we have here is a situation where ONLY Wysiwyg fails to show any formatting inserted into a comment, which includes the HTML inserted by the Quote option.

linc:
This would also block mentions from working.

initvector:
Re: https://github.com/vanilla/vanilla/pull/3275

PR #3464, opened 2016-01-13
Improve saving of embedded comments

initvector:
This update adds special considerations for formatting the body of embedded comments.  Since advanced editing capabilities aren't present when adding a comment in the embedded form, we may need to run the body through special handling to ensure the intended display is maintained (e.g. newlines).

Fixes https://github.com/vanilla/vanilla/issues/3430

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -539,7 +539,6 @@ class PostController extends VanillaController {
             $inputFormatter = $this->Form->getFormValue('Format', c('Garden.InputFormatter'));
 
             switch ($inputFormatter) {
-                case 'Html':
                 case 'Wysiwyg':
                     $this->Form->setFormValue(
                         'Body',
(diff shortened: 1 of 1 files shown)


linc:
This adds double the line breaks in `Html` mode.

linc:
screen shot 2016-01-15 at 5 37 22 pm
screen shot 2016-01-15 at 5 37 14 pm
screen shot 2016-01-15 at 5 37 02 pm

initvector:
It looks like `Gdn_Format::html` already has some [config-based newline replacement](https://github.com/vanilla/vanilla/blob/7b6d985b2dbd07d1dc7c9868caed0e53768f873e/library/core/class.format.php#L879), so this is doubling up on the break tags.  The config value this action is based on, `Garden.Format.ReplaceNewlines`, defaults to `true`.

@linc: Do you think it would be better to force newline-to-break-tag only if `Garden.Format.ReplaceNewlines` is `false`?  Or should the special handling of embedded comments be removed for the `Html` format and we'll just rely on the formatter to make the adjustments, at least in the case of newline conversion?

linc:
As that is nearly the only use of that config setting and I can't think of a situation you'd ever set that to `false` AND want comments to behave otherwise, I think just remove it and let it work as it should.

initvector:
Special case for `Html` has been removed.

PR #3466, opened 2016-01-14
Use phpunit.xml config variables for the test environment

tburry:
This will allow other devs to change settings more easily if they wish.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -16,6 +16,11 @@
         <ini name="intl.error_level" value="0"/>
         <ini name="memory_limit" value="-1"/>
         <ini name="date.timezone" value="America/Montreal"/>
+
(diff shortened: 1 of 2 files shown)


linc:
> This will allow other devs to change settings more easily if they wish.

How? By editing that file directly and just not committing it? Or can we override somehow via another file? If it's the latter, some docs for stuff like that would be helpful.

evach-vf:
It's via another file.  You would copy `phpunit.xml.dist` to `phpunit.xml` and make your change in that file.  PHPUnit will then use the config file `phpunit.xml`.  

From https://phpunit.de/manual/current/en/organizing-tests.html#organizing-tests.xml-configuration:

    If phpunit.xml or phpunit.xml.dist (in that order) exist in the current working directory and --configuration is not used, the configuration will be automatically read from that file.

PR #3467, opened 2016-01-15
Don’t register error/exception handlers in functions.error.php

tburry:
Remove the side-effects from this file so that it can be more easily used in unit testing and other libraries.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -48,6 +48,7 @@ if (function_exists('mb_internal_encoding')) {
 
 // Include the core autoloader.
 require_once __DIR__.'/vendor/autoload.php';
+setHandlers();
 
 // Initialize the autoloader.
 Gdn_Autoloader::start();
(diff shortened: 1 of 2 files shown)


tburry:
On the face of it I think this PR is pretty safe, but since it is a bootstrap-level change it will affect every page request. Some basic testing would be appreciated.

Really though the only side-effect would be that if some error happens in our core auto-loaded files then they will no longer be handled. My guess is this is a low probability and an error that occurs in what should just be function/class definitions is a development level error anyway.

initvector:
`Gdn_ErrorHandler` and `Gdn_ExceptionHandler` aren't really functional until after the following bootstrap operation:
```
Gdn::factoryInstall(Gdn::AliasRequest, 'Gdn_Request', null, Gdn::FactoryRealSingleton, 'Create')
```
Would it make more sense to move `setHandlers` beneath that, so additional errors and exceptions aren't potentially encountered in the handler, itself?

tburry:
Sorry I missed that. I think moving it is fine then.

PR #3469, opened 2016-01-17
Remove cache from getUnreadCount in DiscussionModel and some improvements

GO3LIN:
I noticed that the getUnreadCount was not working properly due to the categories cache. So I suggest this patch that improves the method by adding a count for both unread discussions and comments.


linc:
Can you elaborate on what "not working properly" means here?

I don't understand why we're also kludging CountComments functionality onto this method. This idea of returning an object/array of multiple things you _might_ want isn't very good.

Issue #3056, opened 2015-09-04
Unread status on discussions list should change onclick


linc:
We recently removed a `no-store` cache header, which had a significant side effect: Now clicking the "Back" browser button will give you a discussion home without an updated read/unread for whatever thread you were just viewing. The previous behavior was that the page would refresh; now it does not. This "broken" behavior was actually desirable because it simplified navigation.

The easiest solution would be to add an onclick handler that updates the discussion list status to switch the CSS from unread to read, and remove the `X new` element. This would keep the correct cache strategy by "fixing" the page before navigating away from the page so it is in the correct state when you return via "Back".

linc:
hat tip to @erobbins for the useful suggestion while the rest of my users were just kvetching. :laughing:

bleistivt:
Noticed this too.

Quick idea: Adding a second class would allow you to highlight the topic you just came from for a short time (e.g. using a fade effect). Discourse does this and I find it also makes navigation easier.

initvector:
Implementing the suggested fix doesn't work for me, at least in Chrome.  The browser seems to be caching what it receives from the server, but not any changes to the document between when the link is clicked and the next page is loaded.

The following JavaScript, tested with the +Baseline theme, will make the appropriate updates to a row in the discussion list. They are noticeably made when the link is clicked.  However, once the discussion page loads and the back button is clicked, the changes are lost.  The page appears as it did on initial load.

```js
$('.DataList.Discussions .Unread.ItemDiscussion .Title a').click(function (event) {
    var discussionRow = $(event.target).closest('.Unread.ItemDiscussion');

    if (discussionRow.length) {
        $(discussionRow).removeClass('Unread');
        $(discussionRow).find('.HasNew.NewCommentCount').addClass('JustNew').empty();
    }
});
```

linc:
Re-adding this header to the Discussions & Category controllers would be an appropriate fix. Those pages typically don't have form data associated with them, and even if they did in some special circumstance, the tradeoff here seems well worth it.

PR #3471, opened 2016-01-19
Add no-store header to DiscussionsController

initvector:
The default `Cache-Control` header does not include `no-store`, which can cause issues (e.g. inaccurate unread status or new comment counts) when users visit the discussion list via the browser's back button. This update adds `no-store` to the `Cache-Control` header sent by `DiscussionsController`.

Fixes https://github.com/vanilla/vanilla/issues/3056

Issue originates from https://github.com/vanilla/vanilla/pull/2948

### DIFF
--- a/applications/vanilla/controllers/class.categoriescontroller.php
+++ b/applications/vanilla/controllers/class.categoriescontroller.php
@@ -28,6 +28,19 @@ class CategoriesController extends VanillaController {
     /** @var object Category object. */
     public $Category;
 
+    public function __construct() {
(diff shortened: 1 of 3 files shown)


linc:
I think the `initialize()` method would be a better place for this; still will need to invoke the parent as well.

initvector:
The `Cache-Control ` header assignment has been moved to `initialize` in `CategoriesController`, `CategoryController` and `DiscussionsController`.

linc:
I especially :+1: the docs.

PR #3472, opened 2016-01-20
Fixed unread discussion flag when discussion just created

GO3LIN:
Scenario 1:
1. Create a discussion.
2. Call the getUnread method in DiscussionModel

Expected : The newly created discussion doesn't appear.
Seen : The newly created discussion is fetched.

PS : Even if you click to view the discussion. It's not set to viewed, I noticed after investigation that the LastDateViewed field is null after the discussion is created.


linc:
Thanks @GO3LIN, can you sign our contributors' agreement so I can review this? https://vanillaforums.org/contributors

GO3LIN:
@linc You are welcome. I have signed the contributors agreement, I have the same login in your plateform.

initvector:
Seems sane and functional to me.

I was initially concerned a populated `DateLastViewed` on creation might cause issues with determining if a thread had new content or not (showing it was "new" to the user immediately after creation), but that doesn't seem to be the case.  The "new" status is determined by if a discussion's `DateLastComment` is *greater than* `DateLastViewed` in `UserDiscussion`.  It doesn't matter if they're equal.

PR #3478, opened 2016-01-20
Add a user’s role to the gdn_meta to be accessed by javascript.

nervoustwit:


applications/dashboard/models/class.rolemodel.php
```diff
@@ -750,4 +750,41 @@ public function deleteAndReplace($roleID, $newRoleID) {
...
+        }
+
+        // If an empty string is passed as the field, return all the data from gdn_role row.
+        if (!$field) {
+            return $unformattedRoles;
+        }
+
+        // If there is a return key, return an array with the field as the key
+        // and the value of the field as the value.
+        $formattedRoles = array();
```
linc: use array_column.  see profileController::getUserInfo().

applications/dashboard/models/class.rolemodel.php
```diff
@@ -750,4 +750,41 @@ public function deleteAndReplace($roleID, $newRoleID) {
...
     }
+
+    /**
+     * Get a list of a user's roles, optionally return all the role data or just one field name.
+     *
+     * @param $userID
+     * @param string $field optionally the field name from the role table to return.
+     * @return array|null|void
+     */
+    public function getUserRoles($userID, $field="Name") {
```
linc: spaces around `=`

linc: this isn't a very good method name, because we're filtering by permission. the name suggests it will give you ALL roles.

library/core/class.controller.php
```diff
@@ -621,6 +621,13 @@ public function definitionList($wrap = true) {
             $this->_Definitions['Search'] = t('Search');
         }
 
+        if (!array_key_exists('Roles', $this->_Definitions)) {
```
linc: this is not an appropriate place for this, because the controller is part of the framework, while role data is part of the Dashboard app.

nervoustwit:
I created a getter in the controller class in order to access the protected "_Definitions" array.

applications/dashboard/settings/class.hooks.php
```diff
@@ -135,6 +135,14 @@ public function base_render_before($Sender) {
 
         // Allow global translation of TagHint
         $Sender->addDefinition("TagHint", t("TagHint", "Start to type..."));
+
+        // Add user's viewable roles to gdn.meta if user is logged in.
+        if (!array_key_exists('Roles', $Sender->getDefinitions())) {
```
evach-vf: To check if a definition exists, I would add a controller method called `hasDefinition($term)` instead of `getDefinitions()`.
IMO, the resulting condition would read better like `if (!$this->hasDefinition('Roles')) {`

linc: Some things I'm thinking about:

* `hasDefinition` is a really narrow use case.
* `addDefinition` already exists and, while terribly named, already does this technically.

So now I feel like we're fighting the framework.

Let's step back and rethink this. We're needing to detect this because we're hooking so late in the process, because we value allowing a plugin to override this. So it seems to me we should sidestep this entirely and hook earlier in the process so there's plenty of time left for a plugin to do otherwise. Let's hook into when the user's "session" is getting established and just `addDefinition` blindly like we do everywhere else.

applications/dashboard/models/class.rolemodel.php
```diff
@@ -750,4 +750,39 @@ public function deleteAndReplace($roleID, $newRoleID) {
...
+
+    /**
+     * Get a list of a user's roles that are permitted to be seen.
+     * Optionally return all the role data or just one field name.
+     *
+     * @param $userID
+     * @param string $field optionally the field name from the role table to return.
+     * @return array|null|void
+     */
+    public function getUsersViewableRoles($userID, $field = "Name") {
```
linc: maybe `getPublicUserRoles()` would be clearer? avoids User's/Users, uses UserRole (the name of the table), and Public is the inverse of Private/Personal.

applications/dashboard/settings/class.hooks.php
```diff
@@ -407,4 +408,15 @@ public function updateModel_afterStructure_handler($sender) {
             PermissionModel::resetAllRoles();
         }
     }
+
+    public function gdn_dispatcher_afterControllerCreate_handler($sender, $args) {
```
linc: docblock. as penance, docblock the method above this one too.

applications/dashboard/settings/class.hooks.php
```diff
@@ -135,6 +135,7 @@ public function base_render_before($Sender) {
 
         // Allow global translation of TagHint
         $Sender->addDefinition("TagHint", t("TagHint", "Start to type..."));
+
```
linc: why.

nervoustwit:
Late new year's resolution for me. Never label a PR as a Gimme.

PR #3483, opened 2016-01-22
Update class.discussionmodel.php

rbrahmson:
This will allow plugin owners more control over the sorting order, specifically to have multiple sort orders.  Typical example: sort so that bookmarked discussions are on top of other discussions, alerted discussions (see DiscussionAlert plugin) ahead of bookmarked and ahead or the rest which should be in the "standard" order.


linc:
I don't believe `$Sql` is defined here; I think you wanted `$this->Sql`, but it isn't necessary to pass that as an event argument. That object is globally available as `Gdn::sql()` already.

PR #3484, opened 2016-01-22
add (optional) MinLength for sliceParagraph

korelstar:
In my forum, I have several discussions beginning with a short greeting, like:

```
Hi folks,

foo bar is great. Bla bla bla.
```

The `DiscussionController` uses `sliceParagraph` for setting an appropriate `description` in the `meta` tags of the discussion page. Hence, the result is currently `Hi folks,` which is not very useful (and e.g. google webmaster tools complain about a too short description).

Therefore, I propose to add a new parameter `MinLength`. If the found paragraph is shorter than this value, then the string is sliced on a sentence.

In the example, the result will be `Hi folks, foo bar is great.` which has clearly more benefit.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3740,7 +3740,7 @@ if (!function_exists('sliceParagraph')) {
      * @param int $MinLength The intended minimum length of the string (slice on sentence if paragraph is too short)
      * @return string
      */
-    function sliceParagraph($String, $MaxLength = 500, $Suffix = '…', $MinLength = 20) {
+    function sliceParagraph($String, $MaxLength = 500, $Suffix = '…', $MinLength = 50) {
         if ($MaxLength >= strlen($String)) {
             return $String;
         }
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3735,12 +3735,15 @@ if (!function_exists('sliceParagraph')) {
      * The purpose of this function is to provide a string that is reaonably easy to consume by a human.
      *
      * @param string $String The string to slice.
-     * @param int $MaxLength The maximum length of the string.
+     * @param int|array $Limits Either int $MaxLength or array($MaxLength, $MinLength); whereas $MaxLength The maximum length of the string; $MinLength The intended minimum length of the string (slice on sentence if paragraph is too short).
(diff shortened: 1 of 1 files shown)


library/core/functions.general.php
```diff
@@ -3737,9 +3737,10 @@ function setAppCookie($Name, $Value, $Expire = 0, $Force = false) {
      * @param string $String The string to slice.
      * @param int $MaxLength The maximum length of the string.
      * @param string $Suffix The suffix if the string must be sliced mid-sentence.
+     * @param int $MinLength The intended minimum length of the string (slice on sentence if paragraph is too short)
      * @return string
      */
-    function sliceParagraph($String, $MaxLength = 500, $Suffix = '…') {
+    function sliceParagraph($String, $MaxLength = 500, $Suffix = '…', $MinLength = 50) {
```
linc: It bothers me to throw a parameter on the end like this. Now param 4 is related to param 2 with something you probably never want to override in between.

What if we made the second parameter default to [32,500] as `list($min, $max)` or defaults to just $max if an int is passed?

I also note your more useful suggestion is still only 27 characters. :) Maybe a lower min default of 32?

library/core/functions.general.php
```diff
@@ -3749,7 +3750,7 @@ function sliceParagraph($String, $MaxLength = 500, $Suffix = '…') {
         // See if there is a paragraph.
         $Pos = strrpos(SliceString($String, $MaxLength, ''), "\n\n");
 
-        if ($Pos === false) {
+        if ($Pos === false || $Pos<$MinLength) {
```
linc: Spaces on either side of comparison operators are needed.

korelstar:
You're right, @linc : the example does not fit completely. It was only intended to show the general approach. I don't know, which default would be the best. We could use your suggestion of 32 chars and look how it will behave in real forums (I've tested it with 50 chars for two months and I'm not getting any warnings from google webmaster tools).

Regarding the parameter order: since `sliceParagraph` is never called without the second parameter, there should always be a default for `$MinLength`. I've implemented your suggestion with the array parameter. However, I don't know if the documentation syntax for this parameter is correct.

linc:
Good revision, thank you.

PR #3488, opened 2016-01-25
Applicants v2

### LINKED ISSUE #1627: Large Applicants list makes banning/declining/approving difficult
Each time you click "Decline" or "Approve" or "Ban", the Applicants list re-renders.

When your applicants list is 4k, this becomes resource intensive.

Side note:
- I couldn't figure out how to search. sorry.
- If anyone has any short term pointers for dealing with this, I'd love suggestions. I've tried commenting out some of the php (specifically what looks like rendering) but it seems to cancel the actual decline submission. (I'd also take sql code)

linc:
Approving applicants is an awful process and always has been. Either you have to wait for the page to reload with the entire applicant list every time you click a button, or you have to pick either "allow" or "deny" and only check those, then do the others in a second batch. AND, if you choose the former, the two links are RIGHT NEXT TO EACH OTHER and in exactly the same style, making it super easy toclick the wrong one, which has absolutely no undo. It's madness.

What's been changed:

* Redo the Applicants page as a table of user data instead of cramming it all in 1 column.
* Add IP address (linked to search) & link the email address.
* Move the Approve / Decline options to the left.
* Make Approve / Decline actions asyncronous - no more pageloads!
* Clicking either option immediately gives a "wait" animation which is then replaced with a "Approved" or "Declined" message on success, so you can click lots of them in a row without it shifting around or being delayed by stuff like emails being sent.
* Eliminate the checkbox options (now redundant).
* Make the `approve()` & `decline()` methods postbacks.
* Improve the system for plugins to hook into the applicant page, including an optional built-in definition list. Does not break backwards compatibility of these hooks.

Need frontend help with:

* [ ] Inline styles `border-bottom:none;` - probably should be abstracted to `tr.ApplicantMeta td` rule in CSS?
* [ ] Maybe add a style to `.ConfirmApplicantApprove` and `.ConfirmApplicantDeny` so the feedback messages are nicer.

Closes #1627

screen shot 2016-01-24 at 9 04 02 pm

To test this, I strongly recommend creating a dozen or more applicants (at least) to simulate having many users to work thru. Then back up you GDN_User and GDN_UserRole tables so you can reset between tests.

### DIFF
--- a/applications/dashboard/js/applicants.js
+++ b/applications/dashboard/js/applicants.js
@@ -1,0 +1,28 @@
+jQuery(document).ready(function($) {
+
+    // Approve / Decline an applicant asynchronously.
+    $(document).delegate('a.DeclineApplicant, a.ApproveApplicant', 'click', function() {
+        // Prepare our message.
+        var approvedMsg = gdn.definition('ApprovedTranslation', 'Approved');
+        var declinedMsg = gdn.definition('DeclinedTranslation', 'Declined');
+        var message = ($(this).hasClass('ApproveApplicant')) ? approvedMsg : declinedMsg;
+        message = '<em class="ConfirmApplicant'+message+'">'+message+'</em>';
+
+        // Prepare action & effect.
+        var anchor = this;
+        var container = $(anchor).closest('td');
+        var transientKey = gdn.definition('TransientKey');
+        var data = 'DeliveryType=BOOL&TransientKey=' + transientKey;
+
+        // Do the action.
+        $(container).html('<span class="AfterButtonLoading">&#160;</span>');
+        $.post($(anchor).attr('href'), data, function(response) {
+            if (response == 'TRUE') {
+                //gdn.informMessage(message);
+                // Do the effect.
+                $(container).html(message);
+            }
+        });
+        return false;
+    });
+});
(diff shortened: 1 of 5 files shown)


applications/dashboard/js/applicants.js
```diff
@@ -0,0 +1,28 @@
+jQuery(document).ready(function($) {
+
+    // Approve / Decline an applicant asynchronously.
+    $(document).delegate('a.DeclineApplicant, a.ApproveApplicant', 'click', function() {
+        // Prepare our message.
+        var approvedMsg = gdn.definition('ApprovedTranslation', 'Approved');
+        var declinedMsg = gdn.definition('DeclinedTranslation', 'Declined');
```
beckyvb: These translations need to be added.

applications/dashboard/views/user/applicants.php
```diff
@@ -1,63 +1,69 @@
...
-if ($NumApplicants == 0) { ?>
+
+if ($NumApplicants == 0) : ?>
     <div class="Info"><?php echo t('There are currently no applicants.'); ?></div>
-<?php } else { ?>
+<?php else : ?>
     <?php
-    $AppText = plural($NumApplicants, 'There is currently %s applicant', 'There are currently %s applicants');
+    $AppText = plural($NumApplicants, 'There is currently %s applicant', 'There are currently %s applicants.');
     ?>
```
beckyvb: Let's add a period to the singular sentence too, then?

linc:
@beckyvb but can haz front end help? :(

PR #3495, opened 2016-01-27
Add a max-height and scroll to popups.

beckyvb:
Fixes issue where popups get cut off if overflow is hidden on the body.

### DIFF
--- a/js/library/jquery.popup.js
+++ b/js/library/jquery.popup.js
@@ -133,6 +133,11 @@ Copyright 2007 Chris Wanstrath [ chris@ozmm.org ]
       });
       $('#'+settings.popupId).show();
 
+      $('#'+settings.popupId+' .Body').css({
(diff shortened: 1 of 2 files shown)


francisbo:
Looks alright to me :)

js/library/jquery.popup.js
```diff
@@ -133,6 +133,11 @@ Copyright 2007 Chris Wanstrath [ chris@ozmm.org ]
       });
       $('#'+settings.popupId).show();
 
+      $('#'+settings.popupId+' .Body').css({
+         'max-height': pagesize[3] * .8,
+         'overflow-y': 'scroll'
```
R-J: `overflow-y: scroll` **always** shows a scroll bar, even if it is not needed (see login dialog as an example)
`overflow-y: auto` would be better

PR #3498, opened 2016-01-28
Add htmlEsc() convenience function

linc:
From discussion on https://github.com/vanilla/vanilla/pull/2507

I'm not super-convinced we need this but it seems a useful shortening / abstraction. I was toying with this for a specific issue and ended up not needing it, but thought I'd throw it out there in case we think it's useful since I already wrote it.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -2128,7 +2128,19 @@ if (!function_exists('htmlEntityDecode')) {
         }
         return '';
     }
+}
 
+if (!function_exists('htmlEsc')) {
+    /**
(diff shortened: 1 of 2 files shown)


linc:
@tburry interested in your opinion on this since the conversation was with you.

tburry:
Two points.

1. I'm not super into the is_array() portion of the function. Limited use cases that add overhead aren't useful IMO.
2. Is this a function that should be overridden with a plugin/theme? If not put it in functions.general.php.

linc:
Good points, thanks

PR #3509, opened 2016-02-02
Remove calls to removeKeyFromArray()

tburry:
The function is deprecated because it too closely matching the unset() language feature.


linc:
Are you not going to add a `deprecated()` warning yet?

PR #3512, opened 2016-02-03
Fix recognition of plus signs in paths

tburry:
The request object double decodes paths somewhere along the way. Since it used urldecode() instead of rawurldecode() a “+” sign in the path would get changed into a space.

The main place where this would happen is with usernames. Put a plus sign in a username and then browse to their profile page.

Whether or not the double decoding itself is a bug is too much of a change to play with here, but since paths are always supposed to use rawurlencode() I think this is a safe fix for now.

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -133,7 +133,7 @@ class Gdn_Request {
 
             switch ($key) {
                 case 'URI':
-                    $value = !is_null($value) ? urldecode($value) : $value;
+                    $value = !is_null($value) ? rawurldecode($value) : $value;
                     break;
                 case 'SCRIPT':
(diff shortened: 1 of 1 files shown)


linc:
There's no behavior change I can see here for users with spaces or plus signs in their name. With a space, it works as either a space or plus. With an actual plus, you get a User Not Found regardless of whether it's encoded or not. What am I missing?

initvector:
I created a new user (new+guy).  I went to a discussion that user had authored.  I clicked on the user's name to visit their profile (`/profile/new%2Bguy`).  In current master, this throws a "not found" error for the user.  In this branch, the page successfully retrieves the users profile.

PR #3522, opened 2016-02-09
Add messaging to inform user if username exists and check on keyup.

beckyvb:
If somebody is connecting to a Vanilla account that already exists, we need their password to confirm that they can connect to the existing Vanilla account. Currently, the ajax username checking is tied to the blur event and if the user just hits the 'Connect' button after they type in an existing username, the password field never shows up. But the validation sees an empty password field.

This checks the username on keyup to see whether the username exists or not and adds the password field if necessary. It also adds a helpful message indicating so.


linc:
Really nice enhancement.

PR #3523, opened 2016-02-10
Support multiple order by statements in the discussion model (breaks backwards compat of discussion sorting)

beckyvb:
Add a function that checks order by values. 
Add order by values in a loop to support multiple order by values.
Adds Score to allowed sort fields.
Gives events the ability to override these values.

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -356,10 +356,13 @@ class DiscussionModel extends VanillaModel {
 
     /**
      * Returns an array of 'field' => 'direction', values and ensures the safety of field names and directions.
+     *
      * You can safely use return values from this function in the orderBy() SQL function. Use this return value and orderBy()
      * in a loop to support multiple orderby values.
      *
      * If no values are passed, defaults to the fields from getSortField() and the direction from c('Vanilla.Discussions.SortDirection')
+     *
+     * @since 2.3
      *
      * @param string|array $orderFields
      * @param string $orderDirection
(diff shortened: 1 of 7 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -289,10 +292,9 @@ public function get($Offset = '0', $Limit = '', $Wheres = '', $AdditionalFields
         }
 
         // Get preferred sort order
-        $SortField = self::GetSortField();
+        $orderFields = $this->getOrderFields();
 
-        $this->EventArguments['SortField'] = &$SortField;
-        $this->EventArguments['SortDirection'] = c('Vanilla.Discussions.SortDirection', 'desc');
+        $this->EventArguments['OrderFields'] = &$orderFields;
```
linc: This breaks backwards compatibility for any plugin manipulating this, and I don't see any good reason for that.

beckyvb: This pull request is to support multiple order by clauses. Keeping the single string-type sort field as an argument means that we would need to build extra logic (i.e., if the OrderFields exist, use that, otherwise default to the SortField... etc.) Consolidating the sort fields and directions into one array is simpler and more explicit in that we know which direction belongs to which field.

I've done a search to see whether these arguments are even being used in Vanilla and it came up empty. I think it's a pretty safe enhancement.

linc: It's gonna break some open source plugins for sure. There was a whole conversation about using this hook a couple weeks back. Super conflicted about this.

linc: To be clear: conflicted about whether we need some logic here to restore backwards compat, not the need for the change.

linc:
I feel like maybe we should merge this into #3528 as 1 big pull request since we're now destroying the old discussion sorting behavior and wholesale replacing it.

beckyvb:
Merged into https://github.com/vanilla/vanilla/pull/3528

PR #3525, opened 2016-02-10
Add "Drop" type to routes

### LINKED ISSUE #3524: New route type: block
Add a route type that immediately aborts the request rather than continuing to dispatch to an error page.

initvector:
This update adds a new route type: drop.

A request matching a route of this type will immediately be killed.

Closes #3524


kaecyra:
My suggestion is to rename this to "Drop".

initvector:
Renamed

PR #3528, opened 2016-02-12
Discussion sort filter module

beckyvb:
The module rendering can be added anywhere using this or some variation of:
```
$sorter = new DiscussionsSortFilterModule();
echo $sorter;
```
Filters can be added to the model using the `addFilter()` function in a handler for the DiscussionFilters event that's thrown in the Discussion Model, for example:
```
public function discussionModel_discussionFilters_handler() {
    DiscussionModel::addFilterSet('my-discussions', 'My Discussions');
    DiscussionModel::addFilter('bookmarks', 'Bookmarked', ['w.Bookmarked' => true, 'w.UserID' => Gdn::session()->UserID], 'mine', 'my-discussions');
    DiscussionModel::addFilter('participated', 'Participated', ['w.Participated' => true, 'w.UserID' => Gdn::session()->UserID], 'mine', 'my-discussions');

    DiscussionModel::addFilterSet('discussion-types', 'Discussions Types');
    DiscussionModel::addFilter('discussion-type', 'Discussion', ['Type' => null, 'Announce' => 0], 'type', 'discussion-types');
    DiscussionModel::addFilter('announcement-type', 'Announcement', ['Announce >' => 0], 'type', 'discussion-types');
    DiscussionModel::addFilter('question-type', 'Question', ['Type' => 'Question'], 'type', 'discussion-types');
    DiscussionModel::addFilter('poll-type', 'Poll', ['Type' => 'poll'], 'type', 'discussion-types');
}
```
Default sorts are 'New', 'Hot', and 'Top'. Sorts can be added to the module using the `addSort()` function.


tburry:
1. Change the static allowed methods to **allowedSorts** and **allowedFilters**.
2. Make the current sort/filter methods non-static: **setSort()** and **setFilter()** etc.
3. Add a static initializer trait that fires an event and use that for the allowed* methods.
4. Change the categoryUrl() to accept a `{Page}` string instead of a page number and use this for pagers. concat that querystring after the call rather than making it a parameter.

PR #3535, opened 2016-02-16
Calculate allowed discussion types using category column.

beckyvb:
The allowed discussion types are a column on the category table. No need to abstract it further into a permission category.

Also moves the ui elements not associated with the permission grid out of the hidden area on the category settings page.


tburry:
Does this mean that I have to restrict the discussion types on each category one-by-one? The reason why I did it the way before was to allow admins to set the allowed discussion types on the parent category and have it cascade to all the children without having to edit a massive number of categories.

PR #3537, opened 2016-02-17
Fix Typo in Variable Name

R-J:
There is no `$this->Options` and from the context it is obvious that it must be `$this->_Options`.

The only time this array is really used is in class sqldrivers query method, but it is only checked for a value "Slave". So only if someone explicitly queries a slave connection, that error might show up.


DaazKu:
Thanks.

PR #3546, opened 2016-02-19
Let discussion type opt out of new discussion module.

beckyvb:
Allows discussion types to add a property to opt out of being added to the default NewDiscussionModule dropdown.


linc:
It bothers me that we're adding undocumented functionality. There's not even a docblock on `discussionModel::discussionTypes()` and now we're adding a flag that you'd have to read thru the guts of a module to ever know about.

PR #3554, opened 2016-02-24
Pass categoryDropDownOptions to the BeforeFormInputs event.

DaazKu:
Allows to filter the categoryDropdown by adding CategoryData to the options.

PR #3556, opened 2016-02-25
Add a new event that allows to filter down the categoryDropDown input

DaazKu:
This PR is a better alternative than https://github.com/vanilla/vanilla/pull/3554
With this new event we won't need to modify every view that use the function to enable plugins to filter down the categories.


linc:
conflicts

PR #3557, opened 2016-02-25
Add AllowConnect setting to optionally disable SSO connecting to existing accounts

nervoustwit:
Create a functionality whereby a forum can add a line in their config that would stop users who are connecting through SSO from claiming an existing user by entering the username and password.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -380,6 +380,12 @@ EOT;
         $this->View = 'connect';
         $IsPostBack = $this->Form->isPostBack() && $this->Form->getFormValue('Connect', null) !== null;
         $UserSelect = $this->Form->getFormValue('UserSelect');
+
+        /**
+         * When a user is connecting through SSO he is prompted to choose a username. If he chooses an existing user name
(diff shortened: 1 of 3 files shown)


applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -616,6 +625,18 @@ public function connect($Method) {
...
 
+            if(!$connectToExistingUser) {
+                // Since we are not connecting a joining user to an existing user...
+                // make his name the same as the connect name he chose.
+                $connectName = $this->Form->getFormValue("ConnectName");
+                $this->Form->setFormValue("Name", $connectName);
+                $this->Form->setFormValue("UserSelect", "other");
+                // make sure the photo of the existing user doen't show up on the form.
+                $this->Form->setFormValue("Photo", null);
+                // ignore any existing user(s) found.
```
linc: Shouldn't this be impossible?

linc:
We have a config named `Garden.Registration.AutoConnect`. I think, because this is specific to registration and a similar situation, the clearest possible name is `Garden.Registration.AllowConnect`. And then using the shorter 'AllowConnect' as our conditional naming for clarity.

linc:
Besides the naming & my query about an impossible scenario above, this will be good to go.

PR #3558, opened 2016-02-25
Update default master extension in default themes README

R-J:
It was still showing `default.master.php` instead of `default.master.tpl`.


linc:
Cool, thanks

PR #3562, opened 2016-02-29
Add options for overriding email messages.

beckyvb:
The recent HTML email feature stopped using the email translation compound strings in favour of smaller, concatenated translations. This allowed us to split up the email into its components (title, button, etc.) However, it broke backwards compatibility for users with custom email translations. 

This PR addresses that issue in 2 ways:

1. If there is a custom translation for the welcome email body, it uses that.
2. For activity-related emails, adds the ability to prepend a string to the email body, using the 'Garden.Email.Prefix' config setting.

This does not restore backwards compatibility, but addresses the need for users to add custom messages to the emails.

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -876,7 +876,11 @@ class ActivityModel extends Gdn_Model {
                     ->setTitle($ActivityHeadline);
 
                 if ($message = val('Story', $Activity)) {
-                    $emailTemplate->setMessage($message, true);
+                    $prefix = c('Garden.Email.Prefix', '');
+                    if (!empty($prefix)) {
+                        $prefix .= '<br><br>';
+                    }
+                    $emailTemplate->setMessage($prefix.$message, true);
                 }
 
                 $Email->setEmailTemplate($emailTemplate);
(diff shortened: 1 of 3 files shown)


applications/dashboard/models/class.activitymodel.php
```diff
@@ -876,7 +876,11 @@ public function sendNotification($ActivityID, $Story = '', $Force = false) {
                     ->setTitle($ActivityHeadline);
 
                 if ($message = val('Story', $Activity)) {
-                    $emailTemplate->setMessage($message, true);
+                    $prefix = c('Garden.Email.Prefix', '');
+                    if (!empty($prefix)) {
```
linc: I'd say let's not do this ghost-break thing unless it breaks the world without it. If you want to prefix and you want it broken from the rest of the content, add breaks too to the prefix.

applications/dashboard/models/class.usermodel.php
```diff
@@ -3940,6 +3922,45 @@ public function sendWelcomeEmail($UserID, $Password, $RegisterType = 'Add', $Add
         $Email->send();
     }
 
+    protected function getEmailWelcome($registerType, $user, $data, $password = '') {
+        $appTitle = c('Garden.Title');
+
+        // Backwards compatability. See if anybody has overridden the EmailWelcome string.
+        $emailWelcome = '%2$s has created an account for you at %3$s. Your login credentials are: Email: %6$s Password: %5$s Url: %4$s';
+        if (preg_replace('/\s+/', '', t('EmailWelcome')) != preg_replace('/\s+/', '', $emailWelcome)) {
```
linc: Won't this trigger for every non-English site?

PR #3563, opened 2016-03-01
Build new nav module for rendering/sorting navs.

beckyvb:


applications/dashboard/modules/class.navmodule.php
```diff
@@ -1,349 +1,118 @@
-<?php
+<?php if (!defined('APPLICATION')) exit();
+use NestedCollection;
```
initvector: It looks like `NestedCollection` is a trait, not a namespace, so the `use` keyword here, outside of a class, probably isn't going to have the intended effect.  `NavModule` already includes the `NestedCollection` trait.

PR #3587, opened 2016-03-08
Only add the download attribute if the download is not an image.

beckyvb:
Fixes issue where image attachments were being downloaded on click rather than being opened in a new tab. Downloading an attachment on click is great if it's a document, but is unexpected if it's an image.

### DIFF
--- a/plugins/editor/views/attachments.php
+++ b/plugins/editor/views/attachments.php
@@ -44,7 +44,7 @@ $editorkey = $this->data('_editorkey');
                 <?php
                 // Only add the download attribute if it's not an image.
                 if (!$attachment['ImageHeight']) {
-                    echo 'download=" '.htmlspecialchars($attachment['Name']).'"';
+                    echo 'download="'.htmlspecialchars($attachment['Name']).'"';
                 }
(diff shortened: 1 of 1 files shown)


plugins/editor/views/attachments.php
```diff
@@ -38,9 +38,18 @@
...
-               data-height="<?php echo $attachment['ImageHeight']; ?>" href="<?php echo $pathParse['Url'] ?>"
-               download="<?php echo htmlspecialchars($attachment['Name']); ?>"
-               target="_blank"><?php echo htmlspecialchars($attachment['Name']); ?></a>
+               data-height="<?php echo $attachment['ImageHeight']; ?>"
+               href="<?php echo $pathParse['Url'] ?>"
+               target="_blank"
+                <?php
+                // Only add the download attribute if it's not an image.
+                if (!$attachment['ImageHeight']) {
+                    echo 'download="'.htmlspecialchars($attachment['Name']).'"';
```
initvector: Should you also pull `data-width` and `data-height` in here while you're at it?

Issue #269, opened 2010-03-26
Multiple users


jdbrew:
I have an installation of Vanilla 2 running and have had multiple users using the system at the same point in time. viewing profile information and changing main menu tabs results in displaying other users preferences and some times showing their name instead of mine in the the title bar.

Logging out also logs them out.

PR #3591, opened 2016-03-10
Fix NBBC sometimes breaking search results

valeryan:
This PR address the NBBC addon sometimes breaking search results. This fixes [vanilla/addons#268](https://github.com/vanilla/addons/issues/268) and will work better than [Wrap Format() return in html_entity_decode() #269](https://github.com/vanilla/addons/issues/269). 

I move the logic to remove list items from plainText() into its own function and then enhanced the regex to remove html list items and encode list items.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -993,6 +993,20 @@ class Gdn_Format {
     }
 
     /**
+     * Replaces opening html list tags with an asterisk and closing list tags with new lines.
+     * Accepts both encoded and decoded html strings.
+     *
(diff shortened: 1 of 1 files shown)


beckyvb:
This is great, thanks! I'd like to see a clearer method description (i.e., 'Replaces opening html list tags with an asterisk and closing list tags with new lines. Accepts both encoded and decoded html strings.')

valeryan:
I updated the method description as per your suggestion. Thanks

PR #3597, opened 2016-03-11
Allow HEAD requests in ProxyRequest->request()

R-J:
Setting  $Method => 'HEAD' in request() results in a timeout since the request method doesn't include the head. So it must be handled extra.

### DIFF
--- a/library/core/class.proxyrequest.php
+++ b/library/core/class.proxyrequest.php
@@ -518,6 +518,12 @@ class ProxyRequest {
             }
         }
 
+        // Allow HEAD
+        if ($RequestMethod == 'HEAD') {
+            curl_setopt($Handler, CURLOPT_HEADER, true);
+            curl_setopt($Handler, CURLOPT_NOBODY, true);
+        }
+
         // Any extra needed headers
         if (sizeof($SendExtraHeaders)) {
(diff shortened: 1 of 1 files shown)


linc:
What's the utility of this?

R-J:
I tried writing a plugin that implements [webmentions](http://webmention.net/). When trying to get the header of a page (the webmention specification forces me to inspect the header), I got stuck. 

ProxyRequest::curlHeader() is [borked](https://github.com/vanilla/vanilla/pull/3601). It is called for each line in the response header and it stores the header lines like that: `$this->ResponseHeaders[$Key] = $Value;`. So if there is more than one `<link>` (or any other) element in the header, only the last one will make it into the final ResponseHeaders array.

In order to overcome this shortage, I've tried to call ProxyRequest::request with the HEAD method to get the unparsed HEAD as the return of this function. I've invested some time to find a way to make a HEAD request without changing the code, but came to the conclusion that HEAD requests are impossible to make with class ProxyRequest. Implementing the possibility was quite easy and so I did it.

HEAD requests are the "polite" way to get information about another site. You only request what you need. By looking at "Expires", "Cache-Control", "Last-Modified" and/or "ETags", you can judge if a page has changed which often is information enough. You do not need to fetch/"GET" the complete page. So semantically, whenever you do not need the actual page but only meta information, the HEAD request would be "correct".

PR #3599, opened 2016-03-11
Register user using connect name when provided

DaazKu:
Let begin by saying
---
The logic of the connect function _should_ be refactored/simplified!

A bit of context
---
When Garden.Registration.NameUnique is set to true and a user tries to connect through SSO using a name that already exist in the forum we give 2 options:
Appropriate yourself the user by providing its password or use a different name.

There is a bug with the latter option... We are always told that the new chosen name exist.

That new name is provided via the field "ConnectName" which is what we use to check that this new name is not already existing. All this is correct. 
But we then create the user using the "Name" field. Which is the name that already exist in the DB and got us to provide a ConnectName in the first place.

In short
---
When registering a user on connect we need to use "ConnectName" if it is provided.


linc:
This looks correct to me. Passing to Ryan to also check.

initvector:
Tested and verified.  Looks good to me.

PR #3600, opened 2016-03-14
Correct wrong Vanilla version number in class ProxyRequest

R-J:
The HTTP_USER_AGENT has been hard coded as "Vanilla/2.0". I've changed it to always use the version from the config

### DIFF
--- a/library/core/class.proxyrequest.php
+++ b/library/core/class.proxyrequest.php
@@ -419,7 +419,7 @@ class ProxyRequest {
         curl_setopt($Handler, CURLOPT_HEADER, false);
         curl_setopt($Handler, CURLINFO_HEADER_OUT, true);
         curl_setopt($Handler, CURLOPT_RETURNTRANSFER, true);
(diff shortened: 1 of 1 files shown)


linc:
I don't feel this is correct. The point of `HTTP_USER_AGENT` isn't to provide accurate version data, but to indicate the _type_ of client. This will cause Vanilla to report itself as many different agents depending on the point revision a particular site is using. This makes things like analytics or server-side logic based on agent detection more difficult. It's also potentially a security issue - now you're revealing exact version data to third-party servers.

I really don't think we should go further than the second number at most. Unless there is a compelling reason to do even that, I'd rather leave it as-is until the jump to 3.0.

R-J:
I just stumbled upon "Vanilla 2.0" and thought it has been forgotten. If the version number isn't relevant, shouldn't it be dropped instead of being wrong?

tburry:
Most web browsers report their full version number in part of the string. If we imagine the 2.* release something we want to report completely differently than the 1.* or future releases then we could do something like this:

```
Vanilla/2.0 Garden/<full.app.version>
```

PR #3603, opened 2016-03-14
Fix the editor upload feature when it is called via ajax.

DaazKu:
Add the "canUpload" definition to the base_render hook so that if you request the editor via ajax that definition will be present in the page


initvector:
If we're adding this with `base_render_before`, does it even still need to be in `gdn_form_beforeBodyBox_handler` or is that now a redundant definition?

DaazKu:
It seems removing the line in ``gdn_form_beforeBodyBox_handler`` have no effect what so ever because it will always be executed by base_render_before.

Good call.

PR #3605, opened 2016-03-15
Implement an alter table threshold

tburry:
Add an optional alter table row threshold which represents the maximum number of rows a table is allowed to have for an in-application alter to be allowed to run. If a table is too large then the admin is expected to copy the SQL from the structure endpoint into their RDBMS.

By default this threshold is turned off, but may be turned on by config.

### DIFF
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -251,9 +251,16 @@ class UtilityController extends DashboardController {
             $this->setData('CapturedSql', array());
         }
 
+        $issues = Gdn::structure()->getIssues();
+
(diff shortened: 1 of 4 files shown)


initvector:
The functionality works well, but I can't help feeling it's maybe too silent.

Is there any way we can make it more apparent to an admin that the necessary actions weren't automatically performed and will need to be manually executed?  It looks like the only way this will be noticed is if you're specifically looking for it, since it's only sending a message to the logger.  Unless an admin knows to check their event log after each update, they're going to assume everything is good to go.  This can be an issue if they have pending, potentially critical, database changes.

tburry:
I added an issues list on /utility/structure @initvector. Please note that I rebased this PR to bring in the dbencode() changes. You will have to reset your local branch.

PR #3607, opened 2016-03-15
Allow all quotes to be collapsed.

DaazKu:
There are some inconsistencies about the quotes markup.
You can have "Quote", "Quote UserQuote", "UserQuote".

This PR aim at making all quotes being able to be collapsed.
There is also the possibility of quotes inside a discussion that was not covered before this PR.

Before merging this PR it may be wise to raise an issue about the quotes' inconsistencies in order to fix the source of the problem. 
Something like [this PR](https://github.com/vanilla/vanilla/pull/3606) could be done


linc:
Increment the minor point revision to break cache.

PR #3608, opened 2016-03-15
Make deeply nested quotes more readable

DaazKu:
Set a min-width of 200px and set scroll to auto on the x axis. That way, even if it is a pain to scroll all quotes to to the right to view the inner quotes, it is possible to view all the content of nested quotes in a "human" readable way.


applications/dashboard/scss/legacy/_style.scss
```diff
@@ -2262,6 +2262,8 @@ blockquote.Quote, blockQuote.UserQuote {
   background: rgba(0, 0, 0, 0.05);
```
beckyvb: blockQuote should be blockquote

PR #3611, opened 2016-03-16
Use dropdown module to render discussion and category options

beckyvb:
Add function for converting strings to slugs for auto-generating dropdown item keys.
Deprecate discussion options functions.

### DIFF
--- a/applications/vanilla/views/categories/helper_functions.php
+++ b/applications/vanilla/views/categories/helper_functions.php
@@ -54,46 +54,34 @@ if (!function_exists('CategoryString')):
     }
 endif;
 
-if (!function_exists('GetOptions')):
+if (!function_exists('getOptions')):
     /**
-     * Render options that the user has for this category.
+     * Render options that the user has for this category. Returns an empty string if the session isn't valid.
+     *
+     * @param $category The category to render the options for.
+     * @return DropdownModule|string A dropdown with the category options or an empty string if the session is not valid.
+     * @throws Exception
      */
-    function getOptions($Category) {
-        if (!Gdn::session()->isValid())
-            return;
+    function getOptions($category) {
+        if (!Gdn::session()->isValid()) {
+            return '';
+        }
+        $sender = Gdn::controller();
+        $categoryID = val('CategoryID', $category);
 
-        $Sender = Gdn::controller();
+        $dropdown = new DropdownModule();
+        $tk = urlencode(Gdn::session()->TransientKey());
+        $following = (int)val('Following', $category);
 
-
-        $Result = '';
-        $Options = '';
-        $CategoryID = val('CategoryID', $Category);
-
-        $Result = '<div class="Options">';
-        $TKey = urlencode(Gdn::session()->TransientKey());
-
-        // Mark category read.
-        $Options .= '<li rel="MarkRead">'.anchor(t('Mark Read'), "/category/markread?categoryid=$CategoryID&tkey=$TKey").'</li>';
-
-        // Follow/Unfollow category.
-        if (!val('Following', $Category))
-            $Options .= '<li rel="Hide">'.anchor(t('Unhide'), "/category/follow?categoryid=$CategoryID&value=1&tkey=$TKey").'</li>';
-        else
-            $Options .= '<li rel="Hide">'.anchor(t('Hide'), "/category/follow?categoryid=$CategoryID&value=0&tkey=$TKey").'</li>';
+        $dropdown->addLink(t('Mark Read'), '/category/markread?categoryid='.$categoryID.'&tkey='.$tk);
+        $dropdown->addLink(t($following ? 'Hide' : 'Unhide'), '/category/follow?categoryid=$categoryID&value='.$following.'&tkey='.$tk);
 
         // Allow plugins to add options
-        $Sender->EventArguments['Options'] = &$Options;
-        $Sender->fireEvent('CategoryOptions');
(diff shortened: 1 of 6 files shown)


applications/vanilla/views/categories/helper_functions.php
```diff
@@ -54,46 +54,34 @@ function categoryString($Rows) {
...
-            $Options .= '<li rel="Hide">'.anchor(t('Unhide'), "/category/follow?categoryid=$CategoryID&value=1&tkey=$TKey").'</li>';
-        else
-            $Options .= '<li rel="Hide">'.anchor(t('Hide'), "/category/follow?categoryid=$CategoryID&value=0&tkey=$TKey").'</li>';
+        $dropdown->addLink(t('Mark Read'), '/category/markread?categoryid='.$categoryID.'&tkey='.$tk, 'mark-read');
+        $dropdown->addLink(t($hide ? 'Unhide' : 'Hide'), '/category/follow?categoryid='.$categoryID.'&value='.$hide.'&tkey='.$tk, 'hide');
 
         // Allow plugins to add options
-        $Sender->EventArguments['Options'] = &$Options;
-        $Sender->fireEvent('CategoryOptions');
-
```
beckyvb: Removing this potentially breaks backwards compatibility. However, the event isn't being used in Vanilla at all. To maintain backwards compatibility, It'd take some parsing of the options string in order to break it down into a dropdown element. This isn't something I want to devote time to, as I'm skeptical that it's even being used.

library/core/functions.general.php
```diff
@@ -3788,3 +3788,40 @@ function increaseMaxExecutionTime($maxExecutionTime) {
...
+     *
+     * Based off Symfony's Jobeet tutorial, and found here:
+     * http://stackoverflow.com/questions/2955251/php-function-to-make-slug-url-string
+     *
+     * @param string $text The text to convert.
+     * @return string mixed|string The slugified text.
+     */
+    function slugify($text) {
+        // replace non letter or digits by -
+        $text = preg_replace('~[^\pL\d]+~u', '-', $text);
```
DaazKu: Nitpicking here but I found these Regex hard to read with ~ as delimiter. I would go for / since there are no slash in these anyway.

linc: I'd also prefer we stuck to slashes or backticks, for the record.

applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -274,12 +296,80 @@ function getDiscussionOptions($Discussion = null) {
...
+            $categoryID = val('CategoryID', $sender->Discussion);
+        }
+
+        $discussionID = $discussion->DiscussionID;
+        $categoryUrl = urlencode(categoryUrl(CategoryModel::categories($categoryID)));
+
+        $permissionCategoryID = val('PermissionCategoryID', $discussion, val('PermissionCategoryID', $discussion));
+
+        // Permissions
+        $canEdit     = DiscussionModel::canEdit($discussion, $timeLeft);
```
linc: We don't line up equal signs (or array assignments) vertically.

applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -288,14 +378,16 @@ function writeAdminCheck($Object = null) {
...
+ * Output discussion options.
+ *
+ * @since 2.1
+ */
+if (!function_exists('writeDiscussionOptions')):
     function writeDiscussionOptions($Discussion = null) {
-        $Options = getDiscussionOptions($Discussion);
+        deprecated('writeDiscussionOptions', 'getDiscussionOptionsDropdown', 'March 2016');
+
+        $Options = GetDiscussionOptions($Discussion);
```
linc: caps derp

applications/vanilla/views/discussions/helper_functions.php
```diff
@@ -458,66 +458,14 @@ function writeFilterTabs($Sender) {
     }
 endif;
 
-if (!function_exists('OptionsList')):
-    function optionsList($Discussion) {
```
linc: docs!

applications/vanilla/views/discussions/helper_functions.php
```diff
@@ -458,66 +458,14 @@ function writeFilterTabs($Sender) {
...
-                    '<ul class="Flyout MenuItems">'.
-                    $Sender->Options.
-                    '</ul>'.
-                    '</span>';
-
-                return $Result;
-            }
-
+if (!function_exists('optionsList')):
+    function optionsList($discussion) {
```
linc: docs!

library/core/functions.general.php
```diff
@@ -3788,3 +3788,40 @@ function increaseMaxExecutionTime($maxExecutionTime) {
...
+    /**
+     * Converts a string to a slug-type string.
+     *
+     * Based off Symfony's Jobeet tutorial, and found here:
+     * http://stackoverflow.com/questions/2955251/php-function-to-make-slug-url-string
+     *
+     * @param string $text The text to convert.
+     * @return string mixed|string The slugified text.
+     */
+    function slugify($text) {
```
linc: +rep for this name.

applications/vanilla/views/categories/helper_functions.php
linc: Probably OK since it was broken until 2015 anyway, but I left a note on the PR that fixed it: https://github.com/vanilla/vanilla/pull/2971

PR #3616, opened 2016-03-17
Remove the redirect_uri from the call to Facebook's sharing api.

beckyvb:
Setting the redirect_uri causes redirects within the popup instead of closing the popup, which is weird.


initvector:
Is it possible this URL might be visited outside a popup window (e.g. mobile, disabling pop-ups, etc.)?

PR #3617, opened 2016-03-17
make category all view's count elements consistent with other views

valeryan:
Every other view that displays comment count, discussion count, view count, etc... uses the combination of `PluralTranslate()` and `BigPlural()` to show numbers formatted as 1.2K instead of 1200. This update to category/all makes it consistent with other views, making styling more predictable for these elements.


linc:
thanks!

Issue #3493, opened 2016-01-25
Quotes: Button does not work after comment edit & save


linc:
Likely glitch in livequery replacement.

PR #3622, opened 2016-03-18
Make sure the cached editor Element is still in the DOM

DaazKu:
... before returning it.

After a post edit the editor is sometime refreshed. Let's make sure we have the correct editor in hand before returning it!


DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/3493

PR #3636, opened 2016-03-24
Misspelled fetchPageInfo Fix.

MikhailRoot:
fetchPageInfo  - global function defined  in /library/core/functions.general.php - its purpose to content of html page specified by URL parameter.
For some reason it's name is misspeled as FetchPageInfo - that's why in vanilla comments integration blog posts doesn't get right content and title. In my case i think


linc:
See https://github.com/vanilla/vanilla/pull/3637

PR #3649, opened 2016-03-28
Address At.js Issues

initvector:
At.js currently has some problems, particularly with WYSIWYG mode.

1. When grabbing the text from a WYSIWYG editor, multiple lines are concatenated into one with no separator, before the pattern matching is done.  This means if you want to start a line with a mention, it's a no-go, because the mention character ends up in the middle of two "words" (e.g. Hello world@vanilla is my name.)
2. After a mention is inserted, the caret/cursor remains inside the mention markup, at least on Webkit-based browsers ([see this similar tale of woe](http://miles-by-motorcycle.com/fv-b-8-665/you-can---t-select-that-----webkit-browser-selections-and-ranges-in-chrome-and-safari)).  The `span` element containing the username also ends up containing all text that follows, even additional mentions. (e.g. How are you `@System`? And how about you `@vanilla`? becomes `How are you <span class="vanilla-mention-@">@System? And how about you <span class="vanilla-mention-@">@vanilla</span></span>`)

### DIFF
--- a/js/library/jquery.atwho.js
+++ b/js/library/jquery.atwho.js
@@ -559,19 +559,31 @@
       };
 
       Controller.prototype.content = function() {
+        var result = {
+          content: null,
+          offset: 0
+        };
+
         if (this.$inputor.is('textarea, input')) {
-          return this.$inputor.val();
+          result.content = this.$inputor.val();
         } else {
-          return this.$inputor.text();
+          var textNode = $(document.createElement('div'));
+          var html = this.$inputor.html();
+          var breaks = /<br(\s+)?(\/)?>/;
+          result.offset = html.match(breaks) ? html.match(breaks).length : 0;
+          textNode.html(html.replace(breaks, " "));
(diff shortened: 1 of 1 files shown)


initvector:
I've tried to address item #1 (mentions being lost in the middle of concatenated lines) by replacing breaks with spaces prior to searching for patterns.  Because this offsets the way that content the maps to the actual content of the caret/cursor position in the `contenteditable` element, I've also had to implement adjustments to the caret/cursor placement.

initvector:
The least-ugly hack for item 2 (nested mentions) is to append a zero-width, no-break space (invisible) character to the end of the mention.  This gives us a content buffer, allowing us to place the caret/cursor between the end of the mention's `span` tag and the end of the editor's `body` tag.

The primary side effect is, when deleting the character, the cursor won't appear to do anything, because the deleted character is invisible.  It will appear to require one additional keypress (or millisecond, if you're holding the key) to delete the visible content.

PR #3653, opened 2016-03-30
Pass the form CSS class by reference in the comment form.

beckyvb:
Fixes issue where role titles were no longer appearing in the comment form css.


initvector:
Confirmed.  When the Role Titles plug-in is enabled, a CSS class is added to the comment form in the format of Role_\[Role.Name\] (e.g. `Role_Administrator`).

PR #3654, opened 2016-03-30
Application info improvements

tburry:
Standardize and fix some of the application info loading and calculation.

- [x] Load application info through the loadApplicationInfo() method.
- [x] Add the Dir and IconUrl properties to the application info array.
- [x] Try and use enabledApplications() whenever possible. This method is faster as it doesn't have to scan the entire applications directory.

I recommend looking through the individual commits on this PR to help understand it better.


tburry:
Closing this one because of big refactor in another direction.

PR #3655, opened 2016-03-30
Prevent messages from being displayed when you do not have view permission

DaazKu:
Do not display messages in the categories  you do not have the Vanilla.Discussions.View permission.

### DIFF
--- a/applications/dashboard/models/class.messagemodel.php
+++ b/applications/dashboard/models/class.messagemodel.php
@@ -141,6 +141,12 @@ class MessageModel extends Gdn_Model {
             $Prefs[] = 0;
         }
 
+        $category = null;
+        if ($CategoryID !== null) {
+            $categoryModel = new CategoryModel();
+            $category = $categoryModel->getID($CategoryID, DATASET_TYPE_ARRAY);
+        }
+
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.messagemodel.php
```diff
@@ -162,13 +168,13 @@ public function getMessagesForLocation($Location, $Exceptions = array('[Base]'),
...
-                } elseif ($MApplication == $Application && $MController == $Controller && $MMethod == $Method)
+                } elseif ($MApplication == $Application && $MController == $Controller && $MMethod == $Method) {
                     $Visible = true;
-
-                if ($Visible && !self::inCategory($CategoryID, val('CategoryID', $Message), val('IncludeSubcategories', $Message))) {
-                    $Visible = false;
                 }
 
+                $Visible = $Visible && self::inCategory($CategoryID, val('CategoryID', $Message), val('IncludeSubcategories', $Message));
+                $Visible = $Visible && $Session->checkPermission('Vanilla.Discussions.View', true, 'Category', $category['PermissionCategoryID']);
```
linc: Did you test this on messages not in a category? If `$category` is `null` you're trying to access an array element that doesn't exist here and testing permission for no reason.

PR #3657, opened 2016-03-30
Default core and porter uploads to GPLv2 license

hgtonight:
The new requirement for a license field in the addons repository makes it impossible to upload Vanilla and the porter. This fix changes that.


linc:
Haha, thanks!

PR #3672, opened 2016-04-01
Pass URL query through embed

initvector:
The current embedding solution does not allow the passing of URL query parameters between the parent window and the embed frame.  For instance, if your parent window's location hash contained parameters for the request to be passed down to the embedded site, that information would be lost.  This update aims to allow for the exchanging of that information.


tburry:
Does this have to be done with a change to embed.js? Can you just work with the query in embed_local.js and have embed.js just take whatever it needs?

tburry:
I think I'll need to go through this one next week. My preference is to just use the querystring with the path variable rather than splitting it up.

initvector:
Closing this out in favor of #3881

PR #3673, opened 2016-04-01
Use a valid SPDX license in composer.json

hgtonight:
Got this notice when validating the composer.json file:

License "GPLv2" is not a valid SPDX license identifier, see https://spdx.org/licenses/ if you use an open license.


initvector:
Thanks!

PR #3674, opened 2016-04-01
Re-add option to dismiss announcements

initvector:
It looks like we lost the ability to dismiss an announcement when we switched over how we build discussion dropdown menus.  This update re-adds the ability to dismiss an announcement.

### DIFF
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -346,7 +346,7 @@ if (!function_exists('getDiscussionOptionsDropdown')):
             $timeLeft = ' ('.Gdn_Format::seconds($timeLeft).')';
         }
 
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -339,12 +339,15 @@ function getDiscussionOptionsDropdown($discussion = null) {
...
         $canRefetch = $canEdit && valr('Attributes.ForeignUrl', $discussion);
+        $canDismiss = c('Vanilla.Discussions.Dismiss', 1) && $discussion->Announce == '1' && $discussion->Dismissed != '1';
+
 
         if ($canEdit && $timeLeft) {
             $timeLeft = ' ('.Gdn_Format::seconds($timeLeft).')';
         }
 
-        $dropdown->addLinkIf($canEdit, t('Edit').$timeLeft, url('/post/editdiscussion/'.$discussionID), 'edit')
+        $dropdown->addLinkIf($canDismiss, t('Dismiss'), url("vanilla/discussion/dismissannouncement?discussionid={$discussion->DiscussionID}"), 'dismiss', 'DismissAnnouncement Hijack')
```
beckyvb: For consistency, I'd just use $discussionID here.

PR #3677, opened 2016-04-04
Refine default discussion name when fetchPageInfo timed out

DaazKu:
Depends on https://github.com/vanilla/vanilla/pull/3676


applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -431,7 +431,12 @@ public function comment($DiscussionID = '') {
             if (!($Title = $this->Form->getFormValue('Name'))) {
                 $Title = val('Title', $PageInfo, '');
                 if ($Title == '') {
-                    $Title = t('Undefined discussion subject.');
+                    if (!empty($PageInfo['Exception']) && $PageInfo['Exception'] === "Couldn't connect to host.") {
+                        $Title = t('Undefined discussion subject (Cause: timeout)');
```
linc: Let's just set `$Title = t('Undefined discussion subject.');` up front and then append  `.' '.t('Page timed out.');` if those conditions are true. That way we're not creating translation redundancy.

PR #3679, opened 2016-04-06
Add a hook for manipulating images uploaded in Advanced Editor.

nervoustwit:
This would allow plugins to hook into the process before up images are uploaded via the Advanced Editor to perform any operations to the images (i.e. resizing, watermarking, colourization, etc.)

### DIFF
--- a/plugins/editor/class.editor.plugin.php
+++ b/plugins/editor/class.editor.plugin.php
@@ -729,7 +729,9 @@ class EditorPlugin extends Gdn_Plugin {
             $this->EventArguments['TmpFilePath'] =& $tmpFilePath;
             $this->EventArguments['FileExtension'] = $fileExtension;
             $this->EventArguments['ValidImage'] = $validImage;
-            $this->fireEvent("BeforeSaveUploads");
+            $this->EventArguments['AbsoluteFileDestination'] =& $absoluteFileDestination;
+            $this->EventArguments['DiscussionID'] =& $discussionID;
+            $this->fireEvent('BeforeSaveUploads');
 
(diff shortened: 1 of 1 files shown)


plugins/editor/class.editor.plugin.php
```diff
@@ -725,6 +725,12 @@ public function postController_editorUpload_create($Sender, $Args = array()) {
             ];
             $validImage = !empty($imageType) && in_array($imageType, $validImageTypes);
 
+            $this->EventArguments['CategoryID'] = Gdn::request()->post('CategoryID');
+            $this->EventArguments['TmpFilePath'] =& $tmpFilePath;
+            $this->EventArguments['FileExtension'] = $fileExtension;
+            $this->EventArguments['ValidImage'] = $validImage;
+            $this->fireEvent("BeforeSaveUploads");
```
initvector:
```php
$this->fireEvent('BeforeSaveUploads');
```

plugins/editor/class.editor.plugin.php
```diff
@@ -725,6 +725,12 @@ public function postController_editorUpload_create($Sender, $Args = array()) {
             ];
             $validImage = !empty($imageType) && in_array($imageType, $validImageTypes);
 
+            $this->EventArguments['CategoryID'] = Gdn::request()->post('CategoryID');
```
initvector: In addition to these, we also have `$discussionID` and `$absoluteFileDestination` available which may be beneficial to this hook.  Maybe pass `$absoluteFileDestination` by reference.

PR #3692, opened 2016-04-09
Remove plugins from library/core

tburry:
Plugins in core is being deprecated so the couple that are there must be refactored into regular hooks classes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,11 @@ install:
   - tests/travis/nginx/install-nginx.sh
 
 script:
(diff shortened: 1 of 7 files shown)


initvector:
It looks like moving the `AnalyticsTick` event from `Gdn_Statistics` to `SettingsController` will break at least one plug-in (Online).  It should also be updated, along with any other such events, before this gets merged and deployed.

tburry:
Oh shoot good catch. What about the change I made instead?

initvector:
That'll do!

PR #3693, opened 2016-04-09
Manage all addons through a central class

tburry:
- Manage all addons with one **AddonManager** class.
- Build a static cache of all addons for better speed.
- Keep the old manager classes around, but use the **AddonManager** within them.

This is a mega change and will require some serious QA. However, there are a substantial number of unit tests for this functionality. Any issues should be reproducible with unit tests.

*Note: Some current addons/theme will break under this new scheme and must be fixed. In order to protect against doing a deploy with broken addons you can run the unit tests on the data server and it will test all the symlinked addons.*


tburry:
Remember to run the unit tests on your local environment. There are some backwards compatibility issues with theme and locale keys that should be fixed before this PR is deployed.

PR #3704, opened 2016-04-12
Add and update lang attributes on HTML tags

### LINKED ISSUE #3698: Ensure every HTML document has a lang attribute
**Client Finding:** The language of the page MUST be identified using the HTML lang attribute. The primary language of the page is not defined. The <html> element does not have a lang attribute.

**Client Recommendation:** Ensure every HTML document has a lang attribute and its value is valid.

initvector:
This update ensures all pages have a `lang` attribute on the HTML tag that reflects the site's currently configured locale.

Closes #3698


tburry:
Shouldn't these be output based on Gdn::locale() rather than the config directly?

initvector:
Good call.  Direct config checks have been swapped out in favor of calls to `Gdn::locale()->language()`.

tburry:
Small wrinkle here: It looks like Chinese may not always use the same [code](http://www.w3schools.com/tags/ref_language_codes.asp). Recommendation:

1. Do a check for "zh_TW" and return "zh-Hant".
2. Check to see how Gdn::locale()->language() is used. If it is used in ways that MUST have a two digit code then make a parameter to return the $iso6391 code with the above exception.

Issue #3536, opened 2016-02-16
Hide /messages/add link on other users profile if permission is restricted 


LauraMakesArt:
Users still see a link to send a message when viewing other members profiles, even if Conversations>Add is removed for their current role. The link just leads to the generic permission error for the user.  Currently it needs to be themed out.

PR #3712, opened 2016-04-13
Check Conversation.Add permission before showing "Message" button on other users profile

R-J:
This solves https://github.com/vanilla/vanilla/issues/3536
I've dropped the session->isValid check, since the permission would also fail for invalid sessions.


linc:
Thanks!

bleistivt:
This breaks the button, the correct permission (confusingly) is `Conversations.Conversations.Add`

PR #3716, opened 2016-04-13
Remove explicit reCaptcha handling from core, move to core plugin

kaecyra:
I had to add a Garden.Registration.SkipCaptcha setting to allow the unit tests to pass, since they can't submit captcha. There is no UI way to set that setting, but it could also be conceivably used by admins that reaaaaally don't want Captcha.

Also cleaned up the Basic/Captcha tomfoolery in entrycontroller since it was removed very haphazardly prior.


hgtonight:
hallelujah

kaecyra:
Don't cheer too soon, I am definitely rusty ;)

linc:
How did we merge conflict this already

applications/dashboard/views/entry/registerbasic.php
```diff
@@ -24,7 +25,7 @@
             <li>
                 <?php
                 echo $this->Form->label('Username', 'Name');
-                echo $this->Form->textBox('Name', array('Wrap' => true, 'autocorrect' => 'off', 'autocapitalize' => 'off'));
+                echo $this->Form->textBox('Name', array('autocorrect' => 'off', 'autocapitalize' => 'off', 'Wrap' => TRUE));
```
linc: wut

applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -1269,27 +1269,41 @@ public function register($InvitationCode = '') {
...
     }
 
     /**
      * Select view/method to be used for registration (from config).
      *
      * @access protected
-     * @since 2.0.0
-     *
-     * @return string Method name.
+     * @since 2.2.106
```
linc: I've been going with the standard of labeling these with the first _release_ it will be since / deprecated in, which is 2.3 in this case. I think that makes better sense for folks outside the company.

library/core/class.form.php
```diff
@@ -274,29 +274,6 @@ public function captcha() {
             // A plugin handled the captcha so don't display anything more.
```
linc: At this point there's no purpose to the entire `$handled` construct, right? This just exists to fire an event, now.

library/core/class.captcha.php
```diff
@@ -0,0 +1,80 @@
...
+        $controller->fireAs('captcha')->fireEvent('render');
+    }
+
+    /**
+     * Validate captcha
+     *
+     * @param mixed $value
+     */
+    public static function validate($value = null) {
+
```
linc: Blank opening method lines make Lord Brackos weep.

plugins/recaptcha/class.recaptcha.plugin.php
```diff
@@ -0,0 +1,246 @@
+<?php
+
+/**
+ * @copyright 2010-2016 Vanilla Forums Inc
+ * @license Proprietary
+ */
+
+$PluginInfo['recaptcha'] = array(
+    'Name' => 'Recaptcha Support',
```
linc: They call it `reCAPTCHA` so let's stick with that stupid capitalization in user-facing labels.

plugins/recaptcha/class.recaptcha.plugin.php
```diff
@@ -0,0 +1,246 @@
...
+ * Changes:
+ *  0.1        Development
+ *
+ * @author Tim Gunter <tim@vanillaforums.com>
+ * @package vanilla
+ */
+class RecaptchaPlugin extends Gdn_Plugin {
+
+    /**
+     * reCAPTCHA private key
```
linc: Pretty sure the format for property doc blocks is `/** @var string reCAPTCHA private key. */`

plugins/recaptcha/class.recaptcha.plugin.php
```diff
@@ -0,0 +1,246 @@
...
+    public function __construct() {
+        parent::__construct();
+
+        // Get keys from config
+        $this->privateKey = c('Plugins.Recaptcha.PrivateKey');
+        $this->publicKey = c('Plugins.Recaptcha.PublicKey');
+    }
+
+    /**
+     * Override private key in memory
```
linc: Around these here parts we use proper stops after our summaries, buckaroo.

plugins/recaptcha/class.recaptcha.plugin.php
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Hook plugin enabled
+     *
+     * Move old keys to new location.
+     */
+    public function setup() {
+        // Move private key
+        if (c('Garden.Registration.CaptchaPrivateKey')) {
+            saveToConfig('Plugins.Recaptcha.PrivateKey', c('Garden.Registration.CaptchaPrivateKey'));
```
linc: If we had it all to do over, we'd have never created the `Plugins.` config namespace, so let's not make a new one under it.

plugins/recaptcha/class.recaptcha.plugin.php
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Display reCAPTCHA entry field
+     *
+     * THIS METHOD ECHOS DATA
+     *
+     * @param Gdn_Form $sender
+     * @return string
+     */
+    public function gdn_form_captcha_handler($sender) {
+
```
linc: Tears of the Brackos

linc:
I suggest we add to core's structure:

```
if (c('Garden.EnabledPlugins.Recaptcha', null) === null) {
    saveToConfig('Garden.EnabledPlugins.Recaptcha', true);
}
```

Taking advantage of the fact that our disable plugin mechanism uses an explicit `false` instead of removing it from the config to avoid undoing someone turning it off on purpose.

kaecyra:
Todd and I were discussing this, and we're gunna add a TouchConfig in the core structure. So yeah same idea.

applications/dashboard/views/entry/registerbasic.php
kaecyra: There used to be 2 views, registerbasic and registercaptcha.
Due to the "refactor" done by Mark in the before times, Captcha-less registration was not supported but both views still existed. 

Various overrides existed in the code to ensure that even though `Basic` was the only option provided in the UI, `Captcha` is what ended up in the config and caused `registercaptcha` to be used. 

I just properly consolidated this logic. Thereafter, I fixed these views by deleting `registerbasic` and renaming `registercaptcha` to `registerbasic`. This is what was already in `registercaptcha`.

linc:
I always forget touchConfig.

initvector:
Rebased and moved here: https://github.com/vanilla/vanilla/pull/3761

PR #3720, opened 2016-04-14
Don't let objects creep into makeTreeChildren()

linc:
Our object pollution knows no boundaries. Make sure we really have an array before using it like one, even this deep in the categories collection.


tburry:
Why is there a composer lock committed here?

linc:
Fixt.

PR #3721, opened 2016-04-14
Remove @error suppression in a few places

tburry:
We are getting scrutinized for these things.


tburry:
This fixes issues being reported as security concerns by scrutinizer. Looking at the code I don't think it's the case, but I want the issues cleared out nonetheless.

In general we must avoid error suppression whenever possible.

PR #3723, opened 2016-04-14
Add caching to UserModel::getRoles

### LINKED ISSUE #3686: Investigate Dashboard app hook that queries DB for roles on every page load
Possibly causing excessive database load.

initvector:
`UserModel::getRoles` currently attempts to load a user's current roles from the cache.  Failing that, the function queries the database to retrieve the user's current roles.  However, that result is never saved to the cache.  This means the user's roles are never cached.

This update aims to cache a user's roles with `UserModel::userCacheRoles`, a function that doesn't seem to be used anywhere, when queried with `UserModel::getRoles`.


tburry:
Let me go through this one.

tburry:
This change won't help with the issue at hand. The meta that's added uses RoleModel->getByID() and not the methods on the user model and that method hits the database every time.

initvector:
Removed the issue referenced, since it addresses a similar issue, but not the specific issue originally referenced

applications/dashboard/models/class.usermodel.php
```diff
@@ -4370,22 +4372,24 @@ public function userCache($User, $UserID = null) {
...
-     * @param type $UserID
-     * @param type $RoleIDs
-     * @return type
+     * @param int $userID The ID of a user to cache roles for.
+     * @param array $roles A collection of role records associated with the user.
+     * @return bool Was the caching operation successful?
      */
-    public function userCacheRoles($UserID, $RoleIDs) {
-        if (is_null($UserID) || !$UserID) {
+    public function userCacheRoles($userID, $roles) {
```
tburry: These are actually still $roleIDs aren't they?

applications/dashboard/models/class.usermodel.php
```diff
@@ -4370,22 +4372,24 @@ public function userCache($User, $UserID = null) {
...
+        if (is_null($userID) || !$userID) {
```
tburry: Pretty sure the original intention of this was:

```php
 if ($userID !== 0 && !$userID) {
```

It will still cache the guest roles this way. Not sure though, but as it stands the `is_null` check is meaningless.

PR #3734, opened 2016-04-15
Refactor database serialization calls

### LINKED ISSUE #3685: Refactor database attribute serialization
We will be migrating database attribute type fields to JSON at some point. To prepare for this we want to pass all serialize() and unserialize() calls with respect to the database through helper functions so that we can later make the change in one place.

1. Add dbencode() and dbdecode() to functions.general. These functions should initially be aliases for serialize() and unserialize(), but dbdecode() will only have one parameter.
2. Replace existing calls to serialize() and unserialize() with dbencode() and dbdecode() where appropriate. This would be in all places where attribute style arrays are being saved/loaded from the database. This should be most if not all calls.
3. Make a note of additional calls to serialize()/unserialize(). We may want to get rid of these too if we can.

initvector:
This update refactors encoding and decoding of certain values, primarily `Attributes`, stored in the database.

Closes #3685

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -722,6 +722,30 @@ if (!function_exists('safePrint')) {
     }
 }
 
+if (!function_exists('dbencode')) {
+    /**
+     * Encode a value in preparation for database storage.
+     *
+     * @param mixed $value
+     * @return string
+     */
+    function dbencode($value) {
+        return serialize($value);
+    }
+}
+
+if (!function_exists('dbdecode')) {
+    /**
+     * Decode a value retrieved from database storage.
+     *
+     * @param string $value
+     * @return mixed
+     */
+    function dbdecode($value) {
+        return unserialize($value);
+    }
+}
+
 if (!function_exists('decho')) {
     /**
      * Echo debug messages and variables.
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -730,7 +730,13 @@ if (!function_exists('dbdecode')) {
      * @return mixed
      */
     function dbdecode($value) {
-        return Gdn_Format::unserialize($value);
+        try {
+            $decodedValue = unserialize($value);
+        } catch (Exception $e) {
+            $decodedValue = false;
+        }
+
+        return $decodedValue;
     }
 }
 
@@ -742,7 +748,13 @@ if (!function_exists('dbencode')) {
      * @return string
      */
     function dbencode($value) {
-        return Gdn_Format::serialize($value);
+        try {
+            $encodedValue = serialize($value);
+        } catch (Exception $e) {
+            $encodedValue = false;
+        }
+
+        return $encodedValue;
     }
 }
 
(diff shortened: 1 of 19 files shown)


library/core/functions.general.php
```diff
@@ -746,6 +734,18 @@ function dbdecode($value) {
...
 
+if (!function_exists('dbencode')) {
+    /**
+     * Encode a value in preparation for database storage.
+     *
+     * @param mixed $value
+     * @return string
+     */
+    function dbencode($value) {
+        return Gdn_Format::serialize($value);
```
DaazKu: Gdn_Format::serialize($value); is incompatible with unserialize();

initvector: `dbencode` and `dbdecode` are not meant to be drop-in replacements for `serialize` and `unserialize`.  They're only meant to be helpers for getting complex data into and out of the database.  I don't believe there is currently a legitimate case wherein a string or primitive type needs to undergo encoding, such a serialization, to be stored.  To my knowledge, the types of data needing to be encoded are exclusively arrays and objects.  If we're trying to serialize strings and primitives somewhere, I think that's a separate issue that should be addressed.

initvector:
##### Serialization Present in Non-database Areas
* applications/dashboard/controllers/class.entrycontroller.php: Cookie contents stashed with `serialize`
* applications/dashboard/controllers/class.settingscontroller.php: `addUpdateCheck` uses `Gdn_Format::serialize` for value of `UpdateChecks` JavaScript definition
* applications/dashboard/models/class.updatemodel.php: `serialize` used on addon requirements
* library/core/class.cache.php: `serialize` used on values for caching
* library/core/class.configuration.php: `unserialize` used on some config values
* library/core/class.format.php: `image` uses `serialize`
* library/core/class.memcached.php: `serialize` used on values for caching
* library/core/class.passwordhash.php: `unserialize` used in checkPassword's xenforo routine
* library/core/class.pluginmanager.php: availablePlugins, indexSearchPath `serialize` used in hashing dir contents
* library/core/class.router.php: `unserialize` used in `_parseRoute`
* library/core/class.thememanager.php: availableThemes `serialize` used in hashing dir contents

tburry:
Definitely don't want these going through Gdn_Format please. That was the absolute wrong place for these calls at the time and I don't want to continue that.

Also, make the calls to dbencode/dbdecode never error and remove error suppression from any calls. We need to get the "@" sign out of our app.

tburry:
PS: That "needs testing" label screams "write some unit tests" to me. These functions have no dependencies so are a perfect fit for unit tests.

library/core/functions.general.php
```diff
@@ -722,6 +722,42 @@ function safePrint($mixed, $returnData = false) {
...
+if (!function_exists('dbdecode')) {
+    /**
+     * Decode a value retrieved from database storage.
+     *
+     * @param string $value
+     * @return mixed
+     */
+    function dbdecode($value) {
+        try {
+            $decodedValue = unserialize($value);
```
tburry: I think if you put a test for `$value === null` then you are going to catch a very real scenario with null db values and avoid the exception handler.

initvector:
- [X] Add basic unit tests (encoding/decoding objects and arrays)
- [X] Add specific handling for `null` values
- [X] Remove error suppression from calls

PR #3737, opened 2016-04-15
Remove afterControllerCreate hook from Dashboard

### LINKED ISSUE #3686: Investigate Dashboard app hook that queries DB for roles on every page load

initvector:
Remove `afterControllerCreate` hook from the Dashboard application.

Relevant applications and plug-ins using the behavior of this hook will need to have it added to their respective add-on code before this PR is merged.

Closes #3686

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -417,19 +417,4 @@ class DashboardHooks implements Gdn_IPlugin {
             PermissionModel::resetAllRoles();
         }
     }
-
-    /**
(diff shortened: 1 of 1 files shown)


tburry:
@linc, can you schedule this for next sprint? Also, @nervoustwit can you link to this PR when you do your change for <name withheld>?

linc:
This wasn't just added for a plugin, it was also added to allow the use of third-party analytics systems with role as a tracking value. This hook vastly simplifies that process and *reduces* the load caused by previous Javascript workarounds like pinging /profile constantly.

Issue #1914, opened 2014-05-09
Event "CategoryWatch" in CategoryModel is broken


R-J:
Trying to hook the event with `public function CategoryModel_CategoryWatch_Handler($Sender)` doesn't work, because the event is fired that way:

~~~
public static function CategoryWatch() {
...
      Gdn::PluginManager()->EventArguments['CategoryIDs'] =& $Watch;
      Gdn::PluginManager()->FireEvent('CategoryWatch');
~~~

I searched in the forum and found that discussion: http://vanillaforums.org/discussion/comment/173148#Comment_173148
Seems like the error hasn't been reported back then.

bleistivt:
Looks like this was fixed in #3740

R-J:
Yes, it is fixed.

PR #3740, opened 2016-04-18
Fire "CategoryWatch" as an event of CategoryModel

R-J:
The event CategoryWatch has been fired as an event of the plugin manager. This one closes this issue: https://github.com/vanilla/vanilla/issues/1914


R-J:
I guess this will also be needed for `Gdn::pluginManager()->fireEvent('AllowedDiscussionTypes');` in CategoryModel::allowedDiscussionTypes(). https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L74

I have no installation of current master, so I'm not able to test that one.

R-J:
Since that is a very simple change, I guess it should also be included in release 2.3

PR #3753, opened 2016-04-18
Deprecate UserModel->definePermissions() with serialization

tburry:
- [x] Please do a search through all source code to make sure definePermissions() is called correctly.


initvector:
The code looks good to me.  Source code appears to [avoid using serialization](https://github.com/search?utf8=%E2%9C%93&q=org%3Avanilla+definePermissions&type=Code) with this function.

The PR is a few months old, so there are conflicts.  Once these are resolved, I'll continue testing.

linc:
@initvector Can you resolve the conflicts please?

PR #3754, opened 2016-04-18
Change error handling on dbencode/dbdecode

tburry:
- The documentation for serialize does not mention triggering any errors so the exception handling has been removed.
- Change exception handling in dbdecode to @ error suppression to prevent notices in the debug bar.
- Fix the deserialization in discussion tags which has a bunch of mixed data.


tburry:
Pro tip: Sometimes the split code diff view makes reviewing a PR a bit easier.

PR #3756, opened 2016-04-19
Load nestedSortable from js/library path

austins:
Fixes the js from not loading on some instances by letting Garden know to load it from the global js directory.


DaazKu:
Could you be more specific on "which instances" leads to the file not loading?

- Prepending 'js/library'  does not work on the latest version
- using `/js/library` is deprecated

"As is" the correct path should be determined here: https://github.com/austins/vanilla/blob/14dc9aadd14ee27cf12675668e36cc82776cca2e/applications/dashboard/models/class.assetmodel.php#L431

PR #3761, opened 2016-04-20
Refactor Recaptcha into a plugin

initvector:
Rebased https://github.com/vanilla/vanilla/pull/3716
>I had to add a Garden.Registration.SkipCaptcha setting to allow the unit tests to pass, since they can't submit captcha. There is no UI way to set that setting, but it could also be conceivably used by admins that reaaaaally don't want Captcha.
>
>Also cleaned up the Basic/Captcha tomfoolery in entrycontroller since it was removed very haphazardly prior.


linc:
That's fuckin' teeeeamwooooork!

kaecyra:
!(Yup)[http://i.imgur.com/DaMHupe.jpg]

kaecyra:
Yup

PR #3763, opened 2016-04-20
Ensure Garden.Email.Prefix displays even if there is no email message.

beckyvb:
Makes sure that the config setting is respected in Activity emails with no message.


tburry:
I feel like you should make a private helper function when you see the same code in three places like this.

PR #3767, opened 2016-04-21
Add the ability to capture a user's time zone

initvector:
This update adds the ability to record a user's current time zone to their attributes.  Knowing the time zone, instead of only the offset, is helpful in determining and compensating for the occurrence of DST in past and future dates.


tburry:
I'm very uneasy about this. I can't see that regex working across all locales. I know there was another issue relating to this. Can you tell me what the bigger picture is?

linc:
If you create a post or start an event with a date inside DST, if you view it outside of DST all post times and event times will be off by an hour. The issue was reported as effecting future scheduled events in Groups, but in considering it we realized it would also effect any discussion, comment, or notification in the past as well. I don't think mis-reporting timestamps by an hour at essentially random intervals is something we can shrug off, so I asked Ryan to attempt a core patch for dealing with DST once and for all.

tburry:
Yeah Ryan and I had a talk about it, although he mentioned the issue the opposite way. That is if you schedule an event in the future then it will display as off by an hour now, but the correct time when you are closer to the event. That's correct with the current code IMO.

This PR currently needs more robust timezone checking, including:

1. Some amount of research into how other locales might transmit their timezone.
2. Checking for a correct timezone before setting it on a user.
3. Making sure that a timezone with daylight's saving time will format a date outside of daylight's saving time correctly.

initvector:
Just to recap: We keep track of the user's *current* UTC offset.  We use that to localize dates/times.  If that user's region happens to observe DST, they can potentially have two different offsets.  Someone viewing the site in June for an event in December might see it being an hour off, because in June they're UTC-4 and in December they're UTC-5.  Once they crossed from EDT to EST, they'd see the proper date/time again, because their UTC offset at that time would be inline with the event's.

For events, this is probably the biggest issue when you're *close* to the DST switchover. You might be planning to attend an event on March 14th.  All the way up until March 12th, that event will appear to start at 5:00 PM.  However, on March 13th, DST kicks in and your offset drops from -5 to -4.  Now you'd see the event is actually taking place at 6:00.

library/core/class.session.php
```diff
@@ -213,6 +213,24 @@ public function getCookie($Suffix, $Default = null) {
     }
 
     /**
+     * Return the time zone for the current user.
+     *
+     * @return string A valid IANA time zone identifier.
+     */
+    public function getTimeZone() {
+        $timeZone = $this->getAttribute('TimeZone');
```
tburry: Add a short circuit for the guest timezone if the session isn't valid.

library/core/class.session.php
```diff
@@ -213,6 +213,24 @@ public function getCookie($Suffix, $Default = null) {
     }
 
     /**
+     * Return the time zone for the current user.
+     *
+     * @return string A valid IANA time zone identifier.
+     */
+    public function getTimeZone() {
```
tburry: Can this return a proper DateTimeZone object?

library/core/class.session.php
```diff
@@ -213,6 +213,24 @@ public function getCookie($Suffix, $Default = null) {
...
+        $timeZone = $this->getAttribute('TimeZone');
+        $hourOffset = $this->hourOffset();
+
+        if ($timeZone) {
+            return $timeZone;
+        } elseif (is_int($hourOffset)) {
+            return "Etc/GMT/{$hourOffset}";
```
tburry: The our offset needs to be passed through `sprintf('%+d', $hourOffset)` so that the "+" sign is captured.

tburry:
Hey @initvector, I fixed some bugs in this PR and tweaked the implementation so I assigned the review back to you.

PR #3769, opened 2016-04-22
Clean up inspection issues with the ActivityModel

tburry:
- Code casing.
- Doc blocks.
- Unused variables.


initvector:
Attempting to delete an activity throws the following:
>ActivityModel->delete() is not supported.

tburry:
Fixed.

PR #3773, opened 2016-04-22
Support old overridden views by hooking existing event and passing forward to Captcha::render()

kaecyra:


plugins/recaptcha/class.recaptcha.plugin.php
```diff
@@ -154,10 +160,26 @@ public function captcha_settings_handler($sender) {
...
+     *
+     * Older overridden views will not have the new Captcha::render() method.
+     * This hook catches the old registerFormBeforeTerms event and re-fires
+     * Captcha::render() if the captcha has not already rendered.
+     *
+     * @param Gdn_Controller $sender
+     */
+    public function entrycontroller_registerFormBeforeTerms_handler($sender) {
+        if (!$this->rendered) {
+            Captcha::render($sender);
```
tburry: Shouldn't this also set `$this->rendered = true`?

PR #3782, opened 2016-04-25
Handle Birthdays in Profile Extender exporter.

beckyvb:
Handles the special case of birthdays on utility/exportprofiles. Birthdays are saved on the User table.


plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -527,6 +527,12 @@ public function utilityController_exportProfiles_create($sender) {
             ->where('u.Deleted', 0)
             ->where('u.Admin <', 2);
 
+        if (val('DateOfBirth', $fields)) {
```
tburry: As an FYI: val() is a bit of overkill in this case. Using `!empty($fields['DateOfBirth'])` would be better.

PR #3783, opened 2016-04-25
Make /entry/target internal

tburry:
Although this won’t do anything when browsed to it really shouldn’t be an endpoint.


tburry:
Tangentially related to #2800.

PR #3789, opened 2016-04-26
Add function to format IP addresses for display

### LINKED ISSUE #3687: Pass all IP addresses through a view helper
We have the ipAnchor() function which is great. We need another helper function that formats an IP address without putting it into an anchor. The ipAnchor() should then call this for the display portion of its text. We then need to make sure that no IP address goes into the view without going through the helper function or ipAnchor(). Some clever find in files regexing may make things easier.

initvector:
This update adds a function to format IP addresses for display.  In addition, it aims to ensure any IP address that might be displayed are done so with either `ipAnchor` or this new function.

Closes #3687

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -563,11 +563,11 @@ if (!function_exists('formatIP')) {
      * Format an IP address for display.
      *
      * @param string $IP An IP address to be formatted.
-     * @param bool $escape Convert special characters to HTML entities.
-     * @return string The formatted IP address.
+     * @param bool $html Format as HTML.
+     * @return string Returns the formatted IP address.
      */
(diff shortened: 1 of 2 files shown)


library/core/functions.render.php
```diff
@@ -558,6 +558,19 @@ function fixnl2br($Text) {
...
 
+if (!function_exists('formatIP')) {
+    /**
+     * Format an IP address for display.
+     *
+     * @param string $IP An IP address to be formatted.
+     * @param bool $convert Convert special characters to HTML entities.
+     * @return string The formatted IP address.
+     */
+    function formatIP($IP, $convert = true) {
```
tburry: Maybe rename `$convert` to `$escape`?

PR #3798, opened 2016-04-27
Use the correct permission in category management actions

DaazKu:
Managing a category is related to `Garden.Community.Manage` not `Garden.Settings.Manage`

With this fix users with `Garden.Community.Manage` won't get permission problem when deleting(categories and categories' photo) and sorting categories

### DIFF
--- a/applications/vanilla/controllers/class.settingscontroller.php
+++ b/applications/vanilla/controllers/class.settingscontroller.php
@@ -261,7 +261,7 @@ class SettingsController extends Gdn_Controller {
      */
     public function addCategory() {
         // Check permission
-        $this->permission('Garden.Community.Manage');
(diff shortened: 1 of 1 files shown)


linc:
See `Gdn::session()->checkRankedPermission()`. That'd make better sense here so there isn't a silly misconfiguration where an admin gets locked out because they didn't have 'Communuty' checked.

DaazKu:
Since the permission method of the controller also do a redirect on failure I kept using it but I now check for any of the 2 permissions now.

linc:
Throwing an exception would be an acceptable way to do it, too. This is a little less nice since it's not forward-compatible with future permission ranking changes, but since the issue here is really just backwards compatibility I'll leave it.

PR #3827, opened 2016-04-29
Implement Logger wrapper methods

### LINKED ISSUE #3824: Make the Logger global object be a static wrapper for the Vanilla\Logger

tburry:
- Add Logger::addLogger().
- Add Logger::removeLogger().
- Add Logger::getLevels().
- Change Logger::setLogger().

See the testing notes on the linked PR.

Closes #3824.

### DIFF
--- a/library/core/class.logger.php
+++ b/library/core/class.logger.php
@@ -124,12 +124,92 @@ class Logger {
     }
 
     /**
-     * Log an event
+     * Log a debug message.
      *
-     * @param string $event
-     * @param string $level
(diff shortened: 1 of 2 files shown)


initvector:
`Logger::log` and `Logger::event` both have an empty array as the default value of `$context`.  Should the new shortcut functions (e.g. `Logger::debug`) have that default value for their `$context` parameter as well?

tburry:
Good point. Fixed.

PR #3849, opened 2016-05-01
Prefill the username when using a connection method

R-J:
The ConnectName field has not been prefilled because it was set with setFormValue(), but setValue() has to be used.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -623,7 +623,7 @@ EOT;
             }
 
             if (!isset($NameFound) && !$IsPostBack) {
-                $this->Form->setFormValue('ConnectName', $this->Form->getFormValue('Name'));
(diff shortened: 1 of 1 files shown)


DaazKu:
Could I have more info on what setup did you use to find the problem in the first place?
Especially what method of connection are you using and with what configurations.

There are dozens of possibilities here and I cannot seem to find one where the the connect name would have to be prefilled.

R-J:
I stumbled upon this when I was testing the steam plugin: https://vanillaforums.org/discussion/comment/240252/#Comment_240252

I'll try if I can reproduce it.

R-J:
Strange... I've tried it right now with 2.3 beta and 2.2.1 and both are working flawlessly. Maybe that day Steam had a hiccup and I simply interpreted the normal function being a result of my changes...

DaazKu:
Thank you for your double check!

PR #3864, opened 2016-05-02
Add EventListener to comment post button

francisbo:
If you were in preview mode and posted your comment then the button bar would stay missing until you toggle preview mode again.  This adds an event listener to the post comment button that makes the button bar re-appear if you post the comment while in preview mode.

### DIFF
--- a/plugins/editor/js/editor.js
+++ b/plugins/editor/js/editor.js
@@ -1492,6 +1492,10 @@
                     $currentEditorToolbar.show();
                 });
 
+                $(document).on('click', 'input.CommentButton', function() {
+                    $currentEditorToolbar.show();
+                });
+
                 switch (format) {
                     case 'wysiwyg':
                     case 'ipb':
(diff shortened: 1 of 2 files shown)


plugins/editor/js/editor.js
```diff
@@ -1492,6 +1492,10 @@
                     $currentEditorToolbar.show();
                 });
 
+                $(document).on('click', 'input.CommentButton', function() {
+                    $currentEditorToolbar.show();
+                });
+
```
DaazKu: Changing the click event above from

`$(document).on('click', 'a.WriteButton', function() {`
to
`$(document).on('click', 'a.WriteButton,input.CommentButton', function() {`

would have the same effect.
Also it saves one more handler to be attached to the document :D

francisbo: Noted and added into code 👍

tburry:
I'm hoping this can be superseded by the new contentLoad event. Becky will know.

PR #3865, opened 2016-05-02
Opt out of sort/filter rendering

beckyvb:
Allows the discussion module renderer to opt out of showing sorts or filters.

### DIFF
--- a/applications/vanilla/modules/class.discussionssortfiltermodule.php
+++ b/applications/vanilla/modules/class.discussionssortfiltermodule.php
@@ -33,6 +33,12 @@ class DiscussionsSortFilterModule extends Gdn_Module {
     /** @var array The selected filters. */
     protected $selectedFilters;
 
+    /** @var bool Whether to show the sorting options in the view. */
+    protected $showSorts = true;
+
+    /** @var bool Whether to show the filter options in the view. */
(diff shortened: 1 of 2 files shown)


applications/vanilla/modules/class.discussionssortfiltermodule.php
```diff
@@ -55,6 +61,34 @@ public function __construct($categoryID = 0, $selectedSort = '', $selectedFilter
     }
 
     /**
+     * @param $showSorts Whether to show the sorting options in the view.
+     */
+    public function setShowSorts($showSorts) {
+        $this->showSorts = $showSorts;
```
tburry: return $this.

applications/vanilla/modules/class.discussionssortfiltermodule.php
```diff
@@ -55,6 +61,34 @@ public function __construct($categoryID = 0, $selectedSort = '', $selectedFilter
...
+    }
+
+    /**
+     * @param $showFilters Whether to show the filtering options in the view.
+     */
+    public function setShowFilters($showFilters) {
+        $this->showFilters = $showFilters;
```
tburry: return $this.

PR #3867, opened 2016-05-03
Fix br tag causing malformed document

skinofstars:
This effects new installs in some circumstances.

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -545,7 +545,7 @@ class Gdn_Controller extends Gdn_Pluggable {
     public function definitionList($wrap = true) {
         $Session = Gdn::session();
         if (!array_key_exists('TransportError', $this->_Definitions)) {
-            $this->_Definitions['TransportError'] = T('Transport error: %s', 'A fatal error occurred while processing the request.<br />The server returned the following response: %s');
+            $this->_Definitions['TransportError'] = T('Transport error: %s', 'A fatal error occurred while processing the request. The server returned the following response: %s');
         }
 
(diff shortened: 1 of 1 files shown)


library/core/class.controller.php
```diff
@@ -545,7 +545,7 @@ public function data($Path, $Default = '') {
     public function definitionList($wrap = true) {
         $Session = Gdn::session();
         if (!array_key_exists('TransportError', $this->_Definitions)) {
-            $this->_Definitions['TransportError'] = T('Transport error: %s', 'A fatal error occurred while processing the request.<br />The server returned the following response: %s');
+            $this->_Definitions['TransportError'] = T('Transport error: %s', 'A fatal error occurred while processing the request. The server returned the following response: %s');
```
DaazKu: The <br> was probably there for a reason. It should be replaced by `\n` not space.

skinofstars: Any line-break here seems presentationally unnecessary.

DaazKu: `alert(definition('TransportError').replace('%s', status));` ?

DaazKu: Also, what is the actual error? `$this->_Definitions` is passed through `json_encode` and I don't see how this is raising an error.

skinofstars: It doesn't throw a PHP error, it returns a malformed HTML page that browsers can't display.

DaazKu:
Could you be more specific on the circumstances that leads to give you an error?
I'm asking because I did a fresh install of the forum and going on the forum (with Chrome) I don't have any error.

PR #3868, opened 2016-05-03
Allow the Photo field to be set in /profile/edit

tburry:
Allow moderators to set the Photo field to a URL in /profile/edit. This was functionality that was originally removed, but is getting eased back in. Since the photo would be previously filtered out that behavior is being reproduced for non-admins.

There is a cursory check to make sure the Photo field is a valid URL, but built in function isn’t exhaustive. Since this functionality is mod-only I think this is okay for now, but a more robust URL validator will need to be added in the future.


tburry:
This change comes with unit tests for you, the reviewer.

PR #3874, opened 2016-05-04
Change the protection level of Gdn_DatabaseStructure->rowCountEstimates

tburry:
Since subclasses set the property then it should be protected, not private.


DaazKu:
See: https://github.com/vanilla/vanilla/pull/3871

PR #3875, opened 2016-05-04
Category Scaling Part 1

tburry:
Help scale categories for communities with too many categories to iterate every page load. The entire category table is still loaded when necessary so there are only some pages that this optimization will work. Some testing tips:

- The checklist below can be used as a testing checklist too!.
- Make sure you turn off the categories in the side panel when testing this PR.
- In the debug bar, you should see the following message when all categories have been loaded: **CategoryModel::loadAllCategories**.
- If you are seeing the above message all the time then some plugin may be loading all of the categories. It would be helpful to figure out which one and post it here so the plugin can be refactored, if possible.
- It is imperative to test this PR with caching ON.
- There is some messy stuff in this PR and that's unavoidable. The idea is to slowly refactor functionality until we are completely in the clear.

Implementation Checklist:

- [x] Don't load the entire category list when in a discussion.
- [x] Don't show the category dropdown when posting to a category, hence not loading the category list.
- [x] Saving a category shouldn't show stale data when caching is on.
- [x] Sorting categories shouldn't show stale data when caching is on.
- [x] Looking at an individual category shouldn't load the entire category list.

I'm marking this PR as a blocker because I'll be basing another PR off of it.

### DIFF
--- a/applications/vanilla/library/class.categorycollection.php
+++ b/applications/vanilla/library/class.categorycollection.php
@@ -285,6 +285,23 @@ class CategoryCollection {
     }
 
     /**
+     * Get the children of a category.
+     *
+     * @param int $categoryID The category to get the children for.
+     * @return array Returns an array of categories.
+     */
+    public function getChildren($categoryID) {
+        $children = $this
+            ->sql
+            ->select('CategoryID')
+            ->getWhere('Category', ['ParentCategoryID' => $categoryID])
+            ->resultArray();
+        $ids = array_column($children, 'CategoryID');
+        $categories = $this->getMulti($ids);
+        return $categories;
+    }
+
+    /**
      * Get several categories by ID.
      *
      * @param array $categoryIDs An array of category IDs.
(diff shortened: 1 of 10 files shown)


applications/vanilla/library/class.categorycollection.php
```diff
@@ -0,0 +1,417 @@
...
+     *
+     * @param array $category The category to update.
+     * @return bool Returns **true** if the category updated or **false** otherwise.
+     */
+    public function update(array $category) {
+        if (empty($category['CategoryID'])) {
+            throw new Gdn_UserException("Category ID is required.");
+        }
+
+        $category += [
```
DaazKu: I don't think that the update function is used yet but this would not update the `DateUpdated` field on a category that already has it. (Same for the UpdateUserID)

tburry: This is a WIP PR and the class in question is an intermediate refactoring class. Don't get stuck in the weeds here.

PR #3880, opened 2016-05-06
Remove regex "e" modified since it is deprecated in PHP 7

DaazKu:
htmlEntityDecode returns an empty string in PHP 7 because the `e` modifier does not exist anymore.
This cause fetchPageInfo to return empty Title and Description. (Embedding is broken)


DaazKu:
class.phpmailer.php `EncodeQ()` is also using the `e` modifider.

We have that library in core and in the BluehostFix plugin.

PR #3881, opened 2016-05-06
Pass the query string from embedded pages out to the parent frame

tburry:
This is a hopefully more simple version of #3672. The query string is set in the controller hooks and then just joined in the path. Only altering embed_local.js helps with cache busting.


initvector:
Looking good.  Just need a version bump and I think we're all set here.

PR #3913, opened 2016-05-10
Support optional bootstrap-style css classes in Gdn_Form

### LINKED ISSUE #3852: Allow for Gdn_Form to support bootstrapish classes

tburry:
- Setting `“ControlStyle” => “bootstrap”` in theme info will turn the functionality on.
- The attribute() function will no longer output empty attributes.
- Some bootstrap classes have been added to existing views. (form-group, some radio-inline).

Closes #3852.


library/core/class.form.php
```diff
@@ -134,6 +174,68 @@ public function addErrorClass(&$Attributes) {
...
+        }
+    }
+
+    /**
+     * Translate old CSS classes using the style array.
+     *
+     * @param string|string[] $classes The classes to translate.
+     * @return string Returns the translated class string.
+     */
+    private function translateClasses($classes) {
```
beckyvb: This function on overrides the custom css classes the dev may have placed on the element. For example something like `$this->Form->input('DefaultAvatar', 'file', array('class' => 'js-new-avatar-upload Hidden'));` does not render the passed CSS classes any longer.

tburry: Okay I dealt with the issue and added your specific test case in a unit test. Thanks!

PR #3916, opened 2016-05-11
Add TagModel->addDiscussion()

tburry:
This method should add existing tags to a discussion. It should not wipe out tags that have already been assigned to the discussion.

- [x] Add the tags to the discussions.
- [x] Make sure the method doesn't fail when some of the tags are already assigned to the discussion.
- [x] Make sure the counts are properly updated in the tag table.


DaazKu:
Since "name" is not unique across all categories I removed the possibility to pass a list of tag names.

tburry:
@DaazKu, not allowing names is fine. What I recommend is you give me some candidates for additional methods to add to the TagModel in the future that will help developers deal with this limitation. In the past we've been burned by not providing full functionality to libraries and then developers get stuck.

This is not something to do now. I've added #3969 for the future. I think this will be a good exercise.

PR #3948, opened 2016-05-12
Tweak popins

tburry:
- Allow popins to work off the .js-popin class. This should lead to deprecation of the .Popin class.
- Fire the contentLoad event off the popin instead of its content. The reasoning is that the popin itself is always a direct container.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -668,7 +668,7 @@ jQuery(document).ready(function($) {
             });
         });
     };
-    $('.Popin,.js-popin').not('.Message .Popin,.Message .js-popin').popin();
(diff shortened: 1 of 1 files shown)


js/global.js
```diff
@@ -657,7 +657,7 @@ jQuery(document).ready(function($) {
                 url: gdn.url(url),
                 data: {DeliveryType: 'VIEW'},
                 success: function(data) {
-                    $elem.htmlTrigger(data);
+                    $elem.html($.parseHTML(data + '')).trigger('contentLoad');
```
DaazKu: Just a little note that `$.parseHTML` is not required here since `html()` will parse the html.

PR #3955, opened 2016-05-13
Make activities prune after 2 months

tburry:
This helps prevent the activity table from growing out of control. Requires #3954.

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -1699,13 +1699,13 @@ class ActivityModel extends Gdn_Model {
     /**
      * Get the exact timestamp to prune.
      *
-     * @return \DateTimeInterface|null Returns the date that we should prune after.
+     * @return \DateTime|null Returns the date that we should prune after.
      */
     private function getPruneDate() {
         if (!$this->pruneAfter) {
             return null;
         } else {
-            return new \DateTimeImmutable($this->pruneAfter, new DateTimeZone('UTC'));
+            return new \DateTime($this->pruneAfter, new DateTimeZone('UTC'));
         }
     }
 
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -1710,7 +1710,7 @@ class ActivityModel extends Gdn_Model {
     }
 
     /**
-     * Set the deleteAfter.
+     * Set the prune after date.
      *
      * @param string $pruneAfter A string compatible with {@link strtotime()}. Be sure to specify a negative string.
      * @return ActivityModel Returns `$this` for fluent calls.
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.activitymodel.php
```diff
@@ -1678,4 +1686,62 @@ protected function _touch(&$Data) {
...
+        } else {
+            return new \DateTimeImmutable($this->pruneAfter, new DateTimeZone('UTC'));
+        }
+    }
+
+    /**
+     * Set the deleteAfter.
+     *
+     * @param string $pruneAfter A string compatible with {@link strtotime()}. Be sure to specify a negative string.
+     * @return DbLogger Returns `$this` for fluent calls.
```
evach-vf: Should be `@return ActivityModel` instead of `DbLogger`

evach-vf:
I'm curious to know how to change the `$pruneAfter` setting for a specific site.

Would it be from an event handler defined somewhere where we would call the `setPruneAfter` method, or is there a way to inject a property to a class from a config file that I'm not aware of?

applications/dashboard/models/class.activitymodel.php
```diff
@@ -1678,4 +1686,62 @@ protected function _touch(&$Data) {
...
+    private function getPruneDate() {
+        if (!$this->pruneAfter) {
+            return null;
+        } else {
+            return new \DateTime($this->pruneAfter, new DateTimeZone('UTC'));
+        }
+    }
+
+    /**
+     * Set the deleteAfter.
```
kaecyra: Should be pruneAfter

tburry:
@evach-vf here is my thought process when adding stuff like this.

1. Add a property for the feature. Let's see how much it needs modification.
2. Maybe the odd site wants to modify it and they have a custom plugin that can do this. In this case we'd put an "init" event in the constructor.
3. Okay now a significant number of requests are coming in to change the value. Should we increase the prune after or should we add a config setting?
4. Let's say we added a config setting. Now we decide whether or not this needs a UI.
5. Should we have had one setting for just the activity model or should we have some global prune setting?

Adding config settings comes with a cost in the form of complexity and tech debt. I think it's very lazy UX to give endless config settings and I try and fight it. For this particular table I want to try hard to find the right prune time. If we can't prune this table then we are going to have a hard time pruning *any* table in our framework.

By not providing an option right away it forces us to deal with use-cases up the chain. On the other hand, if we add config tweaks right away then we can never remove them.

evach-vf:
Much clearer @tburry, thanks for your thought process on this, makes sense!

bleistivt:
My open source use-case would be that I already do this with a cronjob, that also keeps wall posts since my users hold those dear.

But this can already be disabled using the BeforeSave event.

PR #3956, opened 2016-05-13
Unset TransientKey on signout

### LINKED ISSUE #3780: Expire transientkey at logout
It's currently difficult for us to do things like expiring a transient key at a higher interval than cookies. However, when the user takes the explicit log out action, that's a case where we could tighten our security by invalidating it immediately.

initvector:
Vanilla currently recycles CSRF tokens for users.  This update aims to increase the security of Vanilla's CSRF token/transient key by removing it from the user's record when their session is explicitly ended.  The next time the user signs in, a new token will be generated.

Closes #3780

### DIFF
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -178,6 +178,7 @@ class Gdn_Session {
         }
 
         if ($this->UserID) {
+            Gdn::userModel()->saveAttribute($this->UserID, 'TransientKey', null);
             Logger::event('session_end', Logger::INFO, 'Session ended for {username}.');
         }
 


tburry:
Here is the issue I just remembered. If I am in two browsers and log out of one of them then the transient key gets removed from one, but not the other. I tested this in a comment form and the comment form itself does not properly refresh the transient key either. I have to refresh the entire page.

I'm wondering if we should schedule a more thorough transient key discussion. I was thinking that we could move it to be part of the cookie and that way each browser gets its own transient key. I had read that there is an issue with cookie transient keys though.

I've opened up issue #3965 to plan these things out. Right now this PR should probably be closed without being merged though.

PR #3957, opened 2016-05-13
Always redirect to the first page after saving a discussion

bleistivt:
After editing a discussion you get redirected to the last page (auto offset), which is confusing.


linc:
Nice patch, thanks!

PR #3958, opened 2016-05-13
Fix $Database definition in conversations/structure

R-J:
The variable definition was missing.

### DIFF
--- a/applications/conversations/settings/structure.php
+++ b/applications/conversations/settings/structure.php
@@ -16,6 +16,7 @@ if (!isset($Explicit)) {
     $Explicit = false;
 }
 
+$Database = Gdn::database();
 $SQL = $Database->sql();
 $Construct = $Database->structure();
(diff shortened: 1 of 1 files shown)


tburry:
These are all variables that are defined before inclusion.

R-J:
I haven't noticed that.

I was debugging a problem with utility/structure and saw that I'm able to run it for specific applications only: `public function structure($appName = 'all')`. So I've tried /utility/structure/vanilla,  /utility/structure/conversations and both failed.

So either the method structure() shouldn't offer the use of an argument or using the arguments shouldn't fail.

hgtonight:
Seems like the specific `$appName` functionality should be scrapped or expanded to include plugins. All plugins' structures are executed every structure call, regardless of the "specific" app.

tburry:
We will be scrapping the $appName functionality. I think that's already in PR.

PR #3968, opened 2016-05-13
Update Gdn_Format::text to deal with the love bug

### LINKED ISSUE #2320: Text-only formatting truncates at <
`<` is being interpreted as an HTML opening tag. Makes it hard to `<3` someone.

### LINKED ISSUE #3787: Love bug in emails
Emails are parsing any `<` as HTML and truncates messages including this character. This especially affects any message where a user inserts `<3`. May be a duplicate of https://github.com/vanilla/vanilla/issues/2320

initvector:
`strip_tags` hates what it perceives to be invalid markup.  When you use a "less-than" sign, followed by a non-whitespace character, it's taken to mean an opening tag.  If you have an opening tag that is never fully completed/closed, you have invalid markup.  When users use "<3" in their messages, `strip_tags` immediately truncates the message just before the "heart".

This is an issue for plain-text e-mails. The message text is cutoff, because it is passed through `EmailTemplate::plainTextEmail`, which then passes the content through to `Gdn_Format::text` where it has a run in with `strip_tags`.  This is also an issue for anywhere in Vanilla which uses `Gdn_Format::text` to display user input.

This update implements a simple routine to convert the opening character of what it detects to be an invalid tag (any < not immediately followed by a letter).  The following call to `htmlspecialchars` has been updated to avoid double encoding of entities.

Closes #3787
Closes #2320

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1847,20 +1847,32 @@ EOT;
     /**
      * Takes a mixed variable, formats it for display on the screen as plain text.
      *
-     * @param mixed $Mixed An object, array, or string to be formatted.
-     * @return mixed
+     * @param mixed $mixed An object, array, or string to be formatted.
+     * @return string
      */
-    public static function text($Mixed, $AddBreaks = true) {
-        if (!is_string($Mixed)) {
-            return self::to($Mixed, 'Text');
-        } else {
-            $Result = htmlspecialchars(strip_tags(preg_replace('`<br\s?/?>`', "\n",
-                html_entity_decode($Mixed, ENT_QUOTES, 'UTF-8'))), ENT_NOQUOTES, 'UTF-8');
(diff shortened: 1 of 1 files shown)


library/core/class.format.php
```diff
@@ -1847,20 +1847,32 @@ public static function serialize($Mixed) {
...
+        $result = html_entity_decode($mixed, ENT_QUOTES, 'UTF-8');
+        $result = preg_replace('`<br\s?/?>`', "\n", $result);
+        /**
+         * We need special handling for invalid markup here, because if we don't compensate
+         * things like <3, they'll lead to invalid markup and strip_tags will truncate the
+         * text.
+         */
+        $result = preg_replace_callback('/<(?![a-z])/i', function($invalidOpening) {
+            return '&lt;';
+        }, $result);
```
DaazKu: `$result = preg_replace('/<((?![a-z]))/i', '&lt;$1', $result);` shorter.

PR #3975, opened 2016-05-14
Category Scaling Part 2

tburry:
Optimize the various all category type views to use the CategoryCollection instead of iterating all categories. This is the implementation of #3845.

- [x] Category tree functionality.
- [x] Add CountAll* calculation.
- [x] Add recent post joining.
- [x] Implement proper permission checking in the category tree.
- [x] Refactor some of the existing category tree functions to use the CategoryCollection.
- [x] Make the default theme use the proper category collection.
- [x] Make sure the subtree uses the proper category collection.
- [x] Support marking a category as read.
- [x] Support category hiding/unhiding.

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -61,8 +61,11 @@ class CategoryModel extends Gdn_Model {
         $this->collection = new CategoryCollection();
         // Inject the calculator dependency.
         $this->collection->setConfig(Gdn::config());
-        $this->collection->setCalculator(function (&$category) {
+        $this->collection->setStaticCalculator(function (&$category) {
             self::calculate($category);
+        });
+
+        $this->collection->setUserCalculator(function (&$category) {
             self::calculateUser($category);
         });
     }
@@ -320,7 +323,7 @@ class CategoryModel extends Gdn_Model {
             $category['PhotoUrl'] = '';
         }
 
-        if ($category['DisplayAs'] == 'Default') {
+        if ($category['DisplayAs'] === 'Default') {
             if ($category['Depth'] <= c('Vanilla.Categories.NavDepth', 0)) {
                 $category['DisplayAs'] = 'Categories';
             } elseif ($category['Depth'] == (c('Vanilla.Categories.NavDepth', 0) + 1) && c('Vanilla.Categories.DoHeadings')) {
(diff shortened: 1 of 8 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -153,25 +166,93 @@ private static function loadAllCategories() {
...
+     * Calculate the user-specific information on a category.
+     *
+     * @param array &$category The category to calculate.
      */
-    private static function calculateUser(&$Category) {
-        $Category['Url'] = url($Category['Url'], '//');
-        if ($Photo = val('Photo', $Category)) {
-            $Category['PhotoUrl'] = Gdn_Upload::url($Photo);
+    private function calculateUser(&$category, $addUserCategory = null) {
+        $category['Url'] = url($category['Url'], '//');
```
initvector: Should this be...

```php
url($category['UrlCode'], '//');
```

applications/vanilla/models/class.categorymodel.php
```diff
@@ -1206,15 +1417,15 @@ public static function getAncestors($categoryID, $checkPermissions = true, $incl
             }
             $Max--;
 
+            if ($category['CategoryID'] == -1) {
+                break;
+            }
+
             if ($checkPermissions && !$category['PermsDiscussionsView']) {
```
initvector: It doesn't look like `$category['PermsDiscussionsView']` is always set by the time this check is made.  Currently triggers a notice:

> Undefined index: PermsDiscussionsView

initvector:
The category portion of the breadcrumb seems to have been lost.  For example, when viewing a discussion I currently see:

> Home > SubSubSubCat

I expect to see

> Home > Cat > SubCat > SubSubCat > SubSubSubCat

The last "crumb" also does not link to the category it should be linked to.  Instead, the anchor's `href` is "#".

initvector:
Visiting /categories should display all categories my user has access to, but no categories are displayed.

initvector:
I've intermittently encountered `CategoryModel::_MakeTreeChildren` triggering an error because `$Category['ChildIDs']` was not set. The error is not consistent and was observed when editing a subcommunity.

applications/vanilla/models/class.categorymodel.php
tburry: This is actually not a change in functionality. I think the URL is calculated as a path and then this switches it to a proper URL.

initvector: In its current state, this line triggers a notice when viewing a discussion:

> Undefined index: Url

`Url` isn't available in `$category` at this point, but `UrlCode` is.  This notice isn't seen on current master.  It looks like this particular element may've been previously inserted with `CategoryModel::calculate`.

tburry:
> Visiting /categories should display all categories my user has access to, but no categories are displayed.

It looks like this was caused by setting the max display depth to never (0). I've pushed an updated.

tburry:
> I've intermittently encountered CategoryModel::_MakeTreeChildren triggering an error because $Category['ChildIDs'] was not set. The error is not consistent and was observed when editing a subcommunity.

I've addressed this in a references PR.

tburry:
> The category portion of the breadcrumb seems to have been lost.

This should be fixed now too.

initvector:
Seems solid now, but we want to wait to merge until master is a little more stable.

PR #3978, opened 2016-05-16
Update SplitMerge to overcome Garden.ForceInputFormatter

initvector:
If `Garden.ForceInputFormatter` is enabled, the SplitMerge plug-in may be unable to save the discussions it creates with the proper "Html" format.  Since the content SplitMerge generates for these posts contains HTML, this can lead to posts being rendering escaped markup text.  E.g.

```
This discussion was created from comments split from: <a href="/discussion/1/example-discussion">Example Discussion</a>.
```

This update applies a [method taken from ModerationController](https://github.com/vanilla/vanilla/blob/320e922082771a160ddba16f63435cf3cd35ad8c/applications/vanilla/controllers/class.moderationcontroller.php#L417) to temporarily set `Garden.InputFormat` to "Html" if `Garden.ForceInputFormatter` is enabled.

### DIFF
--- a/plugins/SplitMerge/class.splitmerge.plugin.php
+++ b/plugins/SplitMerge/class.splitmerge.plugin.php
@@ -92,11 +92,19 @@ class SplitMergePlugin extends Gdn_Plugin {
             $Data['Format'] = 'Html';
             $Data['Type'] = 'Discussion';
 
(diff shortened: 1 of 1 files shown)


linc:
Title: Fix where SplitMerge addon may be unable to save split discussion with "Html" format.

PR #3987, opened 2016-05-17
Improve overall usage of Alt style calculation

R-J:
1. The alternating class could be removed on several places because the output was already done by the function cssClass().
2. It wasn't always implemented right so that either variables were not used after declaration, the variable was not toggled before it was used or - even worse - the variable has been toggled but not used.
3. And in most cases, an ugly conditional operator could be replaced with a simple not().

Most changes have been in the dashboard and I'd expect that to get a major change :+1: 

My general thought on this is that it is only cluttering the html. CSS is able to style that now for several years without adding classes and it is [widely supported](http://caniuse.com/#feat=css-sel3).
I would drop it with the next bigger change of the standard frontend.


DaazKu:
You may want to update your git configuration to use your github account.

https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_git_config

PR #3990, opened 2016-05-17
Improve phpunit docs

linc:
* Revise & add to phpunit README.
* Add `/cgi-bin/` to gitignore.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,13 +1,33 @@
-# Running Vanilla's Tests
+# Vanilla's Tests
 
 Vanilla's tests are designed to work on [Travis CI](https://travis-ci.org/), but you can set your local environment up
-to run the tests with minimal setup. Here is what you need.
+to run the tests with minimal setup. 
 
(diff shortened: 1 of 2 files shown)


tburry:
You are give requirements about having the web server and MySQL access and then your instructions run the tests that don't require that. I recommend you read about the --testsuite parameter and then look at the test suites Vanilla uses.

tests/README.md
```diff
@@ -1,13 +1,33 @@
...
+
+## Running
+
+1. Go to your Vanilla install directory.
+2. `phpunit -c phpunit.xml.dist --testsuite Library`
+
+To run without the `--testsuite` flag requires that `/cgi-bin` be mapped, as in the template mentioned above. In nginx, try:
+
+```
+    location ~* "^/cgi-bin/.+\.php(/|$)" {
```
tburry: Don't indent. This is also something that is in the default-site.tpl.conf. I really don't want people that can't understand the nginx template file to think they can just copy and past snippets like this. What if their file structure is different for example?

PR #3992, opened 2016-05-17
Tighten security of target parameter

### LINKED ISSUE #2800: Open redirect in /entry endpoints & the TrustedDomains setting

initvector:
This update aims to tighten the security of the `target` parameter passed to `EntryController` endpoints in the following ways:

1. Remove the existing default behavior in `EntryController::target` to allow anything to pass through if `Garden.TrustedDomains` was not configured.
2. Update `EntryController::_setRedirect` to use `safeRedirect` instead of `redirect`.
3. Add `HomeController::leaving` to act as a buffer for users when redirected to an untrusted domain.  The page states the user is leaving the site and displays the URL they will be sent to.  After a three-second countdown, the user is automatically forwarded to the new URL.
4. Update `trustedDomains` to automatically includes any authentication provider domains.  The current request/site host is also automatically included.
5. Update `safeRedirect` to send users to the new `HomeController::leaving` page if the destination is not in the list of trusted domains.
6. Log any calls to `safeRedirect` wherein the destination does not match a trusted domain.
7. The configuration of trusted domains has been moved from embed settings, under the advanced tab, over to the more visible advanced forum settings.

Closes #2800.


applications/vanilla/controllers/class.settingscontroller.php
```diff
@@ -75,6 +84,13 @@ public function advanced() {
             $ArchiveDateBak = Gdn::config('Vanilla.Archive.Date');
             $ArchiveExcludeBak = (bool)Gdn::config('Vanilla.Archive.Exclude');
 
+            // Format the trusted domains as an array based on newlines & spaces
+            $TrustedDomains = $this->Form->getValue('Garden.TrustedDomains');
+            $TrustedDomains = explode("\n", $TrustedDomains);
+            $TrustedDomains = array_unique(array_filter(array_map('trim', $TrustedDomains)));
```
tburry: Can you use explodeTrim() here?

PR #3994, opened 2016-05-18
Sanitize table name in _create() method

R-J:
I don't know if sanitation has to be done at all, but it was only used for UNIQUE and FULLTEXT keys, not for normal INDEXes, which surely wasn't meant that way.


initvector:
Looks solid to me.  Thanks!

PR #3996, opened 2016-05-18
Correct field length of GDN_Tag.FullName

R-J:
**Please be careful with this!**

There has been an error that made the field FullName hold 255 characters right after installation until the first run of /utility/structure. Then it is cut down to 100 characters.
I don't know if the model makes any restrictions, but if not, the FullName could be user for strings longer 100 characters and thus might get cut when utility/structure is run for the first time.

It might be safer to set all values to 255 based on that error.

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -637,7 +637,7 @@ if (!$FullNameColumnExists) {
         ->put();
 
     $Construct->table('Tag')
-        ->column('FullName', 'varchar(255)', false, 'index')
+        ->column('FullName', 'varchar(100)', false, 'index')
         ->set();
 }
 


linc:
Where does it get cut to 100?

100 is the correct length because of utf8mb4 encoding restrictions.

R-J:
See the code in dashboards structure:

~~~
$Construct->table('Tag');
$FullNameColumnExists = $Construct->columnExists('FullName');
...
$Construct->table('Tag')
    ->primaryKey('TagID')
    ->column('Name', 'varchar(100)', false, 'unique')
    ->column('FullName', 'varchar(100)', !$FullNameColumnExists, 'index')
...
    ->set($Explicit, $Drop);

if (!$FullNameColumnExists) {
...
    $Construct->table('Tag')
        ->column('FullName', 'varchar(255)', false, 'index')
        ->set();
}
~~~
On first run (installation)
1. $FullNameColumnExists is false.
2. Then column FullName is created with varchar(100).
3. The if condition is true (!false) so Fullname is altered to varchar(255)

On update
1. $FullNameColumnExists is true.
2. Then column FullName definition of varchar(100) is checked and since it has been set to 255 on installation, it must be altered.
3. The if condition is not executed

So after a clean installation, FullName holds 255 characters, while /utility/structure call it is reduced to 100.

PR #4002, opened 2016-05-19
Correct minor bug in notifications preferences.

R-J:
The notification preferences view has a minor bug which you can only see if you create an additional notification type beside Email and Popup via `profileController_afterPreferencesDefined_handler`. When a checkbox should be ommitted, three additional(!) checkboxes will be printed instead. Use this for a quick test:

~~~
    public function profileController_afterPreferencesDefined_handler($sender) {
        foreach ($sender->Preferences as $preferenceGroup => $preferences) {
            foreach ($preferences as $name => $description) {
                $nameParts = explode('.', $name);
                $sender->Preferences[$preferenceGroup]['Whatever.'.$nameParts[1]] = $description;
            }
        }
    }
~~~

Taken a look at the view made me shiver. That's the first time I've seen ugly spaghetti code in Vanilla and that's why I did not only solved the problem, but have rewritten that code.

I reduced 4 foreach loops to only 2, 6 if clauses to 3 and reduced the max. nesting level from 7(!) to 3. Additionally I've packed it with comments so I hope this still complicated routine is more readable now...

### DIFF
--- a/applications/dashboard/views/profile/preferences.php
+++ b/applications/dashboard/views/profile/preferences.php
@@ -48,58 +48,66 @@
                     <?php
                     echo wrap(t('Notification'), 'td', array('style' => 'text-align: left'));
 
-                    $CountTypes = 0;
-                    foreach ($this->data("PreferenceTypes.{$PreferenceGroup}") as $PreferenceType) {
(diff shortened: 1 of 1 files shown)


initvector:
Bang-up job!  This is much easier on the eyes now.  Thanks!

PR #4017, opened 2016-05-23
Implement custom Smarty_Security class

### LINKED ISSUE #4014: Smarty_Security::isTrustedPHPFunction case-sensitive
`Smarty_Security::isTrustedPHPFunction` uses `in_array` with the values we set for `php_functions` to determine if a function should be allowed or not.  `in_array` is case-sensitive, which makes `Smarty_Security::isTrustedPHPFunction` case-sensitive.  Existing templates often use upper-case-word naming schemes (e.g. InSection) for functions, so they don't mesh with our [new Smarty security policy](https://github.com/vanilla/vanilla/blob/72c743dfdd4e8a3d43aba3acf766366b72249bf1/library/core/class.smarty.php#L117) (e.g. inSection).

1. We can change the casing of the functions to their original casing.
2. We can append the old casings to the new list.
3. We can extend the `Smarty_Security` class, override `isTrustedPHPFunction` to be case-insensitive and use our custom security class for the new security policy.

Any other ideas/suggestions are welcome.

initvector:
`Smarty_Security::isTrustedPhpFunction` is case-sensitive.  We need it to be case-insensitive, especially since we changed our code style guidelines in the last year.

This update extends `Smarty_Security` and overrides `isTrustedPhpFunction` to force casing of function names to lowercase for testing.  This update also adds a new function to the Smarty security class, `setPhpFunctions`, which transforms all PHP function names to lowercase before setting them on the object.

Closes #4014

### DIFF
--- a/library/core/class.smarty.php
+++ b/library/core/class.smarty.php
@@ -104,7 +104,7 @@ class Gdn_Smarty {
         // Assign the controller data last so the controllers override any default data.
         $Smarty->assign($Controller->Data);
 
-        $security = new Gdn_SmartySecurity($Smarty);
+        $security = new SmartySecurityVanilla($Smarty);
 
         $security->php_handling = Smarty::PHP_REMOVE;
         $security->allow_constants = false;
--- a/library/core/class.smarty.php
+++ b/library/core/class.smarty.php
@@ -104,7 +104,7 @@ class Gdn_Smarty {
         // Assign the controller data last so the controllers override any default data.
         $Smarty->assign($Controller->Data);
 
-        $security = new Smarty_Security($Smarty);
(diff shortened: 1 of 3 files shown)


linc:
My thinking on this:

* Subclassing dependencies isn't ideal, but this has the potential to be a big breaking change for every custom theme out there in our ecosystem otherwise.
* Our interaction with this class is limited to our own Smarty class, so the scope of this is pretty small (less fragile for the future).
* This change is going to get submitted upstream, but we're unclear whether it will be accepted due to project dogma around case sensitivity. It's currently a roadblock to our next deploy, so it needs to be resolved immediate-term.

Issue #3942, opened 2016-05-12
The Flagging plugin has a bug when using MySQL 5.7


raykai:
if you tern the plugin on in v2.2.1 runing on MySQL 5.7 it crashes the forum dashboard.

missing the Flagging addon which has a bug in vanilla v2.2.1

![plugins - vanilla](https://cloud.githubusercontent.com/assets/8138894/15202430/28f964f2-17af-11e6-8a65-c38b085aa015.jpg)

error:

Fatal Error in PHP.trigger_error();

Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'aquaexotik0forum_vanilla.fl.ForeignID' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

The error occurred on or near: 
/library/database/class.database.php

`408:                 // If we get here then the pdo statement prepared properly.
409:                 break;
410: 
411:             } catch (Gdn_UserException $uex) {
412:                 trigger_error($uex->getMessage(), E_USER_ERROR);
413:             } catch (Exception $ex) {
414:                 list($state, $code, $message) = $PDO->errorInfo();
415: 
416:                 // If the error code is consistent with a disconnect, attempt to retry`

Backtrace:

`/library/database/class.database.phpPHP::Gdn_ErrorHandler();
[/library/database/class.database.php:412] PHP::trigger_error();
[/library/database/class.sqldriver.php:1663] Gdn_Database->query();
[/library/database/class.sqldriver.php:614] Gdn_SQLDriver->query();
[/plugins/Flagging/class.flagging.plugin.php:34] Gdn_SQLDriver->get();
[/library/core/class.pluginmanager.php:867] FlaggingPlugin->base_getAppSettingsMenuItems_handler();
[/library/core/class.pluginmanager.php:780] Gdn_PluginManager->callEventHandler();
[/library/core/class.pluggable.php:133] Gdn_PluginManager->callEventHandlers();
[/applications/dashboard/modules/class.sidemenumodule.php:268] Gdn_Pluggable->fireEvent();
[/library/core/class.module.php:213] SideMenuModule->toString();
[/library/core/class.controller.php:1350] Gdn_Module->render();
[/applications/dashboard/views/admin.master.php:34] Gdn_Controller->renderAsset();
[/library/core/class.controller.php:1891] PHP::include();
[/library/core/class.controller.php:1318] Gdn_Controller->renderMaster();
[...]

PR #4020, opened 2016-05-24
Fix the flagging plugin "number of tracked item" query

DaazKu:
- The query returns 0 when `sql-mode` `ONLY_FULL_GROUP_BY` is off.
- The plugin crash when we turn it on when `sql-mode` `ONLY_FULL_GROUP_BY` is turned on.

Fix https://github.com/vanilla/vanilla/issues/3942

This should be fixed in release/2.3 too, when approved.

### DIFF
--- a/plugins/Flagging/class.flagging.plugin.php
+++ b/plugins/Flagging/class.flagging.plugin.php
@@ -28,13 +28,13 @@ class FlaggingPlugin extends Gdn_Plugin {
      * Add Flagging to Dashboard menu.
      */
     public function base_getAppSettingsMenuItems_handler($Sender) {
-        $flaggedItemsResult = Gdn::sql()->select('count(distinct(ForeignURL)) as NumFlagedItems')
+        $flaggedItemsResult = Gdn::sql()->select('count(distinct(ForeignURL)) as NumFlaggedItems')
             ->from('Flag fl')
             ->get()->firstRow(DATASET_TYPE_ARRAY);
 
         $LinkText = t('Flagged Content');
-        if ($flaggedItemsResult['NumFlagedItems']) {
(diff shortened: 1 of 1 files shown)


plugins/Flagging/class.flagging.plugin.php
```diff
@@ -28,10 +28,9 @@ class FlaggingPlugin extends Gdn_Plugin {
      * Add Flagging to Dashboard menu.
      */
     public function base_getAppSettingsMenuItems_handler($Sender) {
-        $NumFlaggedItems = Gdn::sql()->select('fl.ForeignID', 'DISTINCT', 'NumFlaggedItems')
+        $NumFlaggedItems = Gdn::sql()->select('COUNT(DISTINCT(ForeignURL)) as numFlagItems')
             ->from('Flag fl')
-            ->groupBy('ForeignURL')
-            ->get()->numRows();
+            ->get()->firstRow(DATASET_TYPE_ARRAY)['numFlagItems'];
```
linc: Let's not be accessing array elements in the same statement as we're building an SQL query.

plugins/Flagging/class.flagging.plugin.php
```diff
@@ -28,10 +28,9 @@ class FlaggingPlugin extends Gdn_Plugin {
      * Add Flagging to Dashboard menu.
      */
     public function base_getAppSettingsMenuItems_handler($Sender) {
-        $NumFlaggedItems = Gdn::sql()->select('fl.ForeignID', 'DISTINCT', 'NumFlaggedItems')
+        $NumFlaggedItems = Gdn::sql()->select('COUNT(DISTINCT(ForeignURL)) as numFlagItems')
```
linc: We use lowercase MySQL keywords.

Issue #3809, opened 2016-04-28
Pass all IP addresses through an encode function before going to the database.


tburry:
When saving IP addresses to the database you'll now need to call inet_pton() to pack the human readable form. I recommend making ipEncode() and ipDecode() functions to wrap this functionality and have them check for a human-readable IP address vs one that is already in binary. You can use filter_var($IP, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4|FILTER_FLAG_IPV6) to see if a string is a valid human-readable IP address.

PR #4026, opened 2016-05-24
Migrate AllIPAddresses field to new UserIP table

### LINKED ISSUE #3808: Update formatIP to handle packed IPv6 addresses
IP addresses will either be a string packed with inet_pton or an IPv4 address in plain text or some sort of error. You should do the following:

1. Use filter_var to see if you have an IPv4 address.
2. Check for various things like an empty string too.
3. Finally try and unpack the address, but use an exception handler to gracefully degrade.

Plaintext IP addresses should work as is, but HTML addresses should put the IP in the title of a span and make the text of IPv6 addresses "IPv6". Add appropriate classes to the spans.

Depends on #3806.

### LINKED ISSUE #3906: Create the GDN_UserIP table and migrate the AllIPAddress tracking
The `GDN_UserIP` table will have the following fields:

* UserID (PK)
* IPAddress (PK)
* DateInserted
* DateUpdated

Whenever the user's IP address changes (`LastIPAddress != Gdn::request()->IpAddress()`) you need to track the change in this table. You can use a custom insert on duplicate key update DateUpdated query. In other tables the DateUpdated is null on insert, but perhaps making DateUpdated non nullable and initially set to the DateInserted would help certain queries.

initvector:
Move saving IP addresses from `AllIPAddresses` field in the `User` table to a separate, new table: `UserIP`.

Closes #3906
Closes #3808
Addresses half of #3809

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2871,6 +2871,31 @@ class UserModel extends Gdn_Model {
     }
 
     /**
+     * Record an IP address for a user.
+     *
+     * @param int $userID Unique ID of the user.
+     * @param string $IP Human-readable IP address.
+     */
+    public function saveIP($userID, $IP) {
+        $packedIP = ipEncode($IP);
+        $px = Gdn::database()->DatabasePrefix;
+        $currentDateTime = Gdn_Format::toDateTime();
+
+        $query = "insert into {$px}UserIP (UserID, IPAddress, DateInserted, DateUpdated)
+            values (:UserID, :IPAddress, :DateInserted, :DateUpdated)
+            on duplicate key update DateUpdated = :DateUpdated2";
+        $values = [
+            ':UserID' => $userID,
+            ':IPAddress' => $packedIP,
+            ':DateInserted' => $currentDateTime,
+            ':DateUpdated' => $currentDateTime,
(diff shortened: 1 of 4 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -2871,6 +2885,31 @@ public function addInsertFields(&$Fields) {
...
+            on duplicate key update DateUpdated = :DateUpdated2";
+        $values = [
+            ':UserID' => $userID,
+            ':IPAddress' => $packedIP,
+            ':DateInserted' => $currentDateTime,
+            ':DateUpdated' => $currentDateTime,
+            ':DateUpdated2' => $currentDateTime
+        ];
+
+        Gdn::database()->query($query, $values);
```
tburry: Let's try/catch this one for now to prevent sites that haven't done a db update from failing. You can return true/false on success/failure maybe.

tburry: In fact, I can't even run utility/update if this isn't caught.

tburry:
The initial fixes on this PR need to be cleared before any more review can be done.

initvector:
`UserModel::getIPs` has been updated with a try/catch block.  The default return value for that function is an empty array, which is returned on error.

`UserModel::saveIP` has also been updated with a try/catch block and has been given a return value to indicate the success of the operation.

Issue #3872, opened 2016-05-04
Gdn_DatabaseStructure::query method should not be exposed as public.


DaazKu:
https://github.com/vanilla/vanilla/pull/3605 adds a nice feature that allows to check a threshold before applying an alter table statement.

The problem is that you have to call `query($sql, true)` for the feature to be enabled.

This is impossible when doing to following:
```
Gdn::structure()->table('TABLE NAME')
    ...// Add new columns or whatever
    ->setAlterTableThreshold(100)
    ->set(false, false)
```

Callstack:
```
query($sql, $checkThreshold = false) // this is called like so: query($sql)
_modify($Explicit = false)
set($Explicit = false, $Drop = false)
```

To fix the problem we have 2 choices 
- we propagate the `$checkThreshold` variable through the stack (and everywhere else where is it needed)
- we assume that if `(bool)$this->alterTableThreshold` is true we apply the check.

tburry:
There are some alter commands that check the threshold and some that do not. The idea is that some alters will not put too much load on the database on recent versions of MySQL so they are safe to execute.

DaazKu:
It would be nice to be able to restrict an alterTable on row counts alone.
Let's say I'm creating an index on a table that has a million+ rows, a plugin could use the feature.

tburry:
The reason why I don't do the threshold on indexes is because of [fast index creation](https://dev.mysql.com/doc/refman/5.5/en/innodb-create-index-overview.html). We've not seen indexes be a major bottleneck. This is something that probably can be benchmarked.

DaazKu:
At first I got confused because `query` is public and you can use the second parameter to force an alterTable check but after a more thorough analysis of the code I realized that query should not, in theory, be used outside of its class. (There are only 8 external usages of it through vanilla)

I think that the documentation should be updated to say that the feature check only some alter table changes.
We may want to change `query` from public to protected if we feel nitpicky.

tburry:
There are a lot of legacy methods in these core classes that certainly should be protected. I notice this when I revisit older code. Great point.

DaazKu:
Closed by #4038

PR #4038, opened 2016-05-25
Deprecate Gdn_DatabaseStructure::query() for Gdn_SqlDriver.

DaazKu:
Gdn_DatabaseStructure::query() is to be used internally.

- Rename query() to executeQuery and set the method as protected.
- Forward query() to $this->database::query() and deprecate it.

### DIFF
--- a/library/database/class.databasestructure.php
+++ b/library/database/class.databasestructure.php
@@ -377,9 +377,26 @@ abstract class Gdn_DatabaseStructure extends Gdn_Pluggable {
      * @param string $sql The sql to execute.
      * @return Gdn_Dataset
      */
-    public function query($sql) {
-        deprecated(__CLASS__.'::query()', 'Gdn_SQLDriver::query()');
-        return $this->Database->query($sql);
+    public function query($sql, $checkTreshold = false) {
+
+        $class = null;
+        $internalCall = false;
+
+        $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        if (count($backtrace) > 1) {
+            $class = val('class', $backtrace[1]);
+            if ($class) {
+                $internalCall = is_a($class, __CLASS__, true);
+            }
+        }
+
+        if ($internalCall) {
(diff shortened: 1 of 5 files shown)


DaazKu:
Fix https://github.com/vanilla/vanilla/issues/3872

library/database/class.databasestructure.php
```diff
@@ -373,11 +373,41 @@ public function primaryKey($Name, $Type = 'int') {
...
+     * @deprecated
+     * @param string $sql The sql to execute.
+     * @param bool $checkTreshold Should not be used
+     * @return Gdn_Dataset
+     */
+    public function query($sql, $checkTreshold = false) {
+
+        $class = null;
+        $internalCall = false;
+
```
linc: `// Detect the origin of this call.`

library/database/class.databasestructure.php
```diff
@@ -373,11 +373,41 @@ public function primaryKey($Name, $Type = 'int') {
...
+        $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        if (count($backtrace) > 1) {
+            $class = val('class', $backtrace[1]);
+            if ($class) {
+                $internalCall = is_a($class, __CLASS__, true);
+            }
+        }
+
```
linc: `// Give appropriate message based on whether we're using it internally.`

library/database/class.databasestructure.php
```diff
@@ -373,11 +373,41 @@ public function primaryKey($Name, $Type = 'int') {
     /**
      * Send a query to the database and return the result.
      *
+     * @deprecated
```
linc: `@deprecated since 2.3. Was incorrect public. Replaced by executeQuery().`

(To answer the obvious question: when referencing version numbers in code, use the next open source release number that your change may appear in)

linc:
* Great work hashing out the solution with Todd on a referenced issue, thank you.
* A few extra inline comments as noted will make this change a lot more transparent for other folks / our future selves.

PR #4042, opened 2016-05-26
Automatically update CSRF token with AJAX requests

### LINKED ISSUE #3962: Make forms handle CSRF tokens changing without requiring a full page refresh
Right now if the transient key changes while I'm looking at a discussion then I will not be able to post a comment until I refresh the entire page. I suspect this is an issue across the board and will require a generic solution to push the new transient key back into the definitions array.

initvector:
This update does a few things:

1. Registers a new function, `Gdn_Dispatcher::sendHeaders`, as a callback with `header_register_callback` in `Gdn_Dispatcher::start`.  This function only fires an event.
2. Adds a new hook, `gdn_dispatcher_sendHeaders_handler`, to the Dashboard application.  This hook checks to see if a transient key was supplied.  If it was and it is invalid, the valid transient key is sent back to the client in a X-CSRF-Token header.
3. Creates a new JavaScript hook: `ajaxComplete`.  This hook checks the headers of any completed AJAX requests fired by jQuery.  If a X-CSRF-Token header is present, its value is grabbed and used to update the CSRF token value in the `gdn.meta` collection.

Closes #3962

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -269,13 +269,11 @@ class DashboardHooks implements Gdn_IPlugin {
      * @param Gdn_Dispatcher $sender
      */
     public function gdn_dispatcher_sendHeaders_handler($sender) {
-        $headers = array_change_key_case(getallheaders(), CASE_UPPER);
-
         $csrfToken = Gdn::request()->post(
             Gdn_Session::CSRF_NAME,
             Gdn::request()->get(
                 Gdn_Session::CSRF_NAME,
-                val('X-CSRF-TOKEN', $headers)
+                Gdn::request()->getValueFrom(Gdn_Request::INPUT_SERVER, 'HTTP_X_CSRF_TOKEN')
             )
         );
 
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -266,6 +266,25 @@ class DashboardHooks implements Gdn_IPlugin {
     }
 
     /**
+     * @param Gdn_Dispatcher $sender
+     */
+    public function gdn_dispatcher_sendHeaders_handler($sender) {
+        $headers = array_change_key_case(getallheaders(), CASE_UPPER);
+
+        $csrfToken = Gdn::request()->post(
+            Gdn_Session::CSRF_NAME,
+            Gdn::request()->get(
+                Gdn_Session::CSRF_NAME,
+                val('X-CSRF-TOKEN', $headers)
+            )
+        );
+
+        if ($csrfToken && Gdn::session()->isValid() && !Gdn::session()->validateTransientKey($csrfToken)) {
+            safeHeader('X-CSRF-Token: '.Gdn::session()->transientKey());
+        }
+    }
+
+    /**
      * Method for plugins that want a friendly /sso method to hook into.
      *
      * @param RootController $Sender
--- a/js/global.js
+++ b/js/global.js
@@ -149,6 +149,12 @@
         }
     });
 
+    $(document).ajaxComplete(function(event, jqXHR, ajaxOptions) {
+        var csrfToken = jqXHR.getResponseHeader("X-CSRF-Token");
+        if (csrfToken) {
+            gdn.setMeta("TransientKey", csrfToken);
+        }
+    });
 })(window, jQuery);
 
 // Stuff to fire on document.ready().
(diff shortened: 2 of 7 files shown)


applications/dashboard/settings/class.hooks.php
```diff
@@ -266,6 +266,25 @@ public function gdn_dispatcher_appStartup_handler($Sender) {
     }
 
     /**
+     * @param Gdn_Dispatcher $sender
+     */
+    public function gdn_dispatcher_sendHeaders_handler($sender) {
+        $headers = array_change_key_case(getallheaders(), CASE_UPPER);
```
tburry: Try grabbing from Gdn::request() instead.

js/global.js
```diff
@@ -149,6 +149,12 @@
         }
     });
 
+    $(document).ajaxComplete(function(event, jqXHR, ajaxOptions) {
+        var csrfToken = jqXHR.getResponseHeader("X-CSRF-Token");
+        if (csrfToken) {
+            gdn.setMeta("TransientKey", csrfToken);
```
tburry: Also look through forms.

library/core/class.dispatcher.php
```diff
@@ -796,6 +797,13 @@ private function _mapParts($Parts, $ControllerKey) {
             }
         }
     }
+
+    /**
+     * Trigger an event allowing addons to modify response headers.
+     */
+    public function sendHeaders() {
```
tburry: Can this be private? We don't want to make functions public if there is no need.

initvector: Registering as a callback with `header_register_callback` appears to call the function externally.  Switching up its visibility from public causes errors.

js/global.js
initvector: Added with https://github.com/vanilla/vanilla/commit/7cdd4bea6bfb079d1bc520dc0d99ccc38b93407d

library/core/class.dispatcher.php
DaazKu: If the sole purpose of this function is to be added as a callback you should do 
```php
$that = $this;
header_register_callback(function() use ($that) {
    $that->fireEvent('SendHeaders');
});
```
(Not tested but should work.)

initvector: In PHP, `$this` is [automatically bound](http://php.net/manual/en/functions.anonymous.php#example-199) to the function when you create an anonymous function in the context of a class.  We don't need to go the route of setting `$that` to `$this`.

The anonymous function suggestion is solid and has been implemented.

DaazKu: Nice to know!
I probably took that from Javascript haha.

applications/dashboard/settings/class.hooks.php
```diff
@@ -266,6 +266,23 @@ public function gdn_dispatcher_appStartup_handler($Sender) {
...
+    public function gdn_dispatcher_sendHeaders_handler($sender) {
+        $csrfToken = Gdn::request()->post(
+            Gdn_Session::CSRF_NAME,
+            Gdn::request()->get(
+                Gdn_Session::CSRF_NAME,
+                Gdn::request()->getValueFrom(Gdn_Request::INPUT_SERVER, 'HTTP_X_CSRF_TOKEN')
+            )
+        );
+
+        if ($csrfToken && Gdn::session()->isValid() && !Gdn::session()->validateTransientKey($csrfToken)) {
```
tburry: My worry here is that someone posting a fake TK will get back the proper TK in the response.

library/core/class.dispatcher.php
adrianspeyer: FYI, this dispatcher change borked an older forum running on PHP 5.4. The signin/signout leads to a 500 error. I rewrote as:

` header_register_callback($this->fireEvent('SendHeaders'));`

And it worked! This is more just an FYI for those open sourcers trying to live on the edge. Anyway it's time to upgrade to a new version of PHP.  I am sure other stuff will start to fail in 5.4, so it's time to move on up :)

adrianspeyer:
FYI, [this dispatcher change](https://github.com/vanilla/vanilla/pull/4042/commits/e9814de5e58cc0e9b69201143e1f857cfe9fd0d6) borked an older forum running on PHP 5.4. The signin/signout leads to a 500 error. I rewrote as:

header_register_callback($this->fireEvent('SendHeaders'));

And it worked! This is more just an FYI for those open sourcers trying to live on the edge. Anyway it's time to upgrade to a new version of PHP. I am sure other stuff will start to fail in 5.4, so it's time to move on up :)

PR #4044, opened 2016-05-26
Fix /me comment editing

### LINKED ISSUE #1769: Editing a /me comment doesn't work as expected
First click on edit replaces the "MeFormatted" text with the original text
Second click on edit makes the original text editable
Third click on edit will revert edit textarea to original text (not MeFormatted)

Consecutive clicks switch between stage 2 and 3.

Only pressing save while editing shows the comment MeFormatted again.

That second stage where the /me comment is rendered as a normal posting should never be shown.

bleistivt:
On /me comments the message is invisible.
So instead of checking for message visibility, check if no `.EditCommentForm` exists inside the comment.

fixes #1769


linc:
Thanks!

PR #4045, opened 2016-05-26
Make individual comments shareable

### LINKED ISSUE #2034: When using social connect, and sharing link, the url is that of the discussion not of the comment.
Intended Behaviour: click on Twitter or Facebook link from a specific comment.  It should create link to that specific comment, instead, it posts a link to the root discussion. 

Could the # in URL  cause an issue? In short the anchor is missing.

bleistivt:
The share url for comments was the discussion url which would make share buttons below every comment useless.

fixes #2034

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -1536,7 +1536,7 @@ if (!function_exists('getRecord')) {
                     $Discussion = $Model->getID($Row['DiscussionID']);
                     if ($Discussion) {
                         $Discussion->Url = DiscussionUrl($Discussion);
-                        $Row['ShareUrl'] = $Discussion->Url;
+                        $Row['ShareUrl'] = $Row['Url'];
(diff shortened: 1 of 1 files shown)


library/core/functions.general.php
```diff
@@ -1536,7 +1536,7 @@ function getRecord($recordType, $id, $throw = false) {
                     $Discussion = $Model->getID($Row['DiscussionID']);
                     if ($Discussion) {
                         $Discussion->Url = DiscussionUrl($Discussion);
```
linc: Doesn't this make the line above it useless?

bleistivt: Probably. But 2 lines below the discussion is added to the record, so it may be used:
```php
$Row['Discussion'] = (array)$Discussion;
```

linc: Good point, thanks.

Issue #4019, opened 2016-05-24
Vanilla should not override php logging


kaecyra:
Currently, Vanilla has a `Garden.Errors.LogEnabled` setting which defaults to FALSE in config-defaults.php

This setting, for whatever reason in the past, interferes with PHP's error logging mechanisms and makes life difficult on the server admin side. It should be removed in favour of allowing PHP itself to log errors to file if it wants to. 

There is also a `Garden.Errors.LogFile` setting which seems to somehow allow modifying the target of PHP logs. This should be totally removed, again in favour of allowing PHP to decide where logs are sent.

tburry:
## First PR

The point of this PR is to pass Vanilla's errors through to the error log using a more native PHP methodology.

- Look at log_errors instead of Garden.Errors.LogEnabled (remove the config setting altogether).
- Make an error_log wrapper that looks at log_errors and will use the Garden.Errors.LogFile if set.
- Call the error_log wrapper wherever possible in functions.error
- Should we write a format message function that takes a message, optional filename and line number and formats it in the way PHP formats an error message?

## Second PR

The point of this PR is to respect the error_reporting INI setting unless we are debugging.

- Don't set the error_reporting level in the index.php.
- The default set_error_handler should not handle low level errors (E_NOTICE, E_USER_NOTICE, E_DEPRECATED, E_USER_DEPRECATED, E_STRICT, etc.)
- The call to debug() should elevate error_reporting to E_ALL & ~E_STRICT and call set_error_handler again with E_ALL & ~E_STRICT.

kaecyra:
Errors should DEFINITELY be logged in PHP error format, since this directly impacts the ability for us to aggregate errors.

PR #4047, opened 2016-05-27
Allow php logging when enabled in php.ini

evach-vf:
### Fixes issue https://github.com/vanilla/vanilla/issues/4019.

**First part** in allowing php logging to flow normally to the server error logs.

Before this change, in order to enable PHP logging to flow to the error log file, the config key `Garden.Errors.LogEnabled` needed to be set to true.  This makes it harder for a sysadmin to troubleshoot an installation since `php.ini` settings on the server such as `log_errors` is not respected and `error_reporting` is overridden in `index.php`. 

### Changes
- Config `Garden.Errors.LogEnabled` as been removed.  
Detection is now based on `ini_get('log_errors')` value.
- PHP errors and exceptions are formatted the same way as the native PHP error handler would.
- `Non-Fatal errors` are not converted to exception anymore, preventing the request to halt for `E_NOTICE` for example.

### TODO
- The `php.ini` server setting `error_reporting` should not be overridden by default.

### DIFF
--- a/library/core/functions.error.php
+++ b/library/core/functions.error.php
@@ -400,6 +400,14 @@ if (!function_exists('errorLog')) {
      * @param string|\Exception $message
      */
     function errorLog($message) {
+        $errorLogFile = class_exists('Gdn', false) ? Gdn::config('Garden.Errors.LogFile', '') : '';
+
+        // Log only if the PHP setting "log_errors" is enabled
+        // OR if the Garden config "Garden.Errors.LogFile" is provided
+        if (!$errorLogFile && !ini_get('log_errors')) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


library/core/functions.error.php
```diff
@@ -378,46 +392,168 @@ function errorMessage($Message, $SenderObject, $SenderMethod, $Code = '') {
...
+            $formattedMessage = sprintf('%s in %s on line %s', $formattedMessage, $file, $line);
+        } elseif ($file) {
+            $formattedMessage = sprintf('%s in %s', $formattedMessage, $file, $line);
+        }
+
+        return $formattedMessage;
+    }
+}
+
+if (!function_exists('LogException')) {
```
DaazKu: logException

PR #4048, opened 2016-05-27
Defer responsibility to manage TransformDiscussionToComment to other plugins

DaazKu:
Since we have more than one plugin that need to do some cleanup I think it should be their responsibility.

### DIFF
--- a/plugins/SplitMerge/class.splitmerge.plugin.php
+++ b/plugins/SplitMerge/class.splitmerge.plugin.php
@@ -218,11 +218,10 @@ class SplitMergePlugin extends Gdn_Plugin {
                     $CommentModel->Validation->results(true);
                     $CommentID = $CommentModel->save($Comment);
                     if ($CommentID) {
-                        // Move any attachments (FileUpload plugin awareness)
-                        if (class_exists('MediaModel')) {
-                            $MediaModel = new MediaModel();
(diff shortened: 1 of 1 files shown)


plugins/SplitMerge/class.splitmerge.plugin.php
```diff
@@ -218,11 +218,10 @@ public function moderationController_mergeDiscussions_create($Sender) {
...
                     if ($CommentID) {
-                        // Move any attachments (FileUpload plugin awareness)
-                        if (class_exists('MediaModel')) {
-                            $MediaModel = new MediaModel();
-                            $MediaModel->reassign($Discussion['DiscussionID'], 'discussion', $CommentID, 'comment');
-                        }
+                        $Comment['CommentID'] = $CommentID;
+                        $this->EventArguments['SourceDiscussion'] = $Discussion;
+                        $this->EventArguments['TargetComment'] = $Comment;
+                        $this->fireEvent('TransformDiscussionToComment');
```
linc: This would make more sense to `fireAs('ModerationController')`, I think. Hooking into a plugin-fired event otherwise requires the use of `base_` everywhere.

DaazKu: It does not. (You can use plugginName_event_handler)
But in this particular case I thought that no matter from where the event comes from you kind of want to handle it.

PR #4051, opened 2016-05-29
Set a default addon type of 'addon'

linc:
* A hundred addon checks will be made more concise.
* In 90% of use cases this is the value you want.
* It's completely reasonable & predicable that the `AddonManager`'s default assumption is `addon`.

### DIFF
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -533,7 +533,7 @@ class AddonManager {
      * @param string $type One of the **Addon::TYPE_*** constants.
      * @return bool Returns
      */
-    public function isEnabled($key, $type) {
+    public function isEnabled($key, $type = Addon::TYPE_ADDON) {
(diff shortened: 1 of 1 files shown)


linc:
Possible alternative for `functions.general.php`:

```
if (!function_exists('isEnabled')) {
    /**
     * Determine whether or not an addon is enabled.
     *
     * @see AddonManager::isEnabled().
     *
     * @param string $key The addon key.
     * @param string $type One of the **Addon::TYPE_*** constants.
     * @return bool
     */
    function isEnabled($key, $type = null) {
        // Add a useful assumption that we're talking about addons.
        if (is_null($type)) {
            $type = \Vanilla\Addon::TYPE_ADDON;
        }

        return AddonManager::isEnabled($key, $type);
    }
}
```

linc:
I like that newer idea better. https://github.com/vanilla/vanilla/pull/4059

PR #4052, opened 2016-05-30
Terminate unquoted mentions before apostrophes

### LINKED ISSUE #2924: @mention now allows apostrophe, which is bad
@linc's shouldn't add the 's to the mention.

bleistivt:
fixes #2924


linc:
Thanks, this was driving me nuts.

PR #4053, opened 2016-05-30
YouTube embed: More restricitve video ID match

### LINKED ISSUE #3377: Cannot post a youtube channels link in a comment
Youtube channel links automagically get turned into embedded players, but since the link is not a video it breaks.

Steps to reproduce:
[Post a URL such as]
https://www.youtube.com/user/VanillaForumsTV

http://vanillaforums.org/discussion/31330/cannot-post-a-youtube-channels-link-in-a-comment

bleistivt:
This prevents URIs that don't point to videos like https://www.youtube.com/user/VanillaForumsTV being matched.

While Google doesn't guarantee it will always be that way, video IDs have always been 11 characters and it probably won't change anytime soon.
More info: http://webapps.stackexchange.com/a/54448

fixes #3377

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1278,7 +1278,7 @@ class Gdn_Format {
         // https://www.youtube.com/watch?v=p5kcBxL7-qI
         // https://www.youtube.com/watch?v=bG6b3V2MNxQ#t=33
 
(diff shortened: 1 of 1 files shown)


linc:
Does this maintain compatibility with embedding playlists?

bleistivt:
yes, I tested it with all of these:

https://github.com/vanilla/vanilla/blob/cfd3021d25700ea3fc4605e67f384d89e386a93d/library/core/class.format.php#L1267-L1279

linc:
Thanks!

PR #4054, opened 2016-05-30
Add description for remoteAddress method

R-J:
This method is used but not described anywhere. Description is not as easy as e.g. remoteHost since for getting the IP, multiple server variables are checked.

Additionally, I've changed some casening. No functional code changes were done.


linc:
Good docs update, thanks!

PR #4059, opened 2016-05-30
Add isEnabled convenience function

linc:
Whereas: 

* A hundred addon checks will be made more concise.
* Nearly all existing checks of this nature are for an addon, not locale or theme.

Therefore:

I bring you `isEnabled()` with _optional_ type.


tburry:
I'd rather stop with the global functions that have dependencies into the framework. It's bad practice.

PR #4060, opened 2016-05-31
Refresh after announcing a discussion

### LINKED ISSUE #2013: Sink/Close + Announce redirects to /discussion/sink#Head
Sink a discussion, then announce it. It redirects to /discussion/sink#Head.
Close + announce does the same thing.

bleistivt:
`SelfUrl` can't be used for a target url, because it will point to `discussion/sink` or `discussion/close` whenever the dropdown is updated via ajax.

fixes #2013


linc:
Nice debug! Thanks

PR #4062, opened 2016-05-31
Add array exception to Smarty

linc:


DaazKu:
Fixes https://github.com/vanilla/internal/issues/515

Issue #3792, opened 2016-04-27
Markdown's lists need a space before their declaration to be rendered


bleistivt:
```
[spoiler]
- One
- Two
- Three

- Four
- Five
[/spoiler]
```

The resulting HTML looks like this:

```
<p><div class="Spoiler"><br />
- One<br />
- Two<br />
- Three</p>

<ul>
<li>Four</li>
<li>Five<br />
</div></li>
</ul>
```

See: https://vanillaforums.org/discussion/comment/240189/#Comment_240189

DaazKu:
```
blablabla
- item 1
- item 2
```

Same problem.

PR #4066, opened 2016-06-01
Markdown: Allow starting lists without a newline

bleistivt:
The current markdown implementation requires an empty line before a list:

```
This is a list:

* One
* Two
```

While on github and most other implementations, this will also work:

```
This is a list:
* One
* Two
```

The author explains the reasons for this here:
https://github.com/michelf/php-markdown/blob/6b4f89fd19944f0cdc36a8d6655a8909b91671d8/Michelf/Markdown.php#L1107-L1112

For unordered lists however, it is more intuitive not to require the empty line, see #3792


linc:
I like it.

Issue #4064, opened 2016-05-31
Remove invitation functionality from Getting Started


linc:
Possibly you can leave behind the suggestion to invite people, but the actual "send email" feature should be removed. It is not useful, misleading, and buggy.

PR #4071, opened 2016-06-02
Remove the invite feature from the GettingStarted page

DaazKu:
The email feature was intended to be used to invite 10 or so peoples but was used to mass invite.
Since it is bugged and we do not want to promote mass import we are removing it.

Fixes: https://github.com/vanilla/vanilla/issues/4064

### DIFF
--- a/applications/dashboard/views/settings/gettingstarted.php
+++ b/applications/dashboard/views/settings/gettingstarted.php
@@ -182,10 +182,6 @@ function tutLink($TutorialCode, $WriteTitle = TRUE, $ThumbnailSize = 'medium') {
         }
     </style>
     <h1><?php echo t('Getting Started with Vanilla'); ?></h1>
-<?php
-echo $this->Form->open();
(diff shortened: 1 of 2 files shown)


applications/dashboard/views/settings/gettingstarted.php
```diff
@@ -241,31 +241,7 @@ function tutLink($TutorialCode, $WriteTitle = TRUE, $ThumbnailSize = 'medium') {
...
-                        if ($(this).val() == '')
-                            $(this).val("<?php echo $this->TextEnterEmails; ?>");
-                    });
-                });
-            </script>
-            <?php echo $this->Form->button(t('Send Invitations!')); ?>
+            <p><?php echo t('Invite your friends to register to your new forum!'); ?></p>
         </div>
     </div>
 <?php echo $this->Form->close(); ?>
```
kaecyra: The form can be removed also.

PR #4072, opened 2016-06-02
Fix Smarty and dispatcher paths on Windows

DaazKu:
Smarty use a constant named `DS` to know how to handle `DIRECTORY_SEPARATOR`.
Having our own `DS` constant defined to '/' was making smarty crash on windows.

Also, the dispatcher had hard coded '/' as directory separators so it did not match `\controllers\`.

This PR address those issues.

### DIFF
--- a/index.php
+++ b/index.php
@@ -24,7 +24,7 @@ ob_start();
 
 // Define the constants we need to get going.
 
-define('DS', '/');
+define('DS', DIRECTORY_SEPARATOR);
 define('PATH_ROOT', getcwd());
 
 // Include the bootstrap to configure the framework.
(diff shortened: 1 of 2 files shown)


tburry:
Maybe we should rename our constant to DS_STORE since that won't affect Windows.

DaazKu:
I'm not sure I understand the part about windows in your comment. Don't we use DS to know the current system directory separator?

tburry:
It was a joke. DS_Store is a hidden mac file. I'm okay with this PR in principle, but @Linc will need to get someone to QA it.

In general, we want to get `DS` out of Vanilla so replacing it with "/" in core is fine. The constant is so old that it will have to stay though.

DaazKu:
I didn't get the joke without the dot in front od DS_Store. loll

PR #4083, opened 2016-06-07
Fix sorting and sync problems in the Dashboard menu creation (part 2)

linc:
1. The Dashboard menu items still are not sorting correctly. This is visible most easily on the Categories page, but there are several other things amiss everywhere. The problem is we need to build it in this order EXACTLY: Dashboard tier-1 -> Vanilla (all) + Dashboard tier-2 -> Plugins etc. Any deviation from this results in wrongly ordered items. We only had 2 events, and because of the new addon manager order were firing things in the order: Dashboard (all) -> Plugins -> Vanilla (all). This fails when Vanilla initiates as in the `vanillasettings` controller, but it also fails to put Vanilla's menu items before plugin items in second-tier ordering. My solution here rather than add a third, even earlier event is to move the manual build of Dash tier-1 into the `addSideMenu()` methods.

2. The `addSideMenu()` methods in the DashboardController and the VanillaSettingsControllers are bloody copies of each other. It sucks but it's been fine for 6 years so whatev. But then they need to actually be EXACT copies and not have too-clever interactions with the module. If you're gonna hack something, make it easy to read and not break.

### DIFF
--- a/applications/dashboard/controllers/class.dashboardcontroller.php
+++ b/applications/dashboard/controllers/class.dashboardcontroller.php
@@ -59,10 +59,13 @@ class DashboardController extends Gdn_Controller {
     /**
      * Build and add the Dashboard's side navigation menu.
      *
+     * EXACT COPY OF VanillaController::addSideMenu(). KEEP IN SYNC.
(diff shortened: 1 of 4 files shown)


hgtonight:
This is good.

PR #4095, opened 2016-06-13
Fix emoji menu not being properly dismissed

### LINKED ISSUE #3883: Emojis box in WYSIWYG Advanced Editor persists after ":"
After typing a : in the WYSIWYG editor makes the emoji choice box appear. However, adding a space does not make the box disappear as it should.

999999999 20160506140557345 999999999 20160506140557345ozb7k

initvector:
The current regex for matching emojis allows for matches including trailing whitespace and newlines.  This can cause the autocomplete menu from disappearing if users enter a string like " : ".  The menu would appear when the colon was typed and not disappear when the space was typed after that, as it should.

This update alters emoji regex matching to remove the ability to allow emoji to end with spaces.  The pattern has also been simplified by combining like elements.

Closes #3883

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -1814,7 +1814,8 @@ jQuery(document).ready(function($) {
                             flag = '(?:^|\\s)' + flag;
                         }
 
-                        regexp = new RegExp(flag + '([A-Za-z0-9_\+\-]*|[^\\x00-\\xff]*)$', 'gi');
+                        // Some browsers append a linefeed to the end of subtext.  We need to allow for it.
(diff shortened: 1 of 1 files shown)


initvector:
Seeing some inconsistent behavior with Firefox.  Sometimes the emoji menu appears.  Sometimes it doesn't.  The implementation in Firefox might not allow trailing whitespace characters, because it seems to eat those after an emoji has been inserted.  Deleting characters seems to occasionally be a trigger for causing the autocomplete menu not to display.

PR #4110, opened 2016-06-15
Add legacy IP data migration

### LINKED ISSUE #3907: Partially migrate IP address information into GDN_UserIP
Prerequisite: #3906

Migrating all of the IP addresses from GDN_User.AllIPAddresses is going to be too onerous. I'd do the following:

1. Run some update queries when the GDN_UserIP address is first created.
2. Insert Gdn_User.InsertIPAddress into Gdn_UserIP with DateInserted when AllIPAddresses is like %InsertIPAddresss%. The extra condition will prevent the IP addresses of admins that have manually created user accounts.
3. Insert Gdn_User.LastIPAddress into Gdn_UserIP with DateLastActive on duplicate key just update DateUpdated to DateLastActive.
4. Make sure to filter out null, empty, and 0.0.0.0 IP addresses.

initvector:
This update adds capabilities to the Vanilla application's structure to migrate legacy IP data from the `User` table into the `UserIP` table.  Migration can be done with `/utility/update`.

The migration is done in batches of 10,000. The `InsertIPAddress` and `LastIPAddress` fields are analyzed and inserted into the `UserIP` table, if they are valid.  The `AllIPAddresses` column is wiped out after this operation and acts as a marker of which user records have been updated.

In addition, `UserModel::saveIP` has been updated to verify IPs with `filter_var`, because the `$IP` parameter is expected to be a human-readable address.  A new parameter, `$dateUpdated` has also been added to this function to force a specific `DateUpdated` column value.

Closes #3907

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -769,21 +769,26 @@ $Construct
 // If the AllIPAddresses column exists, attempt to migrate legacy IP data to the UserIP table.
 if ($AllIPAddressesExists) {
     $limit = 10000;
-    $offset = 1;
+    $resetBatch = 100;
+
+    // Grab the initial batch of users.
     $legacyIPAddresses = $SQL->select(['UserID', 'AllIPAddresses', 'InsertIPAddress', 'LastIPAddress', 'DateLastActive'])
-        ->from('User')
-        ->where('AllIPAddresses is not null')
-        ->get()
-        ->resultArray();
+        ->from('User')->where('AllIPAddresses is not null')->limit($limit)
+        ->get()->resultArray();
 
     do {
+        $processedUsers = [];
+
+        // Iterate through the records of users with data needing to be migrated.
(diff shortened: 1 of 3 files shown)


linc:
God I am just terrified of this but we gotta rip the bandaid off and do it.

applications/dashboard/settings/structure.php
```diff
@@ -765,6 +766,69 @@
...
+                    $insertIPAddress
+                );
+            }
+
+            // Record the processed user's ID.
+            $processedUsers[] = $userID;
+
+            // Every X records (determined by $resetBatch), clear out the AllIPAddresses field for processed users.
+            if (count($processedUsers) > 0 && (count($processedUsers) % $resetBatch) === 0) {
+                $this->SQL->update('User')->set('AllIPAddresses', null)->whereIn('UserID', $processedUsers)
```
linc: Usually if a SQL chain needs multiple lines, it's preferable to just break it at every `->` and stack them.

PR #4111, opened 2016-06-16
Fix Advanced Editor upload incompatibility with MySQL strict mode

DataTables:
If MySQL is in strict mode it will reject uploads from the advanced
editor with the following error

> PHP Fatal error: Incorrect integer value: '' for column 'ThumbWidth'
  at row 1|Gdn_Database|Query|insert GDN_Media

This is because the advanced editor is trying to insert a string value
into an integer. Setting a default value of 0 for the thumb and image
height and width values corrects this.

* Forum thread on this topic:
  https://vanillaforums.org/discussion/32255/advanced-editor-upload-failing-incorrect-integer-value#latest
 master

### DIFF
--- a/plugins/editor/class.editor.plugin.php
+++ b/plugins/editor/class.editor.plugin.php
@@ -750,10 +750,10 @@ class EditorPlugin extends Gdn_Plugin {
 
             // Determine if image, and thus requires thumbnail generation, or simply saving the file.
             // Not all files will be images.
-            $thumbHeight = 0;
-            $thumbWidth = 0;
-            $imageHeight = 0;
-            $imageWidth = 0;
+            $thumbHeight = null;
+            $thumbWidth = null;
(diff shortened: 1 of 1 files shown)


initvector:
Thanks for the contribution!

The database defaults for the `Media.Thumb*` and `Media.Image*` fields are `null`.  Instead of explicitly setting an integer default in the code, wouldn't it be better to use the same `null` default?

DataTables:
`null` sounds a lot more sensible to me! Do you want me to update the PR or are you on the case?

initvector:
If you don't mind taking care of it, that'd be excellent.  Thanks for the quick response!

DataTables:
No problem - wish I could always respond this quickly...!

I think that should be it done and through on the PR now.

initvector:
Fantastic.  Thanks again!

PR #4118, opened 2016-06-20
Use correct title for Edit Role

R-J:
The title (html tag) was wrong for rolecontrollers edit method. The title in the view was correct since it did not rely on $this->data('Title'). I've also corrected that.

### DIFF
--- a/applications/dashboard/views/role/edit.php
+++ b/applications/dashboard/views/role/edit.php
@@ -1,3 +1,4 @@
+<?php if (!defined('APPLICATION')) exit(); ?>
 <h1><?php echo $this->title(); ?></h1>
 <?php
 echo $this->Form->open();
--- a/applications/dashboard/views/role/edit.php
+++ b/applications/dashboard/views/role/edit.php
@@ -1,10 +1,4 @@
-<?php if (!defined('APPLICATION')) exit(); ?>
-<h1><?php
-    if (is_object($this->Role))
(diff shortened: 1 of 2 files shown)


applications/dashboard/views/role/edit.php
```diff
@@ -1,10 +1,4 @@
-<?php if (!defined('APPLICATION')) exit(); ?>
```
beckyvb: Was removing this intentional?

DaazKu: Intentional or not this is correct.

This way of doing if now deprecated.

beckyvb: It's not needed in class definitions, but still needed in views.

DaazKu: I stand corrected.

R-J: Yes, I've dropped it because I thought I remembered a discussion stating that this is not needed any more. Frankly spoken, I do not understand the difference why it is not needed in classes but needed in views.
But nevertheless, I've corrected the PR

hgtonight: @R-J You need it in any file that prints something directly to the output. Files that don't (ala plugin files, models, etc.), don't need it.

The idea is to prevent output from happening unless it is through the framework (i.e. APPLICATION is defined).

PR #4119, opened 2016-06-20
HeadModule->title() NoSubTitle parameter is useless.

R-J:
The function begins like that:
~~~
public function title($Title = '', $NoSubTitle = false) {
    if ($Title != '') {
        // Apply $Title to $this->_Title and return it;
        $this->_Title = $Title;
        $this->_Sender->title($Title);
        return $Title;
~~~

So if the first parameter `$Title` is specified, the second parameter is not considered. `$this->_Title` is set to `$Title` but in the definition it is clearly stated that it is `The main text for the "title"`.

The title method is never called with a parameter (except for one wrong usage in rolecontroler->edit for which I already made a PR). That's why changing the logic will not change the behavior. By now calling `title('Some Title')` would have created a title "Some Title" and after my Pull Request it would be "Some Title - $Subtitle" which, based on the function head, is the desired result.

Besides of that, $_TitleDivider and $_Subtitle were useless. So I have added a setter method for both so that users can take more influence on the title. I've changed `SubTitle` to `Subtitle` like it is written in class.controller.php

### DIFF
--- a/applications/dashboard/modules/class.headmodule.php
+++ b/applications/dashboard/modules/class.headmodule.php
@@ -34,9 +34,9 @@ if (!class_exists('HeadModule', false)) {
         protected $_Title;
 
         /** @var string A string to be concatenated with $this->_Title. */
-        protected $_SubTitle;
+        protected $_Subtitle;
 
-        /** @var A string to be concatenated with $this->_Title if there is also a $this->_SubTitle string being concatenated. */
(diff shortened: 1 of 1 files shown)


linc:
Let's use the [grammatically correct em dash](http://chicagomanualofstyle.org/qanda/data/faq/topics/HyphensEnDashesEmDashes/faq0002.html) instead of a hyphen if we're adding a separator here.

R-J:
That's a really great feedback!

Issue #3882, opened 2016-05-06
Remove the e modifier from every regexes. (PHP 7 compatibility)


DaazKu:
See: https://github.com/vanilla/vanilla/pull/3880

I did a quick search and there are 2 places where the modifier is still used:

https://github.com/vanilla/addons/blob/master/plugins/BluehostFix/class.phpmailer.php
https://github.com/vanilla/vanilla/blob/master/library/vendors/phpmailer/class.phpmailer.php

Updating the library might be a good idea.

tburry:
FYI: We've slightly customized PHP Mailer so a simple switch can't be done. T

linc:
The PHPmailer adjustment should be backported to `release/2.3`.

I'd recommend against updating phpmailer as part of this task. We can file a separate one for that if we want. PHP7 compat is a higher priority.

linc:
There's already a separate task for moving phpmailer to composer, so that is where you can address an upgrade too if you want: https://github.com/vanilla/vanilla/issues/3895

linc:
Recommend doing this fix first or else you'll have a problem backporting.

DaazKu:
You beat me to link #3895 .

DaazKu:
A more intensive search I found regexes using the e modifier in:

- library/vendors/Auth/Yadis/ParseHTML.php
- library/vendors/phpmailer/class.phpmailer.php
- plugins/BluehostFix/class.phpmailer.php
- plugins/NBBC/nbbc/nbbc.php
- plugins/NBBC/nbbc/src/nbbc_parse.php

DaazKu:
Fixed in:
https://github.com/vanilla/vanilla/pull/4120
https://github.com/vanilla/addons/pull/347

PR #4120, opened 2016-06-20
Replace preg_replace regexes using the e modifier with preg_replace_callback

DaazKu:
The e modifier was removed from PHP7

### DIFF
--- a/library/vendors/Auth/Yadis/ParseHTML.php
+++ b/library/vendors/Auth/Yadis/ParseHTML.php
@@ -82,8 +82,20 @@ class Auth_Yadis_ParseHTML {
 
         // Replace numeric entities because html_entity_decode doesn't
         // do it for us.
-        $str = preg_replace('~&#x([0-9a-f]+);~ei', 'chr(hexdec("\\1"))', $str);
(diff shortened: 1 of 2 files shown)


initvector:
Instead of modifying the PHPMailer class in the vendors directory, would it make more sense to upgrade the library?  It looks like their latest version does not use the `PREG_REPLACE_EVAL` modifier.

I think we have one custom modification, `CountRecipients`, which we might be able to do away with in favor of counting the result of the newer `PHPMailer::getAllRecipientAddresses` function.

DaazKu:
https://github.com/vanilla/vanilla/issues/3882#issuecomment-226817396 :D

JochenBoettcher:
The problem still exists in the file "library\vendors\phpmailer\class.phpmailer.php" on line 1697 in the public function EncodeQ.

At php 7.0 an error is triggered here. In result, all emails with umlauts (äöü) are sent with "=?utf-8?Q??=" as subject.

If I insert the new code described here (new line 1753), no error will be raised anymore, but all umlauts in the subject line will be missing.

Issue #4121, opened 2016-06-20
Spoilers not rendering on Best Of page


beckyvb:
Spoilers are not rendering on the Best Of pages.

PR #4122, opened 2016-06-20
Render spoilers everywhere in Vanilla.

beckyvb:
Consolidate the spoilers css and js into the base_render_before for a DRYer approach.

Closes https://github.com/vanilla/vanilla/issues/4121


applications/vanilla/settings/class.hooks.php
```diff
@@ -367,6 +367,15 @@ public function base_render_before($sender) {
...
         }
+        
+        if (!inSection('Dashboard')) {
+            // Spoilers assets
+            $sender->addJsFile('spoilers.js', 'dashboard');
+            $sender->addCssFile('spoilers.css', 'dashboard');
+            $sender->addDefinition('Spoiler', t('Spoiler'));
+            $sender->addDefinition('show', t('show'));
+            $sender->addDefinition('hide', t('hide'));
+        }
```
hgtonight: This seems like it should be in `/applications/dashboard/settings/class.hooks.php` instead.

beckyvb: Yeah, I had some hesitation there, too. I felt strange about adding it to the dashboard hooks and then saying 'Only add this if we're *not* in the dashboard'. On the other hand, the assets were located in the dashboard. I decided to move the assets to the Vanilla application instead, because that's where they're being used.

initvector:
Looks good!  One minor thing: spoilers aren't rendering properly on `/profile/comments`

PR #4127, opened 2016-06-21
Fix inability to register when no captcha plug-in enabled

### LINKED ISSUE #4087: Registration fails when the reCaptcha plugin is disabled.
screen shot 2016-06-09 at 1 46 17 pm

The plugin is disabled so obviously I shouldn't have to enter a captcha...

initvector:
This update alters the behavior of `Captcha::enabled`.  It now considers if any plug-ins have registered captcha validation handlers before determining if captcha is enabled for a site, instead of solely relying on the `Garden.Registration.SkipCaptcha` config check.

Closes #4087

### DIFF
--- a/library/core/class.captcha.php
+++ b/library/core/class.captcha.php
@@ -23,7 +23,15 @@ class Captcha {
      * @return boolean
      */
     public static function enabled() {
-        return !c('Garden.Registration.SkipCaptcha', false);
+        // Detect any available captcha validation handlers.
+        $validateHandlers = Gdn::pluginManager()->getEventHandlers(
+            '',
+            'validate',
+            'Handler',
+            ['ClassName' => 'captcha']
+        );
+
+        return c('Garden.Registration.SkipCaptcha') ? false : count($validateHandlers) > 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


linc:
I'm worried this is too clever.

* The PluginManager is being deprecated, so adding more dependencies on it isn't great.
* The existing pattern used for this type of situation is to have the plugin(s) explicitly toggle a `Handled` EventArgument.
* Reaching into the guts of PluginManager and using a function that probably shouldn't even be public in the first place seems like something that's gonna break, especially since I can find no other use of that method in the entire application.
* The breaking condition for this is to leave your forum open to spam floods.

initvector:
I've ditched the previous method in favor of the following:

An `isEnabled` event is fired.  Captcha plug-ins just need to indicate they're enabled during this event.  If no plug-ins respond, `Captcha::enabled` returns `false`, so captcha checks are disabled. I didn't go the route of a "Handled" event argument, since we can avoid even getting into `Captcha::validate` by detecting the presence of captcha services in the "enabled" check.

I did add a new call to `Gdn::pluginManager`, but the `Captcha` class is full of them and that seems to be the way to `fireAs` in Vanilla.

linc:
I dig it.

I expect we'll continue to support `fireAs` so that's fine.

PR #4134, opened 2016-06-23
Fix emoji, mention rendering inside Wysiwyg code blocks

### LINKED ISSUE #4085: [Regression] Markup inside WYSIWYG code blocks must not be parsed
Related: https://github.com/vanilla/vanilla/pull/2694

The previous fix use `<code></code>` but the generated tags when using WYSIWYG are `<pre class="CodeBlock"></pre>`

initvector:
We avoid rendering emoji and mentions inside `code` tags.  Advanced Editor uses `pre` tags with a class of "CodeBlock" to simulate a `code` tag.  This causes the contents to bypass the code block safe zone and they are rendered as normal post content.

There is some indication that [`pre.CodeBlock` should contain a `code` element](https://github.com/vanilla/vanilla/blob/5e83d9212f8727d4d83dc5f6dee980fde75f3c5c/plugins/editor/js/buttonbarplus.js#L542). This update adds that element to code blocks generate by wysihtml5 for the Wysiwyg format in Advanced Editor.

Closes #4085


initvector:
There are some styling inconsistencies that should be addressed.  Markdown code blocks, with just the `code` tag, appear with a yellow (#ff9) background.  The pre+code blocks, generated with Wysiwyg, appear with the same background color, but also wrapped in light gray (rgb-style) background color.  I'm not sure if we should style the code-only blocks like the previous pre.CodeBlock blocks, vice versa or neither.

plugins/editor/js/editor.js
```diff
@@ -1682,16 +1682,54 @@
...
+                                                var range = composer.selection.getRange();
+                                                var selectedNodes = range.extractContents();
+                                                var pre = composer.doc.createElement("pre");
+                                                var code = composer.doc.createElement("code");
+
+                                                pre.className = "CodeBlock";
+                                                pre.appendChild(code);
+                                                code.appendChild(selectedNodes);
+                                                range.insertNode(pre);
+                                                composer.selection.selectNode(pre);
```
beckyvb: When creating a new code block, the cursor is outside of the code block. Changing this to `composer.selection.selectNode(code);` fixes this.

initvector: Good call!  Updated.

PR #4135, opened 2016-06-24
Fix plugin->isEnabled() method

R-J:
The method was using the old config syntax "Plugins.Name.Enabled", but now this is stored as "EnabledPlugins.Name".

This should be backported since it breaks plugins functionality.


R-J:
Please forget that. I have been totally on the wrong track...

R-J:
Nope, I guess I'm not. Sorry, I'm getting a little bit confused right now.

R-J:
*mumble, mumble* closed again *mumble, mumble*

PR #4136, opened 2016-06-24
Update embed.js

bolechen:
Fix bug:
In safari,when the url has utf8 charset.
embed mode will broken.


initvector:
Since this issue seems specific to Safari, I'm guessing other browsers are properly encoding it.  If you're always encoding it as a fix for Safari, aren't you potentially double encoding it for other browsers?

initvector:
Two weeks with no response.  Closing.

Issue #4112, opened 2016-06-16
Advanced notifications do not show when Registration. NoEmail = true


DaazKu:
We should remove that check and update the code to hide the "email" columns.

PR #4140, opened 2016-06-28
Display advanced notification settings even when Garden.Registration.NoEmail is enabled.

DaazKu:
Update advanced notification to take Garden.Registration.NoEmail into account instead of not displaying at all.

Fixes https://github.com/vanilla/vanilla/issues/4112

### DIFF
--- a/applications/vanilla/views/vanillasettings/notificationpreferences.php
+++ b/applications/vanilla/views/vanillasettings/notificationpreferences.php
@@ -1,14 +1,4 @@
-<?php if (!defined('APPLICATION')) return;
-
-if ($Sender->data('NoEmail')) {
-    $emailEnabled = false;
-    $colspan = ' colspan="2" ';
-} else {
-    $emailEnabled = true;
-    $colspan = null;
-}
-?>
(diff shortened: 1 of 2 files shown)


applications/vanilla/views/vanillasettings/notificationpreferences.php
```diff
@@ -1,4 +1,14 @@
...
+
+if ($Sender->data('NoEmail')) {
+    $emailEnabled = false;
+    $colspan = ' colspan="2" ';
+} else {
+    $emailEnabled = true;
+    $colspan = null;
+}
+?>
+asdfdsgssdgsgs
```
linc: yeah?

linc:
`Garden.Registration.NoEmail` should not control the availability of email preferences.

That setting is `Garden.Email.Disabled`. Having a `Registration`-namespaced config controlling non-registration email availability doesn't make sense.

linc:
I think my preferred fix here is to simply revert the changes to the `notificationpreferences.php` view and we can revisit that another time in a way that doesn't add more logic to the view.

Issue #4103, opened 2016-06-15
Add Profile Field: Not save with non-latin characters (Vanilla 2.3beta)


KitsuneSolar:
Add Profile Field...

**Example №1**

```
Type: TextBox
Label: YouTube Link
Show on profiles: Yes
```

... Save ... Ok, field is saved and displayed.

**Example №2**

```
Type: TextBox
Label: Ссылка на Ютуб
Show on profiles: Yes
```

... Save ... Nothing happens. The field is not saved and is not shown.

PR #4141, opened 2016-06-28
Fix Profile Extender not saving field names containing non-ASCII characters

DaazKu:
Fix https://github.com/vanilla/vanilla/issues/4103 .

The problem was that the generated slug to index the field was empty since all characters were stripped.


linc:
When fixing an issue with the Backport label, remember to add it to the PR also.

plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -318,7 +318,7 @@ public function settingsController_profileFieldAddEdit_create($Sender, $Args) {
                 // Make unique name from label for new fields
                 $Name = $TestSlug = substr(preg_replace('`[^0-9a-zA-Z]`', '', val('Label', $FormPostValues)), 0, 50);
                 $i = 1;
-                while (array_key_exists($Name, $Fields) || in_array($Name, $this->ReservedNames)) {
+                while (empty($Name) || array_key_exists($Name, $Fields) || in_array($Name, $this->ReservedNames)) {
```
linc: If you have an empty `$Name` that means `$TestSlug` is also empty, which means on the next line you are allowing it to just be a numeric `$Name`, no? That seems like a problem.

linc:
Wouldn't it have been easier to update the regex to allow Unicode?

DaazKu:
Having the slug being only a number works perfectly.

I could allow Unicode characters but I will need to test that change more thoroughly.

DaazKu:
`-_-` as label would still fail.

I'll make something more robust.

Issue #4073, opened 2016-06-03
"Import" page bonks when mysql extension not installed


linc:
The `mysql` PHP  extension is deprecated and now removed on many platforms.

```
Fatal Error in ImportModel._LoadTableLocalInfile();
Call to undefined function mysql_connect()
The error occurred on or near: /applications/dashboard/models/class.importmodel.php

1285:         // We've got to use the mysql_* functions because PDO doesn't support load data local infile well.
1286:         $dblink = mysql_connect(c('Database.Host'), c('Database.User'), c('Database.Password'), false, 128);
```

Minimally we should not throw a fatal error in this circumstance.

DaazKu:
That is the only place in vanilla that the `mysql` extension is used. We can simply replace it by `mysqli` for the time being.
Or check it PDO got fixed.

JonatanM:
I did it but it is not entirely sure

dblink = new mysqli('Database.Host', 'Database.User', 'Database.Password', 'Database.Name');
$Result = mysqli_query($Sql, $dblink);
if ($Result === false) {
$Ex = new Exception(mysql_error($dblink));
mysql_close($dblink);
throw new $Ex;
}

DaazKu:
@JonatanM You should restart the import from scratch after that fix has been done.

JonatanM:
I did it but still continued receiving this error :(

Table GDN_zUser doesn't exist
or
Incorrect integer value: '' for column 'InviteUserID' at row 1

DaazKu:
Are you on windows?

JonatanM:
yes

JonatanM:
Ubuntu 16.04

DaazKu:
What version of Vanilla and Mysql are you running?

JonatanM:
PHP Version 7
mysql 5
Vanilla  2.3b1
Ubuntu 16.04

DaazKu:
As I suspected. MySQL Strict mode problem. I suggest you just deactivate it until we fix the issue.

JonatanM:
Nice @DaazKu you are a ninja uheuheuhe

Sorry for my English, it was a compliment

PR #4145, opened 2016-06-29
Replace MySQL with MySQLi for imports

DaazKu:
Fix: https://github.com/vanilla/vanilla/issues/4073

Mysql is removed from PHP7.0
Mysqli is available since 5.3 and we support 5.4 an up so we are good :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ If you professionally run a large community or enterprise forum, our cloud solut
 * PHP version 5.3 or newer with `--enable-mbstring`.
 * pdo_mysql module must be enabled.
 * MySQL 5 or newer.
-* If you intend to [Migrate to Vanilla](#migrating-to-vanilla) you will also need php with `--with-mysqli`.
+* If you intend to [Migrate to Vanilla](#migrating-to-vanilla) you will also need PHP with `--with-mysqli`.
 
 Vanilla 2.2 requires PHP 5.3. Running `master` branch requires PHP 5.4, as will future official releases.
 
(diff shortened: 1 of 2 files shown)


linc:
This is actually an undocumented depedency: https://github.com/vanilla/vanilla#self-hosting-requirements

So let's get a note in there too.

* PHP `mysqli` extension is required only for imports.

Did you look at whether PDO still has the limitation described in the comment that made us use not-PDO in the first place?

DaazKu:
I'll add the note.
I didn't check. I thought that it would be safer to assume that PDO is still broken (for backward compatibility).
We could create another PR to determine if the issue is fixed, and when was it fixed, if so.

DaazKu:
An exception is raised if we use PDO.

> Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

This cannot be true since we execute a [truncate](https://github.com/vanilla/vanilla/pull/4145/files#diff-6252cf1a96c1fc1b23d094e03a2631ebR1275) right before it.

linc:
Still needs a dependency note in the readme.

linc:
Fix your capitalization. I'm really growing tired of fixing every acronym you type in PR titles and elsewhere. PHP, CSS, MySQL, MySQLi - it matters for legibility and clarity.

DaazKu:
I was not aware of that.
I'll be careful from now on.

Issue #3980, opened 2016-05-16
Warning when saving in /settings/advanced


DaazKu:
The problem is caused by 'Vanilla.Comments.AutoRefresh' not having a default value (old config?!?).

```
Trying to get property of non-object

#0 /Users/alexandre/workspace/repos/vanilla/library/core/class.validation.php(147): Gdn_ErrorHandler(8, 'Trying to get p...', '/Users/alexandr...', 147, Array)
#1 /Users/alexandre/workspace/repos/vanilla/library/core/class.validation.php(387): Gdn_Validation->applyRulesBySchema(Array)
#2 /Users/alexandre/workspace/repos/vanilla/library/core/class.validation.php(335): Gdn_Validation->getSchemaRules()
#3 /Users/alexandre/workspace/repos/vanilla/library/core/class.validation.php(616): Gdn_Validation->defineValidationRules(Array, false)
#4 /Users/alexandre/workspace/repos/vanilla/library/core/class.configurationmodel.php(132): Gdn_Validation->validate(Array)
#5 /Users/alexandre/workspace/repos/vanilla/library/core/class.form.php(2222): Gdn_ConfigurationModel->save(Array, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL)
#6 /Users/alexandre/workspace/repos/vanilla/applications/vanilla/controllers/class.settingscontroller.php(79): Gdn_Form->save()
#7 /Users/alexandre/workspace/repos/vanilla/library/core/class.dispatcher.php(326): SettingsController->advanced()
#8 /Users/alexandre/workspace/repos/vanilla/index.php(44): Gdn_Dispatcher->dispatch()
#9 {main}
```

https://github.com/vanilla/vanilla/blob/cfde7a18e3a7f364cb37fa58756f66e3fe84b2c3/library/core/class.validation.php#L147

linc:
Yes, we no longer respect that config setting because AutoRefresh is not currently allowed.

PR #4147, opened 2016-06-29
Removes references to Vanilla.Comments.AutoRefresh

DaazKu:
This config is not used anymore!


DaazKu:
Fix: https://github.com/vanilla/vanilla/issues/3980

linc:
Let's restrict this to simply removing it from the Dashboard, not its end effect in `discussion/embed.php` and `index.php`. The config is still usable manually, I enjoy tinkering with it on my sites, and in any case this only removed some the system.

PR #4133, opened 2016-06-23
Update IP Address field types to varbinary(16)

initvector:
This update alters the `ipaddress` type alias in `Gdn_MySQLStructure` from `varchar(15)` to `varbinary(16)`.

Data seems to remain intact after a utility-update.  Updating data also seems to be unaffected.

Closes #3805

PR #4150, opened 2016-06-30
Update all IP address fields to be able to store IPv6 values

### LINKED ISSUE #3809: Pass all IP addresses through an encode function before going to the database.

initvector:
This update encodes all IP address values in preparation for storage in `varbinary(16)` fields.

Requires https://github.com/vanilla/vanilla/pull/4133

Closes #3809


linc:
This is scheduled to merge Oct 3.

PR #4160, opened 2016-07-04
Gdn_Dispatcher untagling, part 2

tburry:
This is a fairly big refactor, but is really only moving things out of large methods into smaller methods.

There is a bit of a methodology change too where controllers are discovered before their methods are examined. This mirrors where the dispatcher is going even though it currently doesn’t serve much of a great purpose.

I recommend walking through the code to review this PR, starting at Gdn_Dispatcher->dispatch(). You should be able to see the logic of a dispatch more clearly now:

1. Check to see if the application is in update mode.
2. Check for GDN_Route URL rewrites.
3. Check against private communities.
4. Analyze the request, looking for a controller.
5. Dispatch to that controller.

    a) Instantiate the controller.
    b) Look for the method to dispatch to.
    c) Reflect the method's arguments.
    d) Call the controller method.

There are two risks with this PR that will need special attention:

**Lost events.** I may have missed a copy-paste of a call to fireEvent or missed the arguments necessary. You can compare dispatcher events by logging them to the debug bar:

```php
public function Base_All_Handler($Sender, $Args, $Key) {
        $ajaxBreakEvents = array(
            'Gdn_Controller_Finalize_Handler',
            'Gdn_Dispatcher_Cleanup_Handler',
            'Base_Render_Before'
        );

        if (in_array($Key, $ajaxBreakEvents)) {
            return;
        }

        unset($Args['WildEventStack']);
        Logger::debug($Key, $Args);
    }
```

**Different controller properties.** The dispatcher sets a tonne of controller properties. I'd dump the controller out just before calling its method on master and this branch and compare the properties. They should be the *exact* same or let me know.

### TODO

- [x] Support querystring passthrough.
- [ ] Support case-insensitive delivery type, delivery method.

### DIFF
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -918,7 +918,7 @@ class SettingsController extends DashboardController {
      * @since 2.0.0
      * @access public
      */
-    public function xIndex() {
+    public function index() {
         $this->addJsFile('settings.js');
         $this->title(t('Dashboard'));
 
--- a/conf/constants.php
+++ b/conf/constants.php
@@ -42,6 +42,7 @@ define('DELIVERY_METHOD_XML', 'XML');
 define('DELIVERY_METHOD_TEXT', 'TXT');
 define('DELIVERY_METHOD_PLAIN', 'PLAIN');
 define('DELIVERY_METHOD_RSS', 'RSS');
+define('DELIVERY_METHOD_ATOM', 'ATOM');
 
 // Handler enumerators:
 define('HANDLER_TYPE_NORMAL', 'NORMAL'); // Standard call to a method on the object.
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -225,7 +225,7 @@ class Gdn_Controller extends Gdn_Pluggable {
         $this->RedirectUrl = '';
         $this->RequestMethod = '';
         $this->RequestArgs = false;
-        $this->Request = false;
+        $this->Request = null;
         $this->SelfUrl = '';
         $this->SyndicationMethod = SYNDICATION_NONE;
         $this->Theme = Theme();
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -241,8 +241,8 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
     private function requestVal($key, $get, $post, $default = null) {
         $keys = [$key, lcfirst($key), strtolower($key)];
 
-        if (isset($get[$keys[0]])) {
-            return $get[$keys[0]];
+        if (isset($get[$keys[2]])) {
+            return $get[$keys[2]];
         }
 
         foreach ($keys as $key) {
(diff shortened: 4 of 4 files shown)


tburry:
Depends on #4158.

linc:
Prerequisite is merged and rebase pull was done.

linc:
In the original dispatcher, you look for `DeliveryType` & `DeliveryMethod` in the merged collection of GET and POST:

```
$Controller->deliveryType($Request->getValue('DeliveryType', $this->deliveryType));
$Controller->deliveryMethod($Request->getValue('DeliveryMethod', $this->deliveryMethod));
```

In this version, you only look in GET:

```
// Initialize the result of our request.
        $result = [
           ...
            'query' => array_change_key_case($request->get())
        ];
...

// An explicitly passed delivery type/method overrides the default.
$result['deliveryMethod'] = val('deliverymethod', $result['query'], $deliveryMethod ?: DELIVERY_METHOD_XHTML);
$result['deliveryType'] = val('deliverytype', $result['query'], $deliveryType);
```

This breaks a ton of requests, including every `Hijack` link in the app.

linc:
Also noted by @kaecyra: after #4158, if you follow an internal route, `Gdn::request()` is not parsed. This is a regression currently in production.

linc:
I am moving this to Sprint 17 because I'm not merging this so late in a sprint / right before I'm out.

tburry:
Can you explain what you mean by "Gdn::request() is not parsed"?

kaecyra:
screen shot 2016-07-15 at 10 24 27 am 3

tburry:
I'm dealing with the request parsing in #4218.

tburry:
I've dealt with the delivery methods/type issues in 419c373e75b1885d2f2395c199b783c097515afc @linc. Note the comments on the commit though.

linc:
Thanks. I did notice it was doing more than GET/POST but agree this will likely suffice.

tburry:
Good catch; fixed.

linc:
I've found two issues related to `Gdn_Dispatcher::findController()`.

1. The path `/dashboard/settings` returns a 404. This is because `SettingsController` uses an `xIndex` method to allow VanillaStats to override it. The new `findController()` does not manually set the index value for the method, so `findControllerMethod()` does not discover the `xIndex` use. 

2. The path `/groups` returns a 404. This is because the new `findController()` uses `array_shift()` on the path if it finds a valid application name at the start of it. We've established it's acceptable to name a controller after an application however, and this situation leads to an empty `$parts` and therefore no controller is found.

tburry:
Good finds.

1. I've opted to remove the xIndex support since plugins can override any controller methods and the "x" methods are an extra layer of complexity.
2. I've added a check for a controller that matches an application name. I think this is all correct.

linc:
Enabling addons does not work because it is lowercasing the entire path, which we magically pass as method arguments. In this case, the addonkey and transientkey are both being lowercased in the call to `SettingsController::plugins($Filter = '', $PluginName = '', $TransientKey = '')`.

linc:
Rebased.

PR #4170, opened 2016-07-05
Translate validation error messages in the UserModel

nervoustwit:
Wrapped the error message for captcha and password recovery because a client requested them to be translated. There are other strings in the class that could be translated but the are generally only seen be admins and mods and are difficult to test.


linc:
This is not correct.

* Lots of places in Vanilla use `addValidationResult()` in this way. If this were the correct approach, it should be addressed consistently in the code base, not in one model.
* Read the docs for `addValidationResult()`. The second param explicitly states that the expected value is a translation code. So you are adding a double translation.
* See `Gdn_Validation->resultsAsArray()` for what you're trying to accomplish, perhaps.

If the client is seeing untranslated error messages, there is a bug in how they are being processed & displayed, not in how they are being collected here.

PR #4182, opened 2016-07-08
Add ability to move a category to another of its parents

### LINKED ISSUE #4157: Add basic category move functionality
In the manage categories page there needs to be a way to move a category to another parent. Ideally, we would have a fancy dropdown of all categories, but for now let's do something simple.

On the manage categories page there should be a "Move" option. This option will launch a popup that will prompt the user for a new parent. The options should be a dropdown list with all of the category's parents.

initvector:
This update allows a user to move a category from its current parent to another of its parents/ancestors from the Manage Categories dashboard area.

Closes #4157

### DIFF
--- a/applications/vanilla/controllers/class.vanillasettingscontroller.php
+++ b/applications/vanilla/controllers/class.vanillasettingscontroller.php
@@ -722,7 +722,10 @@ class VanillaSettingsController extends Gdn_Controller {
      * @param int $categoryID Unique ID for the category to move.
      * @throws Exception if category is not found.
      */
-    public function movecategory($categoryID) {
+    public function moveCategory($categoryID) {
+        // Check permission
+        $this->permission(['Garden.Community.Manage', 'Garden.Settings.Manage'], false);
+
         $category = CategoryModel::categories($categoryID);
 
         if (!$category) {
--- a/applications/vanilla/controllers/class.vanillasettingscontroller.php
+++ b/applications/vanilla/controllers/class.vanillasettingscontroller.php
@@ -716,6 +716,37 @@ class VanillaSettingsController extends Gdn_Controller {
         $this->render();
     }
 
+    /**
+     * Move a category to a different parent.
+     *
+     * @param int $categoryID Unique ID for the category to move.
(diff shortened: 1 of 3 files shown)


applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -716,6 +716,37 @@ public function manageCategories() {
...
     }
 
+    /**
+     * Move a category to a different parent.
+     *
+     * @param int $categoryID Unique ID for the category to move.
+     * @throws Exception if category is not found.
+     */
+    public function movecategory($categoryID) {
+        $category = CategoryModel::categories($categoryID);
```
linc: Perhaps a permission check?

applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -716,6 +716,37 @@ public function manageCategories() {
         $this->render();
     }
 
+    /**
+     * Move a category to a different parent.
+     *
+     * @param int $categoryID Unique ID for the category to move.
+     * @throws Exception if category is not found.
+     */
+    public function movecategory($categoryID) {
```
linc: moveCategory

applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -716,6 +716,37 @@ public function manageCategories() {
...
+        $this->setData('Category', $category);
+
+        $parentCategories = CategoryModel::getAncestors($categoryID);
+        array_pop($parentCategories);
+        if (!empty($parentCategories)) {
+            $this->setData('ParentCategories', array_column($parentCategories, 'Name', 'CategoryID'));
+        }
+
+        if ($this->Form->authenticatedPostBack()) {
+            $this->Form->save();
```
linc: Keep in mind that because of Vanilla's very permissive data handling, you're enabling **any** category data to be set via this endpoint currently. Might consider whitelisting `CategoryID` and `ParentCategoryID`.

PR #4194, opened 2016-07-14
Add basic support for flat category display

### LINKED ISSUE #4153: Create a "flat" method on the categories controller
Create a flat method on the categories controller that will display the current category's immediate children as a single flat, paged list. Even though this is a flat list, the categories should still be set on the "CategoryTree" key on the dat array.

- [ ] When viewing a category with display as = "flat" then this endpoint should be called. In this way you don't actually browse to the flat endpoint, but rather it gets called internally when a category is supposed to display this way.
- [ ] The endpoint should respect the tree/list config setting.
- [ ] When viewing a regular category tree/list you should not see the children of a category if it has a display as of "flat" or "categories". The user should have to click in to both of these.

initvector:
Closes #4153

### DIFF
--- a/applications/vanilla/controllers/class.categoriescontroller.php
+++ b/applications/vanilla/controllers/class.categoriescontroller.php
@@ -163,32 +163,37 @@ class CategoriesController extends VanillaController {
             $this->Description(val('Description', $Category), true);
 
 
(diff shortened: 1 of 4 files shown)


initvector:
Gonna attempt to tackle this in a different way.  Closing.

PR #4198, opened 2016-07-20
Add basic support for flat categories

### LINKED ISSUE #4153: Create a "flat" method on the categories controller

initvector:
Closes #4153 

- [x] Display as "Flat" shout show up in the add/edit category page.
- [x] When viewing a flat category headings should be filtered out.
- [x] You should not be allowed to choose a display as type of heading underneath a flat category. In the case where one sneaks in because of the above then that's when it gets filtered out.
- [x] The view should treat categories as depth 1. Right now they are adding Depth2, Depth-2 CSS classes if the category is at that global depth.

### DIFF
--- a/applications/vanilla/views/categories/flat_all.php
+++ b/applications/vanilla/views/categories/flat_all.php
@@ -94,3 +94,6 @@ if ($ChildCategories != '')
 echo $CatList;
 ?>
 </ul>
+<div class="PageControls Bottom">
+    <?php PagerModule::write(); ?>
+</div>
--- a/applications/vanilla/views/categories/flat_all.php
+++ b/applications/vanilla/views/categories/flat_all.php
@@ -1,0 +1,96 @@
+<?php if (!defined('APPLICATION')) exit();
+if (!function_exists('GetOptions'))
+    include $this->fetchViewLocation('helper_functions', 'categories');
+
+echo '<h1 class="H HomepageTitle">'.$this->data('Title').'</h1>';
+if ($Description = $this->Description()) {
+    echo wrap($Description, 'div', array('class' => 'P PageDescription'));
+}
+$this->fireEvent('AfterPageTitle');
+
+$CatList = '';
+$DoHeadings = c('Vanilla.Categories.DoHeadings');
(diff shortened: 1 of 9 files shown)


initvector:
This branch has now been rebased.

tburry:
I unchecked one item because you need to make an additional server check and not just filter the UI.

initvector:
Added simple controller-based verification of new "Display As" value when editing a category

applications/vanilla/views/categories/flat_all.php
```diff
@@ -0,0 +1,99 @@
...
+$MaxDisplayDepth = c('Vanilla.Categories.MaxDisplayDepth') + $this->data('Category.Depth', 0);
+$ChildCategories = '';
+
+$CategoryTree = $this->data('CategoryTree');
+$Categories = CategoryModel::flattenTree($CategoryTree);
+$this->EventArguments['NumRows'] = $Categories;
+
+echo '<ul class="DataList CategoryList'.($DoHeadings ? ' CategoryListWithHeadings' : '').'">';
+foreach ($Categories as $CategoryRow) {
+    $Category = (object)$CategoryRow;
```
tburry: I'm hoping you can try this with WriteListItem() so it's more of a simple view like the table.

initvector:
I added usage of `writeListItem` to the flat_all view.  Doing this made a lot of the code copied from [applications/vanilla/views/categories/all.php](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/views/categories/all.php) redundant.  The view has been trimmed down and its formatting has been slightly updated.

PR #4201, opened 2016-07-25
Use category URL for category RSS link

austins:
Use `categoryUrl()` instead of `$this->SelfUrl` for the category RSS link to include the domain and avoid having an RSS link with the page number in it. This is consistent with the `addRss()` usage on the `/discussions` index.


linc:
Nice enhancement, thanks.

PR #4215, opened 2016-07-27
Fix data capture when logging a discussion edit

### LINKED ISSUE #3945: Editing a Discussion is not Properly Logged
If you edit a discussion, the discussion details are not saved along with the "Event Log" entry.  A log entry is created.  It includes the user who edited the post.  However, the link to the discussion is broken, because the discussion ID is absent.  There is also no post content present, so there is nothing to diff against.

initvector:
When discussion edits are logged, they are attempting to pass along the original data as an object (`$Stored`), when the formatter used by the log view is expecting an array.

This update alters the logging method for discussion edits to match the [existing method for comment edits](https://github.com/vanilla/vanilla/blob/417f98c55869e86068535ad249e808e46b680093/applications/vanilla/models/class.commentmodel.php#L871).

Closes #3945

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -1936,11 +1936,11 @@ class DiscussionModel extends VanillaModel {
                         Gdn::cache()->remove($CacheKey);
                     }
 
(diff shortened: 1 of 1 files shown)


linc:
Backport to release/1607

PR #4217, opened 2016-07-27
Improve approval notice compatibility with some themes

### LINKED ISSUE #4098: The "spam" view is not rendered in some themes.
Location:
vanilla/views/posts/spam.php
Code:
https://github.com/vanilla/vanilla/blob/e0461ce28ea1f7e7ea167228fa1a4af2dde712b9/applications/vanilla/controllers/class.postcontroller.php#L268

That view is rendered in: 
Baseline, Bootstrap, Bitter Sweet

But not in:
Bootstrap 3, (lots of client theme) ...

The correct behaviour should be to display that view after a post is submitted but unapproved.

initvector:
`post.js` requires there be an element with an ID of `#Content` to display some content from the server, like the "Your post is pending approval" message.  The issue is that some themes do not have this element.  Some themes have instead implemented tagging their content elements with classes like `site-content`.  This update adds support for these additional content element identifiers in `post.js`.

Closes #4098

### DIFF
--- a/applications/vanilla/js/post.js
+++ b/applications/vanilla/js/post.js
@@ -149,7 +149,16 @@ jQuery(document).ready(function($) {
                         // Redirect to the new discussion
                         document.location = json.RedirectUrl;
                     } else {
-                        $('#Content,.page-content,.site-content').html(json.Data);
+                        var contentContainer = $("#Content");
+
+                        if (contentContainer.length === 1) {
+                            contentContainer.html(json.Data);
+                        } else {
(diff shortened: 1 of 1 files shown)


initvector:
This method does make Vanilla more accommodating to non-standard conventions of themes, instead of relying on themes being more compliant with Vanilla's established identifiers.  Although, our other options seem limited (e.g. adding `#Content` to all theme's content element) with addressing this issue while avoiding theming issues.

PR #4218, opened 2016-07-27
Allow query strings to pass through simple routes

tburry:
When a route’s destination does not include a querystring then the requests original querystring will pass through a rewrite rule.

Example:

```
/test -> /discussions
```

Going to /test?page=p2 will rewrite to /discussions?page=p2

However consider this example:

```
/page1 -> /discussions?page=p1
```

Going to /page1?page=p2 will rewrite to /discussions?page=p1

There is also a change in Gdn_Request to make Gdn_Request->path(/foo) automatically strip out the trailing slash. This mirrors the functionality of Gdn_Request->pathAndQuery() and really seems like the correct behavior. However, since this is low-level functionality it does carry some risk and should be carefully reviewed.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -806,7 +806,7 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
         if (!empty($matchRoute)) {
             $dest = $matchRoute['FinalDestination'];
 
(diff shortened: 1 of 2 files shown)


tburry:
Hoping to get @kaecyra's opinion here when convenient.

Issue #4220, opened 2016-07-27
Conversations should validate usernames in URL before tokenizing


linc:
`http://vanilla.dev/messages/add/Please%20call%201-888-888-8888%20for%20more%20information. `

<img width="459" alt="screen shot 2016-07-27 at 6 36 34 pm" src="https://cloud.githubusercontent.com/assets/117672/17195097/e9bc8d16-5429-11e6-880d-c79fcbd3b8d2.png">

PR #4221, opened 2016-07-28
Let class messages add() accept a "Subject" parameter

R-J:
By now only the recipient can be "pre-filled" by adding its name to "/messages/add/recipient". This enhancement allows using "/messages/add/recipient/subject of the message"

### DIFF
--- a/applications/conversations/controllers/class.messagescontroller.php
+++ b/applications/conversations/controllers/class.messagescontroller.php
@@ -54,8 +54,9 @@ class MessagesController extends ConversationsController {
      * @access public
      *
      * @param string $Recipient Username of the recipient.
+     * @param string $Subject Subject of the message.
      */
-    public function add($Recipient = '') {
+    public function add($Recipient = '', $Subject = '') {
         $this->permission('Conversations.Conversations.Add');
         $this->Form->setModel($this->ConversationModel);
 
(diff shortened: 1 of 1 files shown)


linc:
What do you think about this issue and how it impacts this feature request? https://github.com/vanilla/vanilla/issues/4220

R-J:
@linc it doesn't interfere with [this PR](https://github.com/vanilla/vanilla/pull/4288) :wink:

linc:
I mean the broader issue of accepting URL parameters as potentially misleading page text.

R-J:
Sorry, I do not understand/see any issue. In what way would /messages/add/recipient/subject be any different to e.g the mailto link?
http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_mailto

linc:
It's a link having the ability to pre-fill text on the page, which could be mistaken for official UI text. Therefore it could misdirect a user to perform an action they shouldn't.

R-J:
You can find my answer [here](https://vanillaforums.org/post/discussion?Name=I%20do%20not%20see%20a%20problem)  :wink:

R-J:
But without joking: adding text to form is easy like the above example shows. It might be prevented if there is a more restrictive GET and POST separation, but maybe not - I don't know much about that.

If it were a bigger issue it would have been addressed more often and I have  never read anything about any dangers about that.

Having "/messages/add/recipient/subject of the message" looks nicer than "/messages/add/recipient?Subject=subject of the message" but I do not think anybody would really notice the difference

PR #4249, opened 2016-08-01
Avoid including root category in /categories/apiv1list

initvector:
The root category is included in the /categories/apiv1list endpoint.

This update checks for the existence of the root category before the data is delivered and purges it if present.


linc:
Are we pretty confident no one is depending on this behavior?

JulieVanilla:
I know that this was flagged by Harvard a couple of months ago, as this was a new change then.

linc:
You're saying this is *reverting* the unexpected change, and they'll be OK with that?

JulieVanilla:
They are annoyed by the root category showing up, so yes, they would be okay with that.

PR #4255, opened 2016-08-02
Fix asset sorting

tburry:
The dispatcher refactor assigns a lowercased application name to the controller rather than the title cased one. This causes the asset sort arrays to not work. Rather that changing the application that’s assigned to the dispatcher I opted to make the asset sort containers case-insensitive.

This PR also includes some variable renaming to be in line with code standards.

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -874,52 +874,56 @@ class Gdn_Controller extends Gdn_Pluggable {
     /**
      *
      *
-     * @param string $AssetName
+     * @param string $assetName
      */
-    public function getAsset($AssetName) {
-        if (!array_key_exists($AssetName, $this->Assets)) {
+    public function getAsset($assetName) {
+        if (!array_key_exists($assetName, $this->Assets)) {
             return '';
         }
-        if (!is_array($this->Assets[$AssetName])) {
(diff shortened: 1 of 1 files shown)


library/core/class.controller.php
```diff
@@ -874,52 +874,56 @@ public function finalize() {
...
-        } elseif (array_key_exists($this->ModuleSortContainer, $Modules) && array_key_exists($AssetName, $Modules[$this->ModuleSortContainer]))
-            $ModuleSort = $Modules[$this->ModuleSortContainer][$AssetName]; // explicit sort
-        elseif (array_key_exists($this->Application, $Modules) && array_key_exists($AssetName, $Modules[$this->Application]))
-            $ModuleSort = $Modules[$this->Application][$AssetName]; // application default sort
+            $moduleSort = false; // no sort wanted
+        } elseif (isset($modules[$sortContainer][$assetName])) {
+            $moduleSort = $modules[$sortContainer][$assetName]; // explicit sort
+        } elseif (isset($modules[$applicationName][$assetName])) {
+            $moduleSort = $modules[$applicationName][$assetName]; // application default sort
+        }
```
tburry: All of the real code changes are in lines 888 through 898. The rest is just variable renames.

Issue #4275, opened 2016-08-04
The category dropdown does not appear when adding a new discussion to non-DisplayAs:Discussion categories


beckyvb:
To replicate, navigate to a category set to DisplayAs: Categories, click on the new discussion button, and see that you have no way to choose the correct category to add the discussion to. 
Further, you cannot post the discussion on that page, as you're trying to post to a non-discussion category.

PR #4276, opened 2016-08-04
Fix category selector for nested categories in /post

beckyvb:
Show the categories in the category dropdown that are children of the category in the url.
Show the category selector in any category that is not a DisplayAs: Discussion category.

Closes https://github.com/vanilla/vanilla/issues/4275

### DIFF
--- a/applications/vanilla/views/post/discussion.php
+++ b/applications/vanilla/views/post/discussion.php
@@ -20,10 +20,14 @@ if (!$CancelUrl) {
     $this->fireEvent('BeforeFormInputs');
 
     if ($this->ShowCategorySelector === true) {
+        $options = ['Value' => val('CategoryID', $this->Category), 'IncludeNull' => true];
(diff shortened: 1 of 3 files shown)


applications/vanilla/views/post/discussion.php
```diff
@@ -20,10 +20,14 @@
     $this->fireEvent('BeforeFormInputs');
 
     if ($this->ShowCategorySelector === true) {
+        $options = ['Value' => val('CategoryID', $this->Category), 'IncludeNull' => true];
+        if ($this->Context) {
+            $options['Context'] = $this->Context;
```
initvector: Looks like this leads to a notice, generated from the form object:
```
htmlspecialchars() expects parameter 1 to be string, array given
```

initvector:
Merged without CI, because build was hanging (10+ minutes) on start.

PR #4285, opened 2016-08-07
Fix the Gdn_Controller->ResolvedPath compatibility bug

tburry:
This should set the ResolvedPath correctly, with the following caveats:

- The property is set in the dispatcher rather than Gdn_Controller->initialize(). This should be more robust.
- The property is set a bit later due to when the controller method is determine. This could cause issues.
- Plugins with controllers will just use the plugin name rather than “plugins/pluginname” in the resolved path.


tburry:
Fixes vanilla/internal#581

PR #4286, opened 2016-08-08
Use urlencoded user name for /messages/add/...

R-J:
Make the "Message" option in profile options work for all characters.


R-J:
By the way: this does not work with names containing "/" but such user names would already lead to trouble at many other places.

DaazKu:
Thanks

PR #4290, opened 2016-08-08
Add themes' custom.js earlier in the head module!

DaazKu:
We had problems with bootstrap3 theme's popups called from plugins on document ready because the theme's custom.js file was added at the end of the head.

I moved the "patch" into the initialize method which is called right after the controllers declare their assets.


tburry:
I've actually thought a bit on this one and I think moving the custom.js is risky. What we should do is modify bootstrap's custom.js to properly put library stuff outside of the document ready so that it's available earlier.

DaazKu:
This will fix the problem https://github.com/kasperisager/vanilla-bootstrap/pull/194

PR #4293, opened 2016-08-09
Filter out empty get parameters when building the current query string

tburry:
This makes the URL shorter and prevents an empty question mark when URL parameters are null.


tburry:
This solves an issue where the current URL was not always highlighting in the side nav module in the new dashboard.

linc:
Won't this make it impossible to explicitly set a `GET` param to zero?

tburry:
Oh poop. Fixt.

PR #4303, opened 2016-08-10
Revise updating role type based on ConfirmEmailRole config

### LINKED ISSUE #4063: Code to update old ConfirmEmailRole is broken by insert array syntax
Given these circumstances:

* no 'Unconfirmed' role type set in the Roles table
* have the old ''Garden.Registration.ConfirmEmailRole' set in config
* run /utility/structure

Vanilla's code to fix that situation is broken since we made a security fix there.

screen shot 2016-05-31 at 1 05 22 pm

initvector:
If `Garden.Registration.ConfirmEmailRole` is set to a role ID (or IDs) that do not exist in the database, a fatal error is encountered when running the dashboard's structure file.  This is due to...

1. The structure.php attempts to run a `Gdn_SQLDriver::replace` call to [update the role records](https://github.com/vanilla/vanilla/blob/341885917626a70ae8ca585c74e57f85263be219/applications/dashboard/settings/structure.php#L189).
2. `Gdn_SQLDriver::replace` tries to lookup the existing role records to update, finds none, merges our sets+wheres and [passes everything off to `Gdn_SQLDriver::insert`](https://github.com/vanilla/vanilla/blob/341885917626a70ae8ca585c74e57f85263be219/library/database/class.sqldriver.php#L1156).
3. `Gdn_SQLDriver::insert` [passes things off to `Gdn_SQLDriver::set`](https://github.com/vanilla/vanilla/blob/341885917626a70ae8ca585c74e57f85263be219/library/database/class.sqldriver.php#L1071)
4. `Gdn_SQLDriver::set` sees our `RoleID` value is actually an array from our original where conditions and [throws an exception](https://github.com/vanilla/vanilla/blob/341885917626a70ae8ca585c74e57f85263be219/library/database/class.sqldriver.php#L1832), because you can't have an array for a particular field value.

This update alters the query being run, simplifying it to use an update call on records designated by the `Garden.Registration.ConfirmEmailRole` value.

Closes #4063

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -186,12 +186,10 @@ if ($RoleTableExists && $UserRoleExists && $RoleTypeExists) {
     }
 
     if (c('Garden.Registration.ConfirmEmailRole')) {
-        $SQL->replace(
-            'Role',
-            array('Type' => RoleModel::TYPE_UNCONFIRMED),
-            array('RoleID' => $types[RoleModel::TYPE_UNCONFIRMED]),
-            true
-        );
+        $SQL->update('Role')
+            ->set('Type', RoleModel::TYPE_UNCONFIRMED)
+            ->whereIn('RoleID', $types[RoleModel::TYPE_UNCONFIRMED])
+            ->put();
 //      RemoveFromConfig('Garden.Registration.ConfirmEmailRole');
     }
 


linc:
https://github.com/vanilla/vanilla/commit/77a397a8464faf7c57aaf1bd2cbeb6cb1b59c847 :(

PR #4304, opened 2016-08-10
Verify unconfirmed roles before attempting to update them

initvector:
Attempt, the second, after needing to revisit following https://github.com/vanilla/vanilla/pull/4303


initvector:
Need to tweak this, just a bit, to add similar handling for the other couple of configs like this above it.

initvector:
I've expanded the fix to include converting roles from `Garden.Registration.ApplicantRoleID` and
`Garden.Registration.DefaultRoles`, in addition to `Garden.Registration.ConfirmEmailRole`

initvector:
I simplified the routine a little more.  This was possible by removing the legacy role config data from config-defaults.php.

linc:
This looks good to me.

PR #4305, opened 2016-08-10
Optimize Gdn_Request workflow, especially url()

DaazKu:
That function is called a lot! 

Before:
screen shot 2016-08-10 at 5 12 59 pm

After:
screen shot 2016-08-10 at 5 16 36 pm

Cachegrind profiler ^

What I did:

Optimize the `url()` function a bit.
Optimize the environments functions on get.
Removed all internal call to request{X} (This simplifies the call stack!)

Todd want to check this out (with reason)


library/core/class.request.php
```diff
@@ -967,83 +981,72 @@ public function url($path = '', $withDomain = false, $ssl = null) {
...
+                $parts[] = $this->_environmentElement('SCRIPT').'?p=';
+                $query = str_replace('?', '&', $query);
+            }
         }
 
         if ($path == '') {
-            $pathParts = explode('/', $this->path());
-            $pathParts = array_map('rawurlencode', $pathParts);
-            $path = implode('/', $pathParts);
+            $path = $this->path();
```
DaazKu: `path()` already `rawurlencode`

library/core/class.request.php
```diff
@@ -389,19 +395,27 @@ public function isPostBack() {
      * @return int
      * @since 2.1
      */
-    public function port($Port = null) {
-        return $this->_environmentElement('PORT', $Port);
+    public function port($port = null) {
```
tburry: This style of micro-optimization is not something I'm really in favor of.

tburry:
In general:

- Optimize the url() function a bit. **like**
- Optimize the environments functions on get. **micro-optimization, don't like**
- Removed all internal call to request{X} (This simplifies the call stack!). **like**

Can you remove the environment function optimizations and rebase?

DaazKu:
I did micro optimizations because `_environmentElement()` was called thousands of times for essentially "nothing" but I totally understand that you don't like it.

Done!

tburry:
Micro-optimizations only give you a gain when something is called hundreds of thousands of times. Think on that 😝.

Issue #588, opened 2010-10-19
Closed DIscussions are not really closed


vincent-:
Closing just causes the "Write Comment" form to be hidden, but you can still write comments, e.g. if you simply change the "discussion-id" (e.g. using FireBug) when writing a comment in another discussion. I am using Version 2.0.9.

PR #4314, opened 2016-08-12
Keep the current delivery type/method between dispatches

tburry:
When calling Gdn_Dispatcher->dispatch() multiple times the previous delivery type/method should be maintained to support backwards compatibility. In all honestly, this behavior is bad because it necessarily relies on object pollution, but since we have a lot of code depending on it then we have to support it.

This is method is slightly different than the previous method and is not without potential issues as was the previous method so let me explain some things.

The previous method managed to Mr. Magoo its way into working in a sensible way with two pieces of behavior:

1. When dispatching to a .json endpoint then Gdn_Dispatcher would override its own delivery type/method that would then be used for subsequent requests.
2. When re-dispatching the current query string is used even if the new URL contains one which I guess isn't used. This allowed an explicit querystring delivery type/method to be maintained.

Since these were are the only two ways we support overriding the delivery properties dispatching works. However, any future override attempts would reveal these issues pretty quickly. Furthermore, trying to redispatch to a URL with a "?" will lose the querystring.

**The Fix**

The dispatcher now stores the last delivery type/method when dispatching which then becomes the default for subsequent dispatches. A Gdn_Dispatcher->reset() method has been added to clean up the dispatcher if a new environment is desired.

This means that the Gdn_Dispatcher now depends on object pollution which I was trying to get rid of before. I'm really not sure how to get rid of this though. One idea is to add a Gdn_Dispatcher->redispatch() method that will explicitly handle redispatches without object pollution. A lot of obscure code will have to be updated though.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -62,13 +62,13 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
     private $controllerData;
 
     /** @var array Any query string arguments supplied to the controller method. */
-    private $controllerMethodArgs = [];
+    private $controllerMethodArgs;
 
     /** @var string|false The delivery method to set on the controller. */
-    private $deliveryMethod = false;
+    private $deliveryMethod;
 
     /** @var string|false The delivery type to set on the controller. */
-    private $deliveryType = false;
+    private $deliveryType;
 
     /**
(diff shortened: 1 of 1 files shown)


tburry:
Closes #588.

PR #4325, opened 2016-08-17
IndexPhotos is buggy on mobile with table layout.

R-J:
Throughout the plugin the functionality is capsuled in such an if construct:
`if (!$this->hasLayoutTables() || IsMobile()) {`

So if the plugin is used from mobile, every output is made even when table layout is used, which doesn't makes sense at all. But I do not know where this comes from. Maybe that fix isn't correct...

I've deleted that helper function for determining if layout = 'table' since I thought that this wouldn't be helpful if this is the only thing that is checked now.


DaazKu:
Thanks

Issue #4197, opened 2016-07-20
Searching by IP causing a duplication in results


LauraMakesArt:
![image](https://cloud.githubusercontent.com/assets/7501633/17004728/06dc9956-4e9c-11e6-9fa6-e32b2624009e.png)
all the same user, and same IP, just duplicated.

PR #4332, opened 2016-08-18
Never duplicate user search results even if they match twice

linc:
IP search can match per IP address entry. It's foreseeable this will happen for other conditions in the future too, so let's group our results on UserID so we never get duplicates.


linc:
Backport to release/1607

linc:
Closes #4197

PR #4334, opened 2016-08-19
Do not append webRoot again on double URL encoding

DaazKu:
Prevent webroot from being appended 2 times to a path.

Example: `webroot='x'` + `anchor('Link', url('/path/to/something.html))` = `'/x/x/path/to/something.html'`

It should be = to `'/x/path/to/something.html'`

Prevent this https://github.com/vanilla/multisite/issues/36 from happening.

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -19,7 +19,7 @@
  * @method string requestMethod($method = null) Get/Set the Request Method (REQUEST_METHOD).
  * @method string requestHost($uri = null) Get/Set the Request Host (HTTP_HOST).
  * @method string requestFolder($folder = null) Get/Set the Request script's Folder.
- * @method string requestAddress($ip = null) Get/Set the Request IP address (first existing of HTTP_X_ORIGINALLY_FORWARDED_FOR, 
+ * @method string requestAddress($ip = null) Get/Set the Request IP address (first existing of HTTP_X_ORIGINALLY_FORWARDED_FOR,
  *                HTTP_X_CLUSTER_CLIENT_IP, HTTP_CLIENT_IP, HTTP_X_FORWARDED_FOR, REMOTE_ADDR).
  */
 class Gdn_Request {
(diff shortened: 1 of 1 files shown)


DaazKu:
This is a core change. The full impact of this change should be reviewed properly.

tburry:
Your example seems off to me. Since anchor calls url() within itself then anchor(..., url(...)) is not valid.

tburry:
Also, why not write a unit test?

DaazKu:
I stumbled on a bug where it was happening.
Sure it is not supposed to be valid but I feel like double calling `url()` on a path (especially if `url()` returns a relative path) should not end up creating an invalid path.
All in all this will just prevent some errors. 

Maybe we should put a trace in there so that we can be notified that webroot was already present in the path?

I'm gonna write the unit test.

DaazKu:
Another issue related: https://github.com/vanilla/vanilla/pull/4494/files

DaazKu:
I'm gonna need some help for the unit test so I'll wait until Todd is back.

Among other thing `Gargen.RewriteUrls` must be set to `true` and I do not know where to put that.

library/core/class.request.php
```diff
@@ -990,8 +990,12 @@ public function url($path = '', $withDomain = false, $ssl = null) {
             $parts[] = '';
         }
 
-        if ($withDomain !== '/' && $this->webRoot() != '') {
-            $parts[] = $this->webRoot();
+        $webRoot = '';
```
tburry:
```php
$webRoot = $this->webRoot();
if ($withDomain !== '/' && !empty($webRoot)) {
    $parts[] = $this->webRoot();
}

File: library/core/class.request.php [hunk 2]
```diff
@@ -1023,7 +1027,13 @@ public function url($path = '', $withDomain = false, $ssl = null) {
                 }
             }
         }
-        $parts[] = ltrim($path, '/');
+
+        // Prevent the addition of webRoot a second time if the path was already encoded.
+        $path = ltrim($path, '/');
```
tburry: This is going to be a bug. What if I install Vanilla into a directory called "/discussions"?

DaazKu: Well chances are really slim. There is not much to do about that so if it is a deal breaker we should just drop this entirely

library/core/class.request.php
DaazKu: $webRoot could be "0"

DaazKu: Calling webroot is costly. That's why I did it like it. Otherwise I could just keep the code like it was.
#microOptimizationBigNope

DaazKu:
I'm gonna close this since there is no 100% reliable way to determine if the first folder of a path should be treated as the webroot or another folder having the same name.

PR #4340, opened 2016-08-23
Redundantly set the data type in the categories module

initvector:
`Gdn_DataSet` has an issue with its type property that currently requires a hack of setting, then resetting, the data type to properly re-cast its data.

This update adds such a fix to the categories module.

### DIFF
--- a/applications/vanilla/modules/class.categoriesmodule.php
+++ b/applications/vanilla/modules/class.categoriesmodule.php
@@ -53,6 +53,7 @@ class CategoriesModule extends Gdn_Module {
         }
 
         $Data = new Gdn_DataSet($Categories);
+        $Data->DatasetType(DATASET_TYPE_ARRAY);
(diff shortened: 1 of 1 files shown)


DaazKu:
I really feel like this "hack" should have been documented.
[Anyone with common sense will remove the duplicated line.](https://github.com/vanilla/vanilla/commit/76dbfb9208626ca0cc323d9994122347a552bad6)

linc:
Ryan already promised to open an issue about it today and cross-reference. The underlying issues is possibly causing other problems elsewhere already.

initvector:
See https://github.com/vanilla/vanilla/issues/4344

initvector:
This update was reverting to an old, existing "hack" after an issue was discovered following its removal.  It could probably be better written as:

```php
$Data = new Gdn_DataSet($Categories, DATASET_TYPE_ARRAY);
$Data->DatasetType(DATASET_TYPE_OBJECT);
```

PR #4343, opened 2016-08-23
Fix photos' URL for /user/summary

DaazKu:
Serve the profile pictures properly.


linc:
Note this is a really limited change because I don't think anything else uses this endpoint except the WordPress plugin (and whatever else third-parties have built to query against it, which it stands to reason will benefit in the same way).

Issue #4069, opened 2016-06-02
Markdown Quoting needs "" around names with spaces. 


LauraMakesArt:
When quoting a post in markdown, it uses @mentions. It does not render quotes around the name, causing it to incorrectly link the name on forums that allow spaces in their names through SSO.

linc:
This need to be done selectively only to usernames that MUST have quotes around them to function properly as a mention (i.e. contains spaces).

PR #4347, opened 2016-08-24
Fix quotes that are done on a user with spaces in its name.

DaazKu:
Enclose usernames with spaces in-between double quotes when using markdown.


DaazKu:
Fix https://github.com/vanilla/vanilla/issues/4069

Issue #2907, opened 2015-07-26
The Gdn_Request class wrongly urldecodes an already urldecoded path.


mtschirs:
The Gdn_Request class wrongly urldecodes an already urldecoded path.

Example:
- [QUERY_STRING] => p=plugin/simplerss/update/a%252fb
- $_GET['p'] => plugin/simplerss/update/a%2fb
- [URI] => plugin/simplerss/update/a/b

Why is this a problem? Instead of calling the method on the SimpleRSSPlugin Controller `Controller_Update($Sender)` with 'a%2fb' in `$Sender->RequestArgs`, Vanilla returns an HTTP 404 error.

Who is at fault?
The `_loadEnvironment` method of the `Gnd_Request` class in **library/core/class.request.php** sets the URI environment element as follows:

```
$Get =& $_GET;
...
$Path = $Get['p'];
...
$this->RequestURI($Path);
```

The `RequestURI` call is then delegated to the `_environmentElement` method which then does: 

```
case 'URI':
    $Value = !is_null($Value) ? urldecode($Value) : $Value;
```

The problem here is that PHP automatically urldecodes the values in `$_GET` and `RequestURI` also urldecodes the given value, thus effectively twice urldecoding the path.

PR #4352, opened 2016-08-24
Avoid REQUEST_URI double decoding at the request level.

DaazKu:
I played a bit with the request object and found out that you cannot double encode a character in the URL right now.

`/profile/UserWith%252FSlash` "UserWith/Slash"
is still parsed as
`UserWith/Slash` instead of `UserWith%2FSlash` because we rawurldecode the value in `Request::_environmentElement()`.
While this is something that can be "correct" when making internal requests we should not double decode the URI in `Request::_loadEnvironment()`.

The dispatcher has been updated to compensate for this change.


linc:
Does this also fix #2907?

DaazKu:
It should yes.

Issue #4308, opened 2016-08-11
Don't strip question marks and periods from usernames


linc:
* We currently block questions marks and periods by stripping them before building our user URLs.
* We don't necessarily block users from getting those names (ex: SSO) but then their profile is a 404.
* Can we safely remove that old character stripping provision now?

DaazKu:
Completely removing the line that strip `?` and `&` works great if you are not using the p parameter at all.

If you use nginx location like so:
```nginx
    rewrite ^(.*) /index.php?p=$uri last;
```
you will encounter a bug that is fixed by the code below
```php
        // Make sure that the name will not be split if the p parameter is set.
        // Prevent p=/profile/a&b to be translated to $_GET['p'=>'/profile/a?', 'b'=>'']
        $UserName = str_replace('&', '%26', $UserName);
```

linc:
Let's use the encoding then to preserve it for now. I'm not ready to jettison this compatibility today.

PR #4353, opened 2016-08-24
Encode slashes in users's name.

DaazKu:
Depends on https://github.com/vanilla/vanilla/pull/4352
Kind of related to this issue: https://github.com/vanilla/vanilla/issues/4308 (SSO names)


linc:
> Encode slashes in usernames when creating profile links.

PR #4361, opened 2016-08-25
Allow drafts to be orphaned and rework drafts view

### LINKED ISSUE #1922: Deleting a discussion does not update the draft counter
if a draft was saved on a discussion and then the discussion is deleted.  the draft is deleted but the draft counter is not reset.

this may have been filed before.

but it will be reset again properly if you save a draft and delete the draft.

### LINKED ISSUE #4187: Comment draft count persisting after discussion deleted
If you save a comment draft on a discussion, then delete the discussion, your comment/draft disappears but your draft count remains unchanged.

### Steps to reproduce
1. Open a comment on an existing discussion.
2. Click "Save Draft" to save your comment.  Do not post it.
3. Note your "My Drafts" count.
4. Delete the discussion your comment draft is associated with.
5. Compare your current "My Drafts" count against the number you recorded in step 3.  They should be the same.
6. Visit the "My Drafts" page.  You should not see your comment draft displayed.

linc:
This changes the behavior of what happens to drafts when their parent (live) discussion is deleted. Previously, we silently deleted them all and neglected to update draft counters. Now, we let them become "orphans" that bring up the Post Discussion form.

Rationale:

* Drafts are a "private zone" that should not be touched by moderators.
* Moderators cannot have the intent to delete drafts because they cannot ever see them.
* A restored discussion now restores drafts. The previous behavior was permanently destructive.
* Does not interfere with counters.

We also rework the drafts view to operate on arrays and use some nicer logic.

Closes #1922
Closes #4187

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -125,7 +125,7 @@ class PostController extends VanillaController {
             $this->Category = (object)$Category;
             $this->setData('Category', $Category);
             $this->Form->addHidden('CategoryID', $this->Category->CategoryID);
-            if (val('DisplayAs', $this->Category) == 'Discussions') {
+            if (val('DisplayAs', $this->Category) == 'Discussions' || $DraftID) {
                 $this->ShowCategorySelector = false;
             } else {
                 // Get all our subcategories to add to the category if we are in a Header or Categories category.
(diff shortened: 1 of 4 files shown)


DaazKu:
I cannot choose the category in which the comment will be restored as a discussion. I ended up posting my "draft" directly into the "root" category.

linc:
I don't follow. The category dropdown is missing for you?

DaazKu:
Yeah. Since `CategoryID = -1` the dropdown is hidden.

PR #4362, opened 2016-08-25
Add very basic table CSS in UGC areas

### LINKED ISSUE #2139: No CSS for tables in user posts
Users can post tables in the messages using HTML and Markdown. The default theme includes "Reset CSS" which makes the tables dumb-looking. This is a two column, two row table:

table

Remove Reset CSS or include a reasonable style for tables in posts.

linc:
Give a little padding to `td` and `th` elements in `.Message` areas. This is some very minimalist styling that will hopefully make tables usable without blowing out anyone's customizations. My thought here is that padding is the most likely thing that anyone would've added in their theme to fix this, and so that will still override this setting.

Closes #2139


beckyvb:
This needs to be added to scss/legacy/style.scss and then compiled into style.css using the build process, which will also update the css map.

PR #4367, opened 2016-08-26
Fire a beforeAnalyzeRequest event so that plugins can modify the request after rewriting

tburry:


linc:
It was a beautiful angel and you killed it.

PR #4371, opened 2016-08-26
Add filtering to flat category module

initvector:
This update adds a filter/search box to the top of the flat category module.  Text entered here is used to filter the displayed categories by name.


initvector:
The JavaScript used here is placed in global.js, due to modules' seeming inability to include assets only when necessary.  Modules like the one in this PR aren't even instantiated until building the page's body markup is well underway.  Some modules provided by plug-ins do include additional JavaScript, but only because they have the ability to hook into events.  Even then, they don't seem to be included on an as-needed basis.

I'm open to moving the module's JavaScript elsewhere.

PR #4375, opened 2016-08-29
Verify that discussions titles render properly when no word characters are found in them.

### LINKED ISSUE #4196: Disallow Emoji being used as subject

DaazKu:
That one was a bit tricky. Emojis were introduced in the Unicode Standard 6.0.

Some system may or may not render them properly.
I decided to implement a client side solution that check if titles are rendered properly by the browser.

Fix #4196


linc:
Do we even have a reproducible example of this yet?

I'm almost OK with this as a general enhancement to bulletproof our rendering, but as a primary strategy for dealing with potential garbage data I don't like it.

DaazKu:
Yeah. Set the title field to '\x01' using javascript.
Or using an Emoji character as the title.

linc:
Uh. I think you better make a new feature branch and cherry pick or something.

DaazKu:
Holy #$@#.. I'm gonna do a new PR.

PR #4382, opened 2016-08-30
IndexPhotos: no padding on profile discussions index

austins:
The IndexPhotos aren't displayed on the profile discussions index, so padding styles shouldn't be applied there.

### DIFF
--- a/plugins/IndexPhotos/design/indexphotos.css
+++ b/plugins/IndexPhotos/design/indexphotos.css
@@ -7,11 +7,11 @@
     width: 40px;
 }
 
-.Discussion.ItemContent {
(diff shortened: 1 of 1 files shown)


linc:
Is there an applicable class you could use instead?

austins:
@linc, instead of `:not(#dashboard_profile_discussions)`, we can target the Discussions class used on both DiscussionsController and CategoriesController:

~~~
body.Vanilla.Discussions .Discussion.ItemContent {
    padding-left: 48px;
}

body.rtl.Vanilla.Discussions .Discussion.ItemContent {
    padding-right: 48px;
    padding-left: 0;
}
~~~

How about this?

linc:
That'll do, thanks.

PR #4390, opened 2016-09-01
Verify that discussions titles render properly when no word characters are found in them.

### LINKED ISSUE #4196: Disallow Emoji being used as subject

DaazKu:
That one was a bit tricky. Emojis were introduced in the Unicode Standard 6.0.

Some system may or may not render them properly.
I decided to implement a client side solution that check if titles are rendered properly by the browser.

Fix #4196

New PR based on https://github.com/vanilla/vanilla/pull/4375

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -53,6 +53,8 @@ class DiscussionController extends VanillaController {
      * @param int $Page The current page of comments
      */
     public function index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
+        $this->_Definitions['BlankDiscussionTopicText'] = t('Blank Discussion Topic');
+
         // Setup head
         $Session = Gdn::session();
         $this->addJsFile('jquery.autosize.min.js');
@@ -529,7 +531,7 @@ class DiscussionController extends VanillaController {
             if ($Target) {
                 $this->RedirectUrl = url($Target);
             }
-            
+
             $this->jsonTarget('', '', 'Refresh');
         } else {
             if (!$Discussion->Announce) {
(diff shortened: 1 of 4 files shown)


initvector:
I have a couple concerns with this approach:

1. The inconsistency between browsers.  A discussion can potentially have two different titles in the discussion list, depending on what software the user is using to navigate the site.  It could be a string of emojis on one of a user's devices.  It could be "Blank Discussion Topic" on another of that same user's devices.
2. I have no doubt this would work, but it seems like a not-insignificant amount of code for what is a very, very low rate of occurrence.  We have a couple empty discussion title checks (see [#1](https://github.com/vanilla/vanilla/tree/7cd00969dff461bde6e714daee8b92947dbf2625/applications/vanilla/views/discussions/table_functions.php#L61), [#2](https://github.com/vanilla/vanilla/tree/7cd00969dff461bde6e714daee8b92947dbf2625/applications/vanilla/views/discussions/helper_functions.php#L139)).  They could probably be expanded to include a similar check to what you've done here, instead of only checking for empty strings.

I propose we keep this check on the server-side.  We can enforce at least one word character in the title when a discussion is created, moving forward.  We can also update the existing empty title checks to include a condition for verifying at least one word character is present, to handle problematic titles that've already been created.

DaazKu:
We already debated on this outside of github and we will go with a full server side solution but I will still put my arguments here:

I'm totally in favour of enforcing titles to have word characters in them.

I disagree with the inconsistency issue though.
- Titles with emojis (unicode 6.0 support) are already inconsistent.
  - Browser that support them
    - Title1: `This is a title :)`
    - Title2: `:)`
  - Browser that does not
    - Title1: `This is a title `
    - Title2: ` ` <- This is the edge case we want to fix

We currently fix the titles only in the discussion list because it is critical that we can click on them.
This means that replacing Title2 by "Blank Discussion Topic" on the server side will result in inconsistencies as well since browsers with emojis support will display the title as the emoji everywhere but in the discussion list.
A good example of this would be: Click on "Blank Discussion Topic" and then check the discussion title to see that it is not written "Blank Discussion Topic" and it is an emoji.

Instead of having inconsistencies only in the browser lacking the unicode support we end up with inconsistencies everywhere.


As for the `not-insignificant amount of code`. Well.. most of the time only the 2 first lines are executed so..

PR #4392, opened 2016-09-02
Fix auto-detection of incoming data type in Gdn_DataSet

### LINKED ISSUE #4344: Gdn_DataSet and/or CategoriesModule sometimes improperly setting dataset type
This one can be tough to consistently recreate.

`Gdn_DataSet` can explicitly take a dataset type [in its constructor](https://github.com/vanilla/vanilla/blob/e1259d147bfa881b24e97decb0674f30206fa568/library/database/class.dataset.php#L64), it can [try to be guessed](https://github.com/vanilla/vanilla/blob/e1259d147bfa881b24e97decb0674f30206fa568/library/database/class.dataset.php#L67) based on the data or it can be manually set with [`Gdn_DataSet::datasetType`](https://github.com/vanilla/vanilla/blob/e1259d147bfa881b24e97decb0674f30206fa568/library/database/class.dataset.php#L112). The latter will attempt to cast the data to the type (object or array) if it differs from the current type.  A problem arises when the data is arrays, but the dataset type property is set to `DATASET_TYPE_OBJECT`, indicating the data is objects (or vice versa).  You can have a dataset of arrays that you attempt to convert to objects, but the conversion doesn't happen, because the `Gdn_DataSet` instance already thinks it has objects.

One area of Vanilla where this is a problem is [`CategoriesModule::getData`](https://github.com/vanilla/vanilla/blob/e1259d147bfa881b24e97decb0674f30206fa568/applications/vanilla/modules/class.categoriesmodule.php#L36).  Under certain circumstances, the data built in this function can be interpreted by `Gdn_DataSet` to be `DATASET_TYPE_OBJECT`, when it would normally be `DATASET_TYPE_ARRAY`.  `DATASET_TYPE_OBJECT` is the default type for `Gdn_DataSet`.  Since we aren't passing a `$DataSetType` to the constructor from the module, it's very likely the 0-index element is failing the `isset` or `is_array` checks.  The way this has been mitigated in the module has been by first setting explicitly the dataset type to array, then setting the type to object, allowing the data casting to be performed, regardless of the original dataset type.

Steps to recreate:

1. Enable the Reporting2 plug-in.
2. Make sure you have the categories sidebar module enabled and visible.
3. Unfollow the Reported Posts category.
4. You should observe the category list in the categories module is corrupted (all rows with name, counts, etc. missing).

This has only been observed to be an issue in the categories module, but may be a problem elsewhere.

initvector:
[`Gdn_DataSet::__construct`](https://github.com/vanilla/vanilla/blob/8efb18d446977d17bfd5738ec2af0bdff64ef00c/library/database/class.dataset.php#L58) attempts to determine the incoming data type based on the first element of its `$Result` parameter, if no explicit value for `$DataSetType` is provided.  It assumes a numerically-indexed array, starting with 0, when retrieving the first element.  If it cannot locate that element, the data type remains as the class default: `DATASET_TYPE_OBJECT`.  There are occassions (see #4340) where a multi-dimensional array might be passed in as `$Result` with no zero index and no explicitly-defined type parameter.  This leads the instance of `Gdn_DataSet` to assume it is dealing with an array of objects, which can lead to issues.  One such issue is the inability to efficiently invoke `Gdn_DataSet::datasetType` to convert that multi-dimensional array into an array of objects, because the class determines conversion isn't necessary, since it already erroneously flagged as an array of objects.

This update uses [reset](http://php.net/reset) on `$Result` in the `Gdn_DataSet` constructor to obtain the first value, regardless of index.  In addition, it also patches [`CategoriesModule::getData`](https://github.com/vanilla/vanilla/blob/8efb18d446977d17bfd5738ec2af0bdff64ef00c/applications/vanilla/modules/class.categoriesmodule.php#L36) to explicitly define its datatype and avoid having to detect or reset it to begin with.

Closes #4344

### DIFF
--- a/applications/vanilla/modules/class.categoriesmodule.php
+++ b/applications/vanilla/modules/class.categoriesmodule.php
@@ -52,9 +52,8 @@ class CategoriesModule extends Gdn_Module {
             }
         }
 
-        $Data = new Gdn_DataSet($Categories);
-        $Data->DatasetType(DATASET_TYPE_ARRAY);
-        $Data->DatasetType(DATASET_TYPE_OBJECT);
+        $Data = new Gdn_DataSet($Categories, DATASET_TYPE_ARRAY);
+        $Data->datasetType(DATASET_TYPE_OBJECT);
         $this->Data = $Data;
     }
 
--- a/library/database/class.dataset.php
+++ b/library/database/class.dataset.php
@@ -64,7 +64,8 @@ class Gdn_DataSet implements IteratorAggregate, Countable {
         if ($DataSetType !== null) {
             $this->_DatasetType = $DataSetType;
         } elseif ($Result) {
-            if (isset($Result[0]) && is_array($Result[0])) {
+            $firstElement = reset($Result);
+            if (is_array($firstElement)) {
                 $this->_DatasetType = DATASET_TYPE_ARRAY;
             }
         }


linc:
It's funny, as soon as you said "uses `reset()`" I knew the fix was going to be correct. That function wasn't on our radar in the early days and we were much poorer for it.

PR #4395, opened 2016-09-02
Create flat strings that can accommodate gender when being translated.

nervoustwit:
Currently we have certain phrases that we concatenate using sprintf() which reduces duplication in our translation array. This causes problems when, in some languages, the subject of the sentences changes other words in the phrase: 

e.g. sprintf(t('You do not have any %s yet.'), t('notifications'));

In French "You do not have any" could be "Vous n'avez aucun" or "Vous n'avez aucune" based on the value of %s.

This PR would allow us to change the complete phrase.

### DIFF
--- a/applications/conversations/views/messages/all.php
+++ b/applications/conversations/views/messages/all.php
@@ -27,7 +27,7 @@ echo '</div>';
                 include $ViewLocation;
             else:
                 ?>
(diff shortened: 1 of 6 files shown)


nervoustwit:
We will need to update the translations on transifex before deploying this change.

PR #4415, opened 2016-09-07
Changed h2s for h1s for a few edit profile pages

slafleche:
Some of the edit profile pages were using an h2 for the title of the page. I changed them to h1s.


slafleche:
I think this is unlikely to cause issues, because there are other edit profile pages that have an h1. on top of that, the h1s and the old h2s both have the "H" class.

PR #4481, opened 2016-09-15
Add BeforeCheckPreference event to ActivityModel

R-J:
While trying to write a plugin which uses Pushbullet for sending notifications to users, I've found no way to queue activities for users who have neither the Email nor the Notification preference set, based on the rigid `if (!$Popup && !$Email) return` preference check
I needed an event before that check so that I could queue an activity based on my custom preferences ("BeforeCheckPreference"). I would have prefered a "BeforeQueue" event right before the `self::$Queue[...] = ...` but that would have required a logic change in that method and I wanted to be least invasive.

What would it take to ensure this change is in your next release (even if it would be a 2.2.bugfix)?

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -1304,6 +1304,10 @@ class ActivityModel extends Gdn_Model {
             }
         }
 
+        $this->EventArguments['Preference'] = $Preference;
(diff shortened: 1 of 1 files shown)


linc:
> What would it take to ensure this change is in your next release (even if it would be a 2.2.bugfix)?

PRs to those branches. This doesn't merge cleanly to them.

Issue #4476, opened 2016-09-14
Dashboard v3: In the plugin page, saving a form from a modal popup refresh the page


DaazKu:
The page should not refresh since the save is done by ajax.
You have to scroll back to the plugin you were because of that....

PR #4493, opened 2016-09-16
Don't reload page after saving addon settings in a popup

beckyvb:
Add a flag to let the modal know not to refresh after form save and apply it to the addon settings.
Also clean js a bit.

Closes https://github.com/vanilla/vanilla/issues/4476

### DIFF
--- a/applications/dashboard/js/dashboard.js
+++ b/applications/dashboard/js/dashboard.js
@@ -345,9 +345,8 @@ var DashboardModal = (function() {
 
         defaultSettings: {
             httpmethod: 'get',
-            afterSuccess: function(json, sender) {
-                // Called after the confirm url has been loaded via ajax
-            },
+            afterSuccess: function(json, sender) {},
+            reloadPageOnSave: true
         },
 
         id: '',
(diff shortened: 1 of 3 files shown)


applications/dashboard/js/dashboard.js
```diff
@@ -345,9 +345,8 @@ var DashboardModal = (function() {
 
         defaultSettings: {
             httpmethod: 'get',
-            afterSuccess: function(json, sender) {
-                // Called after the confirm url has been loaded via ajax
-            },
+            afterSuccess: function(json, sender) {},
```
DaazKu: You could just use `$.noop` there.

applications/dashboard/js/src/modal.dashboard.js
```diff
@@ -22,9 +22,8 @@ var DashboardModal = (function() {
 
         defaultSettings: {
             httpmethod: 'get',
-            afterSuccess: function(json, sender) {
-                // Called after the confirm url has been loaded via ajax
-            },
+            afterSuccess: function(json, sender) {},
```
DaazKu: You could just use $.noop there.

beckyvb: This gives a developer a better clue of how to override the setting.

DaazKu:
```js
$.noop, // function(json, sender)
```

Save a useless function declaration. It is no big deal though. Just something to keep in mind as a good practice.

PR #4496, opened 2016-09-19
Resolve issues with dashboard permissions

beckyvb:
Ensure permission in for the navigation link is the same as the permission for the page itself.
Ensure that a section's landing page will be a page that a user has permission to see.


linc:
This is a nice enhancement.

I don't think we want the Banning page access restricted away from Community Managers, however.

linc:
Apparently I'm the one who escalated it to admin-only, so I have no idea what I'm talking about today.

PR #4497, opened 2016-09-19
Handle clearing problematic user preferences

beckyvb:
Clear the url from a user's preferences if it's one that redirects them to a 404, 401, etc.
Provide an endpoint to manually clear a user's navigation preferences.

### DIFF
--- a/applications/dashboard/controllers/class.dashboardcontroller.php
+++ b/applications/dashboard/controllers/class.dashboardcontroller.php
@@ -59,23 +59,6 @@ class DashboardController extends Gdn_Controller {
     }
 
     /**
-     * An endpoint to clear all the navigation preferences for a user.
-     *
-     * @param string $userID
-     */
-    public function clearNavigationPreferences($userID = '') {
-        if ($userID && $userID !== Gdn::session()->UserID) {
-            $this->permission('Garden.Users.Edit');
-        }
-
-        $userModel = new UserModel();
-        $userModel->clearNavigationPreferences($userID);
-
-        $this->informMessage(t('Navigation preferences cleared.'));
-        redirect('/');
-    }
-
-    /**
      * Sets a user's preference for dashboard panel nav collapsing. Collapsed groups are stored in an
      * list, by their 'data-key' attribute on the nav-header <a> element.
      *
--- a/applications/dashboard/controllers/class.dashboardcontroller.php
+++ b/applications/dashboard/controllers/class.dashboardcontroller.php
@@ -59,6 +59,23 @@ class DashboardController extends Gdn_Controller {
     }
 
     /**
+     * An endpoint to clear all the navigation preferences for a user.
+     *
+     * @param string $userID
+     */
+    public function clearNavigationPreferences($userID = '') {
+        if ($userID && $userID !== Gdn::session()->UserID) {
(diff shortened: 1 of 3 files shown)


applications/dashboard/controllers/class.homecontroller.php
```diff
@@ -97,6 +99,15 @@ public function fileNotFound() {
     }
 
     /**
+     * Clears the url from the user's navigation preferences.
+     */
+    private function clearNavigationPreferences() {
+        $userModel = new UserModel();
+        $uri = Gdn::request()->getRequestArguments('server')['REQUEST_URI'];
```
linc: `Gdn::request()->path()` ?

applications/dashboard/controllers/class.dashboardcontroller.php
```diff
@@ -59,6 +59,23 @@ public function initialize() {
     }
 
     /**
+     * An endpoint to clear all the navigation preferences for a user.
+     *
+     * @param string $userID
+     */
+    public function clearNavigationPreferences($userID = '') {
+        if ($userID && $userID !== Gdn::session()->UserID) {
```
linc: Wanna check for a `POST` here so we can't CSRF folks.

applications/dashboard/controllers/class.dashboardcontroller.php
```diff
@@ -59,6 +59,23 @@ public function initialize() {
...
+            $this->permission('Garden.Users.Edit');
+        }
+
+        $userModel = new UserModel();
+        $userModel->clearNavigationPreferences($userID);
+
+        $this->informMessage(t('Navigation preferences cleared.'));
+        redirect('/');
```
linc: I'd make this redirect conditional on `DELIVERY_TYPE_ALL` so we don't bounce a menu / ajax request and miss the inform message, which never actually renders as-is. Render blank if you want that to show up.

DaazKu:
> My overriding comment is: Do we really need all the detection of clearSectionNavigationPreference() or can we just blow out the preference completely on any error? Would anyone even notice that lesser granularity?

I would. As soon as we will get used to the dashboard remembering where we were, having it not land you where it was supposed to just because you hit a 404 somewhere will be considered a "bug"

beckyvb:
I agree with @DaazKu. I was going to wipe out all the preferences, but I thought it would be pretty bad user experience.

applications/dashboard/controllers/class.homecontroller.php
beckyvb: That's `dashboard/home/filenotfound` here

DaazKu: Gdn::request()->requestURI();

beckyvb: Same, that's `dashboard/home/filenotfound` here (or whichever endpoint we're redirected to)

DaazKu: I know it is the same but I think it is the intended way of using the Request object.
Otherwhise why not just `$_SERVER['REQUEST_URI'];`. Way simpler.

linc:
I will acquiesce to your additional code complexity at the toll of at least 2 inline comments.

Issue #4391, opened 2016-09-02
Stash function sometimes not working in 2.2.1


rbrahmson:
I have a plugin (DiscussionTopic) that uses the stash function for two functionalities:

1. To store a message that is later displayed in the discussion list
2. To store a string that is later used in DiscussionModel_BeforeGet_Handler as an argument to "$Sender->SQL->Where('d.Topic',$Search);" ($Search is the filtering string).

The first option works as intended.  The second one (in the BeforeGet hook) causes a failure.

I have since bypassed the issue in a new version of the plugin by using cookies instead of the stash function.  The source of the original problematic version is in a the Github history [here](https://github.com/rbrahmson/VanillaPlugins/blob/f3273b6b4e4d636fef738a5e177bc820e25c86db/DiscussionTopic/class.discussiontopic.plugin.php) line 1726.

I include an error screen sent by one of the Vanilla forum users who tried the plugin (he is on 2.2.).

P.S. The version with the error does not conform to the Vanilla coding standard in any way.  The current version in Github (with the cookie-method bypass) has fixed many of the standard coding issues (few more to be fixed in the future) and is therefore more readable.

![stash error](https://cloud.githubusercontent.com/assets/6851541/18204342/3fcd1400-70ea-11e6-86c0-959ee1bc4cb0.png)

R-J:
stash() uses Gdn::sql() to work with the Session table.  Discussion model also uses Gdn::sql(). That way, both work on the same object and when the event is fired in the model, some parts of a query have already been built. This unfinished query gets mixed with the query that is built in the stash() method which obviously has to fail.

@rbrahmson:  whenever you hook into a db event and you want to do any db interactions, this is something to look at. See class UserMetaModels __construct method how this can be avoided.

Since the class Session only has two methods which use the db, adding something like that in the constructor might be too much. Instead using a local clone could be better. I'll make a quick PR...

rbrahmson:
Good advice, but I think that the stash function should be using a new sql object to shield it a SQL query being partially built for a different purpose.  Looking forward to see and learn from your PR.

R-J:
No need for the "but" -  I also think it shouldn't use Gdn:sql(). Since every finished query deletes everything that has been added to the query builder before, there might be some unpredictable misbehaviors.  Imagine you have

~~~
// Do some pagination
// Process RequestArgs for page number
// Calculate limit and offset
// Apply them to the query which should be made
Gdn::sql()->limit($offset, $limit);

[...]

PR #4499, opened 2016-09-20
Make stash() use clone of sql class

R-J:
When using `stash()` while the query builder is not finished with an sql, there would either be errors or false results since the Session class uses the singleton Gdn::sql().

The `stash()` method (and its helper method `_getStashSession()`) are the only two methods using db queries in Session. So I've not set $this->sql in `__construct()` but only in `stash()` since it wouldn't be needed elsewhere.

This solves issue https://github.com/vanilla/vanilla/issues/4391

### DIFF
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -749,7 +749,7 @@ class Gdn_Session {
             // current domain then make the domain work.
             $currentHost = Gdn::request()->host();
             if (!stringEndsWith($currentHost, trim($domain, '.'))) {
-                $Domain = '';
+                $domain = '';
             }
 
             safeCookie($name, $sessionID, $expire, $path, $domain);
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -25,9 +25,6 @@ class Gdn_Session {
 
     /** @var object A User object containing properties relevant to session */
     public $User;
-
-    /** @var Gdn_SQLDriver Contains the sql driver which is set in method stash. */
-    public $sql;
     
     /** @var object Attributes of the current user. */
     protected $_Attributes;
@@ -37,6 +34,9 @@ class Gdn_Session {
 
     /** @var object Preferences of the current user. */
     protected $_Preferences;
+
+    /** @var Gdn_SQLDriver Contains the sql driver which is set in method stash. */
+    protected $sql;
 
     /** @var object The current user's transient key. */
     protected $_TransientKey;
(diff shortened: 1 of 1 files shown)


linc:
If the objective is to avoid pollution, don't you want a local variable rather than a class property?

R-J:
I felt uneasy with that, yes. At first I thought about local variables (it would take two: one in stash and one in _getStashSession) but right after writing the same clone snippet twice I thought it might by better to use a global variable. But I'm really undecided what would be better.

I guess the future would/should be dependency injection and that would, as far as I understood make the class session end up similar to what I did: the sql object would be injected as soon as stash needs it. That's why I did it that way. But I have no real knowledge behind what I'm doing it's mostly best guess...

library/core/class.session.php
```diff
@@ -26,6 +26,9 @@ class Gdn_Session {
     /** @var object A User object containing properties relevant to session */
     public $User;
 
+    /** @var Gdn_SQLDriver Contains the sql driver which is set in method stash. */
+    public $sql;
```
DaazKu: protected please.

library/core/class.session.php
```diff
@@ -656,6 +659,10 @@ public function stash($name = '', $value = '', $unsetOnRetrieve = true) {
             return;
         }
 
+        // Create a fresh copy of the Sql object to avoid pollution.
+        $this->sql = clone Gdn::sql();
+        $this->sql->reset();
```
DaazKu: These 2 lines should be done only one time.
Cloning an object on every `stash()`call is not optimal.

R-J: So would it be better be in __construct?

DaazKu: Late initialization is a good idea so let's do a lazy initialization instead.

```php
if ($this->sql === null) {
...
}
```

library/core/class.session.php
R-J: I hope it's the last thing I have to bother you with before I change my PR: should it be `protected $sql` or should I make it   `protected $_sql`?

DaazKu: As stated by http://docs.vanillaforums.com/developer/contributing/coding-standard/#4-3-properties
No underscore. Never hesitate to ask it is not bothering us at all.

library/core/class.session.php
```diff
@@ -25,7 +25,7 @@ class Gdn_Session {
 
     /** @var object A User object containing properties relevant to session */
     public $User;
-
+
```
DaazKu: Trim this and we are good!

DaazKu:
@tburry @linc Can't because of Travis...

tburry:
Not so fast. Don't make an sql property. Just clone it when needed.

DaazKu:
@R-J Could you remove the $sql property, clone on every call to stash and pass the object to the other function? Thanks.

DaazKu:
Thanks @R-J

PR #4504, opened 2016-09-20
Treat discussion's name with no word character as empty

### LINKED ISSUE #4196: Disallow Emoji being used as subject

DaazKu:
Replace https://github.com/vanilla/vanilla/pull/4390
Fix #4196

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -245,7 +245,7 @@ class PostController extends VanillaController {
                 // Let's be super aggressive and disallow titles with no word characters in them!
                 $hasWordCharacter = preg_match('/\w/u', $Name) === 1;
 
-                if (!$hasWordCharacter || (!$Name != '' && Gdn_Format::text($Name) == '')) {
+                if (!$hasWordCharacter || ($Name != '' && Gdn_Format::text($Name) == '')) {
                     $this->Form->addError(t('You have entered an invalid discussion title'), 'Name');
                 } else {
                     // Trim the name.
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -241,13 +241,11 @@ class PostController extends VanillaController {
                     }
                 }
 
-                // Make sure that the title will not be invisible after rendering
-                if (unicodeRegexSupport()) {
-                    $Name = preg_replace('/^[\pC ]*(.*?)[\pC ]*$/u', '$1', $this->Form->getFormValue('Name', ''));
-                } else {
-                    $Name = trim($this->Form->getFormValue('Name', ''));
-                }
-                if ($Name != '' && Gdn_Format::text($Name) == '') {
+                $Name = trim($this->Form->getFormValue('Name', ''));
+                // Let's be super aggressive and disallow titles with no word characters in them!
+                $hasWordCharacter = preg_match('/\w/u', $Name) === 1;
+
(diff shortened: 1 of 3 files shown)


applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -241,13 +241,11 @@ public function discussion($CategoryUrlCode = '') {
                     }
                 }
 
-                // Make sure that the title will not be invisible after rendering
-                if (unicodeRegexSupport()) {
```
initvector: I think we still want to keep the `unicodeRegexSupport` before using UTF regex.

applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -241,13 +241,11 @@ public function discussion($CategoryUrlCode = '') {
...
-                    $Name = preg_replace('/^[\pC ]*(.*?)[\pC ]*$/u', '$1', $this->Form->getFormValue('Name', ''));
-                } else {
-                    $Name = trim($this->Form->getFormValue('Name', ''));
-                }
-                if ($Name != '' && Gdn_Format::text($Name) == '') {
+                $Name = trim($this->Form->getFormValue('Name', ''));
+                // Let's be super aggressive and disallow titles with no word characters in them!
+                $hasWordCharacter = preg_match('/\w/u', $Name) === 1;
+
+                if (!$hasWordCharacter || (!$Name != '' && Gdn_Format::text($Name) == '')) {
```
initvector: Probably don't want the leading `!` on `$Name`.

applications/vanilla/views/discussions/table_functions.php
```diff
@@ -58,7 +58,8 @@ function writeDiscussionRow($Discussion, $Sender, $Session) {
         $Sender->fireEvent('BeforeDiscussionName');
 
         $DiscussionName = $Discussion->Name;
-        if ($DiscussionName == '') {
+        // If there are no word character detected in the title treat it as if it is blank.
+        if (preg_match('/\w/u', $DiscussionName) !== 1) {
```
initvector: Is using an is-identical comparison that much faster?  I think pretty much everything in Vanilla that uses `preg_match` currently just tests for the result being truthy/falsey. If it's a real benefit, I think it's worth using, going forward.  As it stands, these comparisons stick out with similar conditionals in core and I want to make sure it's for a good reason.

applications/vanilla/controllers/class.postcontroller.php
DaazKu: Indeed

applications/vanilla/controllers/class.postcontroller.php
DaazKu: No need. There is no equivalent to /\w/u if unicode is not supported.
And in that case the u modifier will just be ignored.

applications/vanilla/views/discussions/table_functions.php
DaazKu: Exact match is pretty much always better but I wouldn't call it a real benefit.

initvector:
Seems solid, overall.  I am experiencing one issue:

When I enter a discussion name with emoji and word characters, it passes the test, but saves an empty name in the database.  This leads to a discussion being successfully submitted and saved, but rendering with an empty title.  I believe this is due to Vanilla's encoding for the `GDN_Discussion.Name` field.

DaazKu:
This is indeed a database issue. My guess is that you are using an emojis that needs utf8mb4 instead of utf8.

PR #4509, opened 2016-09-21
Fix activity notifications falsely flagging as sent

### LINKED ISSUE #4397: Fix misleading entries in GDN_Activity
Email for notifications starts life as an entry in GDN_Activity. There is a column called "Emailed" that indicates how the attempt to send the email went. It can be one of several values:

```
    /** Activity status: The activity was added before this system was put in place. */
    const SENT_ARCHIVE = 1;

    /** Activity status: The activity sent just fine. */
    const SENT_OK = 2;

    /** Activity status: The activity is waiting to be sent. */
    const SENT_PENDING = 3;

    /** Activity status: The activity could not be sent. */
    const SENT_FAIL = 4;

    /** Activity status: There was an error sending the activity, but it can be retried. */
    const SENT_ERROR = 5;
```

The calling code executes $Email->send() but **DOES NOT CHECK THE RETURN**, only catches exceptions. SENT_OK is also assumed, and is the default case.

The problem is that email sending can fail or be skipped entirely based on several conditions such as:

* `Garden.Email.Disabled`
* Lack of Garden.Email.View permission
* Any number of events fired inside send()

This needs bulletproofing.

initvector:
This update...

1. Adds a new status for the `Emailed` column of the Activity table: `SENT_SKIPPED`.  This is used to indicate when sending an email has been skipped, instead of an error occurring. 
2. Adds try/catch blocks around all the `Gdn_Email::send` calls I could find that didn't already have them, because that function can (and does) throw exceptions, even before this update.
3. Appeases my IDE by tweaking some docs and default parameter types.

Closes #4397

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -37,6 +37,9 @@ class ActivityModel extends Gdn_Model {
     /** Activity status: There was an error sending the activity, but it can be retried. */
     const SENT_ERROR = 5;
 
+    /** Activity status: The recipient was not eligible for an email notification. */
+    const SENT_SKIPPED = 6;
+
     /** Activity status: Sending is in progress. */
     const SENT_INPROGRESS = 31;
 
@@ -65,6 +68,7 @@ class ActivityModel extends Gdn_Model {
     /**
      * Build basis of common activity SQL query.
      *
+     * @param bool $Join
      * @since 2.0.0
      * @access public
      */
@@ -242,8 +246,8 @@ class ActivityModel extends Gdn_Model {
      * @param array $Where A filter suitable for passing to Gdn_SQLDriver::Where().
      * @param string $orderFields A comma delimited string to order the data.
      * @param string $orderDirection One of **asc** or **desc**.
-     * @param int|false $Limit The database limit.
-     * @param int|false $Offset The database offset.
+     * @param int|bool $Limit The database limit.
+     * @param int|bool $Offset The database offset.
      * @return Gdn_DataSet SQL results.
      */
     public function getWhere($Where = [], $orderFields = '', $orderDirection = '', $Limit = false, $Offset = false) {
@@ -345,7 +349,7 @@ class ActivityModel extends Gdn_Model {
      *
      * Events: BeforeGet, AfterGet.
      *
-     * @param int|false $NotifyUserID Unique ID of user to gather activity for or one of the NOTIFY_* constants in this class.
+     * @param int|bool $NotifyUserID Unique ID of user to gather activity for or one of the NOTIFY_* constants in this class.
      * @param int $Offset Number to skip.
      * @param int $Limit How many to return.
      * @return Gdn_DataSet SQL results.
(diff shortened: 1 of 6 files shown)


applications/dashboard/models/class.invitationmodel.php
```diff
@@ -193,7 +193,18 @@ public function send($InvitationID) {
                 ->setTitle(sprintf(t('Join %s'), $AppTitle));
 
             $Email->setEmailTemplate($emailTemplate);
-            $Email->send();
+
+            try {
+                $Email->send();
+            } catch (phpmailerException $e) {
```
tburry: Since phpmailerException extends Exception you don't need this.

applications/dashboard/models/class.usermodel.php
```diff
@@ -3234,7 +3234,18 @@ public function approve($UserID, $Email) {
                     ->setTitle(t('Membership Approved'));
 
                 $Email->setEmailTemplate($emailTemplate);
-                $Email->send();
+
+                try {
+                    $Email->send();
+                } catch (phpmailerException $e) {
+                    if (debug()) {
```
tburry: Since phpmailerException extends Exception you don't need this.

applications/dashboard/models/class.usermodel.php
```diff
@@ -3896,7 +3907,18 @@ public function sendEmailConfirmationEmail($User = null, $Force = false) {
             ->setButton($url, t('Confirm My Email Address'));
 
         $Email->setEmailTemplate($emailTemplate);
-        $Email->send();
+
+        try {
+            $Email->send();
+        } catch (phpmailerException $e) {
+            if (debug()) {
```
tburry: Since phpmailerException extends Exception you don't need this.

applications/dashboard/models/class.activitymodel.php
```diff
@@ -879,17 +883,24 @@ public function sendNotification($ActivityID, $Story = '', $Force = false) {
                     // Only send if the user is not banned
                     if (!val('Banned', $User)) {
                         $Email->send();
+                        $Emailed = self::SENT_OK;
```
tburry: I think that not sending an email to a banned user means that the result should be skipped. We want to be able to be sure that any email with SENT_OK means an actual email was sent.

initvector:
> If the status is SENT_OK then an email must have been sent and we'll be able to look it up in SendGrid.

`Gdn_Email::send` now only has one return statement at the end of its code, indicating a success.  If any errors in Vanilla's email class or phpMailer are reached before then, they should be throwing exceptions.  Assigning the value of `Activity.Emailed` to `ActivityModel::SENT_OK` should only be happening now if no exceptions were encountered when sending the email.

applications/dashboard/models/class.activitymodel.php
initvector: That should already be the case here:

```php
// Only send if the user is not banned
if (!val('Banned', $User)) {
    $Email->send();
    $Emailed = self::SENT_OK;
} else {
    $Emailed = self::SENT_SKIPPED;
}
```

If `User.Banned` is truthy, it should avoid even trying to send the message and immediately go to a status of `ActivityModel::SENT_SKIPPED`.

tburry: Oh shoot. Misread the if statement.

applications/dashboard/models/class.invitationmodel.php
tburry: Ping.

initvector: :cry: https://github.com/vanilla/vanilla/pull/4597

PR #4524, opened 2016-09-22
Make sure new-school form buttons have a value

tburry:
Old forms output buttons with a value equal to their label. The new buttons should fall back to the same.


tburry:
This fixes a bit of a nasty gotcha where some forms look for the existence of a button value and then it's always empty.

PR #4533, opened 2016-09-24
Dashboard settings update error fix.

Balancer:
Misprint in applications/dashboard/settings/structure.php


linc:
Hello! We have a Contributor License Agreement we need to be signed above (by clicking thru) before we can review this.

PR #4536, opened 2016-09-26
Add an OAuth2 base class with a simple plugin to configure a generic connection

nervoustwit:
Create a simple plugin and settings view to extend the core class for OAuth2 SSO connections.


linc:
That new plugin should have lowercase folder + slug.

The base class still needs clear documentation to tell folks it's strictly for extension in a plugin and not to make an instance of.

plugins/OAuth2/views/settings.php
```diff
@@ -0,0 +1,25 @@
+<h1><?php echo $this->data('Title'); ?></h1>
+
+<div class="PageInfo">
+    <h2><?php echo t('Create Single Sign On integration using the OAuth2 protocol!'); ?></h2>
+
+    <p>
```
linc: I think this settings page is already incorrect for the new dashboard. I also question why we are nesting `p` tags here.

plugins/OAuth2/views/settings.php
```diff
@@ -0,0 +1,25 @@
...
+        echo t('<p>If you haven\'t already, create an SSO application.</p>
+                <p>Once your application is created you will receive a unique Client ID, Client Secret and Domain.</p>
+                <p>You will probably need to provide your SSO application with an allowed callback URL, in part, to validate requests. The callback url for this forum is <code>'.$this->data('redirectUrls').'</code></p>');
+        ?>
+    </p>
+</div>
+<?php
+
+echo $this->Form->open(),
+   $this->Form->errors();
```
linc: We don't use commas in echo statements. Right below this you use separate echos for the same pattern. You can do new echos, or concat when it makes sense.

plugins/OAuth2/class.Oauth2.plugin.php
```diff
@@ -0,0 +1,22 @@
...
+    'SettingsPermission' => 'Garden.Settings.Manage',
+    'MobileFriendly' => true
+);
+
+class OAuth2Plugin extends Gdn_OAuth2 {
+
+    protected $settingsView = 'settings/oauth2';
+
+    public function __construct() {
+        $this
```
linc: This should be 1 line.

plugins/OAuth2/class.Oauth2.plugin.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+$PluginInfo['oauth2'] = array(
```
linc: There are 4 doc blocks missing in this file.

plugins/OAuth2/views/settings.php
```diff
@@ -0,0 +1,25 @@
+<h1><?php echo $this->data('Title'); ?></h1>
+
+<div class="PageInfo">
+    <h2><?php echo t('Create Single Sign On integration using the OAuth2 protocol!'); ?></h2>
```
linc: Don't think a bang is appropriate for basic instructions. This string needs to be put in locales, also. Cross-ref the PR.

plugins/OAuth2/views/settings.php
```diff
@@ -0,0 +1,25 @@
...
+
+    <p>
+        <?php
+        echo t('<p>If you haven\'t already, create an SSO application.</p>
```
linc: This should have a slug of something like `OAuth2 setup instructions text` and likewise be added to the locales repo in the file for longer text strings.

plugins/OAuth2/views/settings.php
```diff
@@ -0,0 +1,25 @@
...
+                <p>Once your application is created you will receive a unique Client ID, Client Secret and Domain.</p>
+                <p>You will probably need to provide your SSO application with an allowed callback URL, in part, to validate requests. The callback url for this forum is <code>'.$this->data('redirectUrls').'</code></p>');
```
linc: You cannot concat data structures into a translation string, that makes no sense. Use `sprintf` on this translation to accommodate it.

library/core/class.oauth2.php
```diff
@@ -256,7 +278,17 @@ public function getProviderKey() {
         return $this->providerKey;
     }
 
-
+    /**
+     * Register a call back function so that multiple plugins can use it as an entry point on SSO
```
linc: Missing period at end of sentence. Should also have a blank comment line after this / before the long description.

library/core/class.oauth2.php
```diff
@@ -230,6 +231,27 @@ public function setGetProfileParams($params) {
...
+     * Valid options are ConnectTimeout, Timeout, Content-Type and Authorization-Header-Message.
+     *
+     * @return array
+     */
+    public function getAccessTokenRequestOptions() {
+        return [];
+    }
+
+
+    /** Allow child classes to pass different options to the Profile request API call.
```
linc: Summary doesn't go on the same line as the `/**`.

PR #4549, opened 2016-09-28
Replace Hide Category with Mute Category

### LINKED ISSUE #4394: Clean up "Hide Category" functionality and UX
Currently, this feature is pretty clunky. Hiding categories removes them from the category list, which means there's no easy way to get them back.

Lincoln suggests making the category list ignore the setting, while retaining its effects in /recent to provide a filter. This would mean we'd need to rename the button to reflect the new effect.

initvector:
This update removes the less-than-intuitive "Hide Category" feature and replaces it with "Mute Category".  Muting a category will still display it in the category list (unlike hiding it), but filter its discussions out of the recent discussions page.  Since we're removing this functionality, `CategoryFollowToggleModule` is being deprecated.  The class still exists, but should not longer render any code when included.

Existing fields, functions and variable naming (following, unfollow, watching, etc.) remain unchanged.  Existing filtering for the recent discussions page remains the same as implemented in [`DiscussionModel::getWhere`](https://github.com/vanilla/vanilla/blob/c5616600b41438fc8de53eb034e7c74ea882cdf7/applications/vanilla/models/class.discussionmodel.php#L601)

Closes #4394

### DIFF
--- a/applications/vanilla/controllers/class.categoriescontroller.php
+++ b/applications/vanilla/controllers/class.categoriescontroller.php
@@ -154,10 +154,6 @@ class CategoriesController extends VanillaController {
             $this->CategoryModel->joinRecent($categoryTree);
         }
 
-        if ($watching && $this->CategoryModel->Watching) {
-            $categoryTree = $this->CategoryModel->filterFollowing($categoryTree);
-        }
-
         return $categoryTree;
     }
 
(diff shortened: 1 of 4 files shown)


linc:
Do we have Mute and Unmute in the locales repo? I don't see it.

initvector:
They're sitting in a PR now: https://github.com/vanilla/locales/pull/16

Issue #4538, opened 2016-09-26
Cannot edit photo thumbnail on profile


initvector:
When `Garden.Profile.ShowActivities` is false, it internally redirects `/profile` views to `/profile/discussions`.  `/profile/discussions` does not contain the necessary code to add `cropimage.js` and `cropimage.css`

See: [applications/dashboard/controllers/class.profilecontroller.php:490](https://github.com/vanilla/vanilla/blob/9af22389edf8a89e2415c0171eef88a3f1885abb/applications/dashboard/controllers/class.profilecontroller.php#L490)
See: [applications/dashboard/controllers/class.profilecontroller.php:477](https://github.com/vanilla/vanilla/blob/9af22389edf8a89e2415c0171eef88a3f1885abb/applications/dashboard/controllers/class.profilecontroller.php#L477)
See: [applications/vanilla/settings/class.hooks.php:587](https://github.com/vanilla/vanilla/blob/9af22389edf8a89e2415c0171eef88a3f1885abb/applications/vanilla/settings/class.hooks.php#L587)

DaazKu:
Thanks for those details. Easy fix.

PR #4553, opened 2016-09-29
Fix avatar cropping in profile

DaazKu:
Add missing js files.

Fix https://github.com/vanilla/vanilla/issues/4538.


initvector:
Fantastique.

Issue #4298, opened 2016-08-10
Tagging box "trap"


matthewbienz:
Add ability for keyboard user to be able to add tags to a post, and then submit the post. 

Keyboard users gets stuck in the tagging box. Normally pressing tab moves the user to the next element, but that's not possible here.

linc:
It's also been reported the 'Enter' button submits the discussion when the tagging box is active. This should also be investigated and remedied as part of this.

DaazKu:
I could not reproduce the 'Enter' problem. The code also prevents it so..

PR #4554, opened 2016-09-29
[Tagging] Allow to tab out of the tokenInput field

DaazKu:
Fix https://github.com/vanilla/vanilla/issues/4298


linc:
Nice find.

Issue #4402, opened 2016-09-02
Dashboard v3: Import setting page


beckyvb:
Style the import settings page.

Issue #4401, opened 2016-09-02
Dashboard v3: import's choose file button does nothing.


DaazKu:
When you click on it nothing happens.
That makes me sad :(

PR #4582, opened 2016-10-03
Fix styling of import settings

beckyvb:
Functionality is fixed with this PR, but form styling relies on https://github.com/vanilla/vanilla/pull/4539 The .progress css class is not styled yet but will be in with another PR.

Closes https://github.com/vanilla/vanilla/issues/4402
Closes https://github.com/vanilla/vanilla/issues/4401


DaazKu:
The steps do not work properly. I can give you an export file if you want.

screen shot 2016-10-03 at 3 46 06 pm

DaazKu:
progress will be added in another PR.

PR #4588, opened 2016-10-03
Add container-interop and the Garden namespace to composer.json

tburry:
This is a package we’ll be using shortly.


linc:
Already conflicted.

tburry:
Resolved.

PR #4592, opened 2016-10-04
Decode IP field data in Gdn_Controller::renderData

### LINKED ISSUE #4590: Binary IP addresses need to be formatted for API requests
Now that IP addresses are formatted in binary, they need to be properly formatted when returning an API result. Calling the format IP function on resulting fields will be sufficient. I'd do this in `Gdn_Controller->renderData()`.

This is actually a high priority bug as binary IP addresses completely kills the JSON. What I'd like to see with this pull request is:

- [x] Reproduce the error on localhost.
- [x] Try getting JSON_PARTIAL_OUTPUT_ON_ERROR to work with a fallback for PHP 5.4. This will protect us in case we miss something.
- [x] Add the appropriate IP address encoding.
- [x] When JSON encoding a result, if that result is an error (false) then ouput an JSON encoded error about the encoding rather than a white screen of death. Note that there are some good functions for this in php 5.5, but we'll need to conditionally use them because we must support php 5.4.

initvector:
This update addresses issues with binary values causing JSON encoding to fail in `Gdn_Controller::RenderData`.  It does this by recursively iterating through the `$Data` array and running `ipDecode` against fields with recognizable names (e.g. *IPAddress, *IPAddresses).

*Note: Elements in the `$Data` array can be other arrays ___or objects___, making this not an ideal candidate for `array_walk`/`array_walk_recursive`.  These functions skip over iterating through object properties.  Also, with `array_walk_recursive`, any key that holds an array will not be passed to the callback function. This makes it tough to handle values like "AllIPAddresses", which is an array, itself.*

Closes #4590

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -1470,6 +1470,24 @@ class Gdn_Controller extends Gdn_Pluggable {
 
         $this->sendHeaders();
 
+        // Recursively walk the data, decoding any IP address fields.
+        $ipWalk = function(&$inputValues) use (&$ipWalk) {
+            if (is_array($inputValues) || is_object($inputValues)) {
+                foreach ($inputValues as $key => &$val) {
+                    if (is_string($val) && stringEndsWith($key, 'IPAddress', true)) {
+                        $val = ipDecode($val);
+                    } elseif (is_array($val) && stringEndsWith($key, 'IPAddresses', true)) {
+                        array_walk($val, function(&$ip, $i) {
+                            $ip = ipDecode($ip);
+                        });
+                    } elseif (is_array($val) || is_object($val)) {
+                        $ipWalk($val);
+                    }
+                }
+            }
+        };
+        $ipWalk($Data);
+
         // Check for a special view.
         $ViewLocation = $this->fetchViewLocation(($this->View ? $this->View : $this->RequestMethod).'_'.strtolower($this->deliveryMethod()), false, false, false);
         if (file_exists($ViewLocation)) {
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -1089,56 +1089,6 @@ class Gdn_Controller extends Gdn_Pluggable {
     }
 
     /**
-     * Encode a value as JSON or throw an exception on error.
-     *
-     * @param mixed $value
-     * @param int|null $options
-     * @return string
-     * @throws Exception
-     */
-    private function jsonEncode($value, $options = null) {
-        $advanced = (PHP_VERSION_ID >= 50500);
-
-        if ($options === null) {
-            if ($advanced) {
-                $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR;
-            } else {
-                $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES;
-            }
-        }
-
-        $encoded = json_encode($value, $options);
-
-        if (json_last_error() !== JSON_ERROR_NONE) {
-            if ($advanced) {
-                switch (json_last_error()) {
-                    case JSON_ERROR_RECURSION:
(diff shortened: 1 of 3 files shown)


library/core/class.controller.php
```diff
@@ -1470,6 +1470,24 @@ public function renderData($Data = null) {
...
+                            $ip = ipDecode($ip);
+                        });
+                    } elseif (is_array($val) || is_object($val)) {
+                        $ipWalk($val);
+                    }
+                }
+            }
+        };
+        $ipWalk($Data);
+
```
tburry: My preference would be for this to be put into a function/method that could be unit tested.

library/core/class.controller.php
```diff
@@ -1074,6 +1074,28 @@ public function json($Key, $Value = null) {
...
+     */
+    private function jsonEncode($value, $options = null) {
+        if ($options === null) {
+            $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES;
+        }
+
+        $encoded = json_encode($value, $options);
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw new Gdn_UserException('A JSON encoding error has occurred');
```
tburry: - This probably isn't a Gdn_UserException(). It's more of a fatal Exception in nature.
- You are missing an HTTP compatible error code (probably a 5xx error).
- We also usually make messages sentences with periods.
- You can conditionally show a more useful message on newer versions of PHP.

library/core/class.controller.php
```diff
@@ -1074,6 +1074,28 @@ public function json($Key, $Value = null) {
...
+     * Encode a value as JSON or throw an exception on error.
+     *
+     * @param mixed $value
+     * @param int|null $options
+     * @return string
+     * @throws Gdn_UserException
+     */
+    private function jsonEncode($value, $options = null) {
+        if ($options === null) {
+            $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES;
```
tburry: Conditionally add supported options from newer PHP (see original issue).

library/core/class.controller.php
```diff
@@ -1074,6 +1074,28 @@ public function json($Key, $Value = null) {
...
 
     /**
+     * Encode a value as JSON or throw an exception on error.
+     *
+     * @param mixed $value
+     * @param int|null $options
+     * @return string
+     * @throws Gdn_UserException
+     */
+    private function jsonEncode($value, $options = null) {
```
tburry: Make a global function and unit test.

library/core/functions.general.php
```diff
@@ -3846,3 +3863,22 @@ function urlMatch($pattern, $url) {
...
+
+if (!function_exists('walkAll')) {
+    /**
+     * Recursively walk through all array elements or object properties.
+     *
+     * @param array|object $input
+     * @param callable $callback
+     * @param string $parent Array key or object property to indicate parent node.
+     */
+    function walkAll(&$input, $callback, $parent = null) {
```
tburry: - The standard PHP practice puts the word "recursive" at the end of these such function names.
- I'm not a huge fan of making the $parent parameter something that the original caller can pass. Perhaps an inner closure that exposes it is better.
- Cycles can create infinite recursion. Sad.

library/core/functions.general.php
```diff
@@ -1767,6 +1767,23 @@ function ipDecode($packedIP) {
     }
 }
 
+if (!function_exists('ipDecodeWalk')) {
+    /**
+     * Recursively walk through the data, decoding any IP address fields.
+     *
+     * @param array|object $data
+     */
+    function ipDecodeWalk(&$data) {
```
tburry: ipDecodeRecursive. This is probably a method that can be left on the controller.

initvector:
I think everything has been addressed.

I added a test for `jsonEncodeChecked`, which basically makes sure it's throwing exceptions to indicate a failure to encode data, instead of silently returning `false`/`null`.

Issue #3895, opened 2016-05-08
Change PHPMailer to composer


tburry:
Change the included PHPMailer to a composer dependency. This is not quite a simple swap over as we had to modify PHPMailer when we originally included. Hopefully the latest PHPMailer doesn't have this issue, but it will have to be investigated.

We added an extra error code level to PHPMailer to differentiate between errors resulting from a malformed email and errors resulting from a server error while sending. The latter means we want to retry the send which is why we need the differentiation. I think the relevant code is [here](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/models/class.activitymodel.php#L880), but there may be other examples.

Looking at the latest PHPMailer the code is certainly different and we may be able to get away with an unmodified version. You'll have to research that.

tburry:
Closes #2207.

DaazKu:
Here is the diff:

```diff
diff --git a/class.phpmailer.php b/class.phpmailer.php
index 602f121..f0bfe0f 100644
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -34,7 +34,7 @@
  * @author Andy Prevost
  * @author Marcus Bointon
  * @copyright 2004 - 2009 Andy Prevost
- * @version $Id$
+ * @version $Id: class.phpmailer.php 447 2009-05-25 01:36:38Z codeworxtech $
  * @license http://www.gnu.org/copyleft/lesser.html GNU Lesser General Public License
  */
 
@@ -330,6 +330,7 @@ class PHPMailer {
   const STOP_MESSAGE  = 0; // message only, continue processing
   const STOP_CONTINUE = 1; // message?, likely ok to continue processing
   const STOP_CRITICAL = 2; // message, plus full stop, critical error reached
+  const STOP_SERVER = 5; // message is fine, but the server din't accept the message.
 
   /////////////////////////////////////////////////
   // METHODS, VARIABLES
@@ -604,7 +605,7 @@ class PHPMailer {
     if ($this->SingleTo === true) {
       foreach ($this->SingleToArray as $key => $val) {
         if(!@$mail = popen($sendmail, 'w')) {
-          throw new phpmailerException($this->Lang('execute') . $this->Sendmail, self::STOP_CRITICAL);
+          throw new phpmailerException($this->Lang('execute') . $this->Sendmail, self::STOP_SERVER);
         }
         fputs($mail, "To: " . $val . "\n");
         fputs($mail, $header);
[...]

PR #4603, opened 2016-10-06
Update PHPMailer and extend it to add some functionalities.

DaazKu:
Update PHPMailer to use composer and create a shim class that provide the custom functionalities that we added in our modified version of PHPMailer.

I've decided to completely drop the STOP_SERVER constant since it is not used in our base code anymore.
The risk of it being used externally is extremely low and keeping it would mean we would have to redefine some functions which it a really bad idea. I'm also not declaring it so that if anyone used it they will know real fast that it does not exist anymore.

Fix: https://github.com/vanilla/vanilla/issues/3895


initvector:
Code looks good.  I tested several mail messages and functionality seems solid.

I'd like confirmation that it's okay to axe `STOP_SERVER`.  Carrying it over was a large part of the previous issue's thread and the conclusion there, as of June, seemed to be it should be staying.

tburry:
We need to differentiate between:

1. An error that occurred because the email was malformed in some way. These kinds of emails cannot be resent.
2. An error that occurred because of network issues or latency. These kinds of emails can be retried.

I'm pretty sure that the STOP_SERVER was for the second category.

DaazKu:
We are not checking for that ANYWHERE in our code but I added a function that let us do that.

tburry:
We are checking for it, or rather the absence of STOP_CRITICAL in the activity model. The STOP_SERVER then gets put as the result status in the activity table.

DaazKu:
Thanks for clarifying that up!

DaazKu:
Good to go.

PR #4614, opened 2016-10-07
Fallback to manual increment in memcached if auto-initializing fails

initvector:
`Gdn_Memcached::increment` takes an optional initial value, which it uses to initialize a cache key, if it doesn't already exist.  It passes this value on down to [Memcached::increment](http://php.net/manual/en/memcached.increment.php), which requires the usage of the binary protocol when it receives a value for its `$initial_value` parameter.  `Gdn_Memcached::increment` does some checking to determine if binary protocol is available before passing that parameter.  However, if for whatever reason, that check fails, the initial value just isn't set.

This update alters the switch structure in `Gdn_Memcached::increment` to try the manual initialization of the key value, if the auto-initialization method fails.

I'm also nixing a call to `self::clearCache()`, which shouldn't be necessary with this fix.

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -2036,8 +2036,6 @@ class CategoryModel extends Gdn_Model {
         if ($rebuild) {
             $this->rebuildTree(true);
         }
-
-        self::clearCache();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


initvector:
Closing in favor of https://github.com/vanilla/vanilla/pull/4619

PR #4624, opened 2016-10-12
Update packed IP address handling in BanModel

initvector:
This update augments ban checking/enforcement in `BanModel` by updating its ability to handle packed IP addresses.  IP comparisons in PHP are done with `ipDecode`.  IP comparisons in MySQL make use of `inet6_ntoa`, which is available since [MySQL 5.6.3 (released October 3, 2011)](https://dev.mysql.com/doc/relnotes/mysql/5.6/en/news-5-6-3.html)

### DIFF
--- a/applications/dashboard/models/class.banmodel.php
+++ b/applications/dashboard/models/class.banmodel.php
@@ -95,12 +95,21 @@ class BanModel extends Gdn_Model {
                 $AllBans[$NewBan['BanID']] = $NewBan;
             }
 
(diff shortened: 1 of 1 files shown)


initvector:
Queries using the `inet6_ntoa` function are now wrapped in try/catch blocks.  If MySQL encounters an error when executing these queries, Vanilla will fail gracefully and log an error message.

initvector:
The log message will say something like:

> FUNCTION [database name].inet6_ntoa does not exist

PR #4627, opened 2016-10-12
Fix MySQL version get

linc:
Guess no one's ever cared if this worked, because it returns null currently.


linc:
I've already backported to 2.3.

Issue #4442, opened 2016-09-09
Quote button shows when users don't have posting permissions


matthewbienz:
Users will see the quote option, as the plugin is separate from Reactions

Siddy00:
Hi there.

One of the users of our forum noticed the same issue on version 1.6.10 of the Quotes plugin. We believe we fixed this by making the following changes to the function addQuoteButton (in class.quotes.plugin.php):

Before
```
        if (!Gdn::session()->UserID) {
            return;
        }
```

After
```
        $Session = Gdn::session();

        if (!$Session->UserID) {
            return;
        }

        if (!isset($Args['Discussion']))
            return;

        $CanComment = $Session->checkPermission('Vanilla.Comments.Add', true, 'Category', $Args['Discussion']->PermissionCategoryID);
        if (!$CanComment)
            return;
```

Hope this helps :)

PR #4644, opened 2016-10-14
Make sure to show the quote button only if you have permission to post

DaazKu:
fix: https://github.com/vanilla/vanilla/issues/4442


matthewbienz:
Is this general? Quote will appear if you have the permission to post anywhere, or it'll only appear in categories for which you have permission to post?

DaazKu:
You need to have the 'Vanilla.Comments.Add' permission.

beckyvb:
This only fixed the case where a category doesn't have custom permissions. If a category has custom permissions and the user does not have Comments.Add permission for that category, the user can still see the Quote button.

PR #4649, opened 2016-10-16
Avoid auto-scrolling to top when in comment embed

### LINKED ISSUE #4550: Vanilla Comment embed should not be scrolled to when loaded
Since https://github.com/vanilla/vanilla/pull/4277 we scroll directly to the embedded comments when they load.
We need not to scroll when it is a comment embed.

initvector:
This update avoids automatically scrolling to the top of an embedded site when it's a comment embed.

Closes #4550


linc:
release/1609

PR #4650, opened 2016-10-16
Default to secure cookies if ForceSSL and behind HTTPS

### LINKED ISSUE #4552: Secure Flag on Identity Cookies
If a request is being served over HTTPS on a site with `Garden.ForceSSL` enabled, there isn't a good reason we should not be toggling on the secure flag for the identity cookies Vanilla serves.

I propose the following solution:

If its `$secure` parameter is identical to `null`, [`safeCookie`](https://github.com/vanilla/vanilla/blob/c5616600b41438fc8de53eb034e7c74ea882cdf7/library/core/functions.compatibility.php#L591) should check for a truthy value of `Garden.ForceSSL`.  If enabled, verify the current request is over HTTPS.  If both are true, pass `true` as the `$secure` parameter of [`setcookie`](http://php.net/setcookie).  If `safeCookie` has `false` explicitly passed as its `$secure` parameter, this behavior should be skipped.

initvector:
This update alters `safeCookie` defaults to set the secure flag on cookies if the following conditions are met:

1. The `$secure` parameter is `null`. Previous boolean values will continue to work as they have.
1. `Garden.ForceSSL` is enabled.
1. The current request is served via HTTPS.

Closes #4552

### DIFF
--- a/library/core/functions.compatibility.php
+++ b/library/core/functions.compatibility.php
@@ -585,16 +585,20 @@ if (!function_exists('safeCookie')) {
      * @param integer $expire
      * @param string $path
      * @param string $domain
-     * @param boolean $secure
+     * @param boolean|null $secure
      * @param boolean $httponly
      */
-    function safeCookie($name, $value = null, $expire = 0, $path = null, $domain = null, $secure = false, $httponly = false) {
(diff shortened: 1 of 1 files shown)


initvector:
Should be backported to `release/1609`

PR #4653, opened 2016-10-17
Enhance isUrl() to also identify data urls as urls

R-J:
I've stumbled upon this when I was trying to write a plugin that generates a SVG avatar. Using userPhotoDefaultUrl() to return a data url does not work, since isUrl() doesn't recognize this format and prepends the upload path before the data url.

The isUrl() function can only be overwritten in the bootstrap.before.php which is quite uncomfortable for plugins.

I'm not sure if this would cause any problems anywhere, but as far as I know Vanilla doesn't use inline images.

But in the profile the isUrl() is sometimes used to determine if the profile picture is a "$RemotePhoto". This would give a false positive, though...

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -1884,7 +1884,9 @@ if (!function_exists('isTimestamp')) {
 
 if (!function_exists('isUrl')) {
     /**
-     * Determine whether or not a string is a url in the form http://, https://, or //.
+     * Determine whether or not a string is a url.
+     *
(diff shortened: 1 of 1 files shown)


DaazKu:
The `data:` scheme is an URI not an URL.

But I'm closing this PR because that change is too dangerous anyway.
By calling isURL you expect the string to start by `http` and changing that will probably break a lot of code. Supporting `data:` is a nice enhancement but I do not feel that is it the way to to it.

PR #4680, opened 2016-10-18
Fire "AllowedDiscussionTypes" as an event of the CategoryModel

R-J:


linc:
Assigning to @initvector to see if he thinks this will break anything. Looks like our only internal plugin uses `base_` to call this event which was a wise choice, so this should be fine.

initvector:
Seems safe, sane, simple, solid.  Super!

The hook has [only been around since spring of this year](https://github.com/vanilla/vanilla/commit/e2d66cce308c5a2e24ffacc72ada7a2c2727eeb0), so I don't think it's had a chance to worm its way into too much. :suspect:

PR #4685, opened 2016-10-18
Reinstate Theme Preview

beckyvb:
Reinstates Theme Preview for both mobile and desktop themes.

Functionality is different than before, in that you can preview both a mobile and desktop theme simultaneously. Caveats are that themehooks and theme view overrides cannot be previewed.

Also proposes new theme property 'AllowPreview', which can be set to false for themes that absolutely can't be previewed in this context. Lithe would be one of these theme, since it requires enabling plugins to work.

Closes https://github.com/vanilla/vanilla/issues/4227


linc:
Don't we need an accompanying PR for Lithe then too?

beckyvb:
Indeed! Voilà: https://github.com/vanilla/lithe/pull/55

PR #4696, opened 2016-10-20
Eliminate as much double translations as possible

R-J:
I used grep on the sources and had a closer look at each addError(),
applyRule(), checkBox(), button(), radio(), etc.
I've killed all useless t() calls for them and, when in doubt, I prefixed
some of the inline sprintf() translations with '@' so that they will not
be translated twice.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -22,9 +22,6 @@ class EntryController extends Gdn_Controller {
     /** @var UserModel */
     public $UserModel;
 
-    /** @var string Reusable username requirement error message. */
-    public $UsernameError = '';
-
     /** @var string Place to store DeliveryType. */
     protected $_RealDeliveryType;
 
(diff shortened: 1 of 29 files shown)


R-J:
Phew, git is killing me. I think I try to kill that branch and create a new one with all fixes included instead of polluting this one further on. I'm not sure how to do this but  I guess I can find it out somehow.

initvector:
The safest move might be using [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) to move your relevant commits to the new branch.

In the future, if you want something cleaner, but a little riskier, you can always [`rebase`](https://git-scm.com/docs/git-rebase) and force-push your updates to overwrite the branch's history.  I wouldn't recommend you do it for this branch, now, since you've added several merge-conflict commits that would get lumped in.

Since this seems to have gone off the rails a little, I'm gonna close the PR.  When you get your new branch together and open a new PR, @-mention me and I'll pick it up from there!

Issue #3796, opened 2016-04-27
Import process should clean unused roles relation


DaazKu:
When importing an export we should clean UserRole relation that are invalids.

PR #4712, opened 2016-10-21
Clean invalid relation between users and roles

DaazKu:
Fix https://github.com/vanilla/vanilla/issues/3796

### DIFF
--- a/applications/dashboard/models/class.rolemodel.php
+++ b/applications/dashboard/models/class.rolemodel.php
@@ -783,4 +783,19 @@ class RoleModel extends Gdn_Model {
 
         return $formattedRoles;
     }
+
+    /**
+     * Enforce integrity between users and roles.
+     */
+    public static function cleanUserRoles() {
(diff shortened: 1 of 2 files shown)


applications/dashboard/models/class.rolemodel.php
```diff
@@ -783,4 +783,19 @@ public function getPublicUserRoles($userID, $field = "Name") {
...
+     */
+    public static function cleanUserRoles() {
+        $px = Gdn::database()->DatabasePrefix;
+        Gdn::sql()->query("
+            delete ur
+            from {$px}UserRole as ur
+                left join {$px}Role as r on r.RoleID = ur.RoleID
+                left join {$px}User as u on u.UserID = ur.UserID
+            where r.RoleID is null
+                or u.UserId is null
```
DaazKu: God I hate that Mac is case insensitive...

linc:
This might be something to keep in the back of your mind as a concern on a very large import.

Issue #4719, opened 2016-10-24
Twitter requires twitter:card meta tag to process OpenGraph meta data


MelVanilla:
Twitter docs: https://dev.twitter.com/cards/types/summary-large-image
Image, description, and title will use OG fallbacks: https://dev.twitter.com/cards/markup
A useful tool for evaluating finished product: http://socialdebug.com/

We probably only need to specify 1 tag here to get it working.

linc:
This is a blocker for marketing. I've amended the description with additional information. We can improve our Twitter integration in the future to fully take advantage of their markup. This issue is to make Twitter cards work for forum discussion & comment links only (min necessary).

DaazKu:
https://cards-dev.twitter.com/validator

PR #4722, opened 2016-10-24
Add meta tags for Twitter cards

DaazKu:
Fix https://github.com/vanilla/vanilla/issues/4719

### DIFF
--- a/applications/dashboard/modules/class.headmodule.php
+++ b/applications/dashboard/modules/class.headmodule.php
@@ -412,6 +412,8 @@ if (!class_exists('HeadModule', false)) {
                 $this->addTag('meta', array('name' => 'description', 'property' => 'og:description', 'itemprop' => 'description', 'content' => $Description));
             }
 
+            $hasImage = false;
+
             // Default to the site logo if there were no images provided by the controller.
             if (count($this->_Sender->Image()) == 0) {
                 $Logo = c('Garden.ShareImage', c('Garden.Logo', ''));
@@ -423,12 +425,22 @@ if (!class_exists('HeadModule', false)) {
 
                     $Logo = Gdn_Upload::url($Logo);
                     $this->addTag('meta', array('property' => 'og:image', 'itemprop' => 'image', 'content' => $Logo));
+                    $hasImage = true;
                 }
             } else {
                 foreach ($this->_Sender->Image() as $Img) {
                     $this->addTag('meta', array('property' => 'og:image', 'itemprop' => 'image', 'content' => $Img));
+                    $hasImage = true;
                 }
             }
+
+            if ($hasImage) {
+                $twitterCardType = 'summary_large_image';
+            } else {
+                $twitterCardType = 'summary';
+            }
+
+            $this->addTag('meta', array('name' => 'twitter:card', 'content' => $twitterCardType));
 
             $this->fireEvent('BeforeToString');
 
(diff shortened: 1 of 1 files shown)


DaazKu:
Sorry for the lack of details:

* Are you certain you can combine tags for double-duty like this (name on top of property)? Is any other software doing this?
  - Yes
* Did you BC test this with Facebook?
  - Yes
* You've added `twitter:image` here without doing any of the checks to see if the found image is within their guidelines for dimensions, size, or context (not a logo or avatar). This sort of qualifications rabbit hole is what I was concerned about.
  - Their documentation state that images over 1mb (or 2 depending where you read) will not be taken into account. I'm pretty sure that checking the size of each images on the page load is a big NO NO. I would let the twitter bot just fail here. I can put a comment in the code though.
* Do we want non-discussion content embedding as a Twitter card? That seems problematic, but there's no consideration of it here. You paste a link to the Activity page, it grabs some random image from the page (probably a tiny badge that gets way scaled up?) and "..." as a description. That's a bad experience.
  - Why not. We already do that with open graph?
  - The '...' is added because a description is needed. If we do not add something as the description the card won't be generated. (Post with only an image for example)

linc:
There are also dimension specs and context specs (which I mentioned above). Specifically, making sure the logo isn't the grabbed image would be pretty trivial since we're doing that in a separate logic block already.

I don't see a problem with *not* using a Twitter card when the content doesn't make sense to have one. The difference between this and the OpenGraph tags is that OG is just describing some content that could be highlighted; the Twitter card markup is a command to format it in a particular way. Therefore we should give thoughtful commands and not just shrug and see what happens. It at least makes halfway sense to have a blank post with an image but no text to get a "..." to force a card that shows the image. The Activity page example I used (or any other random page like that) it makes very little sense.

DaazKu:
Twitter cards are now generated only when relevant. 
- Discussions only.
- Generate a summary card if the discussion has no images.
- Put '...' as the description if the discussions has a relevant image but no text.

The dimension constraint are a SHOULD so we will not try to validate them.
The size constraint is a MUST but I think that it would be stupid check an image size remotely on every page load. **I could do a check if the image is local but, again, meh...**

linc:
I agree checking image size is above and beyond what we want to do here.

DaazKu:
A summary_card is a title and a description. It does not make sense to "pad" the description there.
A summary_large_image can be worth seeing without description since the title and the image may convey enough information for the card to be "valid"

The code went from:
```php
if ($hasImage) {
    $twitterCardType = 'summary_large_image';
} else {
    $twitterCardType = 'summary';
}
```
to
```
if ($hasRelevantImage) {
    $twitterCardType = 'summary_large_image';
    if (!$Description) {
        $Description = '...'; // Let's force a description since a card needs one.
    }
} else {
    if ($Description) {
        $twitterCardType = 'summary';
    }
}
```

and I did not notice that it could be undefined after the change..

linc:
> A summary_card is a title and a description. It does not make sense to "pad" the description there.

Does it make sense to then add a blank twitter:description meta tag?

linc:
If you're going to make a non-obvious choice like only forcing a description for one type of card, it makes sense to leave a code comment so the next dev isn't left scratching their head about the asymmetry of the code.

DaazKu:
I updated the logic of the code and added comments.
I also made sure that $Title has a value since it is required by every twitter cards

Issue #4646, opened 2016-10-15
Exclude vendor docs, examples, and tests from Phing build


linc:
When we distribute Vanilla as a zip, we want to shave off every kilobyte we can. Use Phing to manually exclude any vendor files not used on a production system.

This purge can include README.md & composer.json files (a global rule would be great here if that's possible), but _cannot_ include LICENSE files. Those should be left in our distro.

PR #4730, opened 2016-10-26
Exclude non useful files (but licenses) from the phing build

DaazKu:
The only change I'm not 100% sure about is excluding text files.
I feel there should be no code relying on a text file but we never know!

Fixes https://github.com/vanilla/vanilla/issues/4646


DaazKu:
I assigned but I still need to test the build.

PR #4739, opened 2016-10-28
Add Garden\Schema::dumpSpec

### LINKED ISSUE #4709: Write Schema::dumpSpec() (working method name)
The schema will be use to generate its part of the swagger spec. The spec will dump as an array with the following keys:
- [x] description: The description property.
- [x] parameters: The schema definition as an array, conforming to the [OAI parameter object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject). The internal representation of the schema is very close to this already so it will require only a small amount of massaging.

initvector:
This update adds `Garden\Schema::dumpSpec`.  This function is intended to output a representation of the schema data that is compatible with the [OpenAPI v2 specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject).

Closes #4709

### DIFF
--- a/library/Garden/Schema.php
+++ b/library/Garden/Schema.php
@@ -85,47 +85,48 @@ class Schema implements \JsonSerializable {
      * @return array
      */
     public function dumpSpec() {
+        $schema = $this->schema;
+
+        if (is_array($schema)) {
+            foreach ($schema as &$parameter) {
+                if (!is_array($parameter) || !$parameter['type']) {
+                    unset($parameter);
+                    continue;
+                }
+
+                // Massage schema's types into their Open API v2 counterparts, including potential formatting flags.
+                // string, boolean and array pass through without adjustment.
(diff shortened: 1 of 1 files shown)


initvector:
We do not currently have any indicators for where the parameter is expected to be (head, body, path, etc).  [This is required to be defined as the `in` field of a parameter object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7) and will not be a fully compatible with the OpenAPI v2 specification until we have it.

tburry:
Yeah, I'm guessing we'll have to address that as a future issue. Just add another issue to the epic for now. Here are my initial thoughts though.

1. We'll probably need to have a property for the default "in" on the Schema.
2. Since the properties are all arrays you can look for an "in" key right away.

PR #4743, opened 2016-10-31
Update validation in Garden\Schema to handle unexpected parameters

### LINKED ISSUE #4703: Augment schema validation behavior
Right now the schema object lets data pass through that has been provided, but isn't in the schema. This is a security risk. I want you to add several behaviors to the schema object to control validation.
- [x] Add a Schema::get/setValidationBehavior() property that sets the validation behavior. The property will take one of several class constants.
- [x] Schema::VALIDATE_REMOVE means that fields that are not in the schema will just be silently removed.
- [x] Schema::VALIDATE_NOTICE will do the same as VALIDATE_REMOVE, but will also trigger an E_USER_NOTICE for each removed field.
- [x] Schema::VALIDATE_EXCEPTION will cause fields that are not in the schema to be added as errors in the thrown ValidationException.
- [x] Make sure these behaviors work for nested fields (e.g. object properties).

### LINKED ISSUE #4707: Write unit tests for Schema::get/setValidationBehavior()
Write unit tests for the functionality added in #4703.

initvector:
This update makes a couple changes to `Garden\Schema`:

1. Adds `filterData` function, which removes fields from a `$data` array that aren't explicitly a configured part of the schema.
2. Adds `VALIDATE_*` class constants to assign and detect validation behavior.

In addition, test have been added for each new validation behavior for both basic and nested schemas.

Closes #4703 
Closes #4707

### DIFF
--- a/library/Garden/Schema.php
+++ b/library/Garden/Schema.php
@@ -150,34 +150,32 @@ class Schema implements \JsonSerializable {
      *
      * @param array $data
      * @param array $schema
+     * @param Validation $validation
      * @param string $path
      * @throws ValidationException when configured to do so and an unexpected parameter is encountered
      * @return Schema
      */
-    protected function filterData(array &$data, array $schema, $path = '') {
-        $validation = new Validation();
+    protected function filterData(array &$data, array $schema, Validation &$validation, $path = '') {
         $filtered = false;
 
         foreach ($data as $key => $val) {
             if (array_key_exists($key, $schema)) {
                 continue;
-            } elseif ($filtered === false) {
-                $filtered = true;
             }
 
             $errorMessage = sprintft('Unexpected parameter: %1$s.', $path.$key);
-            $validation->addError(
-                'unexpected_parameter',
-                $key,
-                [
-                    'parameter' => $key,
-                    'message' => $errorMessage,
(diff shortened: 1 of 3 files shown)


tests/Library/Garden/BasicSchemaTest.php
```diff
@@ -346,6 +346,57 @@ public function provideBooleanData() {
...
+        $this->assertArrayNotHasKey('admin', $data);
+        $this->assertArrayNotHasKey('role', $data);
+    }
+
+    /**
+     * Test throwing an exception when removing unexpected parameters from validated data.
+     *
+     * @expectedException \Garden\Exception\ValidationException
+     */
+    public function testValidateException() {
```
tburry: You could do all these three tests with an @dataProvider attribute on the original test.

library/Garden/Schema.php
```diff
@@ -132,6 +146,49 @@ public function dumpSpec() {
     }
 
     /**
+     * Filter fields not in the schema.  The action taken is determined by the configured validation behavior.
+     *
+     * @param array $data
```
tburry: Methods in the Garden namespace should be fully documented. Add descriptions to the parameters and return.

PR #4746, opened 2016-11-02
Add AfterConversationMessageBodyFormat event to PMs

Caylus:
Discussions/Comments fire a AfterCommentBodyFormat event, private messages don't fire any event yet after formatting.
That makes it hard for plugins to add HTML to private messages in a consistent manner.

This change will make sure discussions/comments and private messages exhibit the same behaviour.


Caylus:
Not needed after all!

PR #4750, opened 2016-11-03
Tighten conversation max recipients and use a helper to determine it

linc:
This addresses a potential spam vector where a bot could send hundreds of private messages if a site does not have an upper threshold set manually for the number of users a private message can be sent to.

* Refactor our max recipient checking into a helper method.
* Use a default max of FIVE for new users.
* Use a default max of FIFTY for verified users.
* Moderators may still send unlimited messages.

Future work must still be done to make our flood control tigther, but this is a good first step.

Backport to `release/2.3`.


linc:
Paging @adrianspeyer to review the concept.

adrianspeyer:
Makes a lot of sense. Especially the flood control, which seemed not to function for this. Solid!

tburry:
I think we'll get into a bit of hot water if we can't override both of the default max numbers.

linc:
You want the current global override to be per-default instead?

tburry:
Yeah I think so. I don't want the case where people override the verified default to something smaller and that ends up increasing the non-verified default. Very strange behavior there.

linc:
Added a second config for verified users to graduate to.

applications/conversations/controllers/class.messagescontroller.php
```diff
@@ -60,12 +60,16 @@ public function add($Recipient = '', $Subject = '') {
         $this->permission('Conversations.Conversations.Add');
         $this->Form->setModel($this->ConversationModel);
 
-        // Set recipient limit
-        if (!checkPermission('Garden.Moderation.Manage') && c('Conversations.MaxRecipients')) {
-            $this->addDefinition('MaxRecipients', c('Conversations.MaxRecipients'));
-            $this->setData('MaxRecipients', c('Conversations.MaxRecipients'));
+        // Detect our recipient limit.
+        $maxRecipients = self::getMaxRecipients();
```
initvector: `ConversationModel::getMaxRecipients`

applications/conversations/models/class.conversationmodel.php
```diff
@@ -296,6 +296,30 @@ public function getID($ConversationID, $datasetType = false, $options = []) {
     }
 
     /**
+     * Get how many recipients current user can send a message to.
+     *
+     * @return int|bool A maximum number of recipients or FALSE for unlimited.
+     */
+    public static function getMaxRecipients() {
+        // Moderators can add as many as they want.
+        if (checkPermission('Garden.Moderation.Manage')) {
```
initvector: What about using `Gdn_Session::checkRankedPermission` instead?

Issue #4752, opened 2016-11-03
Searching users by IP in the dashboard isn't returning all the users with the IP


beckyvb:
Also, the user count is not correct.

PR #4753, opened 2016-11-03
Return all users with a valid IP address and update searchCount query

beckyvb:
* Joins User table and UserIP table with a left join.
* Pulls out the IP searching function into its own function.
* Update search count.

Previously, we only returned users who had an entry in the UserIP table. This adds a left join to the query to return all users, whether in the UserIP table or not. Also, fixes searchCount for IP searches, as the count was always empty.

Closes https://github.com/vanilla/vanilla/issues/4752

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2429,7 +2429,7 @@ class UserModel extends Gdn_Model {
             $this->SQL->join('UserRole ur2', "u.UserID = ur2.UserID and ur2.RoleID = $RoleID");
         } elseif (isset($IPAddress)) {
             $fields = ['LastIPAddress'];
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -2495,6 +2483,32 @@ public function search($Filter, $OrderFields = '', $OrderDirection = 'asc', $Lim
...
 
+
+    /**
+     * Searches the UserIP table for users with a given IP Address. Extends the search to GDN_User fields
+     * passed in the $fields param.
+     *
+     * @param string $ip The IP Address to search for.
+     * @param array $fields The additional fields to check in the UserTable
+     */
+    public function searchByIp($ip, $fields = []) {
```
initvector: This method could benefit from limiting its visibility to private.

PR #4759, opened 2016-11-07
Fix dashboard pager next button when there are 0 results

DaazKu:
Use PagerModule::hasMorePages()


beckyvb:
This breaks pagination for 'more'-type pagers with multiple pages.  See dblogger for an example of this.

PR #4761, opened 2016-11-08
Fix heading categories sometimes being wrapped in an anchor

### LINKED ISSUE #4634: Category DisplayAs attribute not always respected when "Heading"
If a category row is rendered with `writeTableRow` or `writeListItem`, it's title is always wrapped in an anchor.  This should not be the case for categories configured to display as "Heading".

initvector:
When a category is configured to display as "Heading", it isn't supposed to have an anchor.  This is true for most of Vanilla, but [`writeListItem`](https://github.com/vanilla/vanilla/blob/7ae304aa7cb89496bc17925d05798352056b4b6c/applications/vanilla/views/categories/helper_functions.php#L130) and [`writeTableRow`](https://github.com/vanilla/vanilla/blob/7ae304aa7cb89496bc17925d05798352056b4b6c/applications/vanilla/views/categories/helper_functions.php#L225) do not currently account for this adjustment.  Everything gets an anchor.

This update alters the aforementioned functions to only apply an anchor to the title of categories when they are not configured to display as "Heading".

Closes #4634

### DIFF
--- a/applications/vanilla/views/categories/helper_functions.php
+++ b/applications/vanilla/views/categories/helper_functions.php
@@ -146,7 +146,9 @@ if (!function_exists('WriteListItem')):
                 echo '<div class="Options">'.getOptions($Row).'</div>';
                 echo '<'.$H.' class="CategoryName TitleWrap">';
                 echo CategoryPhoto($Row);
-                echo anchor(htmlspecialchars($Row['Name']), $Row['Url'], 'Title');
+
+                $safeName = htmlspecialchars($Row['Name']);
+                echo $Row['DisplayAs'] === 'Heading' ? $safeName : anchor($safeName, $Row['Url'], 'Title');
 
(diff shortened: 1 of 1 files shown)


beckyvb:
The only thing I worry about here is that the ui doesn't stop people from nesting categories under headings so we may have issues with users finding that some of their categories have suddenly become unreachable. I don't know if we should make the heading an anchor if it has children, or enforce the semantics of a heading and force users to reorder their categories. Thoughts?

initvector:
I think that if it's configured to display as "Heading", it should always display as a heading (no anchor).  Conditional handling may help transitioning to the new way of structuring categories, but it's just as likely to keep people doing things the old way, which I think we want to move away from.  It could also mean that "Heading" becomes confusing for users, because it means two different things, depending on how their categories are structured.  My vote is for enforcing the semantics.

PR #4772, opened 2016-11-09
Add new UA/RU characters to fix issues with transliteration

aZzZzZoT:
Russian letters: И
Ukrainian letters: ї, і, є, Ї, І, Є, И


aZzZzZoT:
No. This was not intended.  I suggested these changes when there was a new version of this file.

linc:
I think your editor swapped out spaces for tabs on the doc block and added an erroneous space on the last line of it. Sorry to pick at little stuff but that's gotta get fixed too.

 
screen shot 2016-11-10 at 10 54 58 am

PR #4776, opened 2016-11-09
Fix standards in userphoto view

linc:
No functional change.

### DIFF
--- a/applications/dashboard/views/modules/userphoto.php
+++ b/applications/dashboard/views/modules/userphoto.php
@@ -11,7 +11,7 @@ if (!$User) {
 
 $Photo = $User->Photo;
 if ($Photo) {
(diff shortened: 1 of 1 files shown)


applications/dashboard/views/modules/userphoto.php
```diff
@@ -1,41 +1,43 @@
...
+if (!$User) {
     return;
+}
 
 $Photo = $User->Photo;
 if ($Photo) {
-    if (!IsUrl($Photo)) {
-        $Photo = Gdn_Upload::url(changeBasename($Photo, 'p%s'));
-    }
+    $Photo = (isUrl($Photo)) ?: Gdn_Upload::url(changeBasename($Photo, 'p%s'));
```
initvector: Wouldn't this set `$Photo` to `true` if it was already a URL?

linc: how maek ternary

Issue #4595, opened 2016-10-04
Dashboard v3 Design Issues


tburry:
### General
- [x] The chevrons/colapse triangles in the side menu are not aligned correctly when openned or closed
- [x] Some icons like setting icons in buttons and dropdown triangles are still not the correct color
- [x] The text in some of the input fields isn't centered vertically
- [x] Add 12 additional pixels between any icon only buttons i.e. Edit and Delete
- [x] There should be 60px of padding between the last item on the page and the footer
### Homepage
- [x] H2 headers should have a divider above them to create more of a visual separation
- [x] The select transition should fade in and not slide up
### Themes
- [x] The select transition should be a fade in not a slide up
- [x] There should be 60px of padding between the last item on the page and the footer
- [x] Why are some of the theme titles blue and some dark? Just ad outgoing link arrow icon like "View site" button
### Avatars
- [x] The select transition should be a fade in not a slide up
- [x] Let's even out the columns a bit more to be sure that the "Remove" & "Change" buttons stay at the end
### Email
- [x] The content/settings that shows up at the bottom once the toggle has been enabled should be 50% opacity and not hidden
- [x] The modal window for the "Upload new email logo" should be half the width
### Custom Domain Name
- [x] The bullet points should have more line height
- [x] There should be 6px min of padding between the "http://" and the text field
- [x] Why is there a divider after "New Custom Domain"?
- [x] Change the "Continue" label to "Apply"
### Roles & Permissions
- [x] There should be 60px of padding between the last item on the page and the footer
- [x] Can we add another 12px of padding/space between the "Edit" and "Delete" icon?
- [x] Can we give the "Default" info it's own column?
### Registration
- [x] The checkbox at the top should be a toggle control
- [x] Is it me or do the radio buttons look oval
### Profile Fields
- [x] Make the modal window for the "Add Field" button half the width
### Spoof
- [x] Why are the "Your Email" and Your Password" fields yellow? Change to Grouped Button Hover color.
- [x] We should change the button from "Go" to something like "Apply"
### jsconnect
- [x] Can we even out the columns for the "Add Connection " modal window?
### Categories
- [x] Chevrons for each parent category isn't aligned correctly
- [x] The add category page should have a back chevron
### Advanced
- [x] Let's put the the toggle control at the very top of the list
### Social
- [x] The settings icon isn't the correct color
### Tagging
- [x] The placeholder text in the search box is to long and crashes into the clear "X" icon
[...]

PR #4781, opened 2016-11-10
Add a new functionality to disable settings without hiding them

beckyvb:
Adds 'foggy' functionality to make elements in a foggy block appear fuzzy and be disabled.

* Adds foggy css.
* Adds a new jsonTarget method for triggering events: 'Trigger'.
* Fixes the disabled state for file-upload.
* Implemented email styles settings page with foggy.

Partially addresses (but doesn't close) https://github.com/vanilla/vanilla/issues/4595


initvector:
1. I can click through the "fog".  For example, if I double-click on the "Page Color", it highlights text in the associated form control.  Is there an easy way to block click through?
2. I see some strange artifacts on some form controls when they're behind the "fog".  They're small, dark blurs appearing beneath the bottom-left corner of the controls. They're not there when those same controls are not behind the "fog".

PR #4802, opened 2016-11-15
Ensure svg icons have no white space around them

beckyvb:
Fixes issue where an inline svg has a space inserted before it.

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -45,9 +45,9 @@ if (!function_exists('dashboardSymbol')) {
             $alt = 'alt="'.htmlspecialchars($alt).'" ';
         }
         $r = <<<EOT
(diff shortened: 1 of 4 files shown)


library/core/functions.render.php
```diff
@@ -45,9 +45,9 @@ function dashboardSymbol($name, $alt = '', $class = '') {
             $alt = 'alt="'.htmlspecialchars($alt).'" ';
         }
         $r = <<<EOT
-    <svg {$alt}class="icon $class icon-svg-$name" viewBox="0 0 17 17"><use xlink:href="#$name" /></svg>
+<svg {$alt}class="icon $class icon-svg-$name" viewBox="0 0 17 17"><use xlink:href="#$name" /></svg>
```
DaazKu: At this point why not just 
```php
$r = '<svg {$alt}class="icon $class icon-svg-$name" viewBox="0 0 17 17"><use xlink:href="#$name" /></svg>';
```
???

PR #4804, opened 2016-11-16
Drop attribute "type" for style tags

R-J:
Vanilla uses html5 and in html5 the default type is "text/css". Based on [that information](http://www.w3schools.com/tags/att_style_type.asp), this type is the only supported type anyway. So even if browsers wouldn't be compatible with html5, they would treat a simple `<style>` tag correctly.

### DIFF
--- a/applications/dashboard/views/email/email-basic.tpl
+++ b/applications/dashboard/views/email/email-basic.tpl
@@ -6,7 +6,7 @@
 <!--<![endif]--><meta name="viewport" content="width=device-width, initial-scale=1.0">
 <!--[if (gte mso 9)|(IE)]>
     {literal}
-    <style type="text/css">
+    <style>
         table {
             border-collapse: collapse;
         }
--- a/applications/dashboard/views/email/email-basic.tpl
+++ b/applications/dashboard/views/email/email-basic.tpl
@@ -6,7 +6,7 @@
 <!--<![endif]--><meta name="viewport" content="width=device-width, initial-scale=1.0">
 <!--[if (gte mso 9)|(IE)]>
     {literal}
-    <style>
+    <style type="text/css">
         table {
             border-collapse: collapse;
         }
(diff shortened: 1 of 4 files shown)


applications/dashboard/views/email/src/email-basic.php
```diff
@@ -192,7 +192,7 @@
     </style>
     <!--[if (gte mso 9)|(IE)]>
     [[literal]]
-    <style type="text/css">
+    <style>
```
beckyvb: Some email clients are not yet HTML5 compliant, so I don't feel comfortable removing this yet.

applications/dashboard/views/email/email-basic.tpl
```diff
@@ -6,7 +6,7 @@
 <!--<![endif]--><meta name="viewport" content="width=device-width, initial-scale=1.0">
 <!--[if (gte mso 9)|(IE)]>
     {literal}
-    <style type="text/css">
+    <style>
```
beckyvb: Some email clients are not yet HTML5 compliant, so I don't feel comfortable removing this yet.

R-J:
Mail templates now have the type attribute again

applications/dashboard/views/email/src/email-basic.php
```diff
@@ -6,7 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <!--<![endif]-->
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <style>
+    <style type="text/css">
```
beckyvb: This one isn't necessary. This file gets compiled into a Smarty template (`email/email-basic.tpl`) where this block of css gets inlined and the block disappears. The `email/email-basic.tpl` file is the actual email template that gets used.

beckyvb:
Thank you :)

PR #4809, opened 2016-11-16
Respect Vanilla.RootCategory.DisplayAs on all-categories page

### LINKED ISSUE #4792: Make the all categories endpoint obey the root category display as setting
The all categories page needs to obey the Vanilla.RootCategory.DisplayAs config setting. That means the /categories root page will display as flat if that is the setting.

initvector:
This update ensures the public all-categories page respects the new `Vanilla.RootCategory.DisplayAs` config setting, introduced in https://github.com/vanilla/vanilla/pull/4800

Closes #4792


initvector:
Conflicts have been resolved

Issue #4547, opened 2016-09-28
Vanilla.Categories.DoHeadings cannot be set anymore


DaazKu:
The configuration cannot be set but it is still respected in a bunch of places.
We should remove references to it.

This might also cause some weird behaviour when updating from an old version.

initvector:
See: [Instances of Vanilla.Categories.DoHeadings in Vanilla](https://github.com/vanilla/vanilla/search?utf8=%E2%9C%93&q=Vanilla.Categories.DoHeadings) (GitHub's code search isn't great, but it'll get you in the general area.)

Also necessary to review themes for references to this config setting.

PR #4811, opened 2016-11-16
Support headings when writing a category table

beckyvb:
The only way to get a table-view category list to display headings is to use the `Vanilla.Categories.DoHeadings` config setting. This rewrites the writeCategoryTable function to respect a category's `DisplayAs: Headings` property and render in the same way as they did with `Vanilla.Categories.DoHeadings` on.

Partially addresses: https://github.com/vanilla/vanilla/issues/4547

### DIFF
--- a/applications/vanilla/views/categories/table.php
+++ b/applications/vanilla/views/categories/table.php
@@ -5,5 +5,5 @@
 $this->fireEvent('AfterDescription');
 $this->fireEvent('AfterPageTitle');
 $categories = $this->data('CategoryTree');
-writeCategoryTable($Categories);
+writeCategoryTable($categories);
 ?>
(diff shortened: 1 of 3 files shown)


applications/vanilla/views/categories/table.php
```diff
@@ -1,24 +1,9 @@
...
-            WriteCategoryTable($Category['Children'], 2);
-            ?>
-        </div>
-    <?php
-    }
-} else {
-    WriteCategoryTable($Categories, 1);
-}
+$categories = $this->data('CategoryTree');
+writeCategoryTable($Categories);
```
initvector:
```php
writeCategoryTable($categories);
```

PR #4812, opened 2016-11-17
Check if headings are present in category list instead of checking config

beckyvb:
Eliminates checks to the DoHeadings config setting in category list views.

Partially addresses: https://github.com/vanilla/vanilla/issues/4547


beckyvb:
This commit: https://github.com/vanilla/vanilla/commit/446ce39808bae0286db1fe9b781b568192f377e2 changed the way we calculate category depths. Before this, headings had Depth1 and Depth2 css classes, whereas afterwards the had headings had Depth0 and Depth1 classes. The Depth* classes should follow the depth patterns of non .CategoryListWithHeadings classes. As such, I think we can safely remove CategoryListWithHeadings, as anybody who relied on it would have noticed by now that the left padding had shifted.

PR #4819, opened 2016-11-17
Make ban's delete button issue a post request with valid transient key

DaazKu:
The delete button will now work :)


beckyvb:
`js-hijack` is the appropriate class to be used with `js-modal-confirm`

PR #4820, opened 2016-11-17
Contain DoHeadings logic for DisplayAs calculations in Category Model

beckyvb:
* Removes unused $maxDisplayDepth variable from Categories Controller. 

* Moves the logic for temporarily saving the DoHeadings config to the model where DisplayAs is calculated.

* Moves the DisplayAs calculation into its own function.

Partially addresses https://github.com/vanilla/vanilla/issues/4547

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -414,6 +414,14 @@ class CategoryModel extends Gdn_Model {
         }
     }
 
+    /**
+     * Maintains backwards compatibilty with `DisplayAs: Default`-type categories by calculating the DisplayAs property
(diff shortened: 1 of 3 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -403,6 +403,18 @@ private static function calculate(&$category) {
...
+        if (!val('CssClass', $category)) {
+            $category['CssClass'] = 'Category-'.$category['UrlCode'];
+        }
+
+        if (isset($category['AllowedDiscussionTypes']) && is_string($category['AllowedDiscussionTypes'])) {
+            $category['AllowedDiscussionTypes'] = dbdecode($category['AllowedDiscussionTypes']);
+        }
+    }
+
+    public static function calculateDisplayAs(&$category) {
```
initvector: Doc block, please!

PR #4823, opened 2016-11-17
Add js-hijack on messages delete button

DaazKu:
Otherwise you endup on `/dashboard/message/delete/12/{TransientKey}` with TRUE as the content of the page


DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/4825

PR #4836, opened 2016-11-19
Fix standards in render functions

linc:
* Doc blocks: Fix locations, add missing ones, and add/fix params/returns of many.
* Fix some casing; update some array syntax.
* `GetValue()` => `val()`.
* Deprecate `discussionLink()` (use `discussionUrl()`).


DaazKu:
Fix the conflicts and it is good to go!

Issue #4561, opened 2016-09-30
MeBox Messages dropdown should have subjects clickable, not participants


kaecyra:
In the messages dropdown in the MeBox, the subject of the message should be clickable, not the participants. This is confusing UX.

linc:
What I'd really like here is for the entire box to be clickable, and not display the username as tho it is a separate profile link.

DaazKu:
Won't that break theming or other things?

linc:
I doubt folks have been theming the Me Module flyout menu. @beckyvb ?

beckyvb:
The flyout menu does get themed.

linc:
We're gonna risk adding an anchor tag in there to make this work.

PR #4846, opened 2016-11-21
Make li items clickable in the MeBox Conversations list

DaazKu:
Make `.PopList .Item[rel]` clickable and have the `cursor: pointer` property.

We can't wrap everything in an anchor so we use javascript instead.
Bonus is that the markup doesn't change!

Fixes https://github.com/vanilla/vanilla/issues/4561


beckyvb:
This creates an inconsistency between the different dropdowns in the MeModule. I think it's an enhancement to have the entire block clickable, but we should either do it across the MeModule Item-type dropdowns to maintain consistency or not at all. 

Also, I'm not wild about using javascript where an anchor should be used. However, changing the markup would mess up a lot of themes, so in this case it isn't a bad solution IMO.

beckyvb:
My feedback would be to reconsider whether this enhancement is necessary and if so, then apply it to all the Item-type dropdowns in the MeModule.

linc:
Lets use the Javascript approach to all 3 menus, then. I don't think we can afford to break themes currently. We'll do it correctly when we revise the base theme.

PR #4857, opened 2016-11-23
Fix Profile Extender checkbox display on registration

initvector:
Profile Extender uses `include_once` for its registration page controls.  When the first is executed, used for non-checkbox fields, it ensures the second, used for checkbox fields, is not.

This update swaps out the `include_once` calls to `include`.


linc:
release/1609

Issue #4426, opened 2016-09-08
Need a unified Single Sign On page in the Dashboard


linc:
- Work in a parallel way to our Social section: control a class of plugins.
- Centralize management of connections from our various plugins.
- Expose and explain the rest of our SSO config options (see #3586 for some, `EntryController::connect()` for more).

Requires mockups first.

Depends on #3695 (Dashboard v3).

linc:
We'll need to discuss this.

PR #4860, opened 2016-11-25
Rework social connections page to use standard addons view

linc:
* Fixes broken icons on social connections page.
* Social Connections are just addons. Display them with the same view & remove the old one.
* Refactor `getConnections()` to work with the addons view & the addonManager.
* Rename this section to "Social Connect Addons" and put it under the Addons section because logic.

New to-do generated by this PR: Re-implement the "not configured" warning in a standard way for all addons using the media module. It should be considerably more bold than the current one was anyway - like all caps & red next to the name.


linc:
It was a delightful Thanksgiving Day puzzle.

linc:
Gets us a little closer to #4426.

applications/dashboard/controllers/class.socialcontroller.php
```diff
@@ -53,35 +53,38 @@ public function manage() {
...
+                continue;
             }
 
-            $Connections[$PluginKey] = array_merge($Connections[$PluginKey], $PluginInfo, array(
-                'Enabled' => Gdn::addonManager()->isEnabled($PluginKey, \Vanilla\Addon::TYPE_ADDON),
-                'Configured' => $Configured
-            ), array(
-                'Icon' => sprintf("/plugins/%s/icon.png", $PluginInfo['Folder'])
-            ));
+            if (!array_key_exists($addonName, $connections)) {
```
linc: Probably never. I was over-cautiously reusing that check from the last version. But come to think of it there was a lot squirrelly about that version so I probably shouldn't have.

Issue #4551, opened 2016-09-29
SSO does not obey email verification


tburry:
If I SSO via entry/connect and the provider does not pass an email address then the user will be asked to enter one. When this happens, the email address should be verified in the same way that registering normally is. If the email address is provided then the email address should not have to be verified.

PR #4861, opened 2016-11-25
Validate email if it was not set by the provider

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/4551

Also fix a bug where the username field would appear on the form if the email validation (or any validation) failed.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -444,7 +444,9 @@ EOT;
 
         // Allow a provider to not send an email address but require one be manually entered.
         if (!UserModel::noEmail()) {
-            if (!$this->Form->getFormValue('Email') || $this->Form->getFormValue('EmailVisible')) {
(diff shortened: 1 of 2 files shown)


linc:
@initvector Open to suggestions on the wording of this logic nightmare to improve its clarity.

DaazKu:
I decided to let the conditions as is but explain the "process" behind them.

linc:
I added a commit that does three things:

1. Condenses the complex `$displayConnectName` logic to the top of the view with slightly revised comments. This should result in _no functional change_.
2. Makes some minor coding standards revisions.
3. Eliminates an empty set of `li` HTML tags when `$displayConnectName` is `false`.

I feel like this gets us a half-step closer to pulling logic out of this view by at least no longer mixing the worst of it in the middle of the HTML, and makes it a bit easier to follow.

Please have a close look at what I've done to be sure I haven't introduced a bug in this commit. When you're satisfied, this can be merged.

PR #4869, opened 2016-11-27
Create addon groups and an AddonsController

linc:
* Remove special `SocialConnect` addon info key.
* Create new `Groups` addon key and move `SocialConnect` & `SSO` plugins to use it (with those values).
* Add "Single Sign On" as menu navigation for Addons section of Dashboard.
* Hide addons with `Groups` key from main "Plugins" list.
* Refine names & descriptions of addons in our new groups.
* Replace `SocialController` with new `AddonsController` so each group we make doesn't spawn new controllers that could have overlapping namespaces (see already: SSO). This is also a more logical namespace.
* Modify old `SocialController->getConnections()` and its `manage()` view to work generically with any new addons groups we want to make in the future.


linc:
@beckyvb - I'd like @tburry to approve the creation of this new controller before this is merged.

applications/dashboard/settings/class.hooks.php
```diff
@@ -279,7 +279,8 @@ public function dashboardNavModule_init_handler($sender) {
             ->addGroup(t('Forum Settings'), 'forum', '', ['after' => 'users'])
             ->addGroup(t('Reputation'), 'reputation', '', ['after' => 'forum'])
             ->addGroup(t('Addons'), 'add-ons', '', ['after' => 'reputation'])
-            ->addLinkIf('Garden.Settings.Manage', t('Social Connect'), '/social/manage', 'add-ons.social', '', $sort)
+            ->addLinkIf('Garden.Settings.Manage', t('Social Connect'), '/addons/socialconnect', 'add-ons.social', '', $sort)
+            ->addLinkIf('Garden.Settings.Manage', t('Single Sign On'), '/addons/sso', 'add-ons.sso', '', $sort)
```
beckyvb: It would be nice if we could loop through a list of static addon groups defined in the AddonsController, rather than having to manually add each one of them here.

tburry:
I'm not a fan of this architecture.

1. Really, all addons should be available on the addons page. I think it's a mistake to hide some.
2. A unified SSO page is good, but it needs to handle multiple connections from certain addons.
3. We are trying to get our dashboard controllers into /settings/controller namespaces and the AddonSettingsController is reserved for addon management that is already in other issues.

Something like this groups functionality really should be discussed with me before going into implementation.

PR #4870, opened 2016-11-27
Add Vanilla\Permissions class

### LINKED ISSUE #4828: Create a Vanilla\Permissions helper class
Right now permissions are loaded from a combination of GDN_UserRole and GDN_Permission and then compiled into an internal array and cached. The compiling is done across the user model and permission model and there is some functionality to manipulate the array in memory via the Gdn_Session class. What we want is a central object to handle the permissions array.

- The class will be called Vanilla\Permissions.
- The class will contain the internal array structure used in Gdn_Session->_Permissions and have the methods to manipulate it. The structure should remain the same for BC.

Here are some properties/methods.

- [x] Properties: $permissions, $isAdmin, $isBanned. Access these through helper methods. Use this example convention for boolean properties: isAdmin(), setAdmin().
- [x] Constructor: __construct($permissions = []).
- [x] Method: add(string $permission, int|array $id). Give a user $permission for the specified ID(s). This is sort of the same as Gdn_Session->setPermission().
- [x] Method: remove(string $permission, int|array $id = null). Opposite of add().
- [x] Method: set(string $permission, bool $value). Set a global (not per-category) permission value.
- [x] Method: has(string $permission, int $id = null). This will be implemented from Gdn_Session->checkPermission() with a reduced functionality given the reduced parameters.
- [x] Method: hasAny(array $permissions, int $id = null). Returns true if the user has any of the permissions for the given $id (if any).
- [x] Method: hasAll(array $permissions, int $id = null). Returns true if the user has all of the permissions for the given $id (if any).
- [x] Method: loadAndCompile(array $permissions). See UserModel->compilePermissions().
- [x] Method: merge(Permissions $permissions). Merge the permissions from another permissions object into this one. See PermissionModel::addPermissions().
- [x] Method: overwrite(string $permission, bool|array $value). Overwrite $permission with $value, including all associated category IDs.

All of these members have analogues in our code of which I noted where I could. Notice though that I took the opportunity to clean up the signatures a bit so some things are 100% a copy/paste job.

initvector:
This update adds the `Vanilla\Permissions` class, which will serve as a central object for handling permissions.  More detail can be found in the associated issue.

Closes #4828

### DIFF
--- a/library/Vanilla/Permissions.php
+++ b/library/Vanilla/Permissions.php
@@ -172,10 +172,10 @@ class Permissions {
      * @return $this
      */
     public function merge(Permissions $source) {
-        $this->permissions = array_merge_recursive(
+        $this->setPermissions(array_merge_recursive(
             $this->permissions,
             $source->getPermissions()
-        );
+        ));
 
         return $this;
     }
--- a/library/Vanilla/Permissions.php
+++ b/library/Vanilla/Permissions.php
@@ -1,0 +1,122 @@
+<?php
+/**
+ * @copyright 2009-2016 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+namespace Vanilla;
+
+class Permissions {
+
+    /** @var array */
+    private $permissions = [];
+
+    /** @var bool */
+    private $isAdmin = false;
+
+    /** @var bool */
+    private $isBanned = false;
+
+    /**
+     * Permissions constructor.
+     *
+     * @param array $permissions
+     */
+    public function __construct($permissions = []) {
+    }
+
+    /**
+     * @param string $permission
+     * @param int|array|null $id
+     */
+    public function add($permission, $id = null) {
+    }
+
+    /**
+     * @param string $permission
+     * @param int|null $id
+     */
+    public function has($permission, $id = null) {
+    }
+
+    /**
+     * @param array $permissions
+     * @param int|null $id
+     */
+    public function hasAll($permissions, $id = null) {
+    }
+
+    /**
+     * @param array $permissions
+     * @param int|null $id
+     */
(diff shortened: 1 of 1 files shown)


library/Vanilla/Permissions.php
```diff
@@ -0,0 +1,277 @@
...
+     */
+    public function __construct($permissions = []) {
+        $this->setPermissions($permissions);
+    }
+
+    /**
+     * Add a permission.
+     *
+     * @param string $permission
+     * @param int|array $IDs
```
linc: IDs of what?

Let's step up the context on some of these doc blocks, especially when we're paving new ground.

library/Vanilla/Permissions.php
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * @copyright 2009-2016 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+namespace Vanilla;
+
+class Permissions {
+
+    /** @var array */
```
linc: What form is this array expected to be in?

library/Vanilla/Permissions.php
```diff
@@ -0,0 +1,277 @@
+<?php
+/**
+ * @copyright 2009-2016 Vanilla Forums Inc.
+ * @license GPLv2
```
linc: GNU GPLv2

library/Vanilla/Permissions.php
```diff
@@ -0,0 +1,277 @@
...
+ */
+
+namespace Vanilla;
+
+class Permissions {
```
linc: doc block

initvector:
Docs have been improved.

PR #4874, opened 2016-11-29
Add ability to hide category permissions when editing a role

### LINKED ISSUE #4827: Role edit UI scaling options
Currently, when editing a role you will see all of the categories that have custom permissions. This breaks when there are too many categories with custom permissions (boo). In order to solve this we need some way of suppressing the categories on the role edit UI.

- [x] Add a config setting that will remove the UI for all but the default categories on the add/edit role page.
- [x] Editing a role like this should not remove per-category permissions if they exist.

initvector:
This update adds a new config value: `Vanilla.HideRoleCategoryPermissions`.  This config controls whether or not the per-category permissions are displayed when editing a role.  More information can be found on the original issue.

Closes #4827

### DIFF
--- a/applications/dashboard/controllers/class.rolecontroller.php
+++ b/applications/dashboard/controllers/class.rolecontroller.php
@@ -115,6 +115,8 @@ class RoleController extends DashboardController {
             $this->title(t('Edit Role'));
         }
 
+        $showCategories = c('Vanilla.CategoriesOnRoleEdit', true);
+
         $this->setHighlightRoute('dashboard/role');
         $PermissionModel = Gdn::permissionModel();
         $this->Role = $this->RoleModel->getByRoleID($RoleID);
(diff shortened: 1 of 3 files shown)


applications/dashboard/controllers/class.rolecontroller.php
```diff
@@ -133,6 +135,22 @@ public function edit($RoleID = false) {
         if ($this->Form->authenticatedPostBack() === false) {
             // Get the role data for the requested $RoleID and put it into the form.
             $Permissions = $PermissionModel->getPermissionsEdit($RoleID ? $RoleID : 0, $LimitToSuffix);
+
```
tburry: Since we are adding this config setting as an optimization then it should be handled inside PermissionModel->getPermissionsEdit(), perhaps with an extra parameter. We don't want to process all of the categories if we don't have to or else the operation will never be able to be fully optimized.

applications/dashboard/controllers/class.rolecontroller.php
```diff
@@ -157,6 +175,41 @@ public function edit($RoleID = false) {
                 $this->Form->setFormValue('PersonalInfo', false);
             }
 
+            // If the form didn't have category permissions, they'll need to be added before saving.
```
tburry: Again, we want the setting to not re-add the categories or else the optimization is lost. When saving a role's permissions I just want to save two rows in GDN_Permission:

1. The global permission row.
2. The default category permissions row.

tburry:
Couple more comments.

1. Name the config setting Vanilla.HideRoleCategoryPermissions.
2. Make a property of the RoleController->hideCategoryPermissions that initializes to that value in the constructor.

applications/dashboard/controllers/class.rolecontroller.php
```diff
@@ -13,13 +13,23 @@
  */
 class RoleController extends DashboardController {
 
+    /** @var bool Should categories be hidden when editing a role? */
+    private $hideCategories;
```
tburry: hideCategoryPermissions, just hideCategories is ambiguous.

PR #4877, opened 2016-11-30
Refactor Gdn_Session permission checking to use Vanilla\Permissions

### LINKED ISSUE #4834: Refactor Gdn_Session to use Vanilla\Permissions
Depends on #4828.

Refactor Gdn_Session to use Vanilla\Permissions.

- [x] Give Gdn_Session a private $permissions property with getter/setter.
- [x] Refactor methods to use the new permissions object.
- [x] Gdn_Session->_Permissions can probably be removed even though its protected.
- [x] Please keep BC in mind. Be careful.

initvector:
This update moves to replace internal permission checking in `Gdn_Session` with an instance of `Vanilla\Permissions`.  More details may be found in the original issue.

Closes #4834

### DIFF
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -29,9 +29,6 @@ class Gdn_Session {
     /** @var object Attributes of the current user. */
     protected $_Attributes;
 
-    /** @var object Permissions of the current user. */
-    protected $_Permissions;
-
     /** @var object Preferences of the current user. */
     protected $_Preferences;
 
@@ -53,7 +50,6 @@ class Gdn_Session {
         $this->UserID = 0;
         $this->User = false;
         $this->_Attributes = array();
-        $this->_Permissions = array();
         $this->_Preferences = array();
         $this->_TransientKey = false;
 
@@ -67,7 +63,8 @@ class Gdn_Session {
      * @param array $perms The permissions to add.
      */
     public function addPermissions($perms) {
-        $this->_Permissions = PermissionModel::addPermissions($this->_Permissions, $perms);
+        $newPermissions = new Vanilla\Permissions($perms);
+        $this->permissions->merge($newPermissions);
     }
 
     /**
@@ -201,7 +198,6 @@ class Gdn_Session {
         $this->UserID = 0;
         $this->User = false;
         $this->_Attributes = array();
-        $this->_Permissions = array();
         $this->_Preferences = array();
         $this->_TransientKey = false;
         $this->timeZone = null;
@@ -214,7 +210,7 @@ class Gdn_Session {
      * @return array
      */
     public function getPermissions() {
-        return is_array($this->_Permissions) ? $this->_Permissions : array();
+        return $this->permissions->getPermissions();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


library/core/class.session.php
```diff
@@ -132,42 +134,38 @@ public function checkPermission($Permission, $FullMatch = true, $JunctionTable =
             $JunctionID = '';
         }
 
-        $Permissions = $this->getPermissions();
-        if ($JunctionTable && !c('Garden.Permissions.Disabled.'.$JunctionTable)) {
-            // Junction permission ($Permissions[PermissionName] = array(JunctionIDs))
+        if ($JunctionID === '') {
+            $JunctionID = null;
```
tburry: This seems odd given just above you set $JunctionID = '' specifically.

library/core/class.session.php
```diff
@@ -132,42 +134,38 @@ public function checkPermission($Permission, $FullMatch = true, $JunctionTable =
...
-                        && is_array($Permissions[$Permission])
-                        && in_array($JunctionID, $Permissions[$Permission]);
+                if ($JunctionID !== null) {
+                    return $this->permissions->has($Permission, $JunctionID);
                 } else {
-                    $Result = array_key_exists($Permission, $Permissions)
-                        && is_array($Permissions[$Permission])
-                        && count($Permissions[$Permission]);
+                    // Backwards-compatible support for a null JunctionID meaning "permission on any ID"
+                    return array_key_exists($Permission, $allPermissions) && count($allPermissions[$Permission]);
```
tburry: Support for this check should be put into Vanilla\Permissions, perhaps adding a constant of Permissions::ANY = 'any' for an ID would be something that can be done.

library/core/class.session.php
```diff
@@ -132,42 +134,38 @@ public function checkPermission($Permission, $FullMatch = true, $JunctionTable =
...
-                return arrayInArray($Permission, $Permissions, $FullMatch);
+                if ($FullMatch) {
+                    return $this->permissions->hasAll($Permission);
+                } else {
+                    return $this->permissions->hasAny($Permission);
+                }
             } else {
-                return in_array($Permission, $Permissions) || array_key_exists($Permission, $Permissions);
+                // Backwards-compatible permission check
+                return $this->permissions->has($Permission) || array_key_exists($Permission, $allPermissions);
```
tburry: I'd prefer this odd extra check to be removed. Can you not make Vanilla\Permissions handle this?

library/core/class.session.php
```diff
@@ -29,15 +29,15 @@ class Gdn_Session {
...
-    protected $_Permissions;
-
     /** @var object Preferences of the current user. */
     protected $_Preferences;
 
     /** @var object The current user's transient key. */
     protected $_TransientKey;
 
+    /** @var  Vanilla\Permissions */
+    private $permissions;
```
tburry: Add a getter/setter.

initvector: Gonna hold off on this for now, since `Gdn_Session` already has a `getPermissions` and a `setPermission`.  We'll likely deprecate/replace these methods in the future.

PR #4885, opened 2016-12-02
Refactor UserModel to use Vanilla\Permissions

### LINKED ISSUE #4826: Refactor UserModel to use Vanilla\Permissions
The UserModel can now use the Vanilla\Permissions object. It should be BC so will still return the internal array from its methods, but the object will box that array and replace the other helper function calls.

- [x] Implement against UserModel->definePermissions().
- [x] Implement against UserModel->getSession().
- [x] Look for other uses.
- [x] Add userModel_loadPermissions event when permissions have been loaded from the DB, but before they are cached. This will allow plugins to manipulate the permissions before caching. I've pasted a proposed implementation below:

```php
$Data = Gdn::permissionModel()->cachePermissions($UserID);
$permissions = new Permissions();
$permissions->loadAndCompile($Data);

$this->EventArguments['UserID'] = $UserID;
$this->EventArguments['Permissions'] = $permissions;
$this->fireEvent('loadPermissions');

$permissionsArray = $permissions->getPermissions();
$PermissionsSerialized = dbencode($permissionsArray);
if (Gdn::cache()->activeEnabled()) {
    Gdn::cache()->store($UserPermissionsKey, $permissionsArray);
} else {
    // Save the permissions to the user table
    if ($UserID > 0) {
        $this->SQL->put('User', ['Permissions' => $PermissionsSerialized], ['UserID' => $UserID]);
    }
}

return $Serialize ? $PermissionsSerialized : $permissionsArray;
```

initvector:
This update swaps out model-specific permission handling for the `Vanilla\Permissions` class.  More detail can be found in the original issue.

Closes #4826

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -1103,11 +1103,17 @@ class UserModel extends Gdn_Model {
         }
 
         $Data = Gdn::permissionModel()->cachePermissions($UserID);
-        $Permissions = UserModel::compilePermissions($Data);
+        $permissions = new Vanilla\Permissions();
+        $permissions->compileAndLoad($Data);
 
-        $PermissionsSerialized = dbencode($Permissions);
+        $this->EventArguments['UserID'] = $UserID;
+        $this->EventArguments['Permissions'] = $permissions;
+        $this->fireEvent('loadPermissions');
+
+        $permissionsArray = $permissions->getPermissions();
+        $PermissionsSerialized = dbencode($permissionsArray);
         if (Gdn::cache()->activeEnabled()) {
-            Gdn::cache()->store($UserPermissionsKey, $Permissions);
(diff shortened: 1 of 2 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -249,8 +249,10 @@ public function checkPermission($User, $Permission, $Options = []) {
         }
 
         // TODO: Check for junction table permissions.
-        $Result = in_array($Permission, $Permissions) || array_key_exists($Permission, $Permissions);
-        return $Result;
+        $permissions = new Vanilla\Permissions($Permissions);
+
+        // Secondary check added for backwards compatibility
+        return $permissions->has($Permission) || array_key_exists($Permission, $permissions->getPermissions());
```
tburry: Why do we need this secondary check?

PR #4888, opened 2016-12-02
Handle Groups and Events display

DaazKu:
Update the log model to display Group and Event record properly!

### DIFF
--- a/applications/dashboard/models/class.logmodel.php
+++ b/applications/dashboard/models/class.logmodel.php
@@ -95,6 +95,17 @@ class LogModel extends Gdn_Pluggable {
                     $Result .= "<br />".t('Banned');
                 }
                 break;
+            case 'Group':
+                $Result =
+                    '<b>'.$this->formatKey('Name', $Data).'</b><br />'.
+                    $this->formatKey('Description', $Data);
+                break;
+            case 'Event':
+                $Result =
+                    '<b>'.$this->formatKey('Name', $Data).'</b><br />'.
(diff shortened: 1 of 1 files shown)


DaazKu:
That would be best. But since we are pretty much doing the moderation over next year ¯\\_\(ツ)\_\/¯.
I think we should start collecting these kind of improvement feedback  somewhere centralized.

linc:
* We don't want to impose closed-source name spaces on open source.
* It'll be safer to put this in Groups where the code will survive a refactor.
* It will be less work to preserve an event than refactor this entire block out later.

DaazKu:
Makes sense. I'll do the update!

applications/dashboard/models/class.logmodel.php
```diff
@@ -67,48 +67,42 @@ public static function endTransaction() {
...
-            case 'Event':
-                $Result =
-                    '<b>'.$this->formatKey('Name', $Data).'</b><br />'.
-                    $this->formatKey('Body', $Data).'<br />'.
-                    t('Location').': '.$this->formatKey('Location', $Data);
-                break;
-            default:
-                $Result = '';
+        $Result = '';
+        Gdn::pluginManager()->fireAs('LogModel')->fireEvent('FormatContent', [
```
linc: See later in this file for how to properly use event firing in this class:

                $L = self::instance();
                $L->EventArguments['Log'] =& $LogRow;
                $L->fireEvent('BeforeInsert');

Issue #4089, opened 2016-06-09
Special categories (Social Groups, Subcommunities) should not be deletable


DaazKu:
We have some support request from client that deleted the Social Groups category since they did not know what it was about. These kind of "container" should not be deletable or even presented in the dashboard.

linc:
@tburry Can you look at this as part of the category types refactor in progress?

linc:
Step 1 vanilla/vanilla#4899

linc:
Unsure about the assertion that Subcommunities should be undeletable.

The PRs submitted sufficiently fixes this as far as core is concerned.

PR #4899, opened 2016-12-04
Add a CanDelete flag to categories

linc:
We've previously patched "special" categories being deleted by admins (thereby breaking addon functionality) in a variety of ways and it perennially re-breaks. It's time for a toggle on the database record that can be used consistently across Vanilla's addons.

`CanDelete` is what we've called this flag in other scenarios.

See https://github.com/vanilla/vanilla/issues/4089


applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -460,6 +460,10 @@ public function deleteCategory($CategoryID = false) {
         // Get category data
         $this->Category = $this->CategoryModel->getID($CategoryID);
 
+        // Block deleting special categories.
+        if (val('CanDelete', $this->Category) === 0) {
```
tburry: I'd do:

```php
if (val('CanDelete', $this->Category, true)) {
```

You want this to be resilient for the case where the database isn't updated. I also think the database brings the value in as a string doesn't it?

applications/vanilla/settings/structure.php
```diff
@@ -40,6 +40,7 @@
     ->column('DateMarkedRead', 'datetime', null)
     ->column('AllowDiscussions', 'tinyint', '1')
     ->column('Archived', 'tinyint(1)', '0')
+    ->column('CanDelete', 'tinyint', '1')
```
tburry: The role table has a similar column. In its case its called "Deletable".

tburry:
I'm thinking of the case where a plugin adds a category that can't be deleted and then gets disabled. Should such plugins be responsible for their categories?

linc:
I considered that; I think they should be responsible for restoring the CanDelete flag to their category on disable.

applications/vanilla/settings/structure.php
linc: I chose `CanDelete` because it's a column for Groups and for Badges and is used in logic 50+ other places. `Deletable` sounds nicer but it's used exactly that once and looked an awful lot like `Deleted` which is another column we use.

PR #4900, opened 2016-12-05
Allow form errors in ConfigurationModule

R-J:
Make it easier to do validations when using the ConfigurationModule. This will allow using a validation before the call to the initialize method:
~~~
$configurationModule = new ConfigurationModule($sender);
if ($sender->Form->authenticatedPostBack()) {
    $sender->Form->validateRule('plugin.field1', 'ValidateRequired');
}
$configurationModule->initialize(...
~~~
That would prevent saving a wrong info to the config and it will give developers the possibility to validate forms as usual.

### DIFF
--- a/applications/dashboard/modules/class.configurationmodule.php
+++ b/applications/dashboard/modules/class.configurationmodule.php
@@ -104,6 +104,10 @@ class ConfigurationModule extends Gdn_Module {
         $Form = $this->form();
 
         if ($Form->authenticatedPostBack()) {
+            if ($Form->errorCount() > 0) {
+                return;
+            }
+
             // Grab the data from the form.
             $Data = array();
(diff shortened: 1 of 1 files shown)


initvector:
The other end of that condition block already has [a check for form errors before saving](https://github.com/vanilla/vanilla/blob/3f7ef8e63cf0832eac4da0e312b45a111088ac3d/applications/dashboard/modules/class.configurationmodule.php#L140).  What kind of scenario would this earlier check prevent?

R-J:
In Vanilla 2.3 it wasn't there: https://github.com/vanilla/vanilla/blob/release/2.3/applications/dashboard/modules/class.configurationmodule.php#L140

That's why I tried to solve an issue which has already been solved - I should have taken a look at master first...

PR #4913, opened 2016-12-07
Add checking for CategoryID for permissions

initvector:
This update alters `CategoryModel::checkPermission` to consider `CategoryID`, in addition to `PermissionCategoryID`.  It also swaps out some existing per-category permission checks to use this revised function.


tburry:
I'm putting this on hold for now because we don't need it for our current use-case.

tburry:
I pushed this off to next spring. You can do some testing then @initvector.

linc:
@initvector Would you resolve the conflicts?

@DaazKu Wanna kick the tires on this? This is the sort of thing that generates support requests if we miss anything.

initvector:
Conflicts: Resolved :godmode:

DaazKu:
I can tackle this tomorrow probably!

Issue #4919, opened 2016-12-08
Cannot save HTML email styles setttings unless there is a photo


beckyvb:
You are forced to add an email logo to save the form in email settings.

PR #4920, opened 2016-12-08
Fix necessity to add image to email styles form

beckyvb:
Rely on UploadImage to validate the image and throw an error if the upload is incorrect and remove adding a form error if no image is found.

Closes https://github.com/vanilla/vanilla/issues/4919


DaazKu:
screen shot 2016-12-14 at 4 30 45 pm

Issue #4697, opened 2016-10-20
Selecting "All" for users when confirming SPAM flagging does not ban all listed users


initvector:
When flagging records from the site's SPAM log, you're met with a confirmation prompt.  Part of this prompt includes a list of the users associated with the selected records, as well as an ["All" checkbox](https://github.com/vanilla/vanilla/blob/5e54d4cc227ac3a084535d24a92d936f850a75b6/applications/dashboard/views/log/confirm.php#L60).  This checkbox doesn't actually do anything.  `LogController::deleteSpam` only cares about the [individually selected user checkboxes](https://github.com/vanilla/vanilla/blob/5e54d4cc227ac3a084535d24a92d936f850a75b6/applications/dashboard/controllers/class.logcontroller.php#L131).

PR #4922, opened 2016-12-09
Fix selectAll in the spam queue

DaazKu:
Fix https://github.com/vanilla/vanilla/issues/4697

Make the "All" checkbox actually check all other checkboxes!

### DIFF
--- a/applications/dashboard/views/log/confirm.php
+++ b/applications/dashboard/views/log/confirm.php
@@ -57,14 +57,31 @@
                     <?php
                     if (count($this->data('Users')) > 1) {
                         echo '<div class="CheckBoxCell">';
-                        echo $this->Form->CheckBox('SelectAll', t('All'));
+                        echo wrap(
+                            $this->Form->checkBox(
+                                'SelectAll',
+                                t('All'),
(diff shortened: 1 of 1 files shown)


applications/dashboard/views/log/confirm.php
```diff
@@ -57,14 +57,31 @@
...
 
                     foreach ($this->data('Users') as $User) {
                         $RecordUser = Gdn::userModel()->getID($User['UserID'], DATASET_TYPE_ARRAY);
                         echo '<div class="CheckBoxCell">';
-                        echo $this->Form->CheckBox('UserID[]', htmlspecialchars($User['Name']), array('value' => $User['UserID'], 'class' => 'checkbox checkbox-inline'));
+                        echo wrap(
+                            $this->Form->checkBox(
+                                'UserID[]',
+                                htmlspecialchars($User['Name']),
+                                array('value' => $User['UserID'], 'display' => 'after', 'class' => 'js-check-me checkbox checkbox-inline')
```
initvector: While you're already in here, we can go ahead and update this array to use the short syntax.

Issue #4700, opened 2016-10-20
BBCode quotes assume post is comment


initvector:
If a site's configured input format is BBCode and you click the "Quote" button on a discussion, the input field for your comment will be populated with a tag that only carries over the ID of the post.  There's no indication whether or not the ID is associated with a comment or discussion.  For example:

```
[quote="zer0kool;1337"]mess wit da best die like da rest[/quote]
```

When the code is parsed, it is assumed we're dealing with a comment ID, so [a comment URL is constructed](https://github.com/vanilla/vanilla/blob/b2ca7c6410157897beeb1b245d3df6372a8d0524/library/core/class.bbcode.php#L132).  In this case, we'd be linked to a comment with an ID of 1337, instead of a discussion with that ID.

PR #4928, opened 2016-12-12
Add distinction between comments and discussion in BBCode quotes

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/4700

The fix is not retroactive.

### DIFF
--- a/library/core/class.bbcode.php
+++ b/library/core/class.bbcode.php
@@ -128,7 +128,13 @@ class BBCode extends Gdn_Pluggable {
         if (isset($params['url'])) {
             $url = trim($params['url']);
 
-            if (is_numeric($url)) {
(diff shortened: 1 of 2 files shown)


library/core/class.bbcode.php
```diff
@@ -128,7 +128,13 @@ function doQuote($bbcode, $action, $name, $default, $params, $content) {
         if (isset($params['url'])) {
             $url = trim($params['url']);
 
-            if (is_numeric($url)) {
+            if (preg_match('/(c|d)-(\d+)/', strtolower($url), $matches)) {
+                if ($matches[1] === 'd') {
+                    $url = "/discussion/$matches[2]";
+                } else {
+                    $url = "/discussion/comment/$matches[2]#Comment_$matches[2]";
```
initvector: For the sake of readability, do you mind wrapping the variables here (and above) in braces?

Issue #4887, opened 2016-12-02
Refactor the MenuModule to use Gdn::session()->hasPermission()


tburry:
The [MenuModule](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/modules/class.menumodule.php) has a call to fetch the permissions as an array and then duplicates permission checking logic. This is a BC breaking bug waiting to happen. Use the proper method to check permissions.

PR #4930, opened 2016-12-12
Check permissions with Gdn_Session::checkPermission in MenuModule

DaazKu:
The menu module does not seem to be used anywhere anymore.

If the reviewer comes to the same conclusion we should file an issue about deprecating/removing it.

Fix https://github.com/vanilla/vanilla/issues/4887

### DIFF
--- a/applications/dashboard/modules/class.menumodule.php
+++ b/applications/dashboard/modules/class.menumodule.php
@@ -158,14 +158,12 @@ if (!class_exists('MenuModule', false)) {
             $Username = '';
             $UserID = '';
             $Session_TransientKey = '';
-            $Permissions = array();
             $Session = Gdn::session();
             $Admin = false;
             if ($Session->isValid() === true) {
                 $UserID = $Session->User->UserID;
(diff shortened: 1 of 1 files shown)


initvector:
The Dashboard application's bootstrap, [which is called by the primary bootstrap](https://github.com/vanilla/vanilla/blob/3d8751711d3c8f21e8bdade9a663aa854182bb0d/bootstrap.php#L230), assigns a copy of this module [to all controllers](https://github.com/vanilla/vanilla/blob/3d8751711d3c8f21e8bdade9a663aa854182bb0d/applications/dashboard/settings/bootstrap.php#L19) as the `Menu` property.

DaazKu:
Yes but is is never actually printed anywhere.

initvector:
The `Menu` property of the current controller is called by [smarty_function_custom_menu](https://github.com/vanilla/vanilla/blob/3f7ef8e63cf0832eac4da0e312b45a111088ac3d/library/SmartyPlugins/function.custom_menu.php#L18).

PR #4931, opened 2016-12-12
Fix embed-comments when installed in subdirectory

korelstar:
I had vanilla installed in a subdirectory of my domain and wanted to embed the forum using the comments approach on an external site. This didn't work without adding the path variable like in this patch.

### DIFF
--- a/js/embed.js
+++ b/js/embed.js
@@ -231,7 +231,7 @@ window.vanilla.embed = function(host) {
         var result = '';
 
         if (embed_type == 'comments') {
-            result = '//' + host + '/discussion/embed/'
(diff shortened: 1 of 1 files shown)


DaazKu:
What version of vanilla are you using?

I tried using the master branch and I go no problems.

I installed vanilla under a "vanilla" folder and set `vanilla_forum_url='http://vanilla.dev/vanilla'` and set the vanilla_discussion_id to the proper DiscussionID and everything worked.

korelstar:
Strange: I tested it again, and now it works without my patch. Maybe, I've updated the vanilla core in the meanwhile.

PR #4932, opened 2016-12-12
Discussions: Show view counter only if analytics are enabled

korelstar:
The view counter is only increased if `Garden.Analytics.Enabled` is `true` (cp. https://vanillaforums.org/discussion/31374/enable-viewcount-on-discussions-but-disable-remote-analytics).

However, if analytics are disabled, it doesn't make any sense to show the view counter of discussions in the list of discussions. Therefore, I propose to introduce a respective query.

### DIFF
--- a/applications/vanilla/views/discussions/table_functions.php
+++ b/applications/vanilla/views/discussions/table_functions.php
@@ -17,9 +17,11 @@ if (!function_exists('WriteDiscussionHeading')) :
             <td class="BigCount CountReplies">
                 <div class="Wrap"><?php echo t('Replies'); ?></div>
             </td>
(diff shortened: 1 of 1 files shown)


linc:
I don't think this is true.

Discussion counts are tabulated locally regardless of whether the remote analytics is keeping track of your total pageviews. It's a separate system.

If in fact this is breaking down somewhere, I'd wanna find that issue rather than scuttling the view count.

korelstar:
Then there is a bug somewhere.

However, I propose to introduce a config flag to hide the view count. What about that?

linc:
If there's a bug preventing the local view count from incrementing, that should be solved. If you want to hide the count, you can do that with CSS or a customized view.

PR #4937, opened 2016-12-13
Support a JSON content type in the request

tburry:
Allow requests to be made with an application/json content type. In these cases the raw input will be json-decoded and passed as the post.

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -870,7 +870,7 @@ class Gdn_Request {
                 break;
 
             case self::INPUT_POST:
-                $argumentData = $_POST;
+                $argumentData = $this->decodePost($_POST, $_SERVER, 'php://input');
                 break;
 
             case self::INPUT_SERVER:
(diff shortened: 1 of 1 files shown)


library/core/class.request.php
```diff
@@ -897,6 +897,30 @@ protected function _setRequestArguments($paramsType, $paramsData = null) {
         $this->_RequestArguments[$paramsType] = $argumentData;
     }
 
+    /**
+     * Decode the environment's post depending on content type.
+     *
+     * @param array $post Usually the {@link $_POST} super-global.
+     * @param array $server Usually the {@link $_SERVER} super-global.
+     * @param array $inputFile Usually **php://input** for the raw input stream.
```
initvector: `$inputFile` should be a string, shouldn't it?

library/core/class.request.php
```diff
@@ -897,6 +897,30 @@ protected function _setRequestArguments($paramsType, $paramsData = null) {
...
+     */
+    private function decodePost($post, $server, $inputFile) {
```
initvector: What do you think about defaulting `$inputFile` to "php://input"?

DaazKu: I personally think that we should keep it like so.
We are defaulting arguments like crazy and, in most case, we have more refactor to do later on because of it.

initvector: My thinking here is `$inputFile` is almost never going to be anything other than "php://input". It's great that we can change that, for things like automated testing, but otherwise it's probably always going to be PHP's raw-body stream.

With this function being private and only called in one place, it doesn't matter much.  It just struck me as a good candidate for a default value.

tburry: You both have great points here. Since the method is private I'll go ahead and set the default, however @DaazKu is very correct about the general point.

tburry:
I made the changes requested in @initvector's review.

PR #4941, opened 2016-12-14
Create support for mapping response keys in SSO

nervoustwit:
In many of our SSO integrations the client will send back responses with keys that do not correspond with vanilla keys. This PR allows us to store a map of key names in GDN_AuthenticationProvider.Attributes that can be added to the provider array as $provider['KeyMap'].

Usage: create a plugin with a  settings form that saves to GDN_AuthenticationProvider. Include fields with names that are prefixed with "ProfileKeys" (e.g. "ProfileKeysPhoto"). When you update your provider settings fill in the ProfileKeysPhoto field with "picture". Now your provider will have an array $provider['KeyMap'] with a key value pair "Photo" => "picture". An SSO response with the value 'picture': 'https://imgur.com/mypic' can now be mapped to "Photo" by an SSO plugin.

This change is backwards compatible. If plugins are not saving translation keys it is ignored.

### DIFF
--- a/library/core/class.authenticationprovidermodel.php
+++ b/library/core/class.authenticationprovidermodel.php
@@ -50,6 +50,7 @@ class Gdn_AuthenticationProviderModel extends Gdn_Model {
         if (is_array($Attributes)) {
             $Row = array_merge($Attributes, $Row);
         }
+        $Row['KeyMap'] = self::getKeyMap($Attributes);
         unset($Row['Attributes']);
     }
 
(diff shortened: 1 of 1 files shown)


nervoustwit:
When the AuthenticationProviderModel creates the provider array it flattens out and merges any nested arrays in the Attributes field. My life would have been easier if it didn't do that.

linc:
This is awaiting @nervoustwit reevaluating his use of the Attributes here.

nervoustwit:
Since the crux of storing and retrieving the key map happens in individual SSO plugins ( in the settings forms and controllers ) I am closing this PR and will create a PRs in SSO plugins.

PR #4942, opened 2016-12-14
Make format options configurable in Gdn_Configuration

beckyvb:
Makes format options configurable in Gdn_Configuration.


library/core/class.configuration.php
```diff
@@ -61,6 +61,9 @@ class Gdn_Configuration extends Gdn_Pluggable {
     /** @var null @var The sort flag to use with ksort. */
     private $sortFlag = null;
 
+    /** @var array @var Format option overrides. */
```
DaazKu: Nuke the second @var. (Same above even if it was there already)

PR #4946, opened 2016-12-15
Improve maintenance mode

initvector:
This update performs a few improvements related to maintenance mode:

1. A new function, `inMaintenanceMode`, has been added.  This function returns true/false, depending on whether or not Garden.UpdateMode (maintenance mode) is enabled.
2. The dispatcher's direct call to check the config value of Garden.UpdateMode has been removed in favor of using `inMaintenanceMode`.
3. `Gdn_DatabaseStructure::__construct` now sets its `$alterTableThreshold` property to 0 (disabled) if the site is in maintenance mode.
4. A new utility endpoint has been added: /utility/maintenance.  This endpoint can enable or disable the maintenance mode status for a site, depending on the value of the `$updateMode` parameter it receives.  Additionally, this endpoint has the ability to receive a specific value for `$updateMode` (currently "2") to indicate the action was performed by an automated process.  This is done for the sake of being able to retrieve the current Garden.UpdateMode setting after maintenance mode has been disabled through this same endpoint.  If a site was in maintenance mode (Garden.UpdateMode = 1) and received an authorized request to /utility/maintenance?updateMode=2 (or /utility/maintenance/2), maintenance mode would still be enabled after an authorized request was made to /utility/maintenance/?updateMode=0.  A second call to the latter would fully disable maintenance mode for the site.

### DIFF
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -13,6 +13,9 @@
  */
 class UtilityController extends DashboardController {
 
+    /** A flag used to indicate the site was put into maintenance mode in an automated fashion. */
+    const MAINTENANCE_AUTO = 2;
+
     /** @var array Models to automatically instantiate. */
     public $Uses = array('Form');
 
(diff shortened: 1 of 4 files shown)


linc:
FYI @kaecyra

kaecyra:
So UpdateMode set by the console should still be '1' ?

linc:
That's my reading of this, yes.

Issue #4830, opened 2016-11-18
Banned Users should not have their LastIPAddress updated on login attempts


DaazKu:
We currently [use LastIPAddress](https://github.com/vanilla/vanilla/blob/5a1952acd41c99a8acb59bfc6cea141f268d2288/applications/dashboard/models/class.banmodel.php#L173) to show the list of users that were banned by an IP rule.

If a banned user tries to login from a different IP its LastIPAddress is updated and the user is not returned in the list of banned users.

PR #4961, opened 2016-12-22
Prevent user's visit information from being updated if banned

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/4830

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2934,8 +2934,8 @@ class UserModel extends Gdn_Model {
      * Updates visit level information such as date last active and the user's ip address.
      * @param $UserID
      * @param int|float $ClientHour
-     * @throws Exception
-     * @return bool True on success false on failure
+     * @throws Exception If the user ID is not valid.
+     * @return bool True on success, false if the user is banned or deleted.
      */
     public function updateVisit($UserID, $ClientHour = null) {
         $UserID = (int)$UserID;
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2933,7 +2933,7 @@ class UserModel extends Gdn_Model {
     /**
      * Updates visit level information such as date last active and the user's ip address.
      * @param int $UserID
-     * @param int|float $ClientHour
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -2932,18 +2932,24 @@ public function saveIP($userID, $IP, $dateUpdated = false) {
 
     /**
      * Updates visit level information such as date last active and the user's ip address.
-     *
-     * @param int $UserID
-     * @param string|int|float $ClientHour
+     * @param $UserID
+     * @param int|float $ClientHour
```
initvector: Should `$ClientHour` ever be a float?

applications/dashboard/models/class.usermodel.php
```diff
@@ -2932,18 +2932,24 @@ public function saveIP($userID, $IP, $dateUpdated = false) {
...
+     * @throws Exception
```
initvector: The parameters are self-explanatory without needing a description, but I'd like to see one for the exception.  Maybe something like:

```
* @throws Exception if the user ID is not valid.
```

applications/dashboard/models/class.usermodel.php
initvector: `null` is now a valid type for this parameter.

applications/dashboard/models/class.usermodel.php
```diff
@@ -2932,18 +2932,24 @@ public function saveIP($userID, $IP, $dateUpdated = false) {
...
+     * @return bool True on success false on failure
```
initvector: There's actually very specific conditions for false here, so this could be misleading.  How about...

```
* @return bool False if the user is banned or deleted, otherwise true.
```

applications/dashboard/models/class.usermodel.php
DaazKu: Actually null is the same as not passing an argument in this case. It is not a "valid" parameter.

DaazKu: $ClientHour can be a float.
https://en.wikipedia.org/wiki/List_of_UTC_time_offsets#UTC.2B13:45 for example.

PR #4964, opened 2016-12-26
Hide category from breadcrumb on New Discussion when categories are disabled

R-J:
Although "Vanilla.Categories.Use" is set to false, category is shown in breadcrumbs and was used for "New Discussion" module. Using no categories at all should hide categories in all aspects for users.

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -83,8 +83,12 @@ class DiscussionController extends VanillaController {
         $Category = CategoryModel::categories($this->Discussion->CategoryID);
         $this->permission('Vanilla.Discussions.View', true, 'Category', val('PermissionCategoryID', $Category, -1));
         if (c('Vanilla.Categories.Use', true)) {
-            $this->setData('CategoryID', $this->CategoryID = $this->Discussion->CategoryID, true);
+            $this->CategoryID = $this->Discussion->CategoryID;
+        } else {
+            $this->CategoryID = null;
         }
+        $this->setData('CategoryID', $this->CategoryID);
+
 
         if (strcasecmp(val('Type', $this->Discussion), 'redirect') === 0) {
             $this->redirectDiscussion($this->Discussion);
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -82,7 +82,9 @@ class DiscussionController extends VanillaController {
         // Check permissions.
         $Category = CategoryModel::categories($this->Discussion->CategoryID);
         $this->permission('Vanilla.Discussions.View', true, 'Category', val('PermissionCategoryID', $Category, -1));
-        $this->setData('CategoryID', $this->CategoryID = $this->Discussion->CategoryID, true);
+        if (c('Vanilla.Categories.Use', true)) {
+            $this->setData('CategoryID', $this->CategoryID = $this->Discussion->CategoryID, true);
+        }
 
         if (strcasecmp(val('Type', $this->Discussion), 'redirect') === 0) {
             $this->redirectDiscussion($this->Discussion);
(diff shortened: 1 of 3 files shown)


applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -82,7 +82,9 @@ public function index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
         // Check permissions.
         $Category = CategoryModel::categories($this->Discussion->CategoryID);
         $this->permission('Vanilla.Discussions.View', true, 'Category', val('PermissionCategoryID', $Category, -1));
-        $this->setData('CategoryID', $this->CategoryID = $this->Discussion->CategoryID, true);
+        if (c('Vanilla.Categories.Use', true)) {
+            $this->setData('CategoryID', $this->CategoryID = $this->Discussion->CategoryID, true);
```
initvector: We're going from a scenario where CategoryID in the controller's `$Data` array is always set to having it only set some of the time.  Since the controller's data array is public, there's a chance something might try to access a potentially-invalid index.

What about something like this?

```php
$this->CategoryID = c('Vanilla.Categories.Use', true) ? $this->Discussion->CategoryID : null;
$this->setData('CategoryID', $this->CategoryID, true);
```

applications/vanilla/modules/class.newdiscussionmodule.php
```diff
@@ -72,7 +72,7 @@ public function addButton($Text, $Url) {
      */
     public function toString() {
         // Set CategoryID if we have one.
-        if ($this->CategoryID === null) {
+        if ($this->CategoryID === null && c('Vanilla.Categories.Use', true)) {
```
initvector: Switch up the config check to be the first condition.  If that fails, there's no point in checking the CategoryID.

applications/vanilla/views/post/comment.php
```diff
@@ -45,7 +45,7 @@
 
                     echo '<span class="'.$CancelClass.'">';
                     echo anchor($CancelText, '/');
-                    if ($CategoryID = $this->data('Discussion.CategoryID')) {
+                    if ($CategoryID = $this->data('Discussion.CategoryID') && c('Vanilla.Categories.Use', true)) {
```
initvector: Check that config, first.

applications/vanilla/controllers/class.discussioncontroller.php
DaazKu: `$this->setData('CategoryID', $this->CategoryID);`
since `$this->CategoryID` is already defined

R-J:
Does it make sense to wait for a decision on [this question](https://github.com/vanilla/vanilla/issues/4973) before doing any changes here`?

initvector:
> Does it make sense to wait for a decision on this question before doing any changes here`?

Nah.  I don't think that needs to block this.  That task would require updating numerous existing references in other areas of the code.  Your couple additional references wouldn't be a big deal when it comes time to address the others.

linc:
@R_J still pursuing this?

applications/vanilla/controllers/class.discussioncontroller.php
R-J: I'm not sure about the `$this->CategoryID = null;`. It is quite verbose since it is defined in as a class var and never set, so it is already null. Changing this:

~~~
    /** @var array Unique identifier. */
    public $CategoryID;
~~~

to that:

~~~
    /** @var array Unique identifier. */
    public $CategoryID = null;
~~~

and leaving out the "else..." part I've added would be something I prefer. It is not as self explaining as the code above, but setting a variable to a "value" (null) it already has does feel awkward to me, too.

applications/vanilla/views/post/comment.php
R-J: I've changed that if condition some more, because I think using `if  ($a = $b)` is confusing and should be avoided. It is a mixture of `$a = $b; if ($b)` but I would always expect to see a "==" or "===" in an if condition. I don't think that is against the coding style, but when looking at it, my first reaction have been that this must be a typo.

R-J:
@linc I have indeed forgotten about that, thanks for reminding me!

PR #4976, opened 2017-01-03
Fix the quote button not appearing in multiple cases

DaazKu:
Fix a [regression bug](https://github.com/vanilla/vanilla/pull/4925/files#diff-65a59ade5d93c6930c77c9accc6d31ceR242) where the quote button would not appear in some case because 'Discussion' is not always an event's argument.

### DIFF
--- a/plugins/Quotes/class.quotes.plugin.php
+++ b/plugins/Quotes/class.quotes.plugin.php
@@ -259,7 +259,7 @@ class QuotesPlugin extends Gdn_Plugin {
             $ObjectID = 'Comment_'.$Args['Comment']->CommentID;
         } elseif ($discussion) {
             $Object = $discussion;
-            $ObjectID = 'Discussion_'.$Args['Discussion']->DiscussionID;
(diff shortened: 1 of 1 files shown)


plugins/Quotes/class.quotes.plugin.php
```diff
@@ -239,23 +239,26 @@ public function base_AfterFlag_handler($Sender, $Args) {
...
         }
 
         if (isset($Args['Comment'])) {
             $Object = $Args['Comment'];
             $ObjectID = 'Comment_'.$Args['Comment']->CommentID;
-        } elseif (isset($Args['Discussion'])) {
-            $Object = $Args['Discussion'];
+        } elseif ($discussion) {
+            $Object = $discussion;
             $ObjectID = 'Discussion_'.$Args['Discussion']->DiscussionID;
```
initvector: Does this need to go too?

```php
$ObjectID = "Discussion_{$Object->DiscussionID}";
```

Issue #4749, opened 2016-11-03
Confirm email setting with no unconfirmed role creates confusion


beckyvb:
If 'Garden.Registration.ConfirmEmail' is set without a role with type 'unconfirmed', the user gets the 'member'-type role which seems like it defeats the purpose of requiring users to confirm their email.

initvector:
Closing in favor of tracking on related PR: #5767

PR #4977, opened 2017-01-04
Assign Unconfirmed role to new users if confirmation is required

R-J:
By now the whole "user needs to confirm mail address" process is quite of buggy. Although the Unconfirmed role is explicitly described as the role a user gets if he needs to confirm his mail address, that role is not assigned. Instead of this, new users only get the Member role, which grants them more rights than the admin would have guessed.

That can only be "fixed" if Garden.Registration.DefaultRoles in the config is set to the unconfirmed role ID. But that is only a workaround, which might lead to unwanted behavior later on.
If the admin decides that his users will no longer need to require the mail confirmation by removing the tick from the "Require users to confirm their email addresses (recommended)" checkbox in /dashboard/settings/registration, new users still will be assigned to the Unconfirmed role after registration based on the DefaultRoles setting in the config.

I'd say that the problem described above can only be solved if the Unconfirmed role is assigned **dynamically** based on the "Require users to confirm..." setting.
The UserModels Insert method tests if the user needs to confirm his mail. I've added a line which merges the $ConfirmRoleIDs (if they exist) to the $Roles.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -916,15 +916,19 @@ class UserModel extends Gdn_Model {
             unset($Fields['Admin']);
         }
 
+        $Roles = val('Roles', $Fields);
+        unset($Fields['Roles']);
+
         // Massage the roles for email confirmation.
         if (self::requireConfirmEmail() && !val('NoConfirmEmail', $Options)) {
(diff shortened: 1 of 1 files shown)


initvector:
Thanks @R-J!

linc:
I believe this fixed https://github.com/vanilla/vanilla/issues/4749

PR #4978, opened 2017-01-04
Assign Unconfirmed role to new users if confirmation is required

R-J:
Backport of https://github.com/vanilla/vanilla/pull/4977 to 2.3

I wouldn't say it is a critical bug, but if admins expect new users to have a Unconfirmed role with **restricted** access rights, but new users immediate gain Member access, it is at least annoying and irritating.

So just in case there will be a minor release, it would be nice if this could be included ;-)


initvector:
Thanks again!

Issue #4666, opened 2016-10-17
Themes should indicate their expected layout in their about.php


beckyvb:
Many themes that are designed with a table layout look broken with a modern layout and vice-versa. The expected layout should be documented in the about.php file and enforced when a theme is enabled.

Further, theme preview should find this info and preview the correct layout.

PR #4979, opened 2017-01-04
Give themes the option to set their layout

beckyvb:
This allows a theme to set its expected layout in its about.php array. This layout will be saved to config when the theme is enabled. 

When previewing a theme, it will preview with the expected layout.

To test, edit the about.php of a theme and add
```
'Layout'      => [
    'Categories'  => 'modern',
    'Discussions' => 'modern'
]
```
Clear your cache and then enable the theme. Ensure that the layout on `/settings/homepage` has changed to what has been specified in the theme. Preview the theme and make sure that no matter what the config is set as, the theme previews with the settings specified in about.php.

Closes https://github.com/vanilla/vanilla/issues/4666


initvector:
:godmode:

Issue #4880, opened 2016-12-01
"Incorrect integer value" error on draft save when no category selected


michaeltyson:
When composing a new discussion, if one clicks "Save Draft" (or waits for the auto save interval) the following error appears on screen:

```
{ "Code": 256, "Exception": "Incorrect integer value: '' for column 'CategoryID' at row 1|Gdn_Database|Query|insert GDN_Draft \n(Body, CategoryID, InsertUserID, UpdateUserID, Name, Format, DateInserted, DateUpdated) \nvalues (:Body, :CategoryID, :InsertUserID, :UpdateUserID, :Name, :Format, :DateInserted, :DateUpdated)", "Class": "Gdn_ErrorException" }
```

Some digging reveals that this is because the CategoryID field is empty (suspect problems at  [models/class.draftmodel.php:161](https://github.com/vanilla/vanilla/blob/ec5b13f6713dae42cd835267cea328d8e8559432/applications/vanilla/models/class.draftmodel.php#L161)).

michaeltyson:
I've worked around the issue for now by introducing a default category setting. In applications/vanilla/controllers/class.postcontroller.php:107, I've added:

```
        if ($CategoryUrlCode == '' && C('Vanilla.Categories.Default')) {
            $CategoryUrlCode = C('Vanilla.Categories.Default');
        }
```

Then in config.php:

```
$Configuration['Vanilla']['Categories']['Default'] = 'General';
```

linc:
Is this in the current `master`, 2.2, or 2.3?

michaeltyson:
Sorry - this is in 2.3 (although it doesn't appear that routine has changed in master)

PR #4982, opened 2017-01-04
Allow draft saves on new posts

beckyvb:
Currently when adding a new draft, the draft ID form value will be retrieved as the string '0' and if there's no category, the category ID will be retrieved as and empty string. The DraftModel expects its falsey values for draft ID to be an empty string and Category ID to be === false. As such, we never reach the if ($Insert) block on line 162.

This PR adjusts the $Insert calculation and category ID check to ensure we properly calculate the category ID on a draft if it doesn't exist.

Closes https://github.com/vanilla/vanilla/issues/4880

### DIFF
--- a/applications/vanilla/models/class.draftmodel.php
+++ b/applications/vanilla/models/class.draftmodel.php
@@ -147,9 +147,9 @@ class DraftModel extends VanillaModel {
         }
 
         // Get the DraftID from the form so we know if we are inserting or updating.
-        $DraftID = val('DraftID', $formPostValues, '');
-        $Insert = $DraftID == '' ? true : false;
-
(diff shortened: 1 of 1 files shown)


applications/vanilla/models/class.draftmodel.php
```diff
@@ -147,9 +147,9 @@ public function save($formPostValues, $settings = []) {
         }
 
         // Get the DraftID from the form so we know if we are inserting or updating.
-        $DraftID = val('DraftID', $formPostValues, '');
-        $Insert = $DraftID == '' ? true : false;
-
+        $DraftID = (int) val('DraftID', $formPostValues, 0);
+        $Insert = $DraftID === 0 ? true : false;
+
```
DaazKu: Remove these spaces :P

PR #4992, opened 2017-01-09
Move or delete all contents when deleting a category

### LINKED ISSUE #4858: Subcategories not deleted when deleting parent category
Sub categories are dead in the air when deleting a category containing sub categories without moving its content.

screen shot 2016-11-24 at 2 15 51 pm

Wording could also be reviewed to make the whole process more explicit, deleting a parent category should specify that all sub categories will be deleted as well.

When checking the option to move discussions, it should specify that sub categories and discussions will be move into the replacement category.

initvector:
Deleting a category, the confirmation prompt only indicates discussions will be deleted or moved.  If the user chooses not to "move discussions", child categories may become orphaned.

This update alters `CategoryModel::deleteAndReplace` to recursively eliminate all children if no replacement category is provided.  If a replacement category is provided, all content should be moved over as expected.  In addition, the phrasing on the category deletion confirmation prompt has been revised to use more generic terms.

Closes #4858 

*Bonus: I pulled out some dead code, fixed some dynamically-declared properties and updated a deprecated call in the `VanillaSettingsController` class.*


linc:
@initvector conflicted

initvector:
Conflicts: Resolved :suspect:

linc:
@DaazKu Don't ninja into an approved PR. :(

DaazKu:
I spoke to Ryan first!

PR #4997, opened 2017-01-10
Use the garden container for dependency injection

tburry:
Add the [garden container](https://github.com/vanilla/garden-container) to Vanilla for dependency injection. This replaces the Gdn_Factory with additional functionality and will add dependency injection capabilities to controllers, plugins and more in the future.

This PR was developed in several stages.

1. Some factory based unit tests were written against the current Gdn_Factory class in order to test whether or not refactoring breaks backwards compatibility. The tests aren't exhaustive, but should catch a lot. c72516f
2. The Gdn_Factory class was refactored to use the Container. bcfcc85
3. Remove the Gdn_Factory from the Gdn class in facvor of just the container. 8046137

There is some obscure functionality of the Gdn_Factory that isn't supported by the Container.

- The factory has old (forgotten) support for configuring property-based dependencies. The Container has better support for dependencies, but in the form of constructors and method calls.

- The Gdn::factoryInstallFromConfig method has been deprecated.

Although the container is mostly backwards compatible with the factory, it does include some differences so testing should be done to ensure we aren't breaking anything.


linc:
Want a @kaecyra confirmation this isn't gonna break his use cases.

linc:
I'm getting `null` from `Gdn::cache()`.

screen shot 2017-02-02 at 2 35 20 pm

linc:
This code in `factoryInstall()` doesn't jive with the new signature of `setFactory()`, which is why I assume I'm not getting a cache layer installed.

```
     case Gdn::FactoryRealSingleton:
                $dic->setShared(true)
                    ->setFactory([$ClassName, $Data]);
                break;
```

tburry:
The EventManager PR wrecked this because of one of its dependencies requiring an explicit initialization. This is an issue with the container itself, but there is a workaround that I added. I'll update the container to leave Interface dependencies alone if there is no rule for them.

linc:
This doesn't work with DebugBar now.

Failure on enable:
> Failed stopping measure 'controller' because it hasn't been started

Or failure if it's already enabled:
> Failed stopping measure 'dispatcher' because it hasn't been started

Issue #4862, opened 2016-11-25
Show "Not Configured" warning on addon media view


linc:
We should use the "isConfigured()` method to generically check all plugins for whether they are configured and display a prominent "NOT CONFIGURED" warning anywhere addons are listed. This check is currently implemented in the socialController but the 'configured' addonInfo flag isn't used/checked in the view.

This would resolve a to-do created by #4860.

PR #5007, opened 2017-01-12
Add Configured messages to addons media view

beckyvb:
If an isConfigured method exists in an addon, adds a 'Configured' message to the addons media item view.

Also adds new semantic badges to admin.css and recompiles and updates the styleguide.

Closes https://github.com/vanilla/vanilla/issues/4862


linc:
@beckyvb conflicted

PR #5010, opened 2017-01-13
Store last post info in the category table

### LINKED ISSUE #3847: Store the aggregate last post in the category table

initvector:
This update reworks the way the aggregate last post fields (`LastTitle`, `LastUrl`, etc.) are generated for categories.  Currently, this data is queried anytime a user views the category list.  These changes cache the data in a `LastPost` field on the table.  If the field is populated, its values are used.  If not, the values are first generated and cached before use.

I opted to put these values into a single DB-encoded `LastPost` field.  I went this route, because we can expand the aggregate values we collect in the future without additional database updates.  Currently, we have at least three "last" fields in the category table (`LastDiscussionID`, `LastCommentID` and `LastDateInserted`).  We could add ~4 more, but I don't know how much value there would be there if we weren't going to be querying those fields.

*But*, if it's deemed better to go with making them individual DB fields, that shouldn't be too tricky.

Closes #3847


initvector:
We're going to be taking a different approach to this.  New issue filed: https://github.com/vanilla/vanilla/issues/5032

Issue #4795, opened 2016-11-14
Add flood control for activity & activity comments


linc:
Add a combined flood control setting covering both.

PR #5015, opened 2017-01-16
Add flood control to Activity and ActivityComment

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/4795

You will need to run `composer update` to update the autoloader.
`ActivityModel` now extends `VanillaModel` in order to have the renamed function `isSpamming`

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -11,7 +11,7 @@
 /**
  * Activity data management.
  */
-class ActivityModel extends Gdn_Model {
+class ActivityModel extends VanillaModel {
 
     /** Activity notification level: Everyone. */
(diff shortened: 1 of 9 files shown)


linc:
What's breaking the test build?

applications/dashboard/models/class.activitymodel.php
```diff
@@ -11,7 +11,7 @@
 /**
  * Activity data management.
  */
-class ActivityModel extends Gdn_Model {
+class ActivityModel extends VanillaModel {
```
linc: Nope. That's a separate application entirely. You've just created a cross-app dependency.

linc:
Closing this so we can rethink our flood control rather than kludging.

PR #5016, opened 2017-01-16
Separate image upload and file upload in editor

beckyvb:
* Adds accept parameter to file inputs based on whether it's an image and Garden.Upload.AllowedFileExtensions
* Adds appropriate mime types to the file input accept parameter.
* Removes drop zones from upload dropdowns, as they were never accessible.
* Changes logic to either add and attachment or insert image but not both.
* Removes logic to reinsert an image into a post by clicking on the attachment.
* Removes CSS based show/hide upload logic.

Closes https://github.com/vanilla/lithe/issues/53


plugins/editor/class.editor.plugin.php
```diff
@@ -613,7 +616,38 @@ public function base_render_before($Sender) {
...
+            }
+        }
+
+        // prepend extensions with a '.'
+        $allowedFileExtensions = array_map(function($str) { return '.'.$str; }, $allowedFileExtensions);
+        $accept = implode(',', array_merge($allowedFileExtensions, $allowedMimeTypes));
+        $Sender->setData('Accept', $accept);
+
+        // prepend extensions with a '.'
+        $allowedImageExtensions = array_map(function($str) { return '.'.$str; }, $allowedImageExtensions);
```
initvector: You're also using the same anonymous function code in two different areas.  You can save it to a variable and pass that to `array_map` for each call.  Also, the function code should probably be broken up across multiple lines.

PR #5019, opened 2017-01-17
Deprecate all dead code in UpdateModel

DaazKu:
Only `runStructure()` is used now.
I think that everything else was replaced by the AddonsManager.


DaazKu:
I added a future issue to cleanup the deprecated code https://github.com/vanilla/vanilla/issues/5020

PR #5027, opened 2017-01-19
Remove slicing from Vanilla

beckyvb:
* Removes the Gdn_Slice, Gdn_SliceProvider and related assets from Vanilla.
* Throw an exception if any of the methods provided by Gdn_SliceProvider are called within an object extending Gdn_Pluggable.
* Remove any lingering 'Slice*' classes from Vanilla. (The 'enablePlugin' method was needed to add the assets, if it was not called within a plugin, the functionality was never there to begin with.)


kaecyra:
Goodnight sweet prince

tburry:
🌹 ⚰️ 🌹

PR #5031, opened 2017-01-20
Configure the container in the bootstrap

tburry:
Configure the container in the bootstrap instead of using the Gdn object. Although the code looks like it has vastly changed, the functionality is still very similar. The process now consists of the following steps.

1. Rather than calling Gdn::factoryInstall(), the bootstrap now configures rules in the container. The extra features of the container means that order doesn't matter as much so all the configuration can be done first before any object is constructed.

2. The old bootstrap was refactored into an inline function called through Container::call() with references to static Gdn::* calls replaced with dependencies.

3. The last part of the bootstrap was left mostly as is.

*This is another **big scary** PR, but is necessary to unlock all of the features of the container.*


linc:
Irony is that dangling semi-colon being what caused the merge conflict.

PR #5036, opened 2017-01-23
Update dependencies and fix lint/hint notices

beckyvb:
* Update the dependencies in the dashboard application to work with node 7. 
* Fix many notices in javascript and scss files. 
* Recompile everything.


initvector:
Works for me!

Just need to resolve those conflicts...

PR #5052, opened 2017-01-25
Replace various usages of deprecated functions

R-J:
This PR replaces calls to deprecated functions, as well as deprecated event names.
Also `$this->getView()` in plugins is changed to  `$sender->fetchView/$sender->fetchViewLocation ` so that plugins views are themeable


DaazKu:
Always test your changes even if you are confident that everything will work!
The flagging plugin is actually broken.

```php
$Sender->fetchView('reportcomment.php', '', 'plugins/Flagging');
// Should be 
$Sender->fetchView('reportcomment', '', 'plugins/Flagging');
```

Issue #5068, opened 2017-01-29
Russian transliteration failing on one character [report]


linc:
https://open.vanillaforums.com/discussion/33055/how-to-fix-a-bug-with-the-letter-i-i-in-the-link#latest

Character is already in the transliteration array in Gdn_Format: `'И' => 'I',`

PR #5069, opened 2017-01-29
Add russian letter "И" to function url()

R-J:
The russian letter "И" was missing from class Formats `protected static $_UrlTranslations`. Closes  https://github.com/vanilla/vanilla/issues/5068


linc:
screen shot 2017-01-29 at 12 09 02 pm

linc:
I think you've perhaps actually found the root of the issue: It was already in there twice.

R-J:
It is not in the 2.3 release, but it double in the master branch. Seems like that issue has been addressed before, but handled wrong.

R-J:
Seems to come from here: https://github.com/vanilla/vanilla/commit/7762cdb854f0ddd61e1eb2075202f4d120f97b99

Issue #5045, opened 2017-01-24
[Refactoring] Move flood control out of the models


DaazKu:
Flood control is currently checked by `checkForSpam` from VanillaModel.
We will instead create a FloodControl class which will handle all check related to flood control.

The current `checkForSpam` method will instead call FloodControl and will be deprecated.

Important notes about `checkForSpam`:
- There is currently an attribute `SpamCheck` that can be set on the model to prevent the flood control to be checked.
- The model's validation's `addValidationResult` was called when a user was spamming.

Related: https://github.com/vanilla/vanilla/pull/5015

DaazKu:
Reopening for sprint purpose.

PR #5085, opened 2017-01-30
Move flood control handling to its own class and add it to activities

DaazKu:
Better version of https://github.com/vanilla/vanilla/pull/5015.
Fixes  https://github.com/vanilla/vanilla/issues/5045 and https://github.com/vanilla/vanilla/pull/5015

- new FloodControl class
- deprecate VanillaModel
- ensure that everything is backward compatible (validation, SpamCheck attribute on the models)


tburry:
I think this should be a trait. I know you suggested that @DaazKu so sorry for that. Let's chat about an attack plan here.

DaazKu:
Closing in favour of https://github.com/vanilla/vanilla/pull/5085

PR #5090, opened 2017-01-31
Improve password requirement message

### LINKED ISSUE #2708: Password requirements
By default, the only password requirement is that it is only six characters. However, a six-character, simple password (like abcdef) will still not allow you to register because the password is "very weak". Perhaps indicating that this should be at least 8 characters would work better? Otherwise, indicating it also requires at least a "weak" strength would be recommended for a better user experience.

### LINKED ISSUE #5047: MinLength does not change "Your password must be at least %d characters long."
Changing MinLength doesn't actually change the phrase "Your password must be at least 6 characters long." or make the meter say that the password is "too short" in the generated visual.

DaazKu:
and use the correct config for min length

`Garden.Registration.MinPasswordLength` was used only user for UI display so I replaced it with  `Garden.Password.MinLength` which is used to validate the password length

Fixes #5047
Fixes #2708

### DIFF
--- a/applications/dashboard/views/entry/registerbasic.php
+++ b/applications/dashboard/views/entry/registerbasic.php
@@ -33,7 +33,7 @@
             <li>
                 <?php
                 echo $this->Form->label('Password', 'Password');
-                echo wrap(sprintf(t('Your password must be at least %d characters long.'), c('Garden.Password.MinLength')).' '.t('For a stronger password, inscrease its length and use a combination of lower/upper case letters, digits and special characters such as @!#$.'), 'div', array('class' => 'Gloss'));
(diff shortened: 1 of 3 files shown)


applications/dashboard/views/entry/registerbasic.php
```diff
@@ -33,7 +33,7 @@
             <li>
                 <?php
                 echo $this->Form->label('Password', 'Password');
-                echo wrap(sprintf(t('Your password must be at least %d characters long.'), c('Garden.Registration.MinPasswordLength')), 'div', array('class' => 'Gloss'));
+                echo wrap(sprintf(t('Your password must be at least %d characters long.'), c('Garden.Password.MinLength')).' '.t('For a stronger password, inscrease its length and use a combination of lower/upper case letters, digits and special characters such as @!#$.'), 'div', array('class' => 'Gloss'));
```
linc: > For a stronger password, increase its length or combine upper and lowercase letters, digits, and symbols.

Issue #5048, opened 2017-01-24
Apply Password constraints to the Password Reset Page


LauraMakesArt:
Currently sending yourself a password reset link allows you to circumvent all of the password creation constraints, but we should apply all the same ones that we do at registration to the reset.

linc:
release/2.4 if it's forked by then

PR #5091, opened 2017-01-31
Validate password constraints on the reset password form

DaazKu:
Depends on #5090
Fixes: #5048

Add validation on the reset password form! Min length, min score, password matching...

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -1747,6 +1747,7 @@ class EntryController extends Gdn_Controller {
      */
     public function passwordReset($UserID = '', $PasswordResetKey = '') {
         $PasswordResetKey = trim($PasswordResetKey);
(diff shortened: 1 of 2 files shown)


linc:
This pull request arbitrarily broke backwards compatibility by changing the property name of 'Confirm' to 'PasswordMatch'. This should either be avoided entirely, compensated for with additional code, or broadly announced - in that order of preference.

DaazKu:
This will not change the fact that we broke backward compatibility with that change but we changed the field name because [entry.js is using PasswordMatch](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/js/entry.js#L100) to do its validation and the [validation rule](https://github.com/vanilla/vanilla/pull/5091/files#diff-08849f9e0a03bff7685d4ddc673d3f65R1790) uses ["{$Field}Match"](https://github.com/vanilla/vanilla/blob/master/library/core/functions.validation.php#L566)

PR #5115, opened 2017-02-02
Hide images in Feed on dashboard homepage

beckyvb:
Hide images in Feed on dashboard homepage.


adrianspeyer:
@initvector It should be 

```
div.FeedDescription div.summary a img {
display: none;
}

```

Not sure how to add this with the new CSS we are using, but the current change Becky had did not work when I turned on the images for the RSS feed, so I had to turn it off.

PR #5143, opened 2017-02-06
Fix three typos in the Helpers section of the styleguide

R-J:


R-J:
Oh, yes setting up that build process might require too much work for me and I really would prefer to simply change the scss file. But since I've already changed the html file, what about adding the changed scss file to this PR? Wouldn't the effect be exactly the same?

R-J:
Since this is only some branch I thought it couldn't harm to do simply add the scss file. Is it okay for you the way it is? Or is there a way to "remove" files from a branch (other than simply editing in back to the way it is in the master branch). Or do you need another PR?

PR #5144, opened 2017-02-06
Use PHP 5.6 variadics in the EventManager

tburry:


library/Garden/EventManager.php
```diff
@@ -287,27 +285,26 @@ private function callMetaHandlers($event, array $args, $result) {
...
             return $value; // gotcha, return value
         }
 
-        $callArgs = $args = array_slice(func_get_args(), 1);
-
+        $result = $value;
         foreach ($handlers as $callback) {
-            $value = call_user_func_array($callback, $callArgs);
-            $callArgs[0] = $value;
+            $result = call_user_func($callback, $result, ...$args);
```
DaazKu: TIL that you can unpack variadic arguments!
http://php.net/manual/en/functions.arguments.php#example-145
@beckyvb @initvector @linc

PR #5146, opened 2017-02-06
Install the addon managers earlier

tburry:
The cache currently fires an event on initialization which means that it requires the plugin manager before its available. This means that the plugin manager’s configuration needs to be moved up.

Before the container, the plugin manager would just return null before being configured. Because of auto-wiring that’s not the case with the container.


tburry:
This should fix the bootstrap on infrastructure.

PR #5147, opened 2017-02-06
Remove events from the cache

tburry:
The cache is meant to be an “early” object that is available before event handlers have been registered. We shouldn’t have plugins thinking they can do something to the cache during its activation.

The correct way to change cache behavior is to make a new cache driver.


tburry:
I assigned this one to you @kaecyra because you originally did the caching stuff. In my estimation these events aren't needed, but I want you to have a look and see if inf needs them.

PR #5150, opened 2017-02-06
Feature/addon manager register autoload

tburry:
Register the AddonManager autoloader right after the first request. In this way the AddonManager will not be created too early and another line from the bootstrap no longer depends on order.

Depends on #5031.


tburry:
Closing because I goofed the rebase.

PR #5153, opened 2017-02-07
Bind addon events to the EventManager directly

tburry:
Depends on #5031.

Register plugin events to the EventManager directly, bypassing the need for Gdn_PluginManager. A side-effect of this PR is that plugins are now created through the container which means that plugin authors can declare dependencies for their plugins too--that's a beautiful thing.


linc:
#5031 is merged.

PR #5155, opened 2017-02-08
Ensure Memcached keys are valid

### LINKED ISSUE #4996: Caching and Whitespace in Category Slugs
Vanilla uses category slugs in some of its category cache keys.  [Whitespace is not allowed in memcached keys.](https://github.com/memcached/memcached/blob/d9dfbe0e2613b9c20cb3c4fdd3c55d1bf3a8c8bd/doc/protocol.txt#L48-L49)  If whitespace characters are included in the category slug, it can errors when communicating with the caching server.

Vanilla should either block whitespace characters from being part of category slugs or filtering them out [when generating slug-based category keys](https://github.com/vanilla/vanilla/blob/3f9eef442246f058b82084325e2584a3a523516e/applications/vanilla/library/class.categorycollection.php#L241).

initvector:
This update alters `Gdn_Cache::makeKey`, which is used by `Gdn_Memcached::set` and `Gdn_Memcached::get`, to ensure the strings generated are [valid Memcached keys](https://github.com/memcached/memcached/blob/d9dfbe0e2613b9c20cb3c4fdd3c55d1bf3a8c8bd/doc/protocol.txt#L41). This is done by stripping out any whitespace or control characters from the keys and making certain they are no longer than 250 characters.

Closes #4996

### DIFF
--- a/library/core/class.cache.php
+++ b/library/core/class.cache.php
@@ -673,17 +673,17 @@ abstract class Gdn_Cache {
 
         // Make sure key is valid: no control characters or whitespace and no more than 250 characters.
         // See https://github.com/memcached/memcached/blob/master/doc/protocol.txt
+        $Result = trim($Result);
         if (unicodeRegexSupport()) {
             // No whitespace or control characters.
-            $replacePattern = '/[\p{Z}\p{C}]/u';
+            $Result = preg_replace('/[\p{Z}\p{C}]+/u', '-', $Result);
         } else {
             // No whitespace.
-            $replacePattern = '/[\s]/';
+            $Result = preg_replace('/\s+/', '-', $Result);
         }
-        $Result = preg_replace($replacePattern, '-', $Result);
 
-        // Clean up any leading, trailing and consecutive dashes.
-        $Result = trim(preg_replace('/-+/', '-', $Result), '-');
+        // Clean up any leading or trailing dashes.
+        $Result = trim($Result, '-');
 
         if (strlen($Result) > 250) {
             $Result = substr($Result, 0, 250);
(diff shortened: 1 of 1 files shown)


library/core/class.cache.php
```diff
@@ -671,6 +671,23 @@ public function makeKey($Key, $Options) {
             $Result = $Prefix.$Key;
         }
 
+        // Make sure key is valid: no control characters or whitespace and no more than 250 characters.
+        if (unicodeRegexSupport()) {
+            // No whitespace or control characters.
+            $replacePattern = '/[\p{Z}\p{C}]/u';
+        } else {
+            // No space, newline, carriage return or tab characters.
+            $replacePattern = '/[\s\n\r\t]/';
```
DaazKu: http://php.net/manual/en/regexp.reference.escape.php

\s -> any whitespace characte

The "whitespace" characters are HT (9), LF (10), FF (12), CR (13), and space (32). However, if locale-specific matching is happening, characters with code points in the range 128-255 may also be considered as whitespace characters, for instance, NBSP (A0).

DaazKu: Also a quick link to 
> https://github.com/memcached/memcached/blob/master/doc/protocol.txt See: Keys

could be nice

library/core/class.cache.php
```diff
@@ -671,6 +671,24 @@ public function makeKey($Key, $Options) {
...
+        if (unicodeRegexSupport()) {
+            // No whitespace or control characters.
+            $Result = preg_replace('/[\p{Z}\p{C}]+/u', '-', $Result);
+        } else {
+            // No whitespace.
+            $Result = preg_replace('/\s+/', '-', $Result);
+        }
+
+        // Clean up any leading or trailing dashes.
+        $Result = trim($Result, '-');
```
DaazKu: That trim only introduces a risk of collision.

PR #5164, opened 2017-02-09
Add the ClassLocator

tburry:
Currently we dispatch by looking up classes with class_exists() and method_exists(). We also use events though which is a Vanilla peculiarity. What I want to do is create a class abstraction on class/method locating so that we can modify the behavior more easily.

### DIFF
--- a/tests/Library/Vanilla/VanillaClassLocatorTest.php
+++ b/tests/Library/Vanilla/VanillaClassLocatorTest.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+namespace VanillaTests\Library\Vanilla;
+
+use Garden\Container\Container;
+
+class VanillaClassLocatorTest extends \PHPUnit_Framework_TestCase {
+
+    public function testFindMethod() {
+        $container = new Container();
+        $eventManager = $container->get('Garden\\EventManager');
+        $container->setInstance('Garden\\EventManager', $eventManager);
+
+        $basicEventHandlers = $container->get('VanillaTests\\Fixtures\\BasicEventHandlers');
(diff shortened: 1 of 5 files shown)


tburry:
@initvector can you finish the doc blocks for these classes and write some unit tests?

tests/Library/Vanilla/VanillaClassLocatorTest.php
```diff
@@ -0,0 +1,29 @@
...
+
+        $vanillaClassLocator = $container->get('Vanilla\\VanillaClassLocator');
+        $someController = $container->get('VanillaTests\\Fixtures\\SomeController');
+        $method = $vanillaClassLocator->findMethod($someController, 'someEndpoint');
+
+        $this->assertNotNull($method);
+        $this->assertSame($basicEventHandlers, $method[0]);
+        $this->assertSame('somecontroller_someendpoint_method', $method[1]);
+    }
+}
```
tburry: Looking at this method I was surprised it passed due to the container configuration. I had to debug it to see how it passed which is a bit confusing there. One thing you could have done was to just mark everything shared in your test container so that you wouldn't have to bind everything.

Here is another version of your method to show you how the container can be used.

```php
    public function testFindMethod() {
        $container = new Container();
        $container
            ->setInstance(ContainerInterface::class, $container)

            ->defaultRule()
            ->setShared(true)

            ->rule(EventManager::class)
            ->addCall('bindClass', [BasicEventHandlers::class])
        ;

        $vanillaClassLocator = $container->get('Vanilla\\VanillaClassLocator');
        $method = $vanillaClassLocator->findMethod($container->get(SomeController::class), 'someEndpoint');

        $this->assertNotNull($method);
        $this->assertSame($container->get(BasicEventHandlers::class), $method[0]);
        $this->assertEquals('somecontroller_someendpoint_method', $method[1]);
    }
```

PR #5165, opened 2017-02-09
Add jsonFilter function and tests

### LINKED ISSUE #5156: Write the jsonFilter() function
As we start making types more restrictive with the Schema and also use JsonSerializable more we will start having issues where json_encode() fails because of one bad value. We already had this happen with dbencode(). To help with this we want a **jsonFilter()** function that can be called on a variable before json_encode().

The things we need to filter now are the following.

- DateTimeInterface: ->format('r'). This is the main one.
- Unpack IP addresses in a similar way to how dbdecode() does it.
- We will add others as separate PRs.

Here is a suggested stub.

```php
function jsonFilter(&$value) {
   $fn = function (&$value, $key = '') use ($fn) {
        // ...
   }

    if (is_array($value)) {
        array_walk_recursive($value, $fn);
    } elseif (...) {
        // ...
    } else {
        $fn($value);
    }
}
```

initvector:
This update adds a global `jsonFilter` function, which is intended to prepare data for conversion to JSON via `json_encode`. Currently supported are:

1. Converting `DateTimeInterface` subclasses (e.g. `DateTime`) from objects to their RFC 2822 representations
2. Converting packed IP addresses to their human-readable forms with `ipDecode`.

Closes #5156

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -2152,6 +2152,23 @@ if (!function_exists('jsonEncodeChecked')) {
     }
 }
 
+if (!function_exists('jsonFilter')) {
+    /**
+     * Prepare data for json_encode.
+     *
+     * @param mixed $value
+     */
+    function jsonFilter(&$value) {
+        if (is_array($value)) {
+            array_walk_recursive($value, __FUNCTION__);
+        } elseif ($value instanceof \DateTimeInterface) {
+            $value = $value->format('r');
+        } elseif ($ip = ipDecode($value)) {
+            $value = $ip;
+        }
+    }
+}
+
 if (!function_exists('now')) {
     /**
      * Get the current time in seconds with a millisecond fraction.
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -2159,18 +2159,24 @@ if (!function_exists('jsonFilter')) {
      * @param mixed $value
      */
     function jsonFilter(&$value) {
-        $fn = function (&$value, $key = '') use (&$fn) {
-            if ($value instanceof \DateTimeInterface) {
+        $fn = function (&$value, $key = '', $parentKey = '') use (&$fn) {
+            if (is_array($value)) {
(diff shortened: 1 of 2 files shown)


library/core/functions.general.php
```diff
@@ -2152,6 +2152,23 @@ function jsonEncodeChecked($value, $options = null) {
...
+     * Prepare data for json_encode.
+     *
+     * @param mixed $value
+     */
+    function jsonFilter(&$value) {
+        if (is_array($value)) {
+            array_walk_recursive($value, __FUNCTION__);
+        } elseif ($value instanceof \DateTimeInterface) {
+            $value = $value->format('r');
+        } elseif ($ip = ipDecode($value)) {
```
tburry: This looks iffy to me. I'd at least do a === test on the return. Is there any way we can test for a binary string too? At least do an is_string().

initvector:
I did a little research, but couldn't find a reliable way to quickly check if the string was a packed IP address. Instead, I have implemented a call to `is_string` before attempting to decode the value into  an IP address. `ipDecode` should only be able to return two values: a string, if the value is already an unpacked IP or the IP is successfully decoded, or `null`, if decoding fails. The comparison now explicitly checks for a `null` result, instead of "not truthy", before assignment.

tburry:
I think we have two options here:

1. Leave this function as is and hope that some random data doesn't get garbled.
2. Add field name whitelisting with the second argument in array_walk_recursive.

The way I see it is it's going to be more difficult to add the IP whitelisting in the future because there may be some data that depends on everything being decoded. If we add it now we may have to be more strict with IP address field names, but it's much safer.

For reference, here is the if statement from ipDecodeRecursive():

```php
if (stringEndsWith($key, 'IPAddress', true) || stringEndsWith($parent, 'IPAddresses', true)) {
    ...
}
```

One thing to note is a simple array of IP addresses isn't handled by this:

```php
[
    'ipAddresses' => [ ... ] 
]
```

PR #5170, opened 2017-02-10
Add forwards-compatible methods to Gdn_Request

### LINKED ISSUE #4162: Add forwards-compatible methods to Gdn_Request
We are eventually going to want to replace Gdn_Request, but it's going to take a while to untangle. Before that I want to add new-style methods to Gdn_Request that the dispatcher will use.

Here are the additions. Setters should be chainable.

```php
const METHOD_HEAD = 'HEAD';
const METHOD_GET = 'GET';
const METHOD_POST = 'POST';
const METHOD_PUT = 'PUT';
const METHOD_PATCH = 'PATCH';
const METHOD_DELETE = 'DELETE';
const METHOD_OPTIONS = 'OPTIONS';

getHost(), setHost($host)
getHostAndPort()
getScheme(), setScheme($scheme)
getIP(), setIP()
getPath(), setPath($path)
getExt(), setExt($ext)
getPathExt(), setPathExt($path)
getFullPath(), setFullPath() // root + path ext
getRoot(), setRoot($root)
getPort(), setPort($port)
getQuery(), setQuery(array $query), mergeQuery(array $query)
getQueryItem($key, $default = null), setQueryItem($key, $value)
getBody(), setBody(array $body) // this is the post data
getUrl(), setUrl($url) // should parse into parts
```

The long-term goal is to make a request object that has just these methods (+few more probably) and isn't tied to superglobals. We can then add the superglobal parsing to a subclass or elsewhere.

initvector:
This update adds forwards-compatible methods to Gdn_Request and tests to support them. For details, see the original issue.

Closes #4162

### DIFF
--- a/library/Garden/Web/RequestInterface.php
+++ b/library/Garden/Web/RequestInterface.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * @author Todd Burry <todd@vanillaforums.com>
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license Proprietary
+ */
+
+namespace Garden\Web;
+
+
+interface RequestInterface {
+    public function getPath();
+
+    public function getMethod();
+
+    public function getQuery();
+
+    public function getBody();
+}
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -380,7 +380,7 @@ class Gdn_Request implements RequestInterface {
      * @return string
      */
     public function getPath() {
-        $path = (string)$this->_environmentElement('URI');
+        $path = (string)$this->_parsedRequestElement('Path');
         if (strpos($path, '/') !== 0) {
             $path = "/{$path}";
         }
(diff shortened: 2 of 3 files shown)


tburry:
Yeah I think your stuff is going to $_Environment when it should be going to $_ParsedRequest.

initvector:
1. `Gdn_Request::getPath` and `Gdn_Request::setPath` have been updated to use the proper internal arrays.
2. An additional test for `Gdn_Request::getUrl` and `Gdn_Request::url` have been added to ensure they generate the same full "current" URL, regardless of the functions used to configure the request.
3. An additional test has been added to verify the root is properly handled (always a leading slash, unless only a slash, in which case the root is empty).
4. `Gdn_Request::webRoot` has been altered to avoid using a local static variable, since the value can be also be updated by `Gdn_Request::setRoot`.

initvector:
I've added nine more equivalency-style tests for `Gdn_Request` to verify the result of the following: `getBody`, `getHost`, `getHostAndPort`, `getIP`, `getPort`, `getQuery`, `getQueryItem`, `getRoot` and `getScheme`

PR #5174, opened 2017-02-11
API v2 routing

### LINKED ISSUE #4161: Bring in Garden's route logic and change dispatching into a routing loop

tburry:
Add resource routing and API controller support to the /api/v2 path.

There is a lot of code in this PR, but I think its pretty safe to look more at its impact on existing sites which would consist mainly of the code added to **Gdn_Dispatcher**.

Closes #4161, #4164.


library/Garden/Web/Data.php
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public function jsonSerialize() {
+        return $this->getData();
+    }
+
+    /**
+     * Render the response to the output.
+     */
+    public function render() {
+        header('', true, $this->getStatus());
```
initvector: Would [`http_response_code`](http://php.net/manual/en/function.http-response-code.php) be more appropriate here? The [docs for `header`](http://php.net/manual/en/function.header.php#refsect1-function.header-parameters) state the response code parameter "only has an effect if the string is *not empty*".

library/Garden/Web/ResourceRoute.php
```diff
@@ -0,0 +1,351 @@
...
+            $pathPart = substr($path, strlen($this->basePath));
+        } else {
+            return null;
+        }
+
+        $pathArgs = explode('/', $pathPart);
+
+        // First look for the controller.
+        $controllerSlug = $this->filterName(array_shift($pathArgs));
+        if (null === ($controllerClass = $this->classLocator->findClass(sprintf($this->controllerPattern, $controllerSlug)))) {
```
initvector: To improve readability a bit, maybe move the `$controllerClass` assignment out before the `if` statement.

library/Garden/Web/Route.php
```diff
@@ -0,0 +1,223 @@
...
+    public function getConstraints() {
+        return $this->constraints;
+    }
+
+    /**
+     * Set the entire conditions array.
+     *
+     * If you are going to set conditions in this way make sure the array keys are all lowercase.
+     *
+     * @param array $constraints An array of conditions tha maps variable names to **Route::MAP_*** constants.
```
initvector: > conditions *that* maps

library/Garden/Web/Route.php
```diff
@@ -0,0 +1,223 @@
...
+     * @return $this
+     */
+    public function setConstraints($constraints) {
```
initvector: It appears `$constraints` should always be an array. This parameter could probably benefit from a type declaration.

library/Garden/Web/ResourceRoute.php
DaazKu: Speaking of readability/clarity, I find that `$var === null` is more readable than `null === $var`.

tburry:
I incorporated the changes. Please note the rebase and kill your local branch if necessary.

Issue #3902, opened 2016-05-08
Optimize category count fields


tburry:
- [ ] Store `LastCategoryID`, `CountAllDiscussions` and `CountAllComments` fields in the category table.
- [ ] Refactor count and latest post updating into the category model.
- [ ] When posting/deleting discussions/comments you need visit each parent category and increment/decrement the CountAll* columns with SQL increments.
- [ ] When posting discussions/comments you need to visit each parent and set the LastCategoryID = the category ID of the discussion/comment.
- [ ] LastCategoryID with a delete is more difficult. Add an issue for that one.

tburry:
Here is some SQL you can use to calculate an entire tree in a loop. I unwrapped the loop for testing on my localhost, but obviously the code would use a loop. You won't be able to structure this with the SQL object so be careful with injection.

```SQL
-- First grab the max depth so you know where to loop.
select max(Depth)
from GDN_Category;

-- Initialize with self count.
update GDN_Category
set CountAllDiscussions = CountDiscussions, CountAllComments = CountComments;

-- Iterate from max depth down to 1

-- Depth 4
update GDN_Category c
join (
	select
		c2.ParentCategoryID,
		sum(CountAllDiscussions) as CountAllDiscussions,
		sum(CountAllComments) as CountAllComments
	from GDN_Category c2
	where c2.Depth = 4 -- start with max depth
	group by c2.ParentCategoryID
) c2
	on c.CategoryID = c2.ParentCategoryID
set
	c.CountAllDiscussions = c.CountAllDiscussions + c2.CountAllDiscussions,
	c.CountAllComments = c.CountAllComments + c2.CountAllComments
where c.Depth = 3; -- current depth - 1

-- Depth 3
update GDN_Category c
join (
	select
		c2.ParentCategoryID,
		sum(CountAllDiscussions) as CountAllDiscussions,
		sum(CountAllComments) as CountAllComments
	from GDN_Category c2
	where c2.Depth = 3
	group by c2.ParentCategoryID
) c2
	on c.CategoryID = c2.ParentCategoryID
set
	c.CountAllDiscussions = c.CountAllDiscussions + c2.CountAllDiscussions,
	c.CountAllComments = c.CountAllComments + c2.CountAllComments
where c.Depth = 2;

-- Depth 2
update GDN_Category c
join (
	select
		c2.ParentCategoryID,
		sum(CountAllDiscussions) as CountAllDiscussions,
		sum(CountAllComments) as CountAllComments
[...]

PR #5186, opened 2017-02-15
Add aggregate count category functions

### LINKED ISSUE #5187: Add aggregate post count functions to CategoryModel

### LINKED ISSUE #5188: Add aggregate post count job for categories at /dba/counts

initvector:
Going to break out some components of #3902 instead of making one monolithic PR.

This update adds a few new functions to `CategoryModel`:

1. `incrementAggregateCounts` - Convenience function to increment aggregate post counts for a category and its ancestors by one. Will be called when adding a post to a category (e.g. new post, moved post).
2. `decrementAggregateCounts` - Convenience function to decrement aggregate post counts for a category and its ancestors by one. Will be called when removing a post from a category (e.g. deleted post, moved post).
3. `adjustAggregateCounts` - Performs the heavy-lifting behind `incrementAggregateCounts` and `decrementAggregateCounts`. Updates relevant category rows in the database and cache.
4. `recalculateAggregateCounts` - Updates all aggregate post counts for all categories in a database.

A job has been added to `/dba/counts` to update all category aggregate post counts.

`LastCategoryID` has been added, but not referenced in this PR.

These changes are going into the `fix/category-optimization` branch, not `master`.

Closes #5188
Closes #5187


applications/vanilla/models/class.categorymodel.php
```diff
@@ -2801,4 +2805,126 @@ public function recalculateTree() {
     public static function flattenTree($categories) {
         return self::instance()->collection->flattenTree($categories);
     }
+
+    /**
+     * Adjust the aggregate post counts for a category.
+     *
+     * @param int|string|object|array $category
+     * @param int $offset
```
linc: Documenting what this variable does would answer the obvious question: "Adjust how?"

applications/vanilla/models/class.categorymodel.php
```diff
@@ -2801,4 +2805,126 @@ public function recalculateTree() {
...
+            $category = self::instance()->collection->get($category);
+        }
+
+        if (is_array($category) || is_object($category)) {
+            $categoryID = val('CategoryID', $category);
+        }
+
+        if (empty($categoryID)) {
+            return;
+        }
```
linc: These above 10 lines feel like a disambiguation method that other bits of code could make use of to just feed this method a reliable piece of data (CategoryID?) rather than literally accepting 4 variable types in every method that we need to reference a category. Making the var type `int|string|object|array` is basically making it `¯\_(ツ)_/¯`.

linc: I guess the real question is: Wherever `decrementAggregateCounts` and `incrementAggregateCounts` are used, do we really not have a reliable data type available? If not, is that perhaps the problem we should be solving as part of this refactor rather than making these helpers so permissive?

applications/vanilla/models/class.categorymodel.php
```diff
@@ -2801,4 +2805,126 @@ public function recalculateTree() {
...
+                            sum(CountAllComments) as CountAllComments
+                        from {$prefix}Category c2
+                        where c2.Depth = :Depth
+                        group by c2.ParentCategoryID
+                    ) c2 on c.CategoryID = c2.ParentCategoryID
+                set
+                    c.CountAllDiscussions = c.CountAllDiscussions + c2.CountAllDiscussions,
+                    c.CountAllComments = c.CountAllComments + c2.CountAllComments
+                where c.Depth = :ParentDepth";
+            Gdn::database()->query($sql, [':Depth' => $depth, ':ParentDepth' => ($depth - 1)]);
```
linc: RIP ever refactoring our queries in to the sql object entirely

initvector: Implementing the method recommended here: https://github.com/vanilla/vanilla/issues/3902#issuecomment-279808127

Issue #4659, opened 2016-10-17
Add filter to category list in dashboard


initvector:
If a flat category has hundreds of children, it can be tedious to navigate via the pager.  It would be beneficial if it were possible to filter or search for categories in the dashboard.

PR #5196, opened 2017-02-16
Add category filtering to settings page

beckyvb:
Adds category filtering to the dashboard settings page.

* ~~Refactors the FlatCategoryModule to CategoryFilterModule, operating on more than just flat categories.~~
* ~~Adds a new view for the category module for rendering in the dashboard.~~
* ~~Adds a rendering function for the input so the filter can exist apart from the results.~~
* Adds a new endpoint for retrieving flat categories with their options.
* Adds functions for rending categories and their options on the frontend.

Relies on https://github.com/vanilla/vanilla/pull/5195
Closes https://github.com/vanilla/vanilla/issues/4659


initvector:
I'm seeing some responsiveness issues on the category management page:

1. If I enter some filter text, the categories immediately disappear, but it might take a second or two for results to take their place.
1. If I clear out the text field, then enter new text, the results of my old search show up for a second or so before the results of my new search take their place.
1. If I enter a search, then clear out the search field before the results are displayed, the results appear over the full category list.

It might be beneficial to have a single search request. If a user updates their filter terms, an existing search could be killed and a new one created. It'd also be good if the full category list remained in view until the results displayed, or maybe some kind of loading message/indicator could take their place.

What about not using AJAX requests for this page and filtering results from the full category list, instead? I know it conflicts with the usage of general-purpose components you're creating here, but the responsiveness would be almost instantaneous.

PR #5100, opened 2017-02-01
Update MeModule settings to use DropdownModule

beckyvb:
Updates MeModule settings to use DropdownModule to make it configurable and sortable.

Also:
* Removes sprites from the dropdown menu.
* Adds the ability to add a list item CSS class to an item.
* Updates MarkAllViewed's way of adding an item to the menu.
* Adds the ability to add a url for the dropdown trigger.

Other than removing the sprites, this update does not affect the markup of the MeModule menu and should not affect any themes.

PR #5198, opened 2017-02-16
Add overrideAll to Vanilla\Permissions

### LINKED ISSUE #5162: Add get/setOverrideAll() to Vanilla\Permissions
The Vanilla\Permissions class should be able to handle all user permissions. We're almost there except for overrides from Admin, Deleted, and Banned.

- [x] Add a property overrideAll with get/set methods. It should be bool|null.
- [x] If null it does nothing to permission checks.
- [x] If true all permission checks return true. If false all permission checks return false.
- [x] Integrate into Gdn_Session with permission checks.

This sub-property won't cache or serialize at this point since it's tied to the user, but plugins that modify permissions will be able to use it.

initvector:
This update adds the ability to universally override permissions in `Vanilla\Permissions`. This ability to override the result of permission checks has been added to `Gdn_Session`. Banned or deleted users will always return `false`. Admin users will always return `true`.

More details available in the original issue.

Closes #5162

### DIFF
--- a/library/Vanilla/Permissions.php
+++ b/library/Vanilla/Permissions.php
@@ -12,6 +12,12 @@ namespace Vanilla;
  * @package Vanilla
  */
 class Permissions {
+
+    /**
+     * @var bool|null The result of all permission checks can be overridden based on this value.
+     */
+    private $overrideAll = null;
+
     /**
      * Global permissions are stored as numerical indexes.
      * Per-ID permissions are stored as associative keys. The key is the permission name and the values are the IDs.
@@ -94,6 +100,15 @@ class Permissions {
     }
 
     /**
+     * Get the flag to determine if all permission checks are being overridden.
+     *
+     * @return bool|null
+     */
+    public function getOverrideAll() {
+        return $this->overrideAll;
+    }
+
+    /**
      * Grab the current permissions.
      *
      * @return array
@@ -110,6 +125,10 @@ class Permissions {
      * @return bool
      */
     public function has($permission, $id = null) {
+        if ($this->overrideAll !== null) {
+            return $this->overrideAll;
+        }
+
         if ($id === null) {
             return !empty($this->permissions[$permission]) || (array_search($permission, $this->permissions) !== false);
         } else {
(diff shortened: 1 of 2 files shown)


library/Vanilla/Permissions.php
```diff
@@ -274,6 +293,21 @@ public function setBanned($isBanned) {
...
 
     /**
+     * Get the flag to determine if all permission checks are being overridden.
+     *
+     * @param bool|null $overrideAll
+     * @return $this
+     */
+    public function setOverrideAll($overrideAll) {
+        if ($overrideAll !== null) {
+            $overrideAll = boolval($overrideAll);
```
DaazKu: This is the kind of thing we should stop doing IMO.
The method clearly state that you must receive a boolean or null.

We need, as developers, to ensure that we respect the docblocks.

initvector: This could be used by any developer making addons. It'd be great if we could count on everyone to always do the right thing, but I think it's good to correct for simple "mistakes".

DaazKu: Down the way it encourage lazy behaviours and we get bugs like this: https://github.com/vanilla/vanilla/pull/5100#pullrequestreview-20389295

initvector: I view this as an easy-to-implement piece of defensive programming. It will ensure that, no matter what value is received by `setOverrideAll`, this class can always be certain its `overrideAll` property will be a boolean or null value. The class, and other parts of Vanilla that use its `getOverrideAll` method, can always use identical comparisons for reliable results.

I don't see a downside to making an open-source application more resilient with minimal effort.

beckyvb: Vanilla makes use of PHP's non-boolean falsey values everywhere, I think this is a nice safeguard. One of the contributing factors of the bug that @DaazKu refers to was because a method returned a falsey value rather than a boolean. So, bugs can be introduced by assuming this is a boolean, when somebody may pass it another type. I agree with @initvector . Coercing a property to be the type it expects is a nice enhancement.

DaazKu: I agree that it makes the code more resilient. I can't deny that. 
I would at least like to see a notice generated. That way we can correct the errors.

PR #5199, opened 2017-02-17
Allow editor to find and delete locally copied scratch files from both s3 and cf

kaecyra:
The cloudfiles plugin stored its scratch files in /uploads/cftemp/ and the editor was specifically checking for that path slug. s3files stores its scratch files in /uploads/cloudtemp/ and so the "cftemp" check was failing and files were not being deleted.

This fixes that problem, not in a great way but ok for now.

### DIFF
--- a/plugins/editor/class.editor.plugin.php
+++ b/plugins/editor/class.editor.plugin.php
@@ -1509,7 +1509,7 @@ class EditorPlugin extends Gdn_Plugin {
             ));
 
             // Remove cf scratch copy, typically in cftemp, if there was actually a file pulled in from CF.
-            if (strpos($local_path, 'cftemp') !== false) {
+            if (strpos($local_path, 'temp') !== false) {
                 if (!unlink($local_path)) {
                     // Maybe add logging for local cf copies not deleted.
                 }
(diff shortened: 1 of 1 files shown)


plugins/editor/class.editor.plugin.php
```diff
@@ -1509,7 +1509,7 @@ public function utilityController_mediaThumbnail_create($sender, $media_id) {
             ));
 
             // Remove cf scratch copy, typically in cftemp, if there was actually a file pulled in from CF.
```
DaazKu: "typically in temp"?

kaecyra: Thanks

plugins/editor/class.editor.plugin.php
```diff
@@ -1508,11 +1508,9 @@ public function utilityController_mediaThumbnail_create($sender, $media_id) {
                 'ThumbPath' => $filepath_parsed['SaveName']
             ));
 
-            // Remove cf scratch copy, typically in cftemp, if there was actually a file pulled in from CF.
-            if (strpos($local_path, 'temp') !== false) {
-                if (!unlink($local_path)) {
-                    // Maybe add logging for local cf copies not deleted.
-                }
+            // Remove cf scratch copy, typically in /uploads/cftemp/ or /uploads/cloudtemp/, if there was actually a file pulled in from CF.
```
evach-vf: We should probably use the constant `PATH_UPLOADS` instead of hardcoding `/uploads ` in the regex since our Cloud storage plugins uses that constant as well.
This would prevent issues for users setting `PATH_UPLOADS` to somewhere else.

Issue #5086, opened 2017-01-31
Update from 2.3 to 2.4 breaks dashboard


R-J:
In 2.4 applications/vanilla/controller/class.settingscontroller.php has been deleted.

The normal upgrade process does not include a deletion step, so the file from 2.3 still exists. As a consequence, all /dashboard/settings links are routed through that router which standard method is "managecategories". That's why several settings are not accessible after a standard update.

It might be better to still supply such a file with only an index method that tells the user something like `'Please delete the file "'.__FILE__.'" manually'` and also give the link to the corresponding views "applications/vanilla/views/settings"

At least the fact that these artifacts exist and must be deleted should be told to admins.

linc:
See also https://github.com/vanilla/vanilla/issues/2518

Issue #2518, opened 2015-02-22
Simplify upgrades: notice to delete removed files if they still exist


linc:
Our upgrade process for open source is "copy these new files over" but every once in a while, we remove a file that can cause problems with that methodology, like the `helper_functions.php` in the default mobile theme, or the `default.master.php` when we went to .tpl. What do we think of adding some basic file deletion to the Dashboard's Structure to nuke those files so we can simplify upgrade instructions?

My only concern would be if someone has hacked those files, but I kinda feel like you voided your warranty in that situation.

---

The consensus appears to be adding a notice that prompts you to delete them yourself.

R-J:
In general: if it is okay for me that all my files are overwritten, it should be okay that some files vanish altogether. But I'd prefer to replace its contents with a short comment

~~~
<?php
// this file is not needed from version x.y onwards
~~~

linc:
@R-J That would defeat the point and further break sites. The files need to be deleted because of view overriding.

x00:
I don't think so.

I think you can tell people to delete files, but that is as far as it should go. 

Once you you are in the business of file management, that is a lot of responsibly on your shoulders. It has consequences such as requiring permission over the files. 

I think Install/Upgrade should be about guiding people. 

If it worked differently, and you were producing a client program for updates an addons, that is one thing becuase it ti not part of web application itself, it is a tool for server management of the framework.

chanhong:
Is it possible to rename or move the file to be deleted and replace it with a holder file with info about the action so that the user have a way to trace back to figure out what to do.

x00:
@chanhong I don't really see the point. What webmaster gives permission to a webapp to do that sort of thing anyway?

x00:
Don't use a sledgehammer to crush a nut. This is a minor inconvience.

chanhong:
ok, then just show in the install instruction that which file to be deleted so that they will have a chance to backup the file themselves and decide what to do with it.

x00:
general feedback mechanism in install/upgrade +1

hgtonight:
Hmm... I don't like the idea of automatically deleting files.

Perhaps an inform message on admin pages saying a file that should be deleted exists would be better? Basically a dismissible message that is inserted in the structure phase?

adrianspeyer:
I agree with @hgtonight, make it clear: "Hey this file does should be removed" maybe part of the utility structure phase?

linc:
I like the message idea & will amend the issue to reflect that.

PR #5200, opened 2017-02-17
Check for the existence of deleted files at /utility/upgrade endpoint

beckyvb:
Informs users of files that exist in their installation that have been deleted since the last release.

Closes https://github.com/vanilla/vanilla/issues/2518
Closes https://github.com/vanilla/vanilla/issues/5086


linc:
I cannot fix the composer.lock conflict until Packagist updates with the library now that I just added it.

Issue #5203, opened 2017-02-17
Vanilla.Categories.Use = false makes impossible to create new discussions


LauraMakesArt:
Having `Vanilla.Categories.Use` set to false causes a conflict. There is no drop down to select a category but vanilla throws a category ID error when attempting to post.

PR #5204, opened 2017-02-17
Make sure to set discusion's categoryID when Vanilla.Categories.Use is false

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/5203

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -180,13 +180,13 @@ class PostController extends VanillaController {
                 ['Archived' => 0, 'AllowDiscussions' => 1],
                 ['AllowedDiscussionTypes' => $this->Data['Type']]
             );
-            $nbsAllowedCategories = count($AllowedCategories);
+            $AllowedCategoriesCount = count($AllowedCategories);
 
-            if ($this->ShowCategorySelector && $nbsAllowedCategories === 1) {
+            if ($this->ShowCategorySelector && $AllowedCategoriesCount === 1) {
                 $this->ShowCategorySelector = false;
             }
 
-            if (!$this->ShowCategorySelector && $nbsAllowedCategories) {
+            if (!$this->ShowCategorySelector && $AllowedCategoriesCount) {
                 $AllowedCategory = array_pop($AllowedCategories);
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -172,17 +172,22 @@ public function discussion($CategoryUrlCode = '') {
...
+            // See if we should fill the CategoryID value.
             $AllowedCategories = CategoryModel::getByPermission(
                 'Discussions.Add',
                 $this->Form->getValue('CategoryID', $this->CategoryID),
                 ['Archived' => 0, 'AllowDiscussions' => 1],
                 ['AllowedDiscussionTypes' => $this->Data['Type']]
             );
-            if (count($AllowedCategories) == 1) {
-                $AllowedCategory = array_pop($AllowedCategories);
+            $nbsAllowedCategories = count($AllowedCategories);
```
linc: nbs? When did we start doing weird abbreviations in var names?

applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -172,17 +172,22 @@ public function discussion($CategoryUrlCode = '') {
 
         touchValue('Type', $this->Data, 'Discussion');
 
-        // See if we should hide the category dropdown.
-        if ($this->ShowCategorySelector) {
+        if (!$UseCategories || $this->ShowCategorySelector) {
```
linc: Didn't we just start removing this logic condition everywhere in prep for deprecating that config?

DaazKu: The config is still used. We need to either no use it anywhere or respect it until we remove it for real.

applications/vanilla/controllers/class.postcontroller.php
DaazKu: How do you want me to call it? `$AllowedCategoriesCount`?
`$AllowedCategories` would be dumb here

linc: `$AllowedCategoriesCount` is fine. Or allowedCount. Or categoryCount. Or the inverse of any of those.

PR #5205, opened 2017-02-17
Add the title of a discussion to a comment in the user comments module.

nervoustwit:
The display of comments in the user comments module looked sparse without a linkable title to the discussion, especially when you display the UserComments next to the UserDiscussions that does have a link to the discussion. This PR adds a title to the comment.


linc:
I forgot we fixed that formatter to be safe. I reverted the addendum.

PR #5211, opened 2017-02-21
Let Form->validate check that the category is properly set

DaazKu:
https://github.com/vanilla/vanilla/blob/9656876e959f55758fb1f776d6e6f5d97fb17c11/applications/vanilla/models/class.discussionmodel.php#L1887 will catch it if the category is not set.

The group application uses the BeforeSaveDiscussion event to inject a categoryID


initvector:
Reverting https://github.com/vanilla/vanilla/pull/5092

PR #5222, opened 2017-02-23
Make the container available in the bootstrap

tburry:
This is a stopgap to #5031 so that other PRs can add DIC rules in the
bootstrap without the entire DIC being configured.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -60,6 +60,15 @@ if (!class_exists('Gdn')) {
     exit("Class Gdn not found.");
 }
 
+// Set up the dependency injection container.
+$dic = new \Garden\Container\Container();
+Gdn::setContainer($dic);
+
+$dic->setInstance('Garden\Container\Container', $dic)
+    ->rule('Interop\Container\ContainerInterface')
+    ->setAliasOf('Garden\Container\Container')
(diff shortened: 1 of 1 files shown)


tburry:
Hey @linc, we really need at least this PR to get stuff moving. Maybe @DaazKu can look?

bootstrap.php
```diff
@@ -60,6 +60,15 @@
...
 }
 
+// Set up the dependency injection container.
+$dic = new \Garden\Container\Container();
+Gdn::setContainer($dic);
+
+$dic->setInstance('Garden\Container\Container', $dic)
+    ->rule('Interop\Container\ContainerInterface')
+    ->setAliasOf('Garden\Container\Container')
+;
```
initvector: I think this semicolon escaped from its line!

DaazKu: Personally I prefer that way of doing when we chain methods.

tburry: The config is a big chain and putting the semi-colon on a different line will help with merges.

linc: Please no. We chain in about 10,000 places in our code base. Don't invent a new coding standard in the PR for 10,0001.

PR #5235, opened 2017-02-24
Move last post updating into CategoryModel

### LINKED ISSUE #3902: Optimize category count fields

initvector:
Per the original issue, this update:

1. Stores `LastCategoryID`, `CountAllDiscussions` and `CountAllComments` fields in the category table.
1. Refactors count and latest post updating into `CategoryModel`.
1. Increments the CountAll columns, when posting/deleting discussions/comments.
1. Sets the `LastCategoryID` to the current category ID when posting a discussion or comment.

Closes #3902

#### Notes
1. This was a little hairy to sort out, so it warrants some additional in-depth testing by the reviewer.
1. ~~You should run /dba/counts before starting, particularly the `Recalculate Category.CountAllDiscussions` and `Recalculate Category.CountAllComments` jobs.~~
1. This PR should be merged into the `fix/category-optimization` branch.


linc:
Why is this reassigned to @DaazKu with month-stale feedback still waiting to be addressed?

initvector:
She's no longer stale. Fresh as a daisy, she is.

Feedback has been addressed and branches (`fix/category-optimization`, `fix/categorymodel-counts-lastpost`) have been rebased as they were several weeks old. Please make sure you remove any local copies of these branches and pull from origin.

DaazKu:
This PRs wants to merge in fix/category-optimization. Shouldn't it be remade to merge in master?

DaazKu:
Replaced by https://github.com/vanilla/vanilla/pull/5388

initvector:
I've updated the PR to automatically build the necessary counts when the columns are added.

In addition, I was running into an issue where counts were sometimes failing (MySQL's `sum` returning a `null` value), so I added a call to `coalesce` on the count fields.

DaazKu:
Nice work! This PR works great!

screen shot 2017-03-30 at 4 07 55 pm

It was 0 and 0 before!

applications/vanilla/models/class.categorymodel.php
```diff
@@ -1035,6 +1039,155 @@ private function gatherLastIDs($categoryTree, &$result = null) {
     }
 
     /**
+     * Given a discussion, update its category's last post info and counts.
+     *
+     * @param int|array|stdClass $discussion The discussion ID or discussion.
+     */
+    public function incrementLastDiscussion($discussion) {
+        // Lookup the discussion record, if necessary. We need at least a discussion to continue.
+        if (is_numeric($discussion)) {
```
DaazKu: `is_numeric(9.1) === true`
`is_numeric('0xf4c3b00c') === true` prior to php7.0

Use `filter_var($value, FILTER_VALIDATE_INT) !== false` or `ctype_digit((string)$value)`

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2267,38 +2274,17 @@ public function updateDiscussionCount($CategoryID, $Discussion = false) {
...
-
-        $this->SQL->update('Category')
-            ->set('CountDiscussions', 'CountDiscussions + 1', false)
-            ->set('LastDiscussionID', val('DiscussionID', $Discussion))
-            ->set('LastCommentID', null)
-            ->set('LastDateInserted', val('DateInserted', $Discussion))
-            ->where('CategoryID', val('CategoryID', $Discussion))
-            ->put();
+    public function incrementNewDiscussion($discussion) {
+        trigger_error(
```
DaazKu: We have the `deprecated()` function just for that!

applications/vanilla/settings/structure.php
```diff
@@ -423,6 +428,15 @@
         ->put();
 }
 
+if (!$captureOnly) {
```
DaazKu: Maybe we could put the exist check in here since they are only used if captureOnly is false?

initvector: If the is-existing check is done this far down, the columns have already been created if they didn't exist in the first place. It'll always return `true`.

DaazKu: Right. Good then! :)

PR #5245, opened 2017-02-27
Bring a wayward semicolon home in bootstrap.php

initvector:
A semicolon found its way onto its own line in `boostrap.php`. This update moves it back where it belongs.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -66,8 +66,7 @@ Gdn::setContainer($dic);
 
 $dic->setInstance('Garden\Container\Container', $dic)
     ->rule('Interop\Container\ContainerInterface')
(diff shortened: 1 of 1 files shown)


DaazKu:
Let's make sure we do not want this to be a new way of doing first!
See https://github.com/vanilla/vanilla/pull/5222#discussion_r103231564

initvector:
I don't think it's done like this, anywhere else in Vanilla. I'm pretty sure the standard is to end the last call in a list of chain methods with a semicolon.

This seems like a mistake and not something intentional.

initvector:
Overruled by https://github.com/vanilla/vanilla/pull/5222#discussion_r103233891

Issue #5128, opened 2017-02-02
Remove /library/vendors/oauth


linc:
Our final forked dependency. We probably still need it, but can it be Composer-ified?

DaazKu:
I'm gonna need a repo for this!

PR #5246, opened 2017-02-27
Move library/vendors/oauth to composer

DaazKu:
- Clean requires
- Remove directory from library/vendors
- Update composer.json

Fixes https://github.com/vanilla/vanilla/issues/5128


linc:
I cannot fix the composer.lock conflict until Packagist updates with the library now that I just added it.

PR #5251, opened 2017-02-27
Add the Cookie class

tburry:
The Vanilla\Web\Cookie class is a utility class for handling reading
and changing cookies throughout the lifecycle of the request.

Currently, we don’t have controlled cookie management. Just a bunch of
ad-hoc calls to **$_COOKIE** and **setcookie()**. This results in
problems such as unnecessary Set-Cookie headers.

What still needs to be done:

- [x] Add unit tests.
- [x] This implementation may have bugs, the above unit tests should
uncover them.

When implemented this class will be shared in the container and then
any class that wants to manipulate cookies can depend on it.

### DIFF
--- a/library/Garden/Web/Cookie.php
+++ b/library/Garden/Web/Cookie.php
@@ -63,6 +63,17 @@ class Cookie {
     }
 
     /**
+     * Calculate a cookie's expiration time.
+     *
+     * @param int $expire
+     * @return int
+     */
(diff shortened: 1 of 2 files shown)


tburry:
In terms of priority this isn't super urgent, but that's only because we have other code that is super-urgent. I'll want this done soon into Q2 2017.

initvector:
Rebased off rev 7376164

Issue #5226, opened 2017-02-24
Remove PDO::MYSQL_ATTR_INIT_COMMAND from the config


tburry:
We currently have a config option for the startup command. This sets the character set and timezone. These should be automated in the Gdn_Database class.

- [x] In Gdn_Database, set the collation and timezone. Use Database.CharacterEncoding for the encoding.
- [x] Remove Database.ConnectionOptions.1002 from config-defaults.
- [x] Replace the other connection options in config-defaults with their proper PDO constants. I think we support a modern enough version of PHP for them all to be there now although you can do your own research.

PR #5255, opened 2017-02-27
Update charset from utf8 to utf8mb4

DaazKu:
Will need a rebase after https://github.com/vanilla/vanilla/pull/5257 is merged.
Fixes https://github.com/vanilla/vanilla/issues/5228
Depends on https://github.com/vanilla/vanilla/pull/5256
From my tests, this update should not affect already created databases with the utf8 charset.

PR #5257, opened 2017-02-28
Set charset and timezone directly in the database object

DaazKu:
Remove Database.ConnectionOptions.1002 and set the encoding and timezone directly in the database object

We default to utf8mb4 because it is compatible with utf8 and won't cause any problems.

Fix https://github.com/vanilla/vanilla/issues/5226

### DIFF
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -153,9 +153,8 @@ class Gdn_Database {
             $PDO = new PDO(strtolower($this->Engine).':'.$Dsn, $User, $Password, $this->ConnectionOptions);
             $PDO->setAttribute(PDO::ATTR_EMULATE_PREPARES, 0);
 
-            if ($this->ConnectionOptions[1002]) {
-                $PDO->query($this->ConnectionOptions[1002]);
-            }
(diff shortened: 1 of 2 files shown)


library/database/class.database.php
```diff
@@ -153,9 +153,8 @@ protected function newPDO($Dsn, $User, $Password) {
             $PDO = new PDO(strtolower($this->Engine).':'.$Dsn, $User, $Password, $this->ConnectionOptions);
             $PDO->setAttribute(PDO::ATTR_EMULATE_PREPARES, 0);
 
-            if ($this->ConnectionOptions[1002]) {
-                $PDO->query($this->ConnectionOptions[1002]);
-            }
+            $encoding = c('Database.CharacterEncoding', 'utf8mb4');
```
tburry: I'd say this is a bit too soon and looks sloppy without everything else going to utf8mb4.

DaazKu: I can move this into that one if you prefer: https://github.com/vanilla/vanilla/pull/5255 but the result will be the same. Using utf8mb4 for the connexion will cause no problem at all even if the client is using utf8.

PR #5263, opened 2017-03-01
Add functions to find the resolved mobile and desktop themes

beckyvb:
We can't trust the config to let us know what theme is actually being displayed, since if the theme set in config doesn't exist, we display the default theme. This adds two functions to find the resolved mobile and desktop themes.

Also:
* Moves the error messages for the desktop themes settings page above the current theme.
* Adds an error message to theme settings page is the theme set in the config is not found.
* Sets the default mobile and desktop theme names as consts in the ThemeManager.

And bonus!:
* Fixes the default mobile theme, which was previously always set to "default", resolving to "Baseline+". It is now "mobile".


library/core/class.thememanager.php
```diff
@@ -22,6 +22,11 @@ class Gdn_ThemeManager extends Gdn_Pluggable {
 
     const ACTION_DISABLE = 2;
 
+    const DEFAULT_DESKTOP_THEME = 'default';
+    const DEFAULT_MOBILE_THEME = 'mobile';
+
+
+
```
DaazKu: That's a lot of extra new lines!

library/core/class.thememanager.php
```diff
@@ -605,7 +639,8 @@ public function mobileTheme() {
         if ($this->hasMobilePreview()) {
             return $this->getMobilePreview();
         }
-        return c('Garden.MobileTheme', 'default');
+        return $this->getEnabledMobileThemeKey();
+
```
DaazKu: That one isn't too bad.

DaazKu:
I feel like this should try to check if the default theme has options if the current theme was not found!
screen shot 2017-03-02 at 5 12 15 pm

PR #5265, opened 2017-03-01
Fix Message toggle not removing/adding message immediately

linc:
Break cache when changing Message status.

Also adds some docs.

### DIFF
--- a/applications/dashboard/controllers/class.messagecontroller.php
+++ b/applications/dashboard/controllers/class.messagecontroller.php
@@ -165,6 +165,12 @@ class MessageController extends DashboardController {
         }
     }
 
+    /**
+     * Enable a message.
+     *
+     * @param $messageID
+     * @throws Exception
+     */
     public function enable($messageID) {
         $this->permission('Garden.Community.Manage');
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/class.messagecontroller.php
```diff
@@ -201,6 +219,7 @@ protected function setEnabled($messageID, $enabled) {
         } else {
             $this->informMessage(sprintf(t('%s disabled.'), t('Message')));
         }
+        Gdn::cache()->remove('Messages');
```
initvector: It looks like this might already be in the message model as [`MessageModel::messages(null)`](https://github.com/vanilla/vanilla/blob/5286ffe89f4b4b03a82469e77f96e2c517923912/applications/dashboard/models/class.messagemodel.php#L253) (used in `MessageModel::delete`).

linc: I used what's in `MessageModel::save()`.

linc: `Gdn::cache()->remove('Messages')` has obvious functionality.

`MessageModel::messages(null)` is ambiguous unless you know what passing a null does.

PR #5266, opened 2017-03-01
Add ability to modify root node type in Garden\Schema

### LINKED ISSUE #5249: Allow global schema arrays
Currently, schemas can only validate associative arrays at the root level. In the APIv2 we will also be having a lot of arrays and maybe even basic types at the root. We are going to need to have some support for changing the root type. I propose this notation:

```php
$schema = new Schema([
    '*:a' => [
        'name:s' => 'The title of the discussion.',
        'body:s' => 'The body of the discussion.'
    ]
]);
```

Specifying a single root field name of "*" is saying that the root type of the schema is that type. So the above example is meant to validate an array of discussions rather than a single discussion.

initvector:
Closes #5249

### DIFF
--- a/library/Garden/Schema.php
+++ b/library/Garden/Schema.php
@@ -583,13 +583,23 @@ class Schema implements \JsonSerializable {
     public function validate(array &$data, Validation &$validation = null) {
         $schema = static::getFields($this->schema);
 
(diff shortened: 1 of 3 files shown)


tburry:
Deprecated.

PR #5269, opened 2017-03-02
Add the PrunableTrait

tburry:
This trait allows models to add record pruning. You use it in a model
and then usually add a call to `$this->prune()` in the insert method.


tburry:
Too late. This got in from another PR.

PR #5274, opened 2017-03-02
Add a hero component to the styleguide and a rendering function

beckyvb:
Adds a hero component to the styleguide and adds rendering functions. Implements this function in the help and tutorial sections of the dashboard.

This is a hero component:
screen shot 2017-03-02 at 3 54 14 pm

It can exist with or without an image or with or without a button.


initvector:
Looks good, but needs some conflict resolution!

PR #5285, opened 2017-03-03
Exclude directory content not the directory itself!

DaazKu:
I !@#$! up by approving https://github.com/vanilla/vanilla/pull/5173

https://git-scm.com/docs/gitignore:
> An optional prefix "!" which negates the pattern; any [...] again. **It is not possible to re-include a file if a parent directory of that file is excluded.*** [...].


tburry:
Haha did not know.

Issue #3971, opened 2016-05-14
Move tag administration into core


tburry:
Move all the plugin into core and add an option to enable/disabled tagging. (This will control the UI. The ability of using tagging by other plugins will always work)

PR #5292, opened 2017-03-06
Move tagging administration into core

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/3971

Convert all the configurations.
Convert the permissions.
Move all the functionalities into core.


applications/vanilla/modules/class.tagmodule.php
```diff
@@ -157,7 +157,7 @@ public function assetTarget() {
      * @return string
      */
     public function inlineDisplay() {
-        if (!c('EnabledPlugins.Tagging')) {
+        if (!c('Vanilla.Tagging.EnableUI')) {
```
linc: The convention is `.Enabled` so let's stick with that.

DaazKu: Enabled is misleading. Tagging is always "enabled" now. The UI component is what is disabled/enabled.

`Vanilla.Tagging.UI.Enabled`?

DaazKu: My focus was bringing the code into core. I did no do any modification that was not needed.

I guess it wouldn't hurt to clean it up a bit though.

PR #5302, opened 2017-03-09
Add focus styling to radios and checkboxes

beckyvb:
Add focus styling to radios and checkboxes.


beckyvb:
Before this PR, the hover and focus styling were identical, and the focus styling did not extend to all the cases where a checkbox or radio needed focus styling. This PR separates focus and hover styling into two different mixins and applies them. The CSS rules you see removed for focus are added later.

PR #5303, opened 2017-03-09
Have modals check for a header-block element for the title

beckyvb:
Our `heading()` function wraps the h1 in a `.header-block` div. The modal looks for and removes the h1 element from the body to add to title. If the h1 is wrapped in a `.header-block` element, the `.header-block` element is left behind, leaving a gap in the body.

This checks whether a `.header-block` exists and removes it after it removes the title. After this, devs can freely use `heading()` to render their headings without having to worry about how it may look in a modal.

### DIFF
--- a/applications/dashboard/js/dashboard.js
+++ b/applications/dashboard/js/dashboard.js
@@ -624,8 +624,11 @@ var DashboardModal = (function() {
             // Pull out the H1 block from the view to add to the modal title
             if (this.settings.modalType !== 'noheader' && $title.length !== 0) {
                 title = $title.html();
-                $title.remove();
-                $elem.find('.header-block').remove();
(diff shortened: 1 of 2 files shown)


applications/dashboard/js/dashboard.js
```diff
@@ -625,6 +625,7 @@ var DashboardModal = (function() {
             if (this.settings.modalType !== 'noheader' && $title.length !== 0) {
                 title = $title.html();
                 $title.remove();
```
slafleche: If the title is in the .header-block, do we really need to remove the title separately?

beckyvb: We need to remove the title if the header block doesn't exist. I updated the block to make this clearer.

PR #5308, opened 2017-03-11
Recursively scan addon classes

### LINKED ISSUE #5233: Recursively scan addon directories for classes
Right now, the **Addon->scanClassPaths()** method returns PHP files only one level deep within an addon. Once we start using namespaces more there is the potential for more nested directories.

- The **Addon->scanClassPaths()** method should recurse into the whitelisted directories and return a flat list of all PHP files found.
- Add the "/src" directory to the list of directories searched.

tburry:
Instead of scanning just the first level of source directories, addons
will now recurse. Also added support for uppercase directories and the
“src” directory in preperation for namespace support.

Closes #5233


library/Vanilla/Addon.php
```diff
@@ -448,26 +449,68 @@ private function scanClasses() {
...
-        foreach ($globs as $glob) {
-            $paths = $this->glob($glob);
-            $result = array_merge($result, $paths);
+        foreach ($dirs as $dir) {
+            foreach ($this->scanDirPhp($dir) as $path) {
+                yield $path;
+            }
         }
 
         return $result;
```
initvector: Non-empty return statements do not appear to be valid in PHP, prior to v7. This is also true with HHVM. The build is currently failing with the following error:

> Generators cannot return values using "return"

From PHP's [Generator Syntax docs page](http://php.net/manual/en/language.generators.syntax.php):

> A generator cannot return a value: doing so will result in a compile error. An empty return statement is valid syntax within a generator and it will terminate the generator.

tburry: Yeah that was a mistake.

PR #5311, opened 2017-03-13
Remove URL scheme from css

lesaff:
I am removing the URL scheme in the css to allow flexibility when forum is ran in an HTTPS mode


linc:
Thanks!

PR #5317, opened 2017-03-14
Replace bundled Garden Schema with project dependency

### LINKED ISSUE #5316: Use external Garden Schema library
Remove bundled Schema library in favor of using [vanilla/garden-schema](https://github.com/vanilla/garden-schema) as a project dependency.

initvector:
This update swaps out the bundled Garden Schema with a dependency on [vanilla/garden-schema](https://github.com/vanilla/garden-schema).

All classes and tests related to Schema have been removed.

Closes #5316


composer.json
```diff
@@ -38,7 +38,8 @@
         "vanilla/htmlawed": "~2.0",
         "vanilla/legacy-oauth": "~1.0",
         "vanilla/legacy-passwords": "~1.0",
-        "vanilla/nbbc": "~2.1"
+        "vanilla/nbbc": "~2.1",
+        "vanilla/garden-schema": "~1.0"
```
tburry: Always require the most recent version when adding a package and use the "^" semantic requirement.

initvector: All set!

PR #5319, opened 2017-03-15
Move flood control to a trait and make the models use it

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/5045

This PR adds 
- CacheInterface
    - Ultimately it will inherit the SimpleCache interface from PSR
    - Add a level of abstraction
- FloodControlTrait
    - Receive a CacheInterface (which is either an adapter of UserAttributes or Gdn_Cache). The latter will be used to fix concurrency issues mentioned in vanilla/vanilla-patches/issues/65
- Some folders in library/core
    - Namely, the helpers folder. We can put utility classes in there!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ addons:
 install:
   - composer self-update
   - composer install -o
+  - composer require phpunit/phpunit ~5
   - tests/travis/nginx/install-nginx.sh
 
 script:
@@ -35,6 +36,6 @@ script:
   - tests/travis/php-lint.sh ./library
   - tests/travis/php-lint.sh ./plugins
   - tests/travis/php-lint.sh ./themes
-  - phpunit -c phpunit.xml.dist
+  - ./vendor/bin/phpunit -c phpunit.xml.dist --coverage-clover=coverage.clover
   - cat /tmp/error.log
   - cat /tmp/access.log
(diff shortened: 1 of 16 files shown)


applications/conversations/models/class.conversationsmodel.php
```diff
@@ -13,6 +13,13 @@
  */
 abstract class ConversationsModel extends Gdn_Model {
 
+    use \Vanilla\FloodControlTrait;
+
+    /**
+     * @var \Vanilla\CacheInterface Object used to store the FloodControl data.
+     */
+    protected $floodControlStorageObject;
```
linc: `$floodControlTracker` or `$floodgate` or even `$floodController` would be nicer. "StorageObject" makes it sound like I woke up in a Java shop.

linc: I vote floodgate, myself.

initvector: Second.

.travis.yml
```diff
@@ -35,6 +36,6 @@ script:
   - tests/travis/php-lint.sh ./library
   - tests/travis/php-lint.sh ./plugins
   - tests/travis/php-lint.sh ./themes
-  - phpunit -c phpunit.xml.dist
+  - ./vendor/bin/phpunit -c phpunit.xml.dist --coverage-clover=coverage.clover
```
DaazKu: Yes they are. Unless I rebase.

initvector: Yeah, they're identical to what's already in core. The diff just threw me off. Nevermind!

initvector:
Need to update the `$floodControlStorageObject` references in the discussion model and then I think this'll be ready to go.

PR #5320, opened 2017-03-16
Optionally block images embedded from untrusted sources displaying.

nervoustwit:
Presently if a forum admin wants to stop users from embedding images from untrusted domains they are forced to disallow embedding any images. This pull request would scan all user generated input before displaying it and if there is an image that is not from forum's CDN or domain or a list of trusted domains, it will display as a link to the image instead of as an embedded image.


library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+        // If no images are found in the HTML, pass the HTML to the renderer as is.
+        if (!$matches) {
+            return $html;
+        }
+
+        foreach ($matches as $image)  {
+            $imageUrl = parse_url($image[1]);
+            $stripImage = false;
+
+            // If the image is not coming from this forum's CDN, strip this image.
```
linc: This is not the correct was to interface with a CDN.

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+        }
+
+        preg_match_all('/\<img\s+src\s*=\s*[\"\'](.*?)[\"\'].*\>/', $html, $matches, PREG_SET_ORDER);
+        // If no images are found in the HTML, pass the HTML to the renderer as is.
+        if (!$matches) {
+            return $html;
+        }
+
+        foreach ($matches as $image)  {
+            $imageUrl = parse_url($image[1]);
```
linc: You can't assume backreference 1 is always set.

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+            $stripImage = false;
+
+            // If the image is not coming from this forum's CDN, strip this image.
+            if (c('S3.Prefix') && c('S3.Zone')) {
+                if ($imageUrl['host'] !== c('S3.Zone').".v-cdn.net" && !stringBeginsWith($imageUrl['path'], '/'.c('S3.Prefix'))) {
+                    $stripImage = true;
+                }
+            }
+
+            // If the host of the image source is a trusted domain, strip this image
```
linc: is _not_ a trusted domain

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+            if (!in_array(val('host', $imageUrl), trustedDomains())) {
+                $stripImage = true;
+            }
+
+            // Replace the image embed tag with a link to the image.
+            if ($stripImage) {
+                $imageName = trim(substr(val('path', $imageUrl), strrpos(val('path', $imageUrl), '\/')), '/');
+                $html = preg_replace('/\<img\s+src\s*=\s*[\"\']('.preg_quote($image[1], '/').')[\"\'].*\>/', '<a href="$1" target="_blank" title="'.t('Image from unconfirmed host').'">'.$imageName.'</a>', $html);
```
linc: I do not think that title helps, and in any case it would need to be added to locales if it was.

I don't think we want to hard-code `target="_blank"` either.

nervoustwit:
This should not be merged before https://github.com/vanilla/vanilla/pull/5344.

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
 
+    /**
+     * Optionally replace any embedded images from untrusted sites with links to the image.
+     *
+     * @param string $html String from which you might want to strip out untrusted images.
+     * @return string The string with images replaced by links.
+     */
+    public static function sanitizeImages($html) {
+        // If the config does not have a SanitizeImages entry or it is set to false, pass the HTML to the renderer as is.
+        if (!c('Garden.Format.SanitizeImages', false)) {
```
tburry: Make this a property that gets initialized by the config. Also, I'd put the check in the call above rather than here. Someone may want to just call this method.

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+        if (!$matches) {
+            return $html;
+        }
+
+        foreach ($matches as $image)  {
+            $imageUrl = (isset($image[1])) ? parse_url($image[1]) : null;
+            $stripImage = true;
+
+            // If the image is coming from this forum's CDN, do not strip this image.
+            if (c('S3.Prefix') && c('S3.Zone')) {
```
tburry: Gdn_Format should not know about this. My suggestion is that you make the S3 plugin add a trusted domain (although see my note below).

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+                if (val('host', $imageUrl) === c('S3.Zone').".v-cdn.net" && stringBeginsWith(val('path', $imageUrl), '/'.c('S3.Prefix'))) {
+                    $stripImage = false;
+                }
+            }
+
+            // If the host of the image source is a trusted domain and not from the CDN, strip this image
+            if (in_array(val('host', $imageUrl), trustedDomains()) && $stripImage === true) {
```
tburry: The trustedDomains() isn't really for this. They are for redirects. You can see that a site may want to enable images from imgur.com, but not allow redirects to it because it is still an external site.

tburry: I recommend adding a method on Gdn_Format called "isSafeUrl()" that can be used for other embeds too down the line. For now you can just make it work off another config setting. I suspect we'll want to add a bunch of standard ones. My worry though is that customers will fill this up with a thousand domains.

library/core/class.format.php
```diff
@@ -1713,6 +1714,49 @@ public static function mentions($Mixed) {
...
+                $stripImage = false;
+            }
+
+            // Replace the image embed tag with a link to the image.
+            if ($stripImage) {
+                $imageName = trim(substr(val('path', $imageUrl), strrpos(val('path', $imageUrl), '/')), '/');
+                $html = preg_replace('/\<img\s+src\s*=\s*[\"\']('.preg_quote($image[1], '/').')[\"\'].*\>/', '<a href="$1">'.$imageName.'</a>', $html);
```
tburry: Don't do a preg_replace() here. Just make the link with your URL variable from above.

nervoustwit:
I added a toggle in the advanced settings in the dashboard to enforce sanitizing image embeds and a text box for listing your whitelisted domains to uncouple this functionality from the safe redirect domains.

nervoustwit:
I am closing this PR since it was decided to solve this problem by using a plugin until we can come up with a more thorough and elegant solution.

https://github.com/vanilla/internal/pull/970

PR #5332, opened 2017-03-20
Fix time_zone SQL bug when configuring connection on some systems

initvector:
This update separates configuring the `names` and `time_zone` values for the database connection. Executing them in one function call can potentially cause a fatal error on some systems.


linc:
Issue introduced in https://github.com/vanilla/vanilla/pull/5257

DaazKu:
> [12:18 PM] Ryan Perry:
    $Configuration['Database']['ConnectionOptions']['1002'] = "set names 'utf8'";
[12:18 PM] Ryan Perry: That's what the server config is on inf. It's just the one statement. That's all that was being run on there, before.
[12:19 PM] Ryan Perry: Now, it's trying to run two of those statements and we're seeing the error.

Issue #4473, opened 2016-09-14
Dashboard v3: The dashboard nav should be updated on plugins activation.


DaazKu:
If the plugin add a new menu item you have to refresh the page to access it.
The nav should be updated so that we do not have to refresh.

PR #5336, opened 2017-03-20
Update the panel navigation on enabling or disabling a plugin

beckyvb:
Updates the panel navigation when a plugin is enabled or disabled. 

Also: 
* Updates calls to consts in the SiteNavModule from `self::` to `static::` for late-binding to allow child classes to override. This allows the DashboardNavModule to override the default section key.
* Moves the rendering call from `handleAddonToggle()` method to the endpoint instead, so it does not have to be the last call in an endpoint that uses it.

Closes https://github.com/vanilla/vanilla/issues/4473


initvector:
Neat-O.

PR #5340, opened 2017-03-21
Make Route::isMapped() public

tburry:
This isn’t optimal, but I need it for Swagger right now.


DaazKu:
If you do not want to expose the method you can always do an exposer with a note that it should be refactored:

```php
RouteExposer extends Route {
     public static isMapped(Route $route, \ReflectionParameter $param, $type = 0) {
         return $route->isMapped($param, $type);
     }
}
```

tburry:
I don't really want to add a different route to the container @DaazKu.

Issue #5334, opened 2017-03-20
Add an optional site setting to send users to /leaving when clicking on a ugc link


beckyvb:
Add an optional site setting to send users to /leaving when clicking on a ugc link. This will hijack the ugc link and send users to a page to have them confirm they want to leave the site.

PR #5344, opened 2017-03-21
Add an optional site setting to send users to /leaving when clicking a ugc link

beckyvb:
* Breaks the embed code replacement into its own function `embedReplacement($url)`, and optimizes and refactors the embed code replacement.
* ~Adds a `matchesTrustedDomain($url)` function to test whether a given url matches a trusted domain in the `linksCallback` function.~
* Adds a redirect to `/leaving` to any non-trusted link if the `Garden.Format.WarnLeaving` config is set.
* Adds a `reverse` param to the toggle to handle "disable"-type config settings.
* Adds controls for the `Garden.Format.WarnLeaving` and `Garden.Format.DisableUrlEmbeds` config settings to the Advanced settings page in the dashboard.
* Adds a title to the `/leaving` page.

Bonus:
* Fixes double-urling in the SafeRedirect function.
* Fixes Wistia and Twitch embeds

Closes https://github.com/vanilla/vanilla/issues/5334

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3110,6 +3110,27 @@ if (!function_exists('safeImage')) {
     }
 }
 
+if (!function_exists('matchesTrustedDomain')) {
+    /**
+     * Tests whether a given url matches one of our trusted domains.
+     *
+     * @param string $url The url to test.
+     * @return bool Whether the url matches a trusted domain.
+     */
+    function matchesTrustedDomain($url = '') {
+        $url = url($url, true);
+        $trustedDomains = trustedDomains();
+
+        foreach ($trustedDomains as $trustedDomain) {
+            if (urlMatch($trustedDomain, $url)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+}
+
 if (!function_exists('safeRedirect')) {
     /**
      * Redirect, but only to a safe domain.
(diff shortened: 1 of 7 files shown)


library/core/functions.general.php
```diff
@@ -3110,6 +3110,32 @@ function safeImage($ImageUrl, $MinHeight = 0, $MinWidth = 0) {
     }
 }
```
DaazKu: https://github.com/vanilla/vanilla/blob/fa1c31f54f469ca53eaf1d59596f7a4585e780c9/library/core/functions.general.php#L3705

The function already exists!

DaazKu: There is a difference between URLmatch and the preg_match thought!
I don't know if we can safely swap one for the other. TBD

DaazKu: Sorry the comment has been made on the line above...

See: https://github.com/vanilla/vanilla/pull/5344/files/990a0b7d9a644e1ecf06392b8408f95e19df65c2#diff-604330a5cab4bf405103e84612897812R3113

beckyvb: > There is a difference between URLmatch and the preg_match thought! I don't know if we can safely swap one for the other. TBD

There is a difference between the two functions, which is why I pulled it out into its own function. The `isTrustedDomain()` function doesn't match `*` wildcards, whereas the `matchTrustedDomain()` function does.

DaazKu: That is really confusing.
The doc block says URL.

And that function seems dangerous:
```php
print_r([
    matchesTrustedDomain('http://vanilla.*'), // false,
    matchesTrustedDomain('vanilla.*'), // true
    matchesTrustedDomain('www.somethingelse.*'), // true
    matchesTrustedDomain('literallyanything.com/somethingwithoutavalidscheme'), // true
    matchesTrustedDomain('*'), // true
])
```

DaazKu: isTrustedDomain was added 2 years ago:
https://github.com/vanilla/vanilla/commit/110419b1dfdb747799eac2097e288647d3e891f1
wilcards support was added 10 months ago but we never updated the isTrustedDomain function. (It is used in one place only)
https://github.com/vanilla/vanilla/commit/d2a593f912b0e6debd6f15b7186d571573325043

We should just update isTrustedDomain.
Use http://php.net/manual/en/function.parse-url.php to get the host and pregmatch on that! (preg_quote and replace `\*` by `.*?`) for the preg_match.

DaazKu:
@linc A little help with Travis?

Issue #5348, opened 2017-03-22
Refactor isTrustedDomain to work the list of trusted domains


DaazKu:
`isTrustedDomain()` was added 2 years ago: https://github.com/vanilla/vanilla/commit/110419b1dfdb747799eac2097e288647d3e891f1

10 months ago, we added a note on the settings page saying wildcards are supported for domains, but didn't alter the `isTrustedDomain()` function to use wildcards:
https://github.com/vanilla/vanilla/commit/d2a593f912b0e6debd6f15b7186d571573325043#diff-d891be17f4782ddf3f31ac7c33c8aa3aR112

It makes no sense that isTrustedDomain does not respect what is in the configuration.

`safeRedirect()` should be using `isTrustedDomain()` to know where to redirect but is instead iterating on the `trustedDomains()` list and using `urlMatch()`. (Which is basically what `isTrustedDomains()` is for)

Also note that `isTrustedDomains()` is only used in one place in all our repositories right now and it is to set the CORS header.

Now the big problem with all that is that wildcards can cause security issues if wrongly used:
`google.*` can match `google.com` and `google.co.uk` but also `google.evildomain.com`

Some of us say that this is the responsibility of the client to use them properly but others are against updating `isTrustedDomains()` if wildcards are allowed in domain names...

The lucky assignee of this issue will have to find out if we remove the wildcard support or if we update the function!

DaazKu:
Related: https://github.com/vanilla/vanilla/pull/5344#discussion_r107491584

PR #5349, opened 2017-03-22
Add CORS support to the APIv2

tburry:
- The dispatcher can be passed a set of allowed origins and will set CORS headers appropriately.
- There is now a PreflightRoute that responds to OPTIONS requests with an empty response.
- The bootstrap wires up the allowed origins with trusted domains.


library/core/functions.general.php
```diff
@@ -3565,6 +3565,26 @@ function trustedDomains() {
     }
 }
 
+if (!function_exists('isTrustedUrl')) {
```
DaazKu: We already have a similar function!
See https://github.com/vanilla/vanilla/issues/5348

tburry: Oh thanks. I'll extend isTrustedDomain to support what I need. Longer term we'll need to make a class.

tburry:
I split out the isTrustedDomain() stuff to #5361. I'll rebase this when that is merged.

PR #5350, opened 2017-03-23
Fix ProfileRequest Params in Class OAuth2

R-J:
There has been a protected property called `$profileRequestParams` which couldn't be set anywhere and there had been a method `setGetProfileParams()` which was setting a never used (or even declared) property `$getProfileParams`.

I assume they should be connected because then they would allow plugin authors to add additional parameters to profile requests.

I've changed their names completely to reflect the already existing `$requestAccessTokenParams` and `setRequestAccessTokenParams()` to `$requestProfileParams` and `setRequestProfileParams()` accordingly.

<hr>

The default settings view has been 'plugins/oauth2'  and in the render() call it has prefixed with 'plugins/' a second time. That prefixing in the render call
a) results to "plugins/plugins/oauth2" for default view, which is obviously wrong and
b) hinders themes and addons to use the settings method of the class because no view could be set.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -384,7 +384,7 @@ class Gdn_OAuth2 extends Gdn_Plugin {
         $redirectUrls = Gdn::request()->url('/entry/'. $this->getProviderKey(), true, true);
         $sender->setData('redirectUrls', $redirectUrls);
 
-        $sender->render('settings', '', 'plugins/'.$view);
+        $sender->render('settings', '', $view);
     }
 
 
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -229,7 +229,7 @@ class Gdn_OAuth2 extends Gdn_Plugin {
      *
      * @return $this Return this object for chaining purposes.
      */
-    public function setRequestProfileParams($params = []) {
+    public function setRequestProfileParams(array $params) {
         $this->requestProfileParams = $params;
         return $this;
     }
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -46,7 +46,7 @@ class Gdn_OAuth2 extends Gdn_Plugin {
     protected $requestAccessTokenParams = [];
 
     /** @var array optional additional get params to be passed in the request for profile */
-    protected $profileRequestParams = [];
+    protected $requestProfileParams = [];
 
     /** @var  @var string optional set the settings view */
     protected $settingsView;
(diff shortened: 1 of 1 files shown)


R-J:
I would like to backport that to Vanilla 2.4 because that class is super useful! Writing a LinkedIn connect was a breeze :+1: 
I even used it for Vanilla 2.3 although that required emulating some non existing methods...

Because I never did so with only git I would be glad if anybody could affirm that I plan to do the right thing...
Basically I would do all the same of what I did, but this time starting from the Vanilla 2.4 branch:
Checkout 2.4, create branch, do changes and commit, push (from branch) to origin with `git push -u origin branch-name`, come back to GitHub and finally make the PR. Would that be correct?

library/core/class.oauth2.php
```diff
@@ -220,13 +222,15 @@ public function setRequestAccessTokenParams($params) {
...
+     * in the getProfile request.
+     *
+     * @param array $params Params to add to the get profile request.
      *
-     * @param string $params.
      * @return $this Return this object for chaining purposes.
      */
-    public function setGetProfileParams($params) {
-        $this->getProfileParams = $params;
+    public function setRequestProfileParams($params = []) {
```
DaazKu: Why did you default $params to an empty array?
The less default parameters the better. It also doesn't make sense to me in this context because it would not be very useful to call `$obj->setRequestProfileParams();`

R-J: This is my new code where this variable is involved:

~~~
protected $requestProfileParams = [];

// (...)

public function setRequestProfileParams($params = []) {
    $this->requestProfileParams = $params;
    return $this;
}

// (...)

$requestParams = array_merge($defaultParams, $this->requestProfileParams);
~~~

Right from the start the default value for requestProfileParams is an empty array.

If the setter method is called like that `$obj->setRequestProfileParams();`, I would expect that the variable is null afterwards, though I'm not sure about that. But I'm sure it wouldn't be an empty array or any array at all. And that's a problem because the variable is used inside an `array_merge()` where the arguments need to be arrays.

Because of that it might be even better to ensure that the value is an array. What do you think about the following variant?

~~~
public function setRequestProfileParams($params) {
    $this->requestProfileParams = (arr)$params;
    return $this;
}
~~~

DaazKu: The function should not be called without a parameter since it is a setter.
The DocBlock says that the parameter needs to be an array. 

That should be enough.

DaazKu: If you want to make make sure that calling setRequestProfileParams() will fail you can do:
```php
public function setRequestProfileParams(array $params) {
    $this->requestProfileParams = $params;
    return $this;
}
```

DaazKu: You can remove the Revisit tag after you updated the code. Thanks!

DaazKu:
@R-J Could you update this please?

PR #5361, opened 2017-03-24
Make isTrustedDomain() work with URLs and use urlMatch()

tburry:
Add the following functionality to isTrustedDomain():

- You can now pass a URL or a domain to isTrustedDomain().
- The domains in the whitelist accept schemes and glob style wildcards.
- The comparison between domains is a bit more efficient.

Note that this breaks some BC where before we accepted regexes in the
domains. IMO freeform regexes should be discouraged more and more. I’m
not sure where they have been used though.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3727,7 +3727,7 @@ if (!function_exists('isTrustedDomain')) {
                 // Store the trusted domain by its host name.
                 if (strpos($domain, '//') === false) {
                     $domain = '//'.$domain;
-        }
+                }
                 $host = preg_replace('`^(\*?\.)`', '', parse_url($domain, PHP_URL_HOST));
                 $trusted[$host] = $domain;
             }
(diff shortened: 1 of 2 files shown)


library/core/functions.general.php
```diff
@@ -3697,32 +3702,45 @@ function isSafeUrl($Url) {
...
         // If we haven't already compiled an array of trusted domains, grab them.
-        if (!is_array($trustedDomains)) {
+        if ($trusted === null) {
+            $trusted = [];
             $trustedDomains = trustedDomains();
+            foreach ($trustedDomains as $domain) {
+                // Store the trusted domain by its host name.
+                if (strpos($domain, '//') === false) {
+                    $domain = '//'.$domain;
+        }
```
DaazKu: This one is trying to escape!

DaazKu:
This PR seems to address the following issue: https://github.com/vanilla/vanilla/issues/5348

PR #5362, opened 2017-03-24
Handle leaving urls already in anchor tags

beckyvb:
Handle leaving urls already in anchor tags. The linksCallback function adds an anchor to plaintext links in a post. If the anchor already exists, it ignores it.

This replaces the url in the case where the link is already in an anchor.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1478,6 +1478,15 @@ EOT;
             $inTag--;
         }
 
+        if (c('Garden.Format.WarnLeaving', false) && isset($matches[4]) && $inAnchor) {
+            // This is a the href url value in an anchor tag.
+            $url = $matches[4];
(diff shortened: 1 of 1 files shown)


library/core/class.format.php
```diff
@@ -1478,6 +1478,15 @@ public static function linksCallback($matches) {
             $inTag--;
         }
 
+        if (c('Garden.Format.WarnLeaving', false) && isset($matches[4]) && $inAnchor) {
+            // This is a the href url value in an anchor tag.
+            $url = $matches[4];
+            $domain = parse_url($url, PHP_URL_HOST);
+            if (!isTrustedDomain($domain)) {
+                return url('/home/leaving?target='.$url).'" class="Popup';
```
DaazKu: you should `urlencode($url)`

Issue #5063, opened 2017-01-27
Facebook plug-in does not adhere to Platform Policy


initvector:
If a user has connected their Facebook account via the Facebook plug-in, and attempts to share a post, the message is automatically populated with a portion of the body of the post.  Automatically populating the message of a post is against Facebook's current platform policy.

See https://developers.facebook.com/policy/#control

> Don't prefill captions, comments, messages, or the user message parameter of posts with content a person or business didn’t create, even if the person can edit or remove the content before sharing. You may use our Share Dialogs to prefill a single hashtag in a post, but don't prefill any content a person or business didn't create via the API.

The plug-in should probably avoid using the API to automatically share posts on a user's behalf and stick to using the standard Share Dialogs.

PR #5371, opened 2017-03-27
Remove facebook automatic social sharing feature

DaazKu:
Closes https://github.com/vanilla/vanilla/issues/5063


DaazKu:
Also fixes https://github.com/vanilla/vanilla/issues/5381

initvector:
The "Social Share" button still makes a post on behalf of the user without any further interaction. I don't think it technically fills in a message, just the link to the post, but it might be nice if we could use the regular ole Facebook share dialog.

DaazKu:
?????? Not in my case

screen shot 2017-03-29 at 5 58 38 pm

PR #5373, opened 2017-03-27
Add ability to collapse block tags into one newline when plain text

initvector:
This update adds a parameter to `Gdn_Format::plainText` to allow collapsing sequential block tags into a single newline. For example, a formatted string that contains `</p></div></div>` would normally be converted to `\n\n\n\n\n\n`. This change will allow that same substring to be converted to `\n\n`.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1039,10 +1039,11 @@ class Gdn_Format {
      *
      * @param string $Body The text to format.
      * @param string $Format The current format of the text.
+     * @param bool $collapse Collapse block tags in a sequence into a single newline replacement?
      * @return string
      * @since 2.1
      */
-    public static function plainText($Body, $Format = 'Html') {
+    public static function plainText($Body, $Format = 'Html', $collapse = false) {
         $Result = Gdn_Format::to($Body, $Format);
         $Result = Gdn_Format::replaceSpoilers($Result);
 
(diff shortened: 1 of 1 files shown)


DaazKu:
Note quite correct:
```php
$body = <<<TEXT
<div>
    <div>
<div>
        <div>
<div>
            <div>
<div>
                <div>
                    HAAA YEAH!
                    This is nice!
                </div>
</div>
            </div>
</div>
        </div>
</div>
    </div>
</div>
TEXT;

echo '<pre>'.Gdn_Format::plainText($body, 'Html', true).'</pre>';
```

Expected: 
```
                     HAAA YEAH!
                     This is nice!
```
Returned:
```
HAAA YEAH!
                     This is nice!
```

DaazKu:
Nvm the first test. The functions was already going that because of the trimming. I would still call it a bug.
```php
$body = <<<TEXT
<div>
    <div>
<div>
        <div>
<div>   test
            <div><div></div><div></div><div></div><div></div><div></div>
<div>
                <div>
                    HAAA YEAH!
                    This is nice!
                    <div>
                        HAAA YEAH!
                        This is nice!
                    </div>
                </div>
                <div>
                </div>
</div>
            </div>
</div>
        </div>
</div>
    </div>
</div>
TEXT;
echo '<pre>'.Gdn_Format::plainText($body, 'Html', true).'</pre>';
```
The above code gives:
```
test
            











                
                    HAAA YEAH!
                    This is nice!
                    
                        HAAA YEAH!
                        This is nice!
```

I would have expected to get

```
    test
                    HAAA YEAH!
                    This is nice!
                    
                        HAAA YEAH!
                        This is nice!
```

initvector:
This update only alters `Gdn_Format::plainText` to give the option to collapse multiple closing block tags into a single newline replacement. Prior to this update, all closing block tags are replaced with a separate set of newline characters, leading to multiple empty lines.

For the above examples to be valid for *this fix*, they would need to contain multiple adjacent closing block tags, which should then be collapsed into a single newline sequence, instead of one for each individual closing tag.

DaazKu:
Update the parameter description that that will be good!

PR #5375, opened 2017-03-28
Fix inline doc for trusted domain wildcards

linc:
From dashboard inline docs:

> Use * as a wildcard.

This is incorrect.

From `isTrustedDomain()`:

 > See if the current trusted domain matches fully against the domain we're testing or if it matches its end (e.g. domain.tld should match domain.tld and sub.domain.tld)

Fixed this in the dashboard.

### DIFF
--- a/applications/vanilla/views/vanillasettings/advanced.php
+++ b/applications/vanilla/views/vanillasettings/advanced.php
@@ -101,7 +101,7 @@ echo $this->Form->errors();
                     );
                     ?>
                     </p>
-                    <p><strong><?php echo t('Note'); ?>:</strong> <?php echo t('Specify one domain per line. Use * for wildcard matches.'); ?></p>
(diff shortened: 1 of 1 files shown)


initvector:
The wildcard thing is true as of https://github.com/vanilla/vanilla/pull/5361 but maybe not how you might expect. "Trusted Domains" can be full URLs and not just domain names (e.g. https://example.com/sso/* can be a "trusted domain"). The title of that page and its contents could probably use an overhaul.

DaazKu:
The problem is that we are using the list for 2 different things: CORS and Redirects.

Redirects do respect the *.
CORS do not care.

Both feature use the same list. This thing is #%%$^ up.

Related:
- https://github.com/vanilla/vanilla/issues/5348
- https://github.com/vanilla/vanilla/pull/5349
- https://github.com/vanilla/vanilla/pull/5361
- https://github.com/vanilla/vanilla/pull/5344#discussion_r107491584

linc:
Yes, #5361 was the ticket I was looking for and missed.

Issue #5312, opened 2017-03-13
Remove the "set names" call.


tburry:
There is a better way to set the connection character encoding. Take a look here: https://phpdelusions.net/pdo

PR #5376, opened 2017-03-28
Set charset through PDO's dsn string

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/5312

### DIFF
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -150,13 +150,6 @@ class Gdn_Database {
      */
     protected function newPDO($Dsn, $User, $Password) {
         try {
-            if (strpos($Dsn, 'charset=') === false) {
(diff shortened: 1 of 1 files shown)


library/database/class.database.php
```diff
@@ -150,11 +150,15 @@ public function commitTransaction() {
      */
     protected function newPDO($Dsn, $User, $Password) {
         try {
+            if (strpos($Dsn, 'charset=') === false) {
```
tburry: I don't think this is the right place for this code. You should be adding the `charset` when the DSN is being constructed.

DaazKu: I actually wanted to do that at first but I feared that changing the location of where the charset is set could cause problems.

DaazKu: I guess I could do that, test and YOLO it.

PR #5378, opened 2017-03-28
Prevent creation of embed comments if embedding is not ON

DaazKu:
Prevent embed discussion creation if embed is not turned on


linc:
This has pretty serious repercussions for someone who is accidentally using Comments while embed is disabled, no?

DaazKu:
It could on a bad setup. That code is designed to work with embeds.

PR #5385, opened 2017-03-29
Condense 4 lines onto one. SUCH WOW -Doge

DaazKu:


tburry:
**Condensate**

1. verb. 1550s, "to make dense," from condens-, past participle stem of Latin condensare (see condense ) + -ate (2). Meaning "to become dense" is from c.1600.
2. verb. to shorten due to obsessive annoyance with the aesthetics of long things.

PR #5390, opened 2017-03-30
Replace dirname(__FILE__) with __DIR__

R-J:
dirname(__FILE__) has been used on several places, but since PHP 5.3 the identical magical constant __DIR__ is available and saves one function call.

There were some id attribute relicts in closing divs (in the log views) which I have removed, too.

### DIFF
--- a/themes/default/views
+++ b/themes/default/views
@@ -1,0 +1,1 @@
+/var/www/vanilla/themes/default/views/
--- a/themes/default/views
+++ b/themes/default/views
@@ -1,1 +1,0 @@
-/var/www/vanilla/themes/default/views/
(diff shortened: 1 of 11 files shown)


themes/default/views
```diff
@@ -0,0 +1 @@
+/var/www/vanilla/themes/default/views/
```
DaazKu: What is this?

DaazKu:
In the future try not to mix 2 different fixes in one PR. There should be one for the IDs and one for the constants.

But my previous question this is all fine and good. Thanks!

themes/default/views
R-J: That shouldn't be there. I linked a folder into the view to test something but of course I shouldn't have enclosed that...

Issue #5258, opened 2017-02-28
Delete conversation is incomplete


linc:
In Conversations hooks, `userModel_beforeDeleteUser_handler()` does this to delete a user's conversation associations:

`$Sender->SQL->delete('UserConversation', array('UserID' => $UserID));`

This only removes the deleted user's records, not the records for any messages they sent.

If their conversations were ALSO the most recent, that means the indices can reset in MySQL. It then will replay those conversationIDs, and those recipients will appear as "random" additions to newly created private messages.

PR #5394, opened 2017-03-31
Delete all UserConversation entries when delete a user's content

DaazKu:
Fix https://github.com/vanilla/vanilla/issues/5258

Clean UserConversation and ConversationMessage properly

### DIFF
--- a/applications/conversations/settings/class.hooks.php
+++ b/applications/conversations/settings/class.hooks.php
@@ -50,18 +50,16 @@ class ConversationsHooks implements Gdn_IPlugin {
             $sql = $Sender->SQL;
             $sql
                 ->from('UserConversation as uc')
-                    ->join('Conversation as c', 'c.ConversationID = uc.ConversationID')
+                ->join('Conversation as c', 'c.ConversationID = uc.ConversationID')
                 ->where(['c.InsertUserID' => $UserID])
                 ->orWhere(['c.UpdateUserID' => $UserID])
-                ->delete()
-            ;
+                ->delete();
             $sql
                 ->from('ConversationMessage as cm')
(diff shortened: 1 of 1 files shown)


linc:
nfl-ref-meeting-after-reviewing-the-play-the-call-on-the-field-stands-no-danglers

initvector:
Won't this potentially delete `UserConversation` and `ConversationMessage` records for every conversation the user participated in, now that we're targeting based on `Conversation.UpdateUserID`? If the user was the last poster in a group conversation, every other person in that conversation might lose access when the user is deleted.

The original issue also mentions a concern about replaying conversation/message IDs.

> If their conversations were ALSO the most recent, that means the indices can reset in MySQL. It then will replay those conversationIDs, and those recipients will appear as "random" additions to newly created private messages.

Is that something can be addressed here?

DaazKu:
> Won't this potentially delete UserConversation and ConversationMessage records for every conversation the user participated in, now that we're targeting based on Conversation.UpdateUserID? If the user was the last poster in a group conversation, every other person in that conversation might lose access when the user is deleted.

Nothing changed here. I'm just cleaning the UserConversation and ConversationMessage tables properly.
Now if we want to change the behaviour that's another thing!

> If their conversations were ALSO the most recent, that means the indices can reset in MySQL. It then will replay those conversationIDs, and those recipients will appear as "random" additions to newly created private messages.

Not an issue with the current fix since the records are properly cleaned.

initvector:
> Nothing changed here. I'm just cleaning the UserConversation and ConversationMessage tables properly.
Now if we want to change the behaviour that's another thing!

Very true! The thing I was worried about is actually existing behavior, introduced in the time before time.

Issue #5284, opened 2017-03-03
Make sure that no indexes could be too large when using utf8mb4


DaazKu:
https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-conversion.html

> InnoDB has a maximum index length of 767 bytes, so for utf8 or utf8mb4 columns, you can index a maximum of 255 or 191 characters, respectively

We have one case where it COULD, in theory, be problematic and it is in the `GDN_UserAuthenticationToken` table:

```sql
`Token` varchar(128)
`ProviderKey` varchar(64)
```

In practice it will probably never happen but who knows.

tburry:
I wonder if our db object supports a different encoding on a per-table basis. That would be the best, but probably not. Otherwise, I'd knock the provider key down. NBD.

linc:
Let's also search around for additional `varchar(255)` definitions in our private repos as part of this.

DaazKu:
Here is what i've found:

```
avatarstock.plugin.php:64 -> Path varchar(255)
class.wordpresimportmodel.php:52 -> UrlCode varchar(255)
class.disqusimporter.plugin.php:232 -> ForeignID varchar(200)
class.flagging.plugin.php:398 -> ForeignURL varchar(255)
class.sitehub.plugin.php:59 -> Url varchar(255)
class.sitehub.plugin.php:97 -> UrlCode varchar(255)
class.slugurls.plugin.pgp:29 -> Slug varchar(200)
class.subcomunities.plugin.php:45 -> Folder varchar(255)
class.vanillapop.plugin.php:[628,634,640,646] -> SourceID varchar(255)
```

DaazKu:
We have 3 solutions that I know of here:

- Manually limit the index size to 191. 
    - This can cause collision...
- Shrink the column sizes.
   - We will need to do some verifications on the data first...
- Set the encoding to utf8 instead of utf8mb4 for these columns.
   - I do not think it is currently possible to do that using `Gdn_DatabaseStructure`.

Charrondev:
This seems to have already been an issue [before](https://github.com/vanilla/vanilla/issues/2082)

DaazKu:
The fix we did: https://github.com/vanilla/vanilla/pull/2712

linc:
Yup. The issue now is just fix (resize) the ones we missed before, mostly outside of core.

DaazKu:
https://github.com/vanilla/vanilla/pull/5402 will prevent data corruption on old installation.
[...]

PR #5402, opened 2017-04-04
Add integrity check when downsizing a varchar column

DaazKu:
Skip the alteration if the column content is larger than the new size.

This will ensure that we do not corrupt data while shrinking column size when fixing https://github.com/vanilla/vanilla/issues/5284

### DIFF
--- a/library/database/class.mysqlstructure.php
+++ b/library/database/class.mysqlstructure.php
@@ -585,7 +585,7 @@ class Gdn_MySQLStructure extends Gdn_DatabaseStructure {
                             Logger::event(
                                 'structure_integrity',
                                 Logger::ALERT,
-                                "Cannot modify {tableName}'s column {column} because it has a value of {maxVarcharLength,number} and the new length is {newLength,number} threshold.",
+                                "Cannot modify {tableName}'s column {column} because it has a value that is {maxVarcharLength,number} and the new length is {newLength,number}.",
                                 [
                                     'tableName' => $this->tableName(),
                                     'column' => $ColumnName,
--- a/library/database/class.mysqlstructure.php
+++ b/library/database/class.mysqlstructure.php
@@ -501,6 +501,7 @@ class Gdn_MySQLStructure extends Gdn_DatabaseStructure {
         $existingColumns = array_change_key_case($this->existingColumns());
         $columns = array_change_key_case($this->_Columns);
         $AlterSql = array();
+        $InvalidAlterSqlCount = 0;
 
         // 1. Remove any unnecessary columns if this is an explicit modification
         if ($Explicit) {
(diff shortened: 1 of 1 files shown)


library/database/class.mysqlstructure.php
```diff
@@ -572,6 +572,32 @@ protected function _modify($Explicit = false) {
                 $Comment = "-- Existing: $ExistingColumnDef, New: $ColumnDef";
 
                 if ($ExistingColumnDef !== $ColumnDef) {
+                    if (strcasecmp($ExistingColumn->Type, 'varchar') === 0&& strcasecmp($Column->Type, 'varchar') === 0
```
initvector: Need a space here:

```
=== 0&& strcasecmp
```

library/database/class.mysqlstructure.php
```diff
@@ -572,6 +572,32 @@ protected function _modify($Explicit = false) {
...
+                            ->firstRow(DATASET_TYPE_ARRAY);
+
+                        if ($charLength['MaxLength'] > $Column->Length) {
+                            $this->addIssue("The table's column was not altered because it contains a varchar of length {$charLength['MaxLength']}.", $Comment);
+
+                            // Log an event to be captured and analysed later.
+                            Logger::event(
+                                'structure_integrity',
+                                Logger::ALERT,
+                                "Cannot modify {tableName}'s column {column} because it has a value of {maxVarcharLength,number} and the new length it {rowThreshold,number} threshold.",
```
initvector: Might want to think about rephrasing this error message. Example:

> Cannot modify GDN_Discussion's column ForeignID because it has a value of 32 and the new length it 24 threshold.

initvector: There is no `rowThreshold` key in `$context`.

initvector:
Looks good, except for one thing: this update potentially kills /utility/structure.

```php
$charLength = $this->executeQuery("select max(char_length(`$ColumnName`)) as MaxLength from `$Px{$this->_TableName}`;")
->firstRow(DATASET_TYPE_ARRAY);
```

When scanning with /utility/structure, `Gdn_MySQLStructure::executeQuery` returns a boolean value, because it is in capture-only mode. However, this code relies on it returning an object (`Gdn_DataSet`), so you're potentially met with the following error:

> Fatal error: Call to a member function firstRow() on boolean

initvector:
Instead of making sure a result set is always returned, how about avoiding that whole block if the `CaptureOnly` property is truthy? As it stands now, the necessary updates won't show up in the "Database Structure Upgrades" list. An event is also being logged every time scan is run, stating that a column can't be modified, even though no attempt is being made to modify the database.

DaazKu:
We do need to execute that query even in capture mode because it modifies the alter query.

Issue #2464, opened 2015-02-01
Tagging discussions from list


linc:
I wanna be able to tag from the discussions list option menus. Fly-by tagging. Opening every discussion to tag hundreds of discussions is the worst. Specifically, this is Add Tag, not edit existing tags. If you need to change them, I think that should still require an Edit.

DaazKu:
Blocked by https://github.com/vanilla/vanilla/pull/5292

PR #5411, opened 2017-04-05
Add the Tag option to the discussion list menu

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/2464

You can only add new tags as specified in the issue.
Adding a tag that already is linked to the discussion does nothing.

### DIFF
--- a/applications/vanilla/views/discussion/tag.php
+++ b/applications/vanilla/views/discussion/tag.php
@@ -17,7 +17,7 @@ echo '</div>';
 ?>
 <?php
 echo '<div class="Buttons Buttons-Confirm">';
-echo    $this->Form->button('Save', array('class' => 'btn btn-primary'));
-echo    $this->Form->button('Cancel', array('type' => 'button', 'class' => 'btn Close'));
+echo    $form->button('Save', ['class' => 'Button Primary']);
+echo    $form->button('Cancel', ['type' => 'button', 'class' => 'Button Close']);
 echo '</div>';
 echo $form->close();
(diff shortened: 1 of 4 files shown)


applications/vanilla/views/discussion/tag.php
```diff
@@ -0,0 +1,16 @@
...
+echo $form->errors();
+
+echo '<div class="Form-Tags P">';
+    // Tag text box
+    echo $this->Form->label('Tags', 'Tags');
+    echo $this->Form->textBox('Tags', array('data-tags' => ''));
+echo '</div>';
+
+?>
+<?php echo $this->Form->close('Save');
```
initvector: This view is functional, but very minimal. It feels a bit cramped. Maybe add a simple sentence indicating what is being done (e.g. "Which tags would you like to add to the discussion?") and making the input box a bit wider. The design of these windows is all over the board, but some have an "OK" and a "Close" button with the "OK" button having the `PrimaryButton` class, allowing it to be styled a bit differently.

DaazKu: I literally copied what is in the admin section less one field.....

DaazKu: I personally like it "mininal" but it wouldn't hurt to add a description I guess.

As for the layout, I suggest that you raise an issue about it so that we can review these dialogs since it is identical to the dialog in the dashboard.

initvector: Styling in the admin area versus the public-facing site is a little different. If you look at windows for other options in the same discussion menu, you'll see there's a little bit more to them.

DaazKu: I added the text and the buttons.
The width is fine.

PR #5417, opened 2017-04-06
Fix FloodControl configuration names for conversations

DaazKu:
Backport to the latest release.
This will need to be pushed on all the clusters.

PM me if you want me to do the push.


initvector:
Backported to `release/2017-Q1-6`

Issue #5399, opened 2017-04-03
Formatting problems with mentions: add tokenization?


DaazKu:
We currently have a formatting problem with mentions.

Usernames containing valid "Formatting format" code are replaced.

For example, using markdown:
```
@"_k_" -> @"<em>k</em>"
```
Same for BBCode
```
@"[b]boldman[/b]" -> @"<b>boldman</b>"
```

Moving the parsing of mentions before the formatting doesn't solve the issue.
Do mentions -> format, using markdown:
```
@"_k_" -> <a rel="nofollow" href="/profile/_name_">@_name_</a> // This one is correct

// That one does a double anchor. They are side to side because htmLawed balances tags and anchors are not allowed inside another anchors.
@"[link](man)" -> <a rel="nofollow" href="/profile/%5Bname%5D%28lol%29">@</a><a rel="nofollow" href="lol">name</a> 
```

I think that the formatting pipeline needs a rework.
My suggestion would be:

- **Parse custom formatting like mentions and substitute them.**
- Do format transformation/formatting
- **Restore substitutions**
- format_filter
- Format links
- Emojis
- htmLawed

@linc told me that we want to redo the formatting pipeline. @tburry might have more information about how we are supposed to rework it.

tburry:
I think it would be better to restrict allowable usernames. If necessary, something like the markdown editor can be made to not mangle mentions, but that's not an optimal solution.

I don't want to get into cases where other formatters have to know what they are operating on. They should always operate on HTML.

DaazKu:
We cannot restrict usernames. We tried that already and we got a lot of problems doing it that way because of imports/SSO.

Usernames should be allowed to contain anything and be escaped properly!

linc:
I'm in favor of tokenizing mentions to sidestep this issue forever. 

Every time we've tried to restrict usernames someone has twisted our arms into an exception, then we say "ok no mentions" and then one day that's bargain turns into a bug ticket instead.

While I'm in favor of all efforts to restrict usernames, I also like making this bulletproof once and for all so we stop this parade of issues.

charrondev:
Non-security fixes will no longer be prioritized for advanced editor by the vanilla team. Rich Editor ships with Vanilla 2.8 as the default. https://github.com/vanilla/vanilla/pull/8302

PR #5423, opened 2017-04-07
Simplify parsing orders in formatters

linc:
There were a number of problematic inconsistencies & overlaps with our text processing. This change set goes method-by-method and refactors these issues and makes it very clear that you will always get the same result: uniformly sanitized output.

Key issues resolved:

* Inconsistent ways of sanitizing final output, making it difficult to ascertain whether it was truly sanitized.
* Not always sanitizing as the ultimate step in HTML transformations, which is critical.
* Redundant checking of `isHTML()` and getting of the HTMLformatter. This is handled in the newer `filterHTML()`.
* Unclear documentation.

These changes should have no noticeable effects on the vast majority of user-generated content.

Removing the whitespace (`?w=1`) will help on this one since I eliminated a couple large logic blocks.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -869,35 +869,22 @@ class Gdn_Format {
      * Does "magic" formatting of links, mentions, link embeds, emoji, & linebreaks.
      *
      * @param mixed $Mixed An object, array, or string to be formatted.
-     * @return string HTML
+     * @return string Sanitized HTML.
      */
     public static function html($Mixed) {
         if (!is_string($Mixed)) {
             return self::to($Mixed, 'Html');
         } else {
-            if (self::isHtml($Mixed)) {
-                // Purify HTML
-                $Mixed = Gdn_Format::htmlFilter($Mixed);
-
-                // nl2br
-                if (c('Garden.Format.ReplaceNewlines', true)) {
-                    $Mixed = preg_replace("/(\015\012)|(\015)|(\012)/", "<br />", $Mixed);
-                    $Mixed = fixNl2Br($Mixed);
-                }
-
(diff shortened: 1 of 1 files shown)


tburry:
Sanitizing as the final step is a mistake IMO. We should sanitize the user input, but then be allowed to add stuff that may not be allowed by htmLawed() such as styles.

linc:
@tburry So you want `processHTML()` to always run after sanitizing? I feel like our magic transformations are perhaps not that thoroughly robust and I worry someone will add "magic" that causes unforeseen interactions - basically letting someone game our magic into an XSS flaw.

DaazKu:
We currently have a formatting order problem: https://github.com/vanilla/vanilla/issues/5399

linc:
@DaazKu I don't think solving that is in the scope of this PR, but I imagine it's going to make the solution to that one easier.

tburry:
Here is my thinking.

- Additional formatters shouldn't have to worry about dirty HTML which means we clean the HTML first.
- Additional formatters shouldn't have to know about Markdown vs BBCode vs whatever so they should happen on HTML always.
- We aren't going to filter HTML twice "just to be sure". Formatters should be proper. Regular expressions can maybe be made stricter for now and miss on some edge cases rather than let bad HTML in.
- I think eventually we'll construct a DOMDocument around the filtered HTML, but  that needs to wait for caching.

linc:
> Formatters should be proper. Regular expressions can maybe be made stricter for now and miss on some edge cases rather than let bad HTML in.

The issue isn't that any particular formatter isn't "proper", it's that the chain as a whole cumulatively gets exploited because of an unexpected interaction from sequential changes.

PR #5425, opened 2017-04-08
Reorganize and rename dashboard settings menus

linc:
Renames:

* Banner => Branding: This page has expanded from just "banner" settings. "Branding" is clearer.
* Homepage => Layout: This page also covers more than its name implied. It's the layout for 3 different scopes. This rename also included reworking the view to semantically make sense (3 parallel options in subheadings; indents; labeling).
* Forum Settings => Discussions: The heart of a forum is discussions. These menu items are all 1-degree from Discussions. It also makes no sense to go to the "Settings" section of your forum and find a section called "Forum Settings".
* Advanced => Posting: Aside from 2 outliers, these settings are really just about posting. "Advanced" is a cop out; always has been.
* Embed => Embedding: Our menus use nouns, not verbs.
* Site Settings => Technical: Again... yes, it's a site. We're already under Settings. This meant nothing.

New:

* "Connections" section.

Moves:

* "Security" is spun off of Advanced to its own page on Technical: These may continue to expand in the future, and were in a confusing location.
* Embedding moves from old "Forum Settings" to Technical: This is pretty self-explanatory.
* Locales moves from Addons to Technical: No one thinks of languages as addons but developers.
* Social Connect moves to Connections as Social Media.

I added 3 redirects for the pages whose address changed: Advanced, Banner, and Homepage.

Sample from my localhost (multiple addons enabled):

screen shot 2017-04-08 at 5 34 45 pm
screen shot 2017-04-08 at 5 35 24 pm

I feel like this all makes a ton more sense.

Before merging I'd like to see:

* [ ] 2 reviews from R&D
* [x] 1 OK from support
* [x] 1 OK from marketing


linc:
I wonder if "Locales" is better suited under "Appearance".

linc:
@adrianspeyer @lauramakesart

adrianspeyer:
My only comment, it just seems weird that MailChimp is under technical, and not addons. Otherwise, looks great!

slafleche:
Shouldn't we also rename "Homepage". It's more of a Display Style page than a homepage style.

linc:
> My only comment, it just seems weird that MailChimp is under technical, and not addons.

Addons' placements are outside the scope of this PR. I agree there's a lot of improvements to be made here, including 'Ranks' being moved to 'Membership'.

linc:
> Shouldn't we also rename "Homepage". It's more of a Display Style page than a homepage style.

"Display Style" is way more esoteric, so that's worse. I'd accept "Layout"?

adrianspeyer:
"Layout" is a good idea!

LauraMakesArt:
This looks fine to me. I like "Layout" as well.

linc:
I've made the "Layout" change, which included modifying that page to make better sense.

linc:
I went a bit further and also made a "Connections" section.

How things stand now:

screen shot 2017-04-10 at 3 52 09 pm
screen shot 2017-04-10 at 3 52 29 pm

applications/dashboard/controllers/class.settingscontroller.php
```diff
@@ -606,12 +606,12 @@ public function bans($Action = '', $Search = '', $Page = '', $ID = '') {
...
 
     /**
-     * Homepage management screen.
+     * Layout management screen.
      *
      * @since 2.0.0
      * @access public
      */
-    public function homepage() {
+    public function layout() {
```
initvector: Need to update `setHighlightRoute` in here.

applications/dashboard/controllers/class.settingscontroller.php
```diff
@@ -715,6 +715,76 @@ public function configuration() {
...
+     * Backwards compatibility. Remove after 2.5.
+     */
+    public function homepage() {
+        redirect('/settings/layout');
+    }
+
+    /**
+     * Backwards compatibility. Remove after 2.5.
+     */
+    public function banner() {
```
initvector: How about slapping a `@deprecated` tag on these? It'll be easier to find things we want to remove in future versions.

applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -84,33 +76,31 @@ public function advanced() {
...
-        // Render default view (settings/advanced.php)
+        // Render default view (settings/posting.php)
         $this->render();
     }
 
+    /**
+     * Backwards compatibility. Remove after 2.5.
+     */
+    public function advanced() {
+        redirect('/vanillasettings/posting');
```
initvector: I think you wanna use `/vanilla/settings/posting` here.

initvector:
I implemented my own suggestions, so I'm going to slap my stamp of approval on this too. 👍

linc:
Thanks!

PR #5427, opened 2017-04-11
Add APIv2 tests to the default test suite and fix the build

tburry:


tburry:
The two issues look like they may be yours @DaazKu.

Issue #5379, opened 2017-03-28
Investigate failing Travis builds


tburry:
Our travis builds are failing a lot now and its becoming a big inconvenience. I'd like the build process to be tweaked to help us do some troubleshooting. Here are the things I want.

- Dump PHP error logs at the end of builds.
- Configure error logs to include stack traces.
- There are currently some notices coming out in the builds that can easily be fixed with an INI directive.
- Come up with some ideas on your own.

I'll consider this issue complete when we get some useful information about our failing builds.

DaazKu:
Seems like this is not fixed.. https://travis-ci.org/vanilla/vanilla/jobs/224308262

DaazKu:
For the specific remaining error: https://github.com/vanilla/vanilla/issues/5470

PR #5445, opened 2017-04-19
Improve unit testings

DaazKu:
- [x] Clarify some details about the setup.
- [x] Add stack trace to the error log.
- [x] Fix the "Vanilla did not install" problem.
  - I'm not 100% sure that I fixed the issue but It was probably related to 	89eed50
- [x] Fix an error about always_populate_raw_post_data being set on php > 5.6. See https://github.com/piwik/piwik/issues/6465 for an explanation

Addresses https://github.com/vanilla/vanilla/issues/5379
Better version of https://github.com/vanilla/vanilla/pull/5442


DaazKu:
This is ready to be merged. I'm not 100% sure that everything will be fixe by this but it will help a lot!

Issue #5405, opened 2017-04-04
/dba/counts sets DateLastComment to 0000-00-00 00:00:00 when discussion has no comment


topcheese13:
As of now /dba/counts is setting DateLastComment to 0000-00-00 00:00:00.

This happens in the ```DBAmodel::getCountSQL()``` function. The following line of code updates the discussion setting the DateInserted as DateLastComment where DateLastComment is null.

Widening this condition to where DateLastComment is null or DateLastComment = 0000-00-00 00:00:00 would fix this issue. Performance wise it might be quicker to use 2 separate queries instead of an "or".

PR #5456, opened 2017-04-20
Fixing case when /dba/counts sets discussion DateLastComment to 0

topcheese13:
/dba/counts sets DateLastComment to 0000-00-00 00:00:00 when current value of DateLastComment is NULL and discussion has no comment.

https://github.com/vanilla/vanilla/blob/master/applications/dashboard/models/class.dbamodel.php#L107

Query on this line becomes the code below, which returns 0 and the case mentioned above and inserts 0000-00-00 00:00:00 into a datetime field.

```
update GDN_Discussion p
                  set p.DateLastComment = (
                     select coalesce(max(c.DateInserted), 0)
                     from GDN_Comment c
                     where p.DiscussionID = c.DiscussionID)
```

Widening the where clause of the update function fixes that problem.

Fixes https://github.com/vanilla/vanilla/issues/5405

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -254,6 +254,7 @@ class DiscussionModel extends Gdn_Model {
                     ->update('Discussion')
                     ->set('DateLastComment', 'DateInserted', false, false)
                     ->where('DateLastComment', null)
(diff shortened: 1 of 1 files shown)


linc:
What's with all the tests failing?

PR #5461, opened 2017-04-21
Backport several fixes to OAuth2 class

R-J:
Backport recent changes to OAuth2 class: https://github.com/vanilla/vanilla/pull/5350


DaazKu:
You need to cherry-pick the commits from https://github.com/vanilla/vanilla/pull/5350/commits.
That way we end up with the same commit hashes which is better if we want to cherry pick other commits into the branch.

PR #5468, opened 2017-04-25
Fix APIv0 test suite container corruption

DaazKu:
APIv0 plays with the DI Container but does not clean itself!


linc:
It's not filling me with confidence that the 7.0 test failed.

DaazKu:
It was due to the infamous random error: 'Vanilla did not install'

PR #5481, opened 2017-04-28
Allow Form->Toggle to not require a label

linc:
Toggle() is used without a label in multiple places.


linc:
One use: /settings/jsconnect

Issue #5483, opened 2017-04-28
BBCode editor alignment broken


Charrondev:
Steps to reproduce:

* Enabled Advanced Editor plugin.
* Set editor type to BBCode
* Try to make a post with [left] [center] [right] tags.
* Observe

This seems to have been introduced with 5e31b076d.

DaazKu:
```bbcode
[right]Right[/right]
[center]Middle[/center]
[center]Left[/center]
```

gets transformed to

```html
<div class="bbcode_right" style="text-align:right">
Right
</div>

<div class="bbcode_center" style="text-align:center">
Middle
</div>

<div class="bbcode_center" style="text-align:center">
Left
</div>
```

and then `Gdn_Format::htmlFilter()` filters it to

```html
<div>
Right
</div>

<div>
Middle
</div>

<div>
Left
</div>
```

DaazKu:
Filtering last is causing problem for BBCode

linc:
Sounds like either BBCode needs to enable a filtering exception, or we need to just globally allow that formatting.

DaazKu:
"style" should be removed. The classes should probably be allowed.

PR #5484, opened 2017-04-28
Fix BBCode align tags not aligning text

DaazKu:
Allow bbcode_{ALIGN} classes and add them to our style.css

Fixes https://github.com/vanilla/vanilla/issues/5483

Backport to 2017-Q1-6 and 2017-Q2-2


linc:
Wouldn't it make better sense to use the existing CSS classes in the BBCode parsing than make these new exceptions?

DaazKu:
We would have to update https://github.com/vanilla/nbbc which is way more work in the short and long term.

PR #5489, opened 2017-05-01
Interpret plain text as JSON

tburry:
I’m hoping to get thoughts on this so let me explain. This PR is to help CORS requests avoid preflights. In order to avoid a preflight (among other things) the content type must be one of the following:

- application/x-www-form-urlencoded
- multipart/form-data
- text/plain

We know we’d have to transmit JSON as text/plain. The question is whether or not to force a querystring parameter for its content type. My thinking is we can treat text/plain as JSON and force an actual text/plain post to use a different content type or a querystring parameter. This would save us the hassle of seeing application/json in every single cross-domain querystring.

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -1142,7 +1142,7 @@ class Gdn_Request implements RequestInterface {
     private function decodePost($post, $server, $inputFile = 'php://input') {
         $contentType = !isset($server['CONTENT_TYPE']) ? 'application/x-www-form-urlencoded' : $server['CONTENT_TYPE'];
 
-        if (stripos($contentType, 'json') !== false) {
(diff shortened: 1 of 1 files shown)


kaecyra:
As long as json_decode is smart enough to fast-fail when the text is obviously not JSON, and without causing a memory use issue. Perhaps that could be tested?

tburry:
json_decode() won't error out. Just return null. My guess is speed isn't going to be the case. My thought also would be we'd add a qs parameter if/when we want to support plain text uploading.

1__psysh___private_var_www_vanilla__php__and_2__yarn___usr_local_var_www_github_hootsuite_addons_hootsuite-app__node_

PR #5492, opened 2017-05-02
Refactor bootstraping and constants declaration

DaazKu:
Create an envsetup file that can be included to use most of vanilla framework right off the bat.

This is useful for unit testing or using Vanilla's classes in a CLI environment.


tburry:
I think this is a good first step to eventually putting the bootstrap into a class. Can you just rename the file "environment.php" though?

environment.php
```diff
@@ -0,0 +1,68 @@
...
+// Do NOT edit this. See config `Garden.GuestTimeZone`.
+date_default_timezone_set('UTC');
+
+// Make sure the mb_* functions are utf8.
+if (function_exists('mb_internal_encoding')) {
+    mb_internal_encoding('UTF-8');
+}
+
+// Include the core autoloader.
+if (@!include_once(__DIR__.'/vendor/autoload.php')) {
```
tburry: No error silence. I recommend you don't include changes like this that aren't germane to the PR topic.

DaazKu: I silenced the error because we handle it.

I'll remove the silence event if it's just screen pollution in this case.

DaazKu: That error is pretty useful when we include this file in an external CLI program.
You are right, this should be in a separate PR, but the change is pretty small so I decided to include it while doing the refactoring.

environment.php
```diff
@@ -0,0 +1,68 @@
...
+    define('APPLICATION', 'Vanilla');
+}
+if (!defined('APPLICATION_VERSION')) {
+    define('APPLICATION_VERSION', '2.4.200');
+}
+if (!defined('DS')) {
+    define('DS', DIRECTORY_SEPARATOR);
+}
+if (!defined('PATH_ROOT')) {
+    define('PATH_ROOT', __DIR__);
```
initvector: Vanilla currently establishes `PATH_ROOT` by using `getcwd`. Switching this up with `__DIR__` is going to change the way it works with symlinks.

For example, say I have a dir with a Git repo: /srv/src. I symlink certain directories and files from there into a separate directory to act as a site root: /srv/html. If I'm setting `PATH_ROOT` with `__DIR__`, it's going to get set to /srv/src. If I'm setting `PATH_ROOT` with `getcwd`, it gets set to /srv/html.

Switching up to `__DIR__` might be an improvement, but there is a subtle difference.

DaazKu: From a php perspective this doesn't change anything.

For your webserver, you should use [realpath_root instead of document_root if you use symlinks](http://stackoverflow.com/questions/23737627/php-opcache-reset-symlink-style-deployment/23904770#23904770).

initvector: I'm not opposed to switching up `getcwd` for `__DIR__`, but it does break setups/configs that currently work in Vanilla. It might be worth putting a blurb in the release notes for whatever version this gets rolled into.

In tests with PHP 5.6 (I didn't test 7+), even when using $realpath_root in the server config, `getcwd` and `__DIR__` can potentially give different values. If I symlink an index.php from a repo into my web root, `__DIR__` will be the real file dir. `getcwd` will give the web root.

evach-vf: @initvector Did you switch both keys in your nginx config?

    fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
    fastcgi_param DOCUMENT_ROOT $realpath_root;

DaazKu: I'm all in for a special section just for that in the release notes!

tburry:
I'm placing a hold tag on this until @evach-vf and @initvector gets the symlink/realpath stuff sorted out on their machines.

In general I think using realpath in nginx is a positive change and __DIR__ is much better to use if possible. This PR incidentally uncovered some good information that will improve our deployments bigly. I just want to make sure all of the infrastructure and dev server stuff is deployed before this PR is merged.

@DaazKu you may want to rebase this PR so that it can benefit from the travis fixes in master and also doesn't get too stale.

DaazKu:
Any Idea of when we can merge this?

DaazKu:
Todd, Ryan and Eric gave the green light.

DaazKu:
Does that do: https://github.com/vanilla/docs/blob/master/content/developer/backend/server-nginx.md#configuration ?

PR #5493, opened 2017-05-02
Fix unit tests for new add-ons

tburry:
New add-ons with an addon.json cannot be found using deprecated methods so they shouldn’t be in deprecated method unit tests.

This is a cherry pick from the APIv2 branch.


tburry:
Closing as duplicate of #5545.

PR #5495, opened 2017-05-03
Improve unit tests compability and usage

DaazKu:
- Make sure that the tests are compatible with both phpunit 5.7 and 6+
- Fix some unit tests that did not have any assertions
- Fix jsonEncodeChecked. (I spoke with Ryan and the fix seems OK)
- Add the possibility to set environment variables to specify database paramters


DaazKu:
About 3db32a2 ... I made a PR https://github.com/php-fig/log/pull/52
We will be able to remove the shim if they merge it.

library/core/functions.general.php
```diff
@@ -2109,35 +2109,29 @@ function jsonEncodeChecked($value, $options = null) {
         $advanced = (PHP_VERSION_ID >= 50500);
 
         if ($options === null) {
-            if ($advanced) {
-                $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR;
-            } else {
-                $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES;
-            }
+            $options = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES;
```
DaazKu: Having partial response is not necessarily good. Having null values instead of the correct data can cause potential problems down the line.
It is also harder to debug!

Having `JSON_PARTIAL_OUTPUT_ON_ERROR` also prevent some errors to be detected because they will never happen. ie JSON_ERROR_UNSUPPORTED_TYPE

DaazKu:
- I don't think we should be supporting multiple versions of PHPUnit at once.
  - I see no harm in making our tests compatible with phpunit ~6. That does not mean that we fully support those version.
- I don't think our JSON should fail altogether. At least split that code off into a separate PR and make your case in writing on the PR.
   - See my reply. I also updated the function.

PR #5496, opened 2017-05-03
Update AddonManager's caching to be optionnal

DaazKu:
Make the AddonManager cache optional.

This is great if you want to use the AddonManager in a context where the cache is useless.


DaazKu:
I think this is a really good idea.
I would rather add a "cacheEnabled" method though.

Made an issue about it https://github.com/vanilla/vanilla/issues/5544

PR #5505, opened 2017-05-04
Let magic formatter go last

linc:
Part of #5423 was putting the security filter after the magic parser. Our security filter doesn't have sufficient exceptions to allow our embeds. While ideally we would dig deep into the guts of Htmlawed and see if we can whitelist certain domains for embed very granularly, for now taking a step back is our best option.


linc:
`release/2017-Q2-2`

initvector:
Backported to [release/2017-Q2-2](https://github.com/vanilla/vanilla/tree/release/2017-Q2-2)

PR #5514, opened 2017-05-05
Tweak unit tests

tburry:
- Add a bootstrap class to allow container re-initialization.
- Add a SiteTestTrait that allows test classes to install a copy of Vanilla and run internally.
- Move the AccessTokenModelTest out of the APIv2 tests.
- Don't save configurations during unit tests. This is more to work around an issue where the configuration object can only save on destruct which isn't appropriate for unit tests and causes some timing issues too.

Basically, these changes will allow us to move closer to isolating tests against separate installs of Vanilla where a test class can install its own copy of Vanilla without using cURL. This PR started out as a simple cherry pick from the apiv2 branch, but ended up including a lot of yak shaving in order to make builds work on Travis.

*Note to reviewers: The changes to non-testing classes are the ones that need reviewing. I added some line comments to explain what I did and why.*

### DIFF
--- a/library/Vanilla/Models/AddonModel.php
+++ b/library/Vanilla/Models/AddonModel.php
@@ -190,7 +190,9 @@ class AddonModel implements LoggerAwareInterface {
         // Look for a setup method.
         $called = $this->callPluginMethod($addon, 'setup');
 
-        if (($structure = $addon->getSpecial('structure')) && (!$called || !in_array($structure, get_included_files()))) {
+        // @TODO This if is a kludge because Vanilla's core applications are inconsistent.
+        // Once the InstallModel is in use this code can be cleaned up by manual structure inclusion in addons.
+        if (($structure = $addon->getSpecial('structure')) && (!$called || !in_array($addon->path($structure, Addon::PATH_FULL), get_included_files())) || $addon->getKey() === 'dashboard') {
             $this->logger->info(
                 "Executing structure for {addonKey}.",
                 ['event' => 'addon_structure', 'addonKey' => $addon->getKey(), 'structureType' => 'file']
--- a/library/core/class.gdn.php
+++ b/library/core/class.gdn.php
@@ -619,7 +619,18 @@ class Gdn {
      *
      * @param \Garden\Container\Container $container
      */
-    public static function setContainer(\Garden\Container\Container $container) {
+    public static function setContainer(\Garden\Container\Container $container = null) {
         self::$container = $container;
+
+        /**
+         * Reset all of the cached objects that are fetched from the container.
+         */
+        self::$_Config = null;
+        self::$_Factory = null;
+        self::$_FactoryOverwrite = true;
+        self::$_Locale = null;
+        self::$_Request = null;
+        self::$_PluginManager = null;
+        self::$_Session = null;
     }
 }
--- a/library/core/class.pluginmanager.php
+++ b/library/core/class.pluginmanager.php
@@ -214,7 +214,7 @@ class Gdn_PluginManager extends Gdn_Pluggable implements ContainerInterface {
 
             $SearchPluginInfo = $this->scanPluginFile($PluginFile);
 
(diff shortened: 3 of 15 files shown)


library/Vanilla/Models/AddonModel.php
```diff
@@ -190,7 +190,9 @@ private function runSetup(Addon $addon) {
         // Look for a setup method.
         $called = $this->callPluginMethod($addon, 'setup');
 
-        if (($structure = $addon->getSpecial('structure')) && (!$called || !in_array($structure, get_included_files()))) {
+        // @TODO This if is a kludge because Vanilla's core applications are inconsistent.
+        // Once the InstallModel is in use this code can be cleaned up by manual structure inclusion in addons.
+        if (($structure = $addon->getSpecial('structure')) && (!$called || !in_array($addon->path($structure, Addon::PATH_FULL), get_included_files())) || $addon->getKey() === 'dashboard') {
```
tburry: Note the comment I added. I realize this is a kludge right now, but this class is currently only being used in unit tests. As soon as its the one used to install Vanilla we can fix the offending inconsistent classes.

library/core/class.gdn.php
```diff
@@ -619,7 +619,18 @@ public static function getContainer() {
...
+        /**
+         * Reset all of the cached objects that are fetched from the container.
+         */
+        self::$_Config = null;
+        self::$_Factory = null;
+        self::$_FactoryOverwrite = true;
+        self::$_Locale = null;
+        self::$_Request = null;
+        self::$_PluginManager = null;
+        self::$_Session = null;
```
tburry: This code was added so that I can clear out the container between test classes.

library/core/class.pluginmanager.php
```diff
@@ -931,7 +931,9 @@ public function scanPluginFile($PluginFile, $VariableName = null) {
 
         }
         unset($Lines);
-        if ($PluginInfoString != '') {
+        if (empty($PluginInfoString)) {
+            return null;
+        } else {
```
tburry: These changes prevent an issue where plugins using the correct addon.json would fail the forwards compatibility test.

DaazKu:
I rebased the changes

tests/TestInstallModel.php
```diff
@@ -186,5 +177,8 @@ public function uninstall() {
         // Reset the config to defaults.
         $this->config->Data = [];
         $this->config->load(PATH_ROOT.'/conf/config-defaults.php');
+
+        // Clear all database related objects from the container.
```
DaazKu: Something missing here?

tests/Bootstrap.php
```diff
@@ -0,0 +1,243 @@
...
+    private $baseUrl;
+
+    /**
+     * Bootstrap constructor.
+     *
+     * A different base URL affects
+     *
+     * @param string $baseUrl The base URL of the installation.
+     */
+    public function __construct($baseUrl = 'http://vanilla.test') {
```
DaazKu: Shouldn't this be `http://vanilla.test:8080` or better `getenv('baseurl')` ?

tests/SiteTestTrait.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+trait SiteTestTrait {
+    /**
+     * @var Container
+     */
+    private static $container;
+
+    protected static function createContainer() {
+        $folder = strtolower(EventManager::classBasename(get_called_class()));
+        $bootstrap = new Bootstrap("http://vanilla.test/$folder");
```
DaazKu: Same as above: `getenv('baseurl')` ?

tests/Bootstrap.php
tburry: That's something that can be added later. Since these internal tests don't actually run against a site though it only affects  URLs.

tburry:
@DaazKu since this trait and these tests are not intended to actually run through a web server then keeping them at a specific URL is more desirable at this time.

PR #5519, opened 2017-05-05
APIv2

tburry:
This is the mega-branch for APIv2 endpoints. It’s long-running and will have master merged into it from time-to-time so it will end up having to be reviewed via testing rather than line-by-line code review.

Don’t worry, there are tests too.


linc:
This is the sort of thing where we want to keep a list of changes being made in the main PR over time. As it stands, this is completely impossible to review.

PR #5523, opened 2017-05-07
Repair logic in the unbindClass() method

linc:
I'm not super familiar with this area, but going by what _appears_ to be the intent of these checks and the notices I'm receiving when scanning the database for updates, I believe this fixes the method to check variable types correctly.


tburry:
Thanks!

Issue #5106, opened 2017-02-02
Application/Plugin unification and enhancement


tburry:
The AddonManager unified applications and plugins into Addons although left backwards compatibility in place with the various objects and pages. There are two main goals with this enhancement:

- Unify plugins and applications now so that improvements to management will only have to be done once.
- Add improvements to the addon management page to help reduce support.

This is a pretty significant epic. I broke up the issues quite well, but I suspect that merging all PRs into a separate branch might be the best move here.

PR #5541, opened 2017-05-08
Refactor the ApplicationManager to use the AddonManager to get applications info

DaazKu:
That way applications can work with both about.php and addon.json.

### DIFF
--- a/library/Vanilla/Addon.php
+++ b/library/Vanilla/Addon.php
@@ -144,6 +144,12 @@ class Addon {
             if (empty($info)) {
                 throw new \Exception("The addon at $subdir has an empty info array.");
             }
+
+            $baseDir = explode('/', ltrim($subdir, '/'))[0];
+            if (in_array($baseDir, ['applications', 'plugins'])) {
+                $info['oldType'] = substr($baseDir, 0, -1);
+            }
+
             return $info;
         }
 
@@ -265,6 +271,8 @@ class Addon {
             return null;
         }
 
+        $oldType = null;
+
         // See which info array is defined.
         if (!empty($PluginInfo) && is_array($PluginInfo)) {
             $array = $PluginInfo;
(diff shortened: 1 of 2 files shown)


initvector:
On hold pending a fix for the build

library/Vanilla/Addon.php
```diff
@@ -144,6 +144,12 @@ private function scanInfo() {
             if (empty($info)) {
                 throw new \Exception("The addon at $subdir has an empty info array.");
             }
+
+            $baseDir = explode('/', ltrim($subdir, '/'))[0];
+            if (in_array($baseDir, ['applications', 'plugins'])) {
```
tburry: I don't want the new Addon class to have any knowledge of the directory it is in.

DaazKu:
I created an [issue about removing the kludge](https://github.com/vanilla/vanilla/issues/5552) as part of [Application/Plugin unification and enhancement](https://github.com/vanilla/vanilla/issues/5106)

library/core/class.applicationmanager.php
```diff
@@ -48,35 +48,26 @@ public function __construct(AddonManager $addonManager = null) {
...
-                        $Folder = substr($Folder, 0, strpos($Folder, DS));
-                        $ApplicationInfo[$ApplicationName]['Folder'] = $Folder;
-                    }
+            $applications = [];
+
+            $addons = $this->addonManager->lookupAllByType(Addon::TYPE_ADDON);
+            foreach ($addons as $addon) {
+                /* @var Addon $addon */
+                if ($addon->getInfoValue('oldType') !== 'application') {
+                    continue;
```
tburry: Right away you have some issues here. You need to make Gdn_ApplicationManager backwards compatible or else some plugin somewhere is going to break. Look at how Gdn_PluginManager does it:

```php
public function availablePlugins() {
    $addons = $this->addonManager->lookupAllByType(Addon::TYPE_ADDON);

    $result = [];
    foreach ($addons as $addon) {
        /* @var Addon $addon */
        if ($addon->getInfoValue('oldType') !== 'plugin') {
            continue;
        }

        $infoArray  = static::calcOldInfoArray($addon);
        $result[$infoArray['Index']] = $infoArray;
    }
    return $result;
}
```

You can do something almost exactly the same and use Gdn_PluginManager::calcOldInfoArray() too.

DaazKu: I already do it...
```
$info['Folder'] = $directories[count($directories) - 1];
$info['Index'] = $info['name'];
```

DaazKu:
The application manager is now 100% backward compatible.

PR #5545, opened 2017-05-08
Fix plugin scanning logic

DaazKu:
Make sure that nothing is returned if $PluginInfo is not defined in the plugin's class.
Test $PluginInfo properly in indexSearchPath


tburry:
Seems like a duplicate of #5493.

PR #5566, opened 2017-05-10
Improve load time of Vanilla Statistics dashboard summaries

DaazKu:
On large sites the generation the dashboardsummaries should go from 15-30secs to under a second.
Fixes https://github.com/vanilla/vanillastatsapp/issues/8

`/utility/update` needs to be ran for the indexes to be created!

I was told to make the indexes approved by @tburry

### DIFF
--- a/plugins/VanillaStats/class.vanillastats.plugin.php
+++ b/plugins/VanillaStats/class.vanillastats.plugin.php
@@ -36,6 +36,9 @@ class VanillaStatsPlugin extends Gdn_Plugin {
     /**  */
     const RESOLUTION_MONTH = 'month';
 
+    /** Upper limit on date ranges for user record querying. */
+    const USER_MAX_DAYS = 90;
+
     /** @var mixed  */
     public $AnalyticsServer;
 
@@ -54,10 +57,14 @@ class VanillaStatsPlugin extends Gdn_Plugin {
 
         $isVanillaAnalyticEnabled = Gdn::addonManager()->isEnabled('vanillaanalytics', Vanilla\Addon::TYPE_ADDON);
         $this->dashboardSummariesEnabled = c('Garden.Analytics.DashboardSummaries', !$isVanillaAnalyticEnabled);
+
+        parent::__construct();
     }
 
     /**
      * Override the default dashboard page with the new stats one.
+     *
+     * @param Gdn_Dispatcher $Sender
      */
     public function gdn_dispatcher_beforeDispatch_handler($Sender) {
         $Enabled = c('Garden.Analytics.Enabled', true);
@@ -157,6 +164,8 @@ class VanillaStatsPlugin extends Gdn_Plugin {
     /**
      * A view containing most active discussions & users during a specific time
      * period. This gets ajaxed into the dashboard homepage as date ranges are defined.
+     *
+     * @param SettingsController $Sender
      */
     public function settingsController_dashboardSummaries_create($Sender) {
         $DiscussionData = [];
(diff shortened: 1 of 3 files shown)


DaazKu:
Maybe we should keep active users for 30 days max no matter how big the DB is.
That way we won't get support request asking why the reports are different from sites to sites.

DaazKu:
I'm trying an alternative route from what @tburry suggested.

I would really like to not lose any functionality to prevent support request or clients frustration.
The dashboard summary is supposed to be shown for the selected date range. It will be confusing if we change this behaviour.

I got rid of the indexes and modified the query so that it run faster.
Before (It could take ~47 seconds for a client with 1.2M comments)
```sql
select 
	InsertUserID as `UserID`,
	count(CommentID) as `CountComments`
	from GDN_Comment `Comment`
where DateInserted >= '2016-01-10 03:30:00'
	and DateInserted <= '2017-01-10 03:30:00'
group by InsertUserID
order by CountComments desc
limit 5
```
After (takes ~2.6 seconds on the same client data)
```sql
select 
	InsertUserID as `UserID`,
	count(CommentID) as `CountComments`
from GDN_Comment `Comment`
where CommentID in (
	select CommentID
	from GDN_Comment `Comment`
	where DateInserted >= '2016-01-10 03:30:00'
		and DateInserted <= '2017-01-10 03:30:00'
)
group by InsertUserID
order by CountComments desc
limit 5
```

This will fix the issue for 99% of our customers.
I also made a test on a forum with 80M comments and it took 13 seconds which is acceptable I guess.

plugins/VanillaStats/js/vanillastats.js
```diff
@@ -753,6 +753,10 @@ var vanillaStats = (function() {
     };
 
     VanillaStats.prototype.getSummaries = function(container) {
+        if (gdn.definition('DashboardSummaries')) {
```
tburry: Is this correct? It seems like it should be the opposite. Also, you want gdn.getMeta().

DaazKu:
@evach-vf Did some test of my new query on another database and it ran slower than the old one. I did more test and found out why...

The force index thing is not sexy but it works perfectly.

The problem with my test data is shown with this explain:
```
id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
1	SIMPLE	Comment	index	IX_Comment_DateInserted,FK_Comment_InsertUserID	FK_Comment_InsertUserID	5	NULL	1154702	Using where; Using temporary; Using filesort
```
`FK_Comment_InsertUserID` was used instead of `IX_Comment_DateInserted`.
It probably has to do with the amount of users you have to group in the query but that query takes 47 seconds.

Using `force index (IX_Comment_DateInserted)`
```
id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
1	SIMPLE	Comment	range	IX_Comment_DateInserted,FK_Comment_InsertUserID	IX_Comment_DateInserted	6	NULL	577351	Using index condition; Using temporary; Using filesort
```

So the problem was the optimizer being dumb.
I tested the force index solution on multiple database with different amount of records and it works perfectly.

I'm also forcing a clamp of 3 months on the range for the userData if there are more than 10M records in the database (with a clear feedback).

DaazKu:
@tburry do not want the force index kludge to be used even if it is the problem so we are gonna clamp the date range to 90 days.

As previously mentioned, the wrong index is used depending of how many many different users are in InsertUserID compared to the total amount of data queried.

On a smaller database we have the same problem.

400k records span over 4 years, 
screen shot 2017-05-24 at 12 55 19 pm
The wrong index is used. The query run in 1.3 seconds. 
If you force the index you end up with ~200ms.

So I think we are just gonna clamp the user summary to 3 months no matter what.

DaazKu:
If the query is still long on certain sites you can update the `IX_Comment_DateInserted` to be `IX_Comment_DateInserted_InsertUserID` which the engine will properly use.
See 
https://github.com/vanilla/vanilla/pull/5566/commits/271a2a49c95b952b4ee23c4a62a6f55b52ca5eba#diff-5701e9e7d4797969d88e489aa058dd83R73

plugins/VanillaStats/class.vanillastats.plugin.php
```diff
@@ -151,6 +154,55 @@ public function settingsController_home_create($sender) {
      * period. This gets ajaxed into the dashboard homepage as date ranges are defined.
      */
     public function settingsController_dashboardSummaries_create($Sender) {
+        $DiscussionData = [];
+        $UserData = [];
+        $isVanillaAnalyticEnabled = Gdn::addonManager()->isEnabled('vanillaanalytics', Vanilla\Addon::TYPE_ADDON);
+
+        if (c('Garden.Analytics.DashboardSummaries', !$isVanillaAnalyticEnabled)) {
```
initvector: This config is checked and its default prepared in almost the same way in two different places. How about moving this logic into the plugin constructor and saving the value in a new property on the class, similar to how the `AnalyticsServer` and `VanillaID` properties are handled?

```php
    ...
    $vanillaAnalyticEnabled = Gdn::addonManager()->isEnabled('vanillaanalytics', Vanilla\Addon::TYPE_ADDON);
    $this->doDashboardSummaries = c('Garden.Analytics.DashboardSummaries', !$isVanillaAnalyticEnabled);
}
```

initvector:
According to @DaazKu, @tburry is onboard with the current solution. Things look good to me. Since the original request for changes is in reference to a previous approach, I'm going to dismiss it.

PR #5570, opened 2017-05-11
Do the UserSpamming check only if the discussion/comment is an insert

DaazKu:
I updated the function name from "is*" to "check*" since it does not only return a state.
I updated the docblock to reflect that change.
Only do the check if the discussion/comment is new


linc:
release/2017-Q1-6

Issue #5212, opened 2017-02-22
NestedCollectionAdapter does not work well with non ascii locales


DaazKu:
[NestedCollectionAdapter::addLink()](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/library/class.nestedcollectionadapter.php#L35) use [slugify()](https://github.com/vanilla/vanilla/blob/master/library/core/functions.general.php#L3849) which strips any non ascii characters. 

Something like `Реакции` (Reactions) will get replaced by `n-a`.
When in the context of the dashboard site nav n-a is not a valid group and the link is then never added to the nav.

nervoustwit:
This is also affecting the dropdown menu on the user profile page where mods now cannot mark a user as a troll or add notes to user in Japanese or Russian because it is passing an empty string to slugify.

PR #5574, opened 2017-05-11
Fix non-ascii items not being added to menus

DaazKu:
`slugify()` does not work with non ascii characters.
Also... keys are used as index in php arrays there is no need to `slugify()` them.

Fixes https://github.com/vanilla/vanilla/issues/5212

### DIFF
--- a/applications/dashboard/library/class.nestedcollectionadapter.php
+++ b/applications/dashboard/library/class.nestedcollectionadapter.php
@@ -36,7 +36,7 @@ class NestedCollectionAdapter {
         if ($permission === false) {
             $permission = true;
         }
-        $key = slugify($group) . '.' . slugify($text);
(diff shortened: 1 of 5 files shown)


DaazKu:
It seems that we are going to have to refactor the nav code a bit more...

> [3:14 PM] Alexandre Chouinard: We create groups with keys like this:
[3:14 PM] Alexandre Chouinard: 'site-settings'
[3:14 PM] Alexandre Chouinard: but we never reference them as such
[3:14 PM] Alexandre Chouinard: we use "Site Settings" instead
[3:14 PM] Alexandre Chouinard: and in reality "Site Settings" is named "Technical"

When adding an item to a group we should use the group key...

DaazKu:
I decided to restore preserve the "spaces to -" conversion.

PR #5576, opened 2017-05-11
Remove HOST header support from Gdn_Request

initvector:
Support for the HOST header should only be utilized in advanced network setups that can properly vet the value. This update removes automatic handling of that header.


kaecyra:
This would break hosted infrastructure. I need to review this more deeply as I'm not sure its correct.

PR #5577, opened 2017-05-12
Add Apple touch icon support

### LINKED ISSUE #2560: Touch Icon addon => core feature
Should be a core feature on Banner page, not an addon.

slafleche:
Moving Touch Icon plugin to core, new branding option for mobile address bar colour

Closes #2560

### DIFF
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -631,8 +631,12 @@ class UtilityController extends DashboardController {
      * @access public
      */
     public function showTouchIcon() {
-        $redirect = Gdn_Upload::url(c('Garden.TouchIcon'));
-        redirect($redirect, 302);
-        exit();
+        $icon = c('Garden.TouchIcon');
+
+        if (!empty($icon)) {
+            redirect(Gdn_Upload::url($icon), 302);
+        } else {
+            throw new Exception('Touch icon not found.', 404);
+        }
     }
 }
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -623,4 +623,17 @@ class UtilityController extends DashboardController {
         $this->deliveryMethod(DELIVERY_METHOD_JSON);
         $this->render();
     }
+
+
+    /**
+     * Redirect to touch icon.
+     *
+     * @since 1.0
+     * @access public
+     */
+    public function UtilityController_ShowTouchIcon($Sender) {
+        $redirect = $this->getIconUrl();
+        redirect($redirect, 302);
+        exit();
+    }
 }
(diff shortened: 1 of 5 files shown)


applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -623,4 +623,17 @@ public function maintenance($updateMode = 0) {
...
     }
+
+
+    /**
+     * Redirect to touch icon.
+     *
+     * @since 1.0
+     * @access public
+     */
+    public function UtilityController_ShowTouchIcon($Sender) {
```
linc: This is wrong. It should be named `showTouchIcon`.

applications/dashboard/modules/class.headmodule.php
```diff
@@ -270,14 +276,56 @@ public function setFavIcon($HRef) {
...
                     null,
                     'favicon'
                 );
             }
         }
 
         /**
+         * Sets the touch icon
+         *
+         * @param string The location of the fav icon relative to the web root. ie. /themes/default/images/layout.css
```
initvector: Missing the parameter name here

```
* @param [type] $var [...]
```

applications/dashboard/modules/class.headmodule.php
```diff
@@ -270,14 +276,56 @@ public function setFavIcon($HRef) {
...
+         */
+        public function setTouchIcon($HRef) {
```
initvector: `$href`

applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -623,4 +623,16 @@ public function maintenance($updateMode = 0) {
...
+    /**
+     * Redirect to touch icon.
+     *
+     * @since 1.0
+     * @access public
+     */
+    public function showTouchIcon() {
+        $redirect = Gdn_Upload::url(c('Garden.TouchIcon'));
+        redirect($redirect, 302);
+        exit();
```
initvector: `redirect` already ends with an `exit`.

applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -623,4 +623,16 @@ public function maintenance($updateMode = 0) {
...
+
+    /**
+     * Redirect to touch icon.
+     *
+     * @since 1.0
+     * @access public
+     */
+    public function showTouchIcon() {
+        $redirect = Gdn_Upload::url(c('Garden.TouchIcon'));
+        redirect($redirect, 302);
```
initvector: It's possible the value in `Garden.TouchIcon` is empty. This should probably return a 404 if there's no valid touch icon URL to redirect to.

applications/dashboard/controllers/class.utilitycontroller.php
slafleche: Do I need to add the exit if the else (when we throw an exception?)

initvector: No. When you throw an exception, it stops executing that function. No code after it will be executed.

PR #5592, opened 2017-05-16
Give more control over the category tree in the categories module

### LINKED ISSUE #5585: CategoriesModule should force display of Subcategories if endDepth is specified.
When declaring a Categories module with startDepth, and endDepth parameters, it should respect them regardless of the category type. Currently, this only works if the category is displayed as a discussions category.

This enhancement would allow admins to use a category structure with more "Nested" types, which is more performant especially on sites with very large amounts of categories.

## **For a given category structure:**

```
Category 1 - display as Discussions
|--->Subcategory 1.1
|--->Subcategory 1.2
Category 2 - display as Nested
|--->Subcategory 2.1
|--->Subcategory 2.2
|--->Subcategory 2.3
```

and the Categories Module:

```
{module name="CategoriesModule" startDepth=1 endDepth=2}
```

Currently this will output a structure like:

```
Category 1
|--->Subcategory 1.1
|--->Subcategory 1.2
Category 2
```

The explicit declaration of these parameters on the module should override a categories "display as" preference and potentially other settings declared such as "categoryDepth" (a global preference, not sure about this one).

The Expected output would look like

```
Category 1
|--->Subcategory 1.1
|--->Subcategory 1.2
Category 2
|--->Subcategory 2.1
|--->Subcategory 2.2
|--->Subcategory 2.3
```

tburry:
- Allow the categories module to provide a collapseCategories property
that specifies whether category categories are displayed.
- All the categories module to provide a root category.
- Use the CategoryCollection to filter category categories rather than
the redundant filterChildren() method.

Fixes #5585.

### DIFF
--- a/applications/vanilla/library/class.categorycollection.php
+++ b/applications/vanilla/library/class.categorycollection.php
@@ -397,8 +397,8 @@ class CategoryCollection {
      * @param int $parentID The ID of the parent category.
      * @param array $options An array of options to affect the fetching.
      *
-     * - maxdepth: The maximum depth of the tree.
-     * - collapsed: Stop when looking at a category of a certain type.
+     * - maxDepth: The maximum depth of the tree.
+     * - collapseCategories: Stop when looking at a categories that contain categories.
      * - permission: The permission to use when looking at the tree.
      * @return array
      */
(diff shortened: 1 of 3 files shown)


tburry:
This is to help support @Charrondev in a theme he's working on.

PR #5594, opened 2017-05-17
Fix find-ban URL on profile page when user has been banned by a rule

slafleche:
When subcommunities are turned on, the link to the banned reasons had the subcommunity uri twice in the address.

### DIFF
--- a/applications/dashboard/modules/class.userbanmodule.php
+++ b/applications/dashboard/modules/class.userbanmodule.php
@@ -51,7 +51,7 @@ class UserBanModule extends GDN_Module {
             // Add a link to identify the corresponding ban rule.
             if ($bit === 2) {
                 $text = sprintf(t('Find the corresponding ban rule for %s'), val('Name', $user));
-                $reasons[$bit] .= ' '.anchor($text, url('settings/bans/find/'.$userID));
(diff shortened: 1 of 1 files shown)


applications/dashboard/modules/class.userbanmodule.php
```diff
@@ -51,7 +51,7 @@ protected function getData() {
             // Add a link to identify the corresponding ban rule.
             if ($bit === 2) {
                 $text = sprintf(t('Find the corresponding ban rule for %s'), val('Name', $user));
-                $reasons[$bit] .= ' '.anchor($text, url('settings/bans/find/'.$userID));
+                $reasons[$bit] .= ' '.anchor($text, url('settings/bans/find/'.$userID, true));
```
initvector: The problem here is that `anchor` doesn't need to have its `$Destination` parameter run through `url`. It'll take care of that, itself. Having the parameter go through URL translation twice means the subcommunity prefix is being applied twice.

A better fix here would be to remove the call to `url` from the `$Destination` parameter. Forcing `$Destination` to be a full URL might technically work here, but it only works because being a full URL stops `anchor` from applying its own URL processing.

```php
$reasons[$bit] .= ' '.anchor($text, 'settings/bans/find/'.$userID);
```

PR #5603, opened 2017-05-18
Fix nomobile logic

DaazKu:
If you did:
- Go to full site
- Go back to mobile

you ended up with the mobile site appearing over the full site because `redirect()` was used instead of `$this->RedirectUrl`


nervoustwit:
Looks good.

PR #5606, opened 2017-05-18
Refactor unit tests

DaazKu:
- Update environment variables to respect naming convention
- Use PHPUnit namespace
- Fix some unit tests that did not have any assertions
- Update jsonEncodeChecked so that it fails properly when needed

Better version of https://github.com/vanilla/vanilla/pull/5495

### DIFF
--- a/tests/TestInstallModel.php
+++ b/tests/TestInstallModel.php
@@ -111,11 +111,7 @@ class TestInstallModel extends InstallModel {
      */
     public function getDbHost() {
         if (empty($this->dbHost)) {
-            if ($dbHost = getenv('TEST_DB_HOST')) {
-                $this->dbHost = $dbHost;
-            } else {
-                $this->dbHost = 'localhost';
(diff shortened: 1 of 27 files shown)


tests/TestInstallModel.php
```diff
@@ -110,8 +110,14 @@ private function getDbInfo() {
...
      */
     public function getDbHost() {
-        $host = isset($_ENV['dbhost']) ? $_ENV['dbhost'] : 'localhost';
-        return $host;
+        if (empty($this->dbHost)) {
+            if ($dbHost = getenv('TEST_DB_HOST')) {
+                $this->dbHost = $dbHost;
+            } else {
+                $this->dbHost = 'localhost';
+            }
```
initvector: Couldn't you condense this down into something like this?

```php
$this->dbHost = getenv('TEST_DB_HOST') ?: 'localhost';
```

PR #5608, opened 2017-05-19
Fix registration invitations expire time 

topcheese13:
Fixes problem where in dashboard/settings/registration upon changing the time when the invitation will expire and saving we get the "Your changes have been saved" message but the page still shows the previous value in the "Invitations will expire" select box until we refresh the page one more time manually.

We were fetching that value from the config before it was updated. Setting $this->InviteExpiration to the postback value will have the view build correctly after the form submission.


linc:
What problem? Is there a bug filed?

We're sure we wanna trust whatever value someone sticks in a form?

topcheese13:
Updated the PR description to be much more descriptive. Any reason why setting that variable to the postback value could be a bad practice ?

linc:
> Any reason why setting that variable to the postback value could be a bad practice ?

I can submit _anything_ as a postback. What if I decide the invitation expires in 2050? Can I send an invite with an expiration of yesterday to block someone from accepting a valid invite?

By setting this prior to the save block, you're effecting what goes to the database, not just what's being set on the page.

PR #5609, opened 2017-05-19
Force SSL when Garden.AllowSSL or Garden.ForceSSL are true

slafleche:
Updated canonicalUrl(), parse(), and url() to force SSL if Garden.AllowSSL or Garden.ForceSSL are true.

### DIFF
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -1415,7 +1415,7 @@ class Gdn_Request implements RequestInterface {
     public function url($path = '', $withDomain = false, $ssl = null) {
         static $allowSSL = null;
         if ($allowSSL === null) {
-            $allowSSL = c('Garden.AllowSSL', c('Garden.ForceSSL'));
+            $allowSSL = c('Garden.AllowSSL', true) && c('Garden.ForceSSL');
         }
 
         static $rewrite = null;
--- a/library/core/class.request.php
+++ b/library/core/class.request.php
@@ -1415,8 +1415,9 @@ class Gdn_Request implements RequestInterface {
     public function url($path = '', $withDomain = false, $ssl = null) {
         static $allowSSL = null;
         if ($allowSSL === null) {
-            $allowSSL = c('Garden.AllowSSL', false);
+            $allowSSL = c('Garden.AllowSSL', c('Garden.ForceSSL'));
         }
+
         static $rewrite = null;
         if ($rewrite === null) {
             $rewrite = val('X_REWRITE', $_SERVER, c('Garden.RewriteUrls', true));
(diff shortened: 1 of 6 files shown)


DaazKu:
`https` should not be forced when when only `Garden.AllowSSL` is true

library/core/class.upload.php
```diff
@@ -168,6 +168,10 @@ public static function parse($Name) {
         $Name = str_replace('\\', '/', $Name);
         $PathUploads = str_replace('\\', '/', PATH_UPLOADS);
 
+        if (c('Garden.AllowSSL') || c('Garden.ForceSSL')) {
+            $Name = preg_replace("/^http:/i", "https:", $Name);
+        }
+
         if (preg_match('`^https?://`', $Name)) {
```
DaazKu: you already have a pregmatch here. You should not preg_replace for every call to parse.

DaazKu: You already have a pregmatch here. You should not preg_replace for every call to parse.

library/core/class.upload.php
```diff
@@ -334,9 +338,15 @@ public function saveAs($Source, $Target, $Options = array()) {
      * @param $Name
      * @return mixed
      */
-    public static function url($Name) {
-        $Parsed = self::parse($Name);
-        return $Parsed['Url'];
+    public static function url($name) {
+        $parsed = self::parse($name);
```
DaazKu: Parse already handle this case no?

library/core/class.request.php
```diff
@@ -1415,16 +1415,17 @@ public function _unsetRequestArguments($paramsType) {
...
+
         static $rewrite = null;
         if ($rewrite === null) {
             $rewrite = val('X_REWRITE', $_SERVER, c('Garden.RewriteUrls', true));
         }
 
         if (!$allowSSL) {
             $ssl = null;
-        } elseif ($withDomain === 'https') {
+        } elseif ($withDomain === 'https' || c('Garden.ForceSSL')) {
```
slafleche: @linc @tburry Is this a good idea? Without this, I still get an HTTP canonical link when I access a HTTP page with ForceSSL to true.

linc: I am fine with this.

slafleche:
Blocked by https://github.com/vanilla/multisite/issues/130

DaazKu:
Why was this blocked by vanilla/multisite#130 ?

library/core/class.request.php
DaazKu: That's not fine. We sometime pass full URLs to url().
Especially while doing redirects. This would update the URL from http to https.

slafleche: Are there cases we would be "forcing ssl" but would redirect to http?

DaazKu: Yes.

> We sometime pass full URLs to url().

These may be external URLs

DaazKu:
We should probably rethink this.
I actually raised an issue that is related to this one: https://github.com/vanilla/vanilla/issues/5700

slafleche:
Reopening issue, since this PR is no good

PR #5610, opened 2017-05-19
Create multiple OAuth2 connections

nervoustwit:
This feature will allow users to create multiple connections in their dashboard to different OAuth2 applications similar to how JSConnect works.


nervoustwit:
Todo 

- [ ] More robust validation so to avoid multiple instances with the same AuthenticaitonKey or Name.
- [ ] Set defaults when creating new Provider.
- [ ] Doc blocks, documentation.
- [ ] Testing.

nervoustwit:
It looks like, because we don't have Unique IDs in the provider table we can't validate if a provider has a unique key or not.

charrondev:
I believe I've asked a couple times, but this PR is stale enough that the requirements will likely have shifted (or implementation plan) by the time we circle back on it. For that reason I'm closing it at this time.

PR #5616, opened 2017-05-24
Fix stripping of custom tag attributes when using BBCode

initvector:
BBCode allows for certain styling, like `color`. The current processing pipeline runs the output from NBBC, Vanilla's BBCode parser, through htmLawed. Vanilla's stricter htmLawed config for other formats strips the custom formatting. BBCode, unlike other formats like Markdown, does not allow for HTML to be used alongside it. This means HTML entities are escaped or encoded, making them unusable. NBBC should be handling cleaning up the post format.

To that end, this update removes htmLawed from the BBCode formatting chain and entrusts NBBC with ensuring the output is properly sanitized.

Please backport to [release/2017-Q1-6](https://github.com/vanilla/vanilla/tree/release/2017-Q1-6).

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -278,13 +278,10 @@ class Gdn_Format {
                 // Standard BBCode parsing.
                 $Mixed = $BBCodeFormatter->format($Mixed);
 
-                // Always filter after basic parsing.
-                $Sanitized = Gdn_Format::htmlFilter($Mixed);
-
                 // Vanilla magic parsing.
-                $Sanitized = Gdn_Format::processHTML($Sanitized);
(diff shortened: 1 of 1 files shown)


initvector:
Gonna try this with a custom htmLawed spec

DaazKu:
I really think we should be updating the htmlawed plugin to add an exception for this instead of not using it.

`span=style(match="/^(color:(#[a-f\d]{3}[a-f\d]{3}?|[a-z]+))?;?$/i"); ` to the spec string could work.

DaazKu:
Welp you closed the issue way before I hit "Comment" XD

Issue #3431, opened 2016-01-06
Prevent access to cache files and libraries


DaazKu:
It is possible to access some files that shouldn't be accessed.

The content of (Non exhaustive list)
- /cache folder
- /uploads/import
- /vendor
  - /composer 
  - /libraries

I made a little proof of concept to show what could be done for apache to at least prevent some php execution and restrict complete folders entirely:

```
    # Prevent access to anything in there
    RewriteRule ^(cache|vendor/composer)/.*$ index.php/notfound [QSA,E=X_REWRITE:1,E=X_PATH_INFO:/notfound,L]

    # Prevent access to php files in there
    RewriteRule ^(vendor)/.+\.php$ index.php/notfound [QSA,E=X_REWRITE:1,E=X_PATH_INFO:/notfound,L]
```

We should try to determine what must not be accessed and do something intelligent about it.

Note from Lincoln: Vanilla is often installed in the root directory of along side other applications so we cannot blindly block everything but a few select script to be called.

PR #5621, opened 2017-05-25
.htaccess hardening

DaazKu:
- Disallow directories that do not need to be exposed by the webserver
- Prevent any script but /index.php from being called by default.
- Clean up existing rules
- Remove deprecated 

Fixes https://github.com/vanilla/vanilla/issues/3431

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -12,7 +12,7 @@
     #RewriteBase /
 
     ####
-    # Deny access to certain directories that SHOULD not be exposed.
+    # Deny access to certain directories that SHOULD NOT be exposed.
     ####
     RewriteRule (^|/)\.git - [L,R=403]
     #RewriteRule ^build/ - [L,R=403] # Already covered by .htaccess
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -15,12 +15,9 @@
     # Deny access to certain directories that SHOULD NOT be exposed.
     ####
     RewriteRule (^|/)\.git - [L,R=403]
-    #RewriteRule ^build/ - [L,R=403] # Already covered by .htaccess
     RewriteRule ^cache/ - [L,R=403]
     RewriteRule ^cgi-bin/ - [L,R=403]
     RewriteRule ^uploads/import/ - [L,R=403]
-    #RewriteRule ^conf/ - [L,R=403] # Already covered by .htaccess
-    #RewriteRule ^tests/ - [L,R=403] # Already covered by .htaccess
     RewriteRule ^vendor/ - [L,R=403]
 
     ####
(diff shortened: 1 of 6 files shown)


DaazKu:
@linc Are you ok with this not being commented by default? https://github.com/vanilla/vanilla/blob/31273c7373aa42b3eee32112c590a75fc90a399f/.htaccess.dist#L33

linc:
Not really, no. That's the sort of thing that confuses the hell out of folks who aren't expecting it and will send them on a days-long odyssey. It should be opt-in for folks who want that level of security.

.htaccess.dist
```diff
@@ -4,33 +4,53 @@
...
     # If vanilla is in a subfolder then you need to specify it after the /.
     # (ex. You put Vanilla in /forum so change the next line to: RewriteBase /forum)
-    # RewriteBase /
+    ####
+    #RewriteBase /
 
-    RewriteCond %{QUERY_STRING} ^p=/?([^&]+)(&([^?]+))?$
-    RewriteRule ^index\.php %1?%3 [E=X_REWRITE:1,L]
+    ####
+    # Deny access to certain directories that SHOULD not be exposed.
```
linc: I think you wanna capitalize the 'NOT' also or instead.

.htaccess.dist
```diff
@@ -4,33 +4,53 @@
...
+    ####
+    RewriteRule (^|/)\.git - [L,R=403]
+    #RewriteRule ^build/ - [L,R=403] # Already covered by .htaccess
```
linc: In what scenario is someone uploading a `build` or `tests` folder that we'd need to account for it?

build/.htaccess
```diff
@@ -0,0 +1 @@
+Deny from all
```
linc: It's common for hosts to ignore .htaccess files that aren't in the root directory. What's the rationale for splitting this off?

.htaccess.dist
DaazKu: If they upload the cloned repository

build/.htaccess
DaazKu: /tests already have an .htaccess I just copied the behaviour.

I added the commented rules in the case someone preferred not to use the .htaccess (which is faster I think)

DaazKu: That being said should I uncomment them?

linc: I'd delete the redundant lines in the main htaccess entirely.

These directories should never, ever make it to production. The sub-.htaccess files being needed is already a backup. Double-backing it up by also having comments in the main htaccess seems like tacit approval of this massive screwup because someone would have to opt into it, at which point what the hell are they even doing?

Issue #5597, opened 2017-05-18
Advanced Embed jumps to iframe if it's not into the viewport


DaazKu:
EDIT: This happens only on the login page because there is an autofocus on the email.

PR #5630, opened 2017-05-29
Disable autoscroll on an embedded signin page

DaazKu:
This fixes an issue with embedding where the iframe would be down the page and the page would scroll down to the embed's email field.

Fixes https://github.com/vanilla/vanilla/issues/5597


linc:
How does adding a config fix this?

PR #5644, opened 2017-06-02
Add support for refresh tokens in OAuth2

nervoustwit:
Some implementations of Oauth2 may want to issue a refresh token that can then be used to get a new access token when the access token is expired. In this PR the OAuth2 class has been changed to save a refresh token to the users attributes if one is passed and when getting the AccessToken OAuth2 will:

- check if the access token is present, if not
- if no refresh token is saved in Gdn_User Attributes it will use the access_token saved there.
- if there is a refresh token saved in Gdn_User Attributes it will request a new access_token from the provider and save the new a refresh token to Gdn_User Attributes.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -559,7 +559,7 @@ class Gdn_OAuth2 extends Gdn_Plugin {
                 // for future requests, if not, store the access_token.
                 $attributes = [
                     'RefreshToken' => val('refresh_token', $response),
-                    'AccessToken' => val('access_token', $response, val('access_token', $response)),
+                    'AccessToken' => val('access_token', $response, val('refresh_token', $response)),
                     'Profile' => $profile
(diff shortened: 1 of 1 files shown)


library/core/class.oauth2.php
```diff
@@ -543,8 +555,11 @@ public function entryEndpoint($sender, $code, $state) {
                     'UniqueID' => $profile['id']]);
 
                 // Save the information as attributes.
+                // If a client has passed a refresh_token, store it as the access_token in the attributes
+                // for future requests, if not, store the access_token.
                 $attributes = [
-                    'AccessToken' => $response['access_token'],
+                    'RefreshToken' => val('refresh_token', $response),
+                    'AccessToken' => val('access_token', $response, val('access_token', $response)),
```
DaazKu: Wut?

PR #5654, opened 2017-06-09
Change html tag "td" in "thead" to "th" in notifications settings page

R-J:
In a table header `<thead>`, you should use `<th>` instead of `<td>` for the header elements.

Small css adjustments were made to keep the formatting. The visible effect here is that `<th>` element has a heavier font weight and therefore the header is now printed bold. But if you ask me, that's an improvement, too.

### DIFF
--- a/applications/dashboard/views/profile/preferences.php
+++ b/applications/dashboard/views/profile/preferences.php
@@ -4,12 +4,13 @@
         width: 500px;
     }
 
-    thead td {
+    table.PreferenceGroup th {
         vertical-align: bottom;
         text-align: center;
     }
 
     table.PreferenceGroup thead .TopHeading {
+        text-align: center;
         border-bottom: none;
     }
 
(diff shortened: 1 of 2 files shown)


chrisvanpatten:
Could this also be updated to include support for the `scope` attribute, and possibly the `headers` attribute?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th

These can be helpful for a11y (screen readers, etc).

R-J:
At least I can make a pull request ;-)
I've already made the changes for the general preferences, but I'm not sure how and if the scope has to be applied to the advanced preferences:

~~~
<table class="PreferenceGroup">
    <thead>
    <tr>
        <th style="border: none;">&nbsp;</th>
        <th class="TopHeading" colspan="2">Discussions</th>
        <th class="TopHeading" colspan="2">Comments</th>
    </tr>
    <tr>
        <th style="text-align: left;">Category</th>
        <th class="PrefCheckBox BottomHeading">Email</th>
        <th class="PrefCheckBox BottomHeading">Popup</th>
        <th class="PrefCheckBox BottomHeading">Email</th>
        <th class="PrefCheckBox BottomHeading">Popup</th>
    </tr>
    </thead>
~~~

I think I will simply add it to each and every `<th>` element, but would that be correct for those elements that have a colspan?

PR #5656, opened 2017-06-09
Extend notification preference table with accessibility attributes

R-J:
First I've changed some `<td>` elements to semantically correct `<th>` elements.

Furthermore I have added `scope` and `headers` attributes to the `th` and `td` elements like @chrisvanpatten suggested [here](https://github.com/vanilla/vanilla/pull/5654). 

If someone has the possibility to test it, I've pasted the table to [codepen](https://codepen.io/anon/pen/qjbMeP)

I'm not sure if the scope in line 25 and 27 of the codepen must be col or colgroup. See [here](https://www.w3schools.com/tags/att_th_scope.asp) for some more info

### DIFF
--- a/applications/dashboard/views/profile/preferences.php
+++ b/applications/dashboard/views/profile/preferences.php
@@ -4,12 +4,13 @@
         width: 500px;
     }
 
-    thead td {
+    table.PreferenceGroup th {
         vertical-align: bottom;
         text-align: center;
     }
 
     table.PreferenceGroup thead .TopHeading {
+        text-align: center;
         border-bottom: none;
     }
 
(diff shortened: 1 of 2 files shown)


chrisvanpatten:
@R-J My interpretation of the spec is that it should only be "colgroup" if you use an actual `<colgroup>` tag. I think the way you did it looks good!

chrisvanpatten:
@R-J Actually, here's an example of an accessible table with the exact same layout that makes use of the `<colgroup>` tag. Maybe worth copying the whole thing? http://www.yourhtmlsource.com/examples/accessibletable.html

R-J:
@chrisvanpatten feel free to do a PR anytime you want, but this time I have been faster ;-)

If you know of other places where there could be made some accessibility improvements, just make it an issue and give some details so that even someone like me, who has no idea of accessibility at all could see what to do.

I know that the Vanilla staff welcomes such efforts and I myself like to tune small parts of code, so I guess that would be a perfect fit!

DaazKu:
Thanks!

PR #5658, opened 2017-06-12
Fix/categories discussions tag information

R-J:
CategoriesController's and DiscussionController's index method fill their $Data['Discussions'] array with a list of discussions. That is needed in the base_render_before() method in applications/vanilla/settings/class.hooks.php. If there is  Discussions info available, tagging information is joined to the discussion:
~~~
if (null !== $sender->data('Discussions', null)) {
    TagModel::instance()->joinTags($sender->Data['Discussions']);
}
~~~

I do not see a need for that in Vanilla, but I wrote a plugin which shows tag information in discussion overviews and I was surprised to find that it worked in /discussions, /categories/something, but not in /categories/discussions

For the sake of consistency, that info should also be available in that method.

### DIFF
--- a/applications/vanilla/controllers/class.categoriescontroller.php
+++ b/applications/vanilla/controllers/class.categoriescontroller.php
@@ -572,7 +572,7 @@ class CategoriesController extends VanillaController {
                 );
             }
         }
-        // $this->setData('Discussions', $Discussions); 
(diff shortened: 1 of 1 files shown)


DaazKu:
To answer your question this feature is used by some internal plugins.
Thanks for the fix!

PR #5672, opened 2017-06-15
Improve feedback when a duplicate addon key is encountered

### LINKED ISSUE #5053: Warn when AddonManager encounters a duplicate addon key
Currently, if an **application** and a **plugin** have the same key name such as `vanillastats`, only the plugin gets scanned and the application is skipped silently, even though it shows as enabled in the dashboard.

It would be useful to warn whenever that happens so that we could troubleshoot the problem without enabling and disabling plugins until we find the culprit.

initvector:
When addons use duplicate directories or keys, there's currently not a lot of feedback to go on for the errors that arise.

This update adds some exception throwing to two key areas when addons are scanned. If a duplicately-named addon directory is encountered, an exception is thrown indicating which addon is causing the issue. If a duplicate addon key is encountered, a similar exception is thrown and the error is logged to the database. The directory scan doesn't include logging, because it's done before addons are loaded and hence we cannot count on a logger being available. The addon key scanning is run on addon pages during cache checks, so there's an increased likeliness a logger will be available.

We're moving in a directory where an addon is an addon is an addon. Themes, plugins and applications should have globally unique keys and their keys should match their directory names.

Closes #5053

### DIFF
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -328,6 +328,10 @@ class AddonManager {
                 if (!array_key_exists($key, $addons)) {
                     $addons[$key] = $addon;
                 } else {
+                    \Logger::error('Duplicate addon: {key}', [
+                        'key' => $key,
+                        'event' => 'duplicate_addon'
+                    ]);
                     throw new \Exception("Duplicate addon: {$key}");
                 }
             } catch (\Exception $ex) {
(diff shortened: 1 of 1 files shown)


library/Vanilla/AddonManager.php
```diff
@@ -359,7 +371,15 @@ private function scanAddonDirs($type) {
         foreach ($this->scanDirs[$type] as $subdir) {
             $paths = glob(PATH_ROOT."$subdir/*", GLOB_ONLYDIR | GLOB_NOSORT);
             foreach ($paths as $path) {
-                $result[basename($path)] = substr($path, $strlen);
+                $basename = basename($path);
+                if (!array_key_exists($basename, $result)) {
+                    $result[basename($path)] = substr($path, $strlen);
```
evach-vf: $result[$basename] = substr($path, $strlen);
instead of 
$result[basename($path)] = substr($path, $strlen);

PR #5674, opened 2017-06-16
Fix path detection in environment setup to work with symlinks again

linc:
Reverts use of `__DIR__` to `getcwd()` in setting our `PATH_ROOT`. 

Using `__DIR__` adds the benefit of caching its value, but removes the ability to use symlinks for much of Vanilla. Since we currently have no plan or strategy to accommodate complex localhost setups (e.g. "How do you then symlink 50 plugins into 50 localhost sites?"), this tradeoff was very bad. It also creates a bunch of communication overhead for the open source release that we had no plan for.

Relevant discussion was previously here: https://github.com/vanilla/vanilla/pull/5492#pullrequestreview-36433978

### DIFF
--- a/environment.php
+++ b/environment.php
@@ -22,7 +22,7 @@ if (!defined('DS')) {
     define('DS', DIRECTORY_SEPARATOR);
 }
 if (!defined('PATH_ROOT')) {
-    define('PATH_ROOT', __DIR__);
+    define('PATH_ROOT', getcwd());
 }
 
 /**


DaazKu:
Actually you can create a file that define PATH_ROOT and include index.php. Then you update nginx to use this indead of index.php

linc:
What is the value proposition of caching that value vs. the communication overhead of what this breaks? Yes, there's a workaround, but you'd still need to implement that PER SITE with a new symlink to a new secret file that needs to go in gitignore and get documented.

DaazKu:
Symlinking the file in 50 folder to create 50 sites is pretty atypical. There's probably not gonna be more than an handful of installation having that "problem". 

Also you do not need to implement that solution per site since you will be able to symlink it. And no need to put it in git ignore since you can setup that special case in .git/info/exclude

Not against the revert though. Just trying to balance things out.

linc:
There are only four reasonable ways I know of to run dozens of Vanilla sites in parallel using a unified code base:

1. Empty shell web folders that symlink in the root files + "collection" folders of all addons etc symlinked into one place. This is what I use.
2. Physically copying all the files via a manual deploy, a la our Pusher (inf).
3. Some sort of automated symlink tooling that doesn't exist. This sounds like a nightmare.
4. A "hybrid" setup where you have a series of configs per site, but you're locked into running all those "sites" in the repo itself with 1 shared `uploads` and 1 shared `cache` folder (maybe you could hack around this with more constants declarations?), and an impossibility of running different sites against different repos (versions). It also complicates things like running WordPress or alternate setups on an instance without modifying your nginx conf every time. I understand multiple devs are doing this but I'm strongly opposed currently.

Using `__DIR__` breaks number 1. Number 2 adds a manual step and we don't have a localhost implementation of it. Number 3 is out. And number 4 makes managing multiple releases in tandem significantly more difficult.

So unless there is serious value in the original switch, I don't see how it outweighs scuttling this useful behavior.

DaazKu:
Funny story: reverting to getcwr() broke https://github.com/vanilla/vanilla-cli addon commands.
Fortunately the fix is to do what I said in my [previous comment](https://github.com/vanilla/vanilla/pull/5674#issuecomment-309049267) about how to fix PATH_ROOT.

PR #5678, opened 2017-06-16
Fix recipient being added a second time to the same conversation

DaazKu:
There is a bug where if you have over 20 recipient and you try to add the 21st again on the conversation it will throw an exception due to the database rejecting the insert query because of the unique constraint on (ConversationID, UserID).

The default limit of getRecipients is 20 so I'm upping the query to 1000 since conversation are not meant to be thread of discussion.

### DIFF
--- a/applications/conversations/models/class.conversationmodel.php
+++ b/applications/conversations/models/class.conversationmodel.php
@@ -329,7 +329,7 @@ class ConversationModel extends ConversationsModel {
      * @param int $Limit The number of recipients to grab.
      * @return Gdn_DataSet SQL results.
      */
-    public function getRecipients($ConversationID, $Limit = 20) {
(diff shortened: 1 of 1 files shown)


applications/conversations/models/class.conversationmodel.php
```diff
@@ -771,7 +771,7 @@ public function addUserToConversation($ConversationID, $UserID) {
         }
 
         // First define the current users in the conversation
-        $OldContributorData = $this->getRecipients($ConversationID);
+        $OldContributorData = $this->getRecipients($ConversationID, 1000);
```
initvector: If you're dialing up the artificial recipient ceiling, why not apply it to all calls or update the default in the function? `ConversationModel::getRecipients` is currently called in four places in Vanilla. It seems like a bad idea to have it continue to return 20, everywhere else.

Issue #5695, opened 2017-06-26
Unify redirect, redirectURL and safeRedirect into a new function called redirectTo


DaazKu:
To improve security site wise we will now validate redirects against the trusted domain list by default. To achieve this we will create the new redirectTo function.

See: https://github.com/vanilla/vanilla/issues/5218#issuecomment-310513372

Issue #5218, opened 2017-02-22
Audit of redirect() and Controller->RedirectUrl


DaazKu:
In most cases, users should not be leaving the forum without being warned.
A review of every redirects needs to be done.

Both call to redirect() and $this->RedirectUrl should be reviewed.

~~It would also be wise to default redirect() to do a safeRedirect since it can be easy to forget that safeRedirect exists and what is the difference between the 2.~~ *This would introduce too many bugs if we do it*

~~We could deprecate safeRedirect and add a third parameter to redirect() which would do a safeRedirect by default.~~ See https://github.com/vanilla/vanilla/issues/5218#issuecomment-310513372

Depends on https://github.com/vanilla/vanilla/issues/5695

DaazKu:
My current plan is to:
- deprecate `redirect()`, `redirectUrl()` and `Gdn_Controller()->RedirectUrl`
- create a `unsafeRedirect()` function which will be the counterpart to `safeRedirect()`
- create `Gdn_Controller->setSafeRedirect()` and `Gdn_Controller->setUnsafeRedirect()` which will take precedence over `Gdn_Controller()->RedirectUrl`
- Replace all call of `redirect()`, `redirectUrl()` to `unsafeRedirect()`
- Replace all assignation to `Gdn_Controller()->RedirectUrl` by `Gdn_Controller->setUnsafeRedirect()`

The PRs for this are almost done. Once this is approved I will start auditing all the unsafe calls.
After this update, awareness of the security implication of a redirect should be way more obvious.

tburry:
Really I think redirect() is well named. I just think that safeRedirect() is maybe not properly named. Maybe we could also just make a parameter to check against whitelisted URLs?

WRT a controller method I'd prefer that the controller not get fatter also given that the API uses a different base class. Another option is a RedirectException that the dispatcher can handle.

DaazKu:
The problem with the parameter is that we'd like it to be "safe" by default.
That way people won't call redirect incorrectly by default which is why we have non trusted redirection problem currently.

DaazKu:
After an internal chat with Todd we ended up with the following plan:

- deprecate `redirect()`, `redirectUrl()` (It's almost an exact copy of redirect()), `safeRedirect()` `Gdn_Controller()->RedirectUrl`
- create `redirectTo($destination='', $statusCode=302, $trustedOnly=true)`
[...]

PR #5696, opened 2017-06-26
Use the new redirectTo function

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/5695

- Deprecate other redirect functions
- Use redirectTo everywhere as a perfect replacement. (most redirects are unsafe for now. We will address that in https://github.com/vanilla/vanilla/issues/5218

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -648,7 +648,7 @@ class DashboardHooks extends Gdn_Plugin {
             $deliveryType = $Sender->getDeliveryType($deliveryMethod);
             if (!$IsApi && !Gdn::request()->isPostBack() && $deliveryType !== DELIVERY_TYPE_DATA) {
                 $url = trim(preg_replace('#(\?.*)sso=[^&]*&?(.*)$#', '$1$2', Gdn::request()->pathAndQuery()), '&');
-                redirectUrl($url);
+                redirectTo(url($url, true), 302, false);
             }
         }
         $this->checkAccessToken();
@@ -725,7 +725,7 @@ class DashboardHooks extends Gdn_Plugin {
 
         // If an event handler didn't handle the signin then just redirect to the target.
         if (!$Handled) {
-            safeRedirect($Target, 302);
+            redirectTo($Target, 302);
         }
     }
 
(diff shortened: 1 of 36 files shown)


applications/dashboard/settings/class.hooks.php
```diff
@@ -725,7 +725,7 @@ public function rootController_sso_create($Sender, $Target = '') {
 
         // If an event handler didn't handle the signin then just redirect to the target.
         if (!$Handled) {
-            safeRedirect($Target, 302);
+            redirectTo($Target, 302);
```
initvector: I know you just switched over the functions, but 302 is the default response code, so you shouldn't need to explicitly state it here.

applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -351,7 +351,7 @@ public function discussion($CategoryUrlCode = '') {
                         $this->fireEvent('AfterDiscussionSave');
 
                         if ($this->_DeliveryType == DELIVERY_TYPE_ALL) {
-                            redirect(discussionUrl($Discussion, 1)).'?new=1';
+                            redirectTo(discussionUrl($Discussion, 1), 302, false).'?new=1';
```
initvector: This isn't behavior you introduced, but it appears incorrect. The `new` param is being appended to the result of `redirectTo`.  It should be appended to the result of `discussionUrl`.

```php
redirectTo(discussionUrl($Discussion, 1).'?new=1', 302, false);
```

library/core/functions.general.php
```diff
@@ -2834,67 +2834,37 @@ function betterRandomString($length, $characterOptions = 'A0') {
...
+     * @param int $statusCode HTTP status code. 302 by default.
+     * @param bool $trustedOnly Non trusted destinations will be redirected to /home/leaving?Target=$destination
      */
-    function redirect($Destination = false, $StatusCode = null) {
-        if (!$Destination) {
-            $Destination = '';
+    function redirectTo($destination = null, $statusCode = 302, $trustedOnly = true) {
+        if ($destination === null) {
+            $url = url('');
+        } else if ($trustedOnly) {
```
initvector: `elseif`

applications/dashboard/settings/class.hooks.php
DaazKu: I'm going to update review all the redirect in my next PR that's why I did not do any further in my update.

initvector:
Backported to [release/2.4](https://github.com/vanilla/vanilla/tree/release/2.4)

PR #5698, opened 2017-06-27
Allow FloodControlHelper to restrict the actions of admins and mods

nervoustwit:
Presently when using he flood control helper we detect if the user is an admin or a moderator and by-pass flood control. There are some specialized instances where we may want to control the frequency of some activity of admins as well, e.g. executing a custom script like downloading.

The PR adds an `$skipAdmins` param to the flood control helper constructor.

https://github.com/vanillaforums/vanilla-clients/pull/269


nervoustwit:
Backport to release/2017-Q2-6

PR #5703, opened 2017-06-28
Deprecate Gdn_Controller->RedirectUrl and add redirectTo()

DaazKu:
Follow up of https://github.com/vanilla/vanilla/issues/5695

The EntryController had a function named redirectTo that was really used to get the target route.
I renamed the function and made it protected since it was meant to be used internally.

Backport to release/2.4


library/core/class.controller.php
```diff
@@ -2142,4 +2164,23 @@ public function title($Title = null, $Subtitle = null) {
...
+    public function redirectTo($destination = null, $trustedOnly = true) {
+        if ($destination === null) {
+            $url = url('');
+        } elseif ($trustedOnly) {
+            $url = safeURL($destination);
+        } else {
+            $url = url($destination);
+        }
+
+        $this->redirectTo = $url;
```
initvector: For this to truly replace the existing direct assignment addons are using for setting the `RedirectUrl` property, shouldn't this function also be setting it? This seems like we're breaking backwards compatibility. If some addon out there is relying on `RedirectUrl` having been set at some point where this function is being introduced, it would begin to fail. I know we're trying to deprecate usage of the `RedirectUrl` property, but there has to be a better way than outright removing a public property being set in so many places.

library/core/class.controller.php
```diff
@@ -2142,4 +2164,23 @@ public function title($Title = null, $Subtitle = null) {
...
     }
+
+    /**
+     * Set the destination URL where the page will be redirected after an ajax request.
+     *
+     * @param string|null $destination Destination URL or path.
+     *      Redirect to current URL if nothing or null is supplied.
+     * @param bool $trustedOnly Non trusted destinations will be redirected to /home/leaving?Target=$destination
+     */
+    public function redirectTo($destination = null, $trustedOnly = true) {
```
initvector: Having this method share a name with a property of the class could be confusing. I recommend changing the method name to `setRedirectTo`.

library/core/class.controller.php
```diff
@@ -1270,7 +1278,16 @@ public function xRender($View = '', $ControllerName = false, $ApplicationFolder
...
-            $this->setJson('RedirectUrl', $this->RedirectUrl);
+            if ($this->redirectTo !== null) {
+                $this->setJson('RedirectTo', $this->redirectTo);
+                $this->setJson('RedirectUrl', $this->redirectTo);
+            } else if ($this->RedirectUrl !== '') {
+                $this->setJson('RedirectTo', $this->RedirectUrl);
+                $this->setJson('RedirectUrl', $this->RedirectUrl);
+            } else {
+                $this->setJson('RedirectTo', '');
+                $this->setJson('RedirectUrl', '');
```
initvector: The `RedirectUrl` blocks can probably be simplified into one.

```php
$redirectUrl = $this->RedirectUrl ?: '';
$this->setJson('RedirectTo', $redirectUrl);
$this->setJson('RedirectUrl', $redirectUrl);
```

Unless there a reason we might want to allow any falsy values.

library/core/class.controller.php
DaazKu: I thought about it but the property if actually protected so it can only be confusing internally.
I also wanted the function to have the same name than it's non AJAX counterpart.

library/core/class.controller.php
DaazKu: If an addon sets RedirectUrl it will be honoured: https://github.com/vanilla/vanilla/pull/5703/files#diff-0b51e9233f2e072d92f57f60e061864dR1284

initvector: If it *sets* the `RedirectUrl`, it's good. But what if it's hooking into an event and is trying to *read* the `RedirectUrl` property that was previously set in any of these dozens of locations?

initvector:
Backported to [release/2.4](https://github.com/vanilla/vanilla/tree/release/2.4)

Issue #3015, opened 2015-08-28
Could not find file 'plugins/Flagging/design/flagging.css' in folder 'vanilla'.


btafoya:
While viewing the demo thread..

http://{site}discussion/1/bam-you-ve-got-a-sweet-forum#latest

Debug Trace shows:
Info    Could not find file 'plugins/Flagging/design/flagging.css' in folder 'vanilla'.

Git Master source.

mtschirs:
Might be related to http://vanillaforums.org/discussion/comment/232071#Comment_232071

Data66Tech:
Thanks, this worked for 2.3 stable for me.

PR #5706, opened 2017-06-29
Remove addCssFile calls referencing non existent or superfluous files

R-J:
The plugin tries to a) add admin.css which is included anyway in the dashboard and b) add flagging.css which has been removed since it wasn't used.
The way that this plugin was add css files wasn't working since release 2.2

This closes https://github.com/vanilla/vanilla/issues/3015


linc:
Thanks!

PR #5707, opened 2017-06-29
Replace deleted users mail placeholder

R-J:
Vanilla replaces deleted users mail address with a placeholder: `user_XY@deleted.email`. Since ".email" became a valid TLD, that isn't an invalid mail address any more.
[RFC 2606](https://tools.ietf.org/html/rfc2606) specifies a better TLD for that: ".invalid". Using this TLD leaves no doubt that the mail address itself is invalid although the format isvalid.


DaazKu:
May I suggest `"deleted_$userID@email.invalid"` ?

PR #5708, opened 2017-06-29
[Editor] Fix alignment issues in BBCode

Charrondev:
Built on #5711 

The editor (specifically bbcode) is not aligning text correctly. It is floating elements when it should only be using `text-align`. As a result elements are not getting their own line.

**Before:**

image

**After:**

image

This will need a backport to `release/2017-Q1-6` for Zenimax.


Charrondev:
I've broken the bbcode classes into there own style with no margins or floating. The AlignLeft/AlignRight classes seem to only be used for the other editors. I'm of the opinion that none of the editors should move this simplified styling though. Right now the HTML editor for example needs these margins and floats because it inserts a line-break after each floated element. The margins combined then take up the height of the line-break.

The BBCode editor does not insert line breaks between aligned items though.

Charrondev:
Backported

PR #5713, opened 2017-06-29
Update redirectTo calls to use the $trustedOnly option where necessary

DaazKu:
Partially addresses https://github.com/vanilla/vanilla/issues/5218

We [consolidated all redirect functions into redirectTo](https://github.com/vanilla/vanilla/issues/5695) in order to have a "safe by default" redirect function.
This PR removes `$trustedOnly = false` to all redirections where it should be safe to do so.

Redirects updated to have `$trustedOnly = true`:
- Any relative `$destination`.
- Any `$destination` coming from Target that are not part of SSO redirects.

Backport to 2.4 (all pr depending on this one too)


initvector:
Backported to [release/2.4](https://github.com/vanilla/vanilla/tree/release/2.4).

PR #5716, opened 2017-06-30
Added clearfix for messages container.

slafleche:
User content can contain floating element, like a right aligned element. The message should clear it's self, to not interfere with other elements.

This is pretty low risk of breaking themes. I've added this same fix to some themes in the past. I've also tested a handful of popular themes.

A note for the future. This PR also explicitly defines a version of node-sass, which caused the output of one color function to vary slightly. This shouldn't happen in the future though.

Fixes: vanilla/addons#521


slafleche:
Backported

PR #5727, opened 2017-07-05
Enable pagination when displaying tagged discussions

nervoustwit:
In the role tracking module we use Tags to designate discussions in which tracked users have posted. That module can track multiple Roles and therefore multiple Tags. If the a user clicks on the link in the Role Tracker to "View All" presently it there is no pagination because we weren't adding up the number of discussions per Tag. This PR will fix that.

### DIFF
--- a/applications/vanilla/controllers/class.discussionscontroller.php
+++ b/applications/vanilla/controllers/class.discussionscontroller.php
@@ -722,40 +722,48 @@ class DiscussionsController extends VanillaController {
         $Tag = stringEndsWith($Tag, '.rss', true, true);
         list($Offset, $Limit) = offsetLimit($Page, c('Vanilla.Discussions.PerPage', 30));
 
-        $MultipleTags = strpos($Tag, ',') !== false;
-
         $this->setData('Tag', $Tag, true);
 
         $TagModel = TagModel::instance();
-        $RecordCount = false;
-        if (!$MultipleTags) {
+        //If more than one tag has been passed via GET, pass them as an array to the model.
+        $MultipleTags = strpos($Tag, ',') !== false;
+        if ($MultipleTags) {
+            $Tags = $TagModel->getWhere(array('Name' => explode(',', $Tag)))->resultArray();
+        } else {
             $Tags = $TagModel->getWhere(array('Name' => $Tag))->resultArray();
-
-            if (count($Tags) == 0) {
-                throw notFoundException('Page');
-            }
-
-            if (count($Tags) > 1) {
-                foreach ($Tags as $TagRow) {
-                    if ($TagRow['CategoryID'] == val('CategoryID', $Category)) {
-                        break;
-                    }
-                }
-            } else {
-                $TagRow = array_pop($Tags);
-            }
-            $Tags = $TagModel->getRelatedTags($TagRow);
-
-            $RecordCount = $TagRow['CountDiscussions'];
-            $this->setData('CountDiscussions', $RecordCount);
-            $this->setData('Tags', $Tags);
-            $this->setData('Tag', $TagRow);
-
-            $ChildTags = $TagModel->getChildTags($TagRow['TagID']);
-            $this->setData('ChildTags', $ChildTags);
         }
 
-        $this->title(htmlspecialchars($TagRow['FullName']));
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -722,40 +722,48 @@ public function tagged() {
...
         list($Offset, $Limit) = offsetLimit($Page, c('Vanilla.Discussions.PerPage', 30));
 
-        $MultipleTags = strpos($Tag, ',') !== false;
-
         $this->setData('Tag', $Tag, true);
 
         $TagModel = TagModel::instance();
-        $RecordCount = false;
-        if (!$MultipleTags) {
+        //If more than one tag has been passed via GET, pass them as an array to the model.
```
DaazKu: Comments should start with a space.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -722,40 +722,48 @@ public function tagged() {
...
+        $MultipleTags = strpos($Tag, ',') !== false;
+        if ($MultipleTags) {
+            $Tags = $TagModel->getWhere(array('Name' => explode(',', $Tag)))->resultArray();
```
DaazKu: Use the short array syntax for new code.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -722,40 +722,48 @@ public function tagged() {
...
+        }
 
-            if (count($Tags) > 1) {
-                foreach ($Tags as $TagRow) {
-                    if ($TagRow['CategoryID'] == val('CategoryID', $Category)) {
-                        break;
-                    }
+        // Get the number of discussions and the page title by either looping
+        // through all the tags or getting it from the result from the TagModel.
+        $RecordCount = false;
```
DaazKu: Don't write new code with old standards. You can just update all the variable to lower cases.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -722,40 +724,48 @@ public function tagged() {
...
+        // Get the number of discussions and the page title by either looping
+        // through all the tags or getting it from the result from the TagModel.
+        $recordCount = false;
+        if (count($tags) > 1) {
+            $fullNames = [];
+            foreach ($tags as $TagRow) {
+                $recordCount += val('CountDiscussions', $TagRow, 0);
+                $fullNames[] = val('FullName', $TagRow);
+                if ($TagRow['CategoryID'] == val('CategoryID', $Category)) {
+                    break;
```
DaazKu: Why is there a break here?

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -722,40 +724,48 @@ public function tagged() {
...
+        } else {
+            $TagRow = array_pop($tags);
+            $recordCount = val('CountDiscussions', $TagRow, 0);
+            $tags = $tagModel->getRelatedTags($TagRow);
+            $this->setData('Tags', $tags);
             $this->setData('Tag', $TagRow);
-
-            $ChildTags = $TagModel->getChildTags($TagRow['TagID']);
+            $ChildTags = $tagModel->getChildTags($TagRow['TagID']);
             $this->setData('ChildTags', $ChildTags);
```
DaazKu: Why aren't we setting child tags when we do multiple tags?

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -722,40 +724,48 @@ public function tagged() {
...
 
-            if (count($Tags) > 1) {
-                foreach ($Tags as $TagRow) {
-                    if ($TagRow['CategoryID'] == val('CategoryID', $Category)) {
-                        break;
-                    }
+        // Get the number of discussions and the page title by either looping
+        // through all the tags or getting it from the result from the TagModel.
+        $recordCount = false;
+        if (count($tags) > 1) {
```
DaazKu: We can probably get rid of this condition and foreach on tags no matter is there is one or more tags.

PR #5729, opened 2017-07-06
Add back category URL kludge in calculateUser

DaazKu:
We undid an undocumented kludge in https://github.com/vanilla/vanilla/pull/5701
Reverting the kludge does not break https://github.com/vanilla/multisite/pull/133

Changing the URL from being absolute to relative causes some problem since that value is accessible from the API.

Edit:
calculate() sets the URL as relative because the routing process needs the value to be.
calculateUser() then rewrites the URL as absolute to fix that. These kind of kludge should be explained in the code. The comment I made should at least help understand the why a bit more.

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -180,9 +180,8 @@ class CategoryModel extends Gdn_Model {
      * @param bool|null $addUserCategory
      */
     private function calculateUser(&$category, $addUserCategory = null) {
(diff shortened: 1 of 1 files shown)


DaazKu:
I'm not 100% sure but this may need to be backported to the latest release 2017-Q2-6

initvector:
Backported to [release/2017-Q2-6](https://github.com/vanilla/vanilla/tree/release/2017-Q2-6)

Issue #5718, opened 2017-06-30
Improves addon's error message when addon.json could not be parsed


DaazKu:
The current exception says `The addon at $subdir has an empty info array.` but that might not be true.

There are 2 possibilities there:
- addon.json was not readable
- addon.json was not valid json (extra , for example)

We should update the message to be more relevant.

PR #5733, opened 2017-07-06
Addon: Differentiate between unreadable file and invalid JSON format

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/5718

This report a more accurate error if you have a syntax error in the addon.json file.
Having an extra comma in the file would throw "The addon at $subdir doesn't have any info." which is not useful at all :)

### DIFF
--- a/library/Vanilla/Addon.php
+++ b/library/Vanilla/Addon.php
@@ -146,7 +146,7 @@ class Addon {
 
             $info = json_decode($addonJSON, true);
             if (empty($info)) {
-                throw new \Exception("The addon at $subdir has non valid JSON in addon.json.");
+                throw new \Exception("The addon at $subdir has invalid JSON in addon.json.");
             }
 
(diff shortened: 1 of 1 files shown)


library/Vanilla/Addon.php
```diff
@@ -139,10 +139,14 @@ private function scanInfo() {
...
-            $info = json_decode(file_get_contents("$dir/addon.json"), true);
+            $addonJSON = file_get_contents("$dir/addon.json");
+            if (!$addonJSON) {
+                throw new \Exception("The addon at $subdir has an unreadable addon.json file.");
+            }
 
+            $info = json_decode($addonJSON, true);
             if (empty($info)) {
-                throw new \Exception("The addon at $subdir has an empty info array.");
+                throw new \Exception("The addon at $subdir has non valid JSON in addon.json.");
```
initvector: > The addon at $subdir has invalid JSON in addon.json

Issue #5020, opened 2017-01-17
Remove deprecated methods from UpdateModel


DaazKu:
Put that in `2017-Q4-6` when it will exists and assign whoever you want :)

linc:
This duplicates https://github.com/vanilla/vanilla/issues/3055

PR #5735, opened 2017-07-07
Remove deprecated/dead code from UpdateModel

DaazKu:
Closes https://github.com/vanilla/vanilla/issues/5020

Remove deprecated and unused methods from UpdateModel


linc:
Doesn't this completely break the addons directory?

DaazKu:
You can actually update disable addons but I did find a reference to it in a [repository])https://github.com/vanilla/vanillainfrastructure/blob/c5172142d7587b1d445ef11052f861ad3f1be177/plugins/vfoptions/class.linkaddoninstaller.php#L21).
I will create an issue. Good call!

DaazKu:
I may have been too quick on the trigger with that one.
I was not aware that repo even existed.

PR #5736, opened 2017-07-07
Adapt OAuth2 class to allow for more flexible workflow.

nervoustwit:
Presently we make an API call to an authenticator to request the accessToken. In that call we return the entire results assigning it to a variable in the immediate scope. This PR would assign it to a protected variable in the class so that we could access other data being sent back.

This is necessary for some plugins that extend this class.


library/core/class.oauth2.php
```diff
@@ -373,9 +377,9 @@ public function settingsEndpoint($sender, $args) {
         $formFields = [
             'AssociationKey' =>  ['LabelCode' => 'Client ID', 'Description' => 'Enter the unique ID of the authentication application.'],
             'AssociationSecret' =>  ['LabelCode' => 'Secret', 'Description' => 'Enter the secret provided by the authentication provider.'],
-            'AuthorizeUrl' =>  ['LabelCode' => 'Authorize Url', 'Description' => 'Enter the endpoint to be appended to the base domain to retrieve the authorization token for a user.'],
-            'TokenUrl' => ['LabelCode' => 'Token Url', 'Description' => 'Enter the endpoint to be appended to the base domain to retrieve the authorization token for a user.'],
-            'ProfileUrl' => ['LabelCode' => 'Profile Url', 'Description' => 'Enter the endpoint to be appended to the base domain to retrieve a user\'s profile.']
+            'AuthorizeUrl' =>  ['LabelCode' => 'Authorize Url', 'Description' => 'Enter the endpoint to retrieve the authorization token for a user.'],
+            'TokenUrl' => ['LabelCode' => 'Token Url', 'Description' => 'Enter the endpoint to retrieve the authorization token for a user.'],
```
DaazKu: Why do we have 2 different fields that do the same thing???

library/core/class.oauth2.php
```diff
@@ -373,9 +377,9 @@ public function settingsEndpoint($sender, $args) {
         $formFields = [
             'AssociationKey' =>  ['LabelCode' => 'Client ID', 'Description' => 'Enter the unique ID of the authentication application.'],
```
DaazKu: Since the description describe the field we can probably remove all the "Enter the" in there.

library/core/class.oauth2.php
```diff
@@ -529,7 +533,7 @@ public function entryEndpoint($sender, $code, $state) {
         } elseif (empty($response['access_token'])) {
             throw new Gdn_UserException('The OAuth server did not return an access token.', 400);
         } else {
-            $this->accessToken($response['access_token']);
+            $this->accessToken(val('access_token', $response));
```
DaazKu: Why changing this one and not the one above?
Also we already have a check the access_token exists above. Calling val() is not necessary. No big deal though.

PR #5742, opened 2017-07-10
Use JWT for storing identity data in cookies

### LINKED ISSUE #5639: Overhaul Cookie Identity to implement JWT
Vanilla's pipe-delimited format for cookie identity is due for an upgrade. The JWT format seems like a good fit.

Some details, per @tburry 
> 1. Our cookies need to handle the old hashing algorithm and JWT side-by-side. No sessions should be invalidated.
> 1. Cookies with the old hashing algorithm should issue a Set-Cookie with the new JWT when encountered.
> 1. Hard code the JWT algorithm to h256 and require that algorithm on incoming cookies as an extra check. This avoids some of the more terrible JWT security holes.
> 1. Make the session setting mechanism pass an array with user ID and optional expiry information so that we can add other information to the cookie via plugins.
> 1. Make sure to use JWT's standard field names when available.
> 1. Make sure that no extraneous Set-Cookie headers are returned.

A good candidate for a PHP JWT library: [firebase/php-jwt](https://github.com/firebase/php-jwt). It has the features we'd need, no dependencies and looks simple to use.

Some light reading: [RFC 7519](https://tools.ietf.org/html/rfc7519)

initvector:
This update upgrades how a user's identity is stored in Vanilla. The current pipe-deliminated format is being removed in favor of using JWT.

Closes #5639


initvector:
Logging in and out in various scenarios should be tested, but more specifically, here's some highlights from the original spec:

- [ ] The legacy, pipe-delimited cookie format should still be supported.
- [ ] Incoming cookies using the legacy format should automatically be upgraded to the JWT format.
- [ ] The signing/checking process should be hardcoded to using HS256.
- [ ] Addons should be allowed to access and modify the payload data when setting/getting a user's identity.
- [ ] JWT standard claim names should be used, where applicable. See [Registered Claim Names](https://tools.ietf.org/html/rfc7519#section-4.1).
- [ ] No extraneous Set-Cookie headers should be sent.

library/core/class.cookieidentity.php
```diff
@@ -98,18 +106,41 @@ public function getIdentity() {
...
+                    $this->_clearIdentity();
+                }
+                break;
+            case self::VERSION:
+            default:
+                $payload = $this->getJWTPayload($name);
+                if ($payload !== null) {
+                    Gdn::pluginManager()
+                        ->fireAs(self::class)
+                        ->fireEvent('getIdentity', ['payload' => &$payload]);
```
DaazKu: Shouldn't that even be fire even when updating the the cookie too?

You could make setIdentity return the payload, loose the break and check is payload is not set before calling `$this->getJWTPayload($name);`.

linc:
When adding a Composer dependency, I'd include discussion about your evaluation of the package being included. Saves the step of the reviewer's eyes going wide as saucers and going to look at _their_ composer.json and asking obvious questions like "why did you choose this one?"

linc:
What was the value in keeping the old cookie format? We didn't want to create a 1-time logout event?

DaazKu:
> We didn't want to create a 1-time logout event?

Basically yes.

screen shot 2017-07-13 at 4 34 55 pm

DaazKu:
Also on the main issue:
> A good candidate for a PHP JWT library: firebase/php-jwt. It has the features we'd need, no dependencies and looks simple to use.

initvector:
> Why did you choose this one?

[jwt.io](https://jwt.io) lists out several libraries for many languages. The firebase library checked some key boxes for me:

1. It's one of the most popular libraries. It's less likely to become abandoned.
1. It has what we need in terms of signing and validating tokens with the required algorithm (HS256).
1. It doesn't have a lot of features we don't need. This means it doesn't have the bloat of a lot of other JWT libraries.
1. No dependencies. This is probably due to the lack of bloat.
1. [It's unit tested and utilizes continuous integration](https://travis-ci.org/firebase/php-jwt).

PR #5768, opened 2017-07-14
Fix ability to require email confirmation when no unconfirmed role exists

### LINKED ISSUE #4749: Confirm email setting with no unconfirmed role creates confusion

### LINKED ISSUE #5767: Saving a role after encountering an error when editing wipes out permissions
If you trigger a validation error when editing a role, the permission fields fail to reload with the rest of the form. If you don't notice this, or if you fix your original error and click "Save" anyway, the role is saved without any permissions.

### Steps to Reproduce
1. Edit a role.
1. Empty its name column.
1. Attempt to save.
1. Notice the permission fields do not appear.
1. Give the role a name.
1. Save again.
1. Return to edit the role.
1. Notice it has no permissions.

initvector:
There is some ambiguity regarding the "require email confirmation" setting and the requirement for an "Unconfirmed"-type role. It may not be clear to users that an unconfirmed role is required for enabling the setting. Similarly, if users enable the setting without an unconfirmed role, the default member permissions may be applied to the user. This negates the purpose of the confirmation setting.

This update does a few things:

1. Fixes an issue where saving a role after encountering an error caused the role's permissions to be wiped out.
1. Disables the form field for requiring email confirmation if no unconfirmed role is available. A warning is also included when the field is disabled.
1. Throws an error when you try to save the email confirmation config if there are no unconfirmed roles available.
1. Adds a check when saving a role to ensure the last unconfirmed role cannot have its type changed. If this happens, an error is displayed on the edit form.

Closes #4749
Closes #5767

Standard disclaimer: While this PR doesn't deal with checking permissions or setting them on users, it does still touch permissions. Please be thorough in reviewing.


applications/dashboard/controllers/class.settingscontroller.php
```diff
@@ -1584,6 +1591,10 @@ public function registration($RedirectUrl = '') {
             $this->ExistingRoleInvitations = arrayCombine($InvitationRoleIDs, $InvitationCounts);
             $ConfigurationModel->forceSetting('Garden.Registration.InviteRoles', $this->ExistingRoleInvitations);
 
+            if ($this->data('ConfirmationSupported') === false && $this->Form->getValue('Garden.Registration.ConfirmEmail')) {
+                $this->Form->addError('One unconfirmed role is required for email confirmation.');
```
linc: > A role with default type "unconfirmed" is required to use email confirmation.

linc: This also need to go in the locales repo, I believe.

applications/dashboard/views/settings/registration.php
```diff
@@ -13,7 +13,12 @@
         t('Email addresses are disabled.', 'Email addresses are disabled. You can only add an email address if you are an administrator.').
         '</div>';
 } ?>
-<div class="form-group">
+<?php if (!$confirmationSupported) {
+    echo '<div class="alert alert-warning padded-top">'.
+        t('No unconfirmed role available for email confirmation.', 'The site needs an unconfirmed role to use email confirmation. Please add an unconfirmed role to enable this setting.').
```
linc: > The site needs a role with default type "unconfirmed" to use email confirmation. Please add one to enable this setting.

Likewise, locales.

linc:
Otherwise, this looks fine.

initvector:
The cited messages have been updated. Master has been merged into this branch, as well, so conflicts have been resolved. Lastly, a PR has been made against the locales repo with the new error messages: https://github.com/vanilla/locales/pull/40

Issue #5547, opened 2017-05-08
Roll the HtmLawed plugin into core


tburry:
I think it's time we remove the HtmLawed plugin and just put its functionality into core. This will prevent the plugin from being removed which causes serious unexpected behavior.

This mirrors what we did with BBCode and Markdown and is immensely sensible.

x00:
+1

tburry:
Does that count as a +2 @x00?

x00:
Sure old skool :p

PR #5779, opened 2017-07-17
Roll HtmlawedPlugin into core

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/5547

- Remove the plugin
- Create the VanillaHtmlFormatter class
- Move `formatRssCustom()` into the render functions' file

Be sure to run composer update so have VanillaHtmlFormatter autoloaded by the container.


library/core/class.vanillahtmlawed.php
```diff
@@ -1,18 +1,13 @@
...
  */
 
-use Garden\Container\Container;
-
 /**
- * Class HTMLawedPlugin
+ * Class VanillaHtmlFormatter
  */
-class HtmLawedPlugin extends Gdn_Plugin {
+class VanillaHtmlFormatter {
```
initvector: This class is in a file named "class.vanillahtmlawed.php". The filename should be updated to reflect the class name.

Issue #3082, opened 2015-09-14
Add notifications for Moderation/Spam Queue


LauraMakesArt:
Option for Administrative roles to receive email notifications when items are sent to moderation.

//EDIT:

Version 1 of this task will be to create two new permission'd notifications:

* `Preferences.Email.ModerationQueue`
* `Preferences.Email.SpamQueue`

It will use the GroupBy / rollup functionality on the ActivityModel to keep multiple actions in the same day together. The notification will not include specifics, but rather be a generic message that there are new items in the X queue. It will track users opted in in the same manner as Advanced Notifications. 

No flood control would be included - if you have a high volume to queue traffic, you can obviously just check more frequently rather than waiting for an email if it's happening that often.

Version 2 of this feature would be:

* Set a per-user flag when a notification for X queue is set.
* Do not send notifications for X queue while the flag is set.
* Unset the flag when the user visits the queue.

This is beyond the scope of this task, because we'd like to make this core functionality to the ActivityModel when we revamp how notifications are handled.

DaazKu:
For that feature to make sense, especially for spam, I would probably do:

- A per user config in advanced notifications that would only appear / be effective if you have Garden.Moderation.Manage or Moderation.ModerationQueue.Manage
- These notifications would have a minimum interval that can be set in either minutes, hours, or days. That way you won't get spammed by emails if you have 60 notifications in a short internal. This will be especially handy for the spam queue.

adrianspeyer:
My main thing: As someone who gets a lot of notifications from his users about things that need to be moderated, I would love a solution where you are notified by email the first time a new discussion/comment gets added to the moderation queue since you were last logged in (active). This way you know you have something to review and you don't get an email for every new one.

DaazKu:
The "since last logged" part is tricky that's why I'm leaning toward a minimum delay (internal) between notifications.

linc:
I've added the tentative scope above for a version 1 of this after speaking with Todd. We're tabling this for the short term.

PR #5784, opened 2017-07-19
Add, per user, spam and moderation queue email notification settings

DaazKu:
Closes https://github.com/vanilla/vanilla/issues/3082

Add Moderation Notifications preferences to user's notification preferences.
Admins and Moderators can now be notified on incoming logs in the spam and moderation queues.

The notifications respect a delay by comparing against `User.DateLastActive` and then `$queue.LastNotification`. This prevents users from being notified for nothing if they log in after their `$queue.LastNotification` setting.

Original details: https://github.com/vanilla/vanilla/issues/3082#issuecomment-314208608


applications/dashboard/settings/class.hooks.php
```diff
@@ -854,4 +854,131 @@ public function gdn_upload_copyLocal_handler($sender, $args) {
...
+                        ->to($users[$userID]['Email'])
+                        ->message(sprintf(
+                            t('Go to your %s to review it.'),
+                            anchor(strtolower($queueName), url("/dashboard/log/$lcQueue", true))
+                        ))
+                        ->send()
+                    ;
+                } catch (Exception $ex) {
+                    // Welp...
+                    trace([
```
initvector: What about logging this, instead of recording a trace? It might be beneficial to keep a record of these failures for debugging.

DaazKu: Not really. The errors returned by PHPMailer are not super useful.
If mailing doesn't work it probably a general problem that will be noticed somewhere else.
It's also an error that will be generated like so: # mod/admin X # queue X interval so it can pollute the logs pretty quickly. That's why I did not bother with logging

DaazKu: TL;DR; having the trace and setting your interval to 1 minute is an easy way to have the exception on hand.

DaazKu:
> First and foremost, you've made a UI change with neither a design step, nor mentioning it / providing a screenshot of it in this PR. You're describing a purely technical change above but you've fundamentally altered an every-user interaction point with the product.

Ok.. So I have to get all UI change approved by Dan first??

> Second, you significantly expanded the scope of the original issue and I have no recollection of discussing this with you before it was committed to.

How did I expand the scope of the original issue?????????? Did you see the title and description of that issue??? There is nothing in there. I had to speak with multiple people, ask for opinions and read threads about it.

See: https://github.com/vanilla/vanilla/issues/3082#issuecomment-314208608

Also, July 10 in the support chat:
> 	Alexandre Chouinard	@here I'm about to implement notifications for the Spam/Moderation queue. Any comments or suggestions about this: https://github.com/vanilla/vanilla/issues/3082#issuecomment-314208608 ?	3:31 PM
Link	
Add notifications for Moderation/Spam Queue · Issue #3082 · vanilla/vanilla
Option for Administrative roles to receive email notifications when items are sent to moderation,
3:31 PM
Adrian Speyer	We had a chat on the staff forum about it: https://staff.vanillaforums.com/discussion/163/notification-of-items-in-moderation-queue	3:32 PM
Valérie Robitaille	Oh i am sure	3:32 PM
Alexandre Chouinard	I read it	3:33 PM
Adrian Speyer	cool	3:33 PM
Alexandre Chouinard	Thanks Adrian	3:33 PM
Adrian Speyer	:)	3:33 PM
Adrian Speyer	My main thing: As someone who gets a lot of notifications from his users about things that need to be moderated, I would love a solution where you are notified by email the first time a new discussion/comment gets added to the moderation queue since you were last logged in (active). This way you know you have something to review and you don't get an email for every new one.	3:35 PM
Adrian Speyer	I'll add the note to Github :)	3:36 PM
Alexandre Chouinard	
The "since last logged" part is tricky that's why I'm leaning toward a minimum delay (internal) between notifications.
3:39 PM
Alexandre Chouinard	Also put that on the task	3:39 PM
Alexandre Chouinard	interval* not internal

Thread dedicated to this issue: https://staff.vanillaforums.com/discussion/163/notification-of-items-in-moderation-queue

> Third, having a per-user knob/dial thing like this is very against the ethos of Vanilla. We make good design decisions; we don't offload them to the user to figure out in their kitchen sink control panel.

I did ask for input on this. I got some but it was mostly silent.
A per-user knob is what makes sense here since this notification setting, to be useful has to be fine tuned depending on the forum's load and the user's availability.

> The only path I see for moving this forward in the near term involves two things:
> 1. Standardize the UI of notifications opt-in, putting them at the end of the current list.

How is putting them at the end better? Do not forget that this is a mod only notification setting.
I would also like to point out that the category list can be quite long on certain forum and that's why I've put it at the top. It is static and small.

> 2. Pick a reasonable notification interval. Or, reconsider this approach to the logic entirely. At worst, we could look at global settings in the Dashboard, but I don't even favor that. What would make this "just work"? That's the challenge here, not the writing of code.

I did ask for opinions on the matter. The solution was implemented from the feedback I gathered on the subject.

DaazKu:
I do recognize that I've done a #$@%#$% job at explaining, in the PR description, any decision that has been made.

linc:
> Ok.. So I have to get all UI changes approved by Dan first??

No. UI changes need a design step if they don't fit neatly into an existing pattern (e.g. adding an item to a menu). That means adding a notification to that list doesn't need a review, but inventing a new form paradigm that exists nowhere else in user-facing UI certainly does. 

A design step at least involves running it by Todd or I, or bringing it up in an R&D meeting. We can tell you whether it requires Dan's input if you're in doubt.

> Did you see the title and description of that issue??? There is nothing in there.
> [...]
> I did ask for input on this. I got some but it was mostly silent.

You have a weekly meeting with me and a bi-weekly team meeting that includes the CPO. I agree the scope was insufficient in the original assignment and I understand your frustration about that. That doesn't change anything about this PR. Unresolved staff discussions and chats do not constitute formal guidance, nor do I understand how you're using it for justification after I've explicitly said it's incorrect as currently implemented.

linc:
> How is putting them at the end better? Do not forget that this is a mod only notification setting.
I would also like to point out that the category list can be quite long on certain forum and that's why I've put the new settings at the top. It is static and small.

I'm amenable to creating a special moderation section for notifications. I think it makes better sense to add them as a new section under the general notifications - they are additive, and not necessarily any more important than the others. My general opinion of UI changes from higher privileges is that, whenever possible, they shouldn't move the original UI elements.

If we do this, consider that the Flagging addon has its own moderator-only notification that should be adapted to this. We may have others internally.

PR #5785, opened 2017-07-20
Add ability to filter category dropdowns by allowed discussion type

### LINKED ISSUE #5337: Polish new discussion category dropdown for Ideas/Questions
-You can currently create discussions in idea categories if you make your thread from the Recent Discussions page and select the category with the drop down. This shouldn't be the case. 

-If you create an Idea from the recent discussion page via the button: 
image

It successfully isolates your idea category and only displays that as an option in the category dropdown. Most Q&A users also want this option for questions.

initvector:
Vanilla's categories have an `AllowedDiscussionTypes` attribute. One of the areas this attribute is used is by `NewDiscussionModule`. It is used by this module to determine if a particular discussion type should be allowed to be created in the current category. If it isn't allowed, the menu option for that discussion type is removed.

When creating a new discussion, the category dropdown control should be filtered using a similar method. If the current discussion type isn't allowed in a category, the category shouldn't be available as an option in the dropdown.

This update adds a new option value to `Gdn_Form::categoryDropDown`: DiscussionType. If this option is provided, its value is used to determine if the discussion type is allowed for each category. If it isn't, the category is not displayed in the dropdown.

Some notes on this fix:

1. It's opt-in. It's been added to core Vanilla's "add discussion" view. If an addon is overriding this view, its own call to `Gdn_Form::categoryDropDown` may need to be updated.
1. Addons can hook into this filtering by using the "AllowedDiscussionTypes" event in `CategoryModel`. This event passes the category's configured types, the category row and its permission category's row.
1. Vanilla's current implementation of AllowedDiscussionTypes is based on permissions. If a category doesn't seem to be filtering properly, check the `AllowedDiscussionTypes` of the category matching its `PermissionCategoryID`.

Closes #5337


library/core/class.form.php
```diff
@@ -496,17 +495,25 @@ public function categoryDropDown($FieldName = 'CategoryID', $Options = []) {
...
             $SafeCategoryData[$CategoryID] = $Category;
         }
+        unset($discussionType, $permissionCategory, $allowedDiscussionTypes);
 
         unset($Options['Filter'], $Options['PermFilter'], $Options['Context'], $Options['CategoryData']);
 
         if (!isset($Options['class'])) {
             $Options['class'] = $this->getStyle('dropdown');
-        } else {
-            $Options['class'] = $this->translateClasses($Attributes['class']);
```
linc: Adding rando stuff to your PR = ಠ_ಠ

Issue #5725, opened 2017-07-05
Make Htmlawed's class whitelist extensible via addon


linc:
Prerequisite: https://github.com/vanilla/vanilla/issues/5547

The `$allowedClasses` list should be made modifiable via addon.

DaazKu:
I think we should only allow addition and not modification.

PR #5786, opened 2017-07-21
Add a way to defined extra allowed classes in VanillaHTMLFormatter

DaazKu:
Closes https://github.com/vanilla/vanilla/issues/5725

~~Fire 'VanillaHtmlFormatter_init' event when an instance is created.~~
This allows addons to add extra allowed classes through `addExtraAllowedClasses()`

Edit:

To test you can add this to a plugin:
```
    /**
     * @param Container $container
     */
    public function container_init_handler($container) {
        $container->get(VanillaHtmlFormatter::class)->addExtraAllowedClasses(['test']);
    }
```

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -179,6 +179,7 @@ $dic->setInstance('Garden\Container\Container', $dic)
 
     ->rule('HtmlFormatter')
     ->setClass(VanillaHtmlFormatter::class)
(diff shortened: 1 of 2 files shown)


DaazKu:
I spoke with Todd a bit more and he suggested that we ditch the dependency on the event manager and make people use the container_init event to add the extra classes to the object.

I will try to do that instead.

PR #5790, opened 2017-07-24
Fix Typo in Twitter Emoji Set

R-J:
Added also a new alias so that changing the file name doesn't break existing `:dissapointed:` emojis


linc:
Well handled, thanks. One knit-pick for the future is that it's helpful to leave a comment for why things like a misspelled alias are in the list. :)

PR #5792, opened 2017-07-24
Fix discussion URLs in advanced notification emails

initvector:
This fix updates how URLs are generated for new discussion alerts with advanced notifications. Currently, the URL is generated with `discussionUrl`, setting its `$withDomain` to `false`, then it is parsed again with `externalUrl` to generate a full URL for use in emails. Both of these function calls use the `url` function and both of them add the site's web root. It is currently possible for a site's web root to be doubly added to discussion URLs.

This update sets the `$withDomain` parameter to "/" during the call to `$discussionUrl`. This returns the Vanilla path, only. The web root is not included. When the string reaches `externalUrl`, the site's web root is only added the single time, so the URLs should be valid under normal usage.

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -976,7 +976,7 @@ class PostController extends VanillaController {
             'HeadlineFormat' => $HeadlineFormat,
             'RecordType' => 'Discussion',
             'RecordID' => $DiscussionID,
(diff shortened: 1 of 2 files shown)


initvector:
The error addressed here seems to have been introduced in https://github.com/vanilla/vanilla/pull/5734. We need to verify this PR doesn't introduce a regression.

PR #5798, opened 2017-07-28
Improvements to API v2

initvector:
This update does several things towards bettering API v2:

1. Adds CRUD endpoints for categories.
1. Adds CRUD endpoints for roles.
1. Adds CRUD endpoints for users.
1. Improves unit test compatibility for records with unique values (e.g. names and emails with users, URL slugs with categories).
1. Adds automated testing for category, role and user endpoints.
1. Makes some minor tweaks to existing code for the sake of readability.

In addition, some updates to other areas of the code base were necessary. These were primarily done for the sake of improving automated testing:

1. Updates `Gdn_DirtyCache::increment` and `Gdn_DirtyCache::decrement` to actually do something.
1. Fixes an issue in `CategoryCollection::flushCache` where the cache revision was not being reset, so the category cache couldn't be updated in the same "request".
1. Fixes an issue in `CategoryModel::rebuildTree` where the shared instance's collection might not be reset.

Configuring permissions for roles and categories is not part of this update. This capability will be added separately.

DELETE, GET (including GET /<id>/edit), POST, PATCH should now be operational for comments, discussions, categories, users and roles. The users controller supports PUT for the banned and verified flags.

### Testing

I recommend testing with a GUI like [Postman](https://www.getpostman.com/). Additionally, you'll likely need to send along an access token in an authorization header with your request. You can find out more in the [Authentication](http://docs.vanillaforums.com/developer/framework/apiv2/authentication) of Vanilla's Writing APIs docs.

The input schemas should let you know which parameters are possible to use and where. Only PATCH, POST and PUT support a (JSON-encoded) request body. Only POST can be used without a resource ID in the URL.


applications/dashboard/controllers/Api/RolesApiController.php
```diff
@@ -0,0 +1,288 @@
...
+    public function delete($id) {
+        $this->permission('Garden.Settings.Manage');
+
+        $in = $this->idParamSchema()->setDescription('Delete a role.');
+        $out = $this->schema([], 'out');
+
+        $this->roleByID($id);
+        $this->roleModel->deleteID($id);
+
+        return null;
```
DaazKu: This return statement is useless.
screen shot 2017-07-28 at 10 34 58 am

DaazKu: I don't think it improves readability. On the contrary if I see return null at the end of a function my first reaction would be to check for another return statement in that function since the default return is null.

applications/dashboard/controllers/Api/RolesApiController.php
```diff
@@ -0,0 +1,288 @@
...
+
+    /**
+     * RolesApiController constructor.
+     *
+     * @param RoleModel $roleModel
+     */
+    public function __construct(RoleModel $roleModel, PermissionModel $permissionModel) {
+        $this->roleModel = $roleModel;
+        $this->permissionModel = $permissionModel;
+        $this->caseScheme = new CapitalCaseScheme();
```
DaazKu: What is the rational on passing some some object through dependency injection not doing it for some others?

applications/dashboard/controllers/Api/RolesApiController.php
```diff
@@ -0,0 +1,288 @@
...
+        return null;
+    }
+
+    /**
+     * Get a schema instance comprised of all available role fields.
+     *
+     * @return Schema Returns a schema object.
+     */
+    protected function fullSchema() {
+        $schema = Schema::parse([
```
DaazKu: We should probably set the $schema variable as static to prevent calling Schema::parse on every call to fullSchema.

Also, that function has been created on every APIController. Maybe we should move it's definition in a parent class.

One solution I'm seeing here is to create an abstract function defineFullSchema() in the parent class. (which would replace this one. Then in the parent constructor do this assignation:
```
$this->fullSchema = $this->defineFullSchema();
```

Every child class will then be able to use $this->fullSchema.
This will also allow to get rid of functions like: `roleSchema()` since we will be able to use `$this->fullSchema` in `schema()` (if I'm not wrong)

DaazKu: Actually we may need to create a different function than `schema()` but it can go on the parent class instead of being defined on all children.

DaazKu:
There's an inconstancy on how functions are named. Most of the new classes are using getX setX but the controllers don't. We should stick with one way of doing things.

applications/dashboard/controllers/Api/UsersApiController.php
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Get a schema instance comprised of all available user fields.
+     *
+     * @return Schema Returns a schema object.
+     */
+    protected function fullSchema() {
+        $schema = Schema::parse([
+            'userID:i' => 'ID of the user.',
+            'name:s' => 'Name of the user.',
+            'password:s' => 'Password of the user.',
```
DaazKu: I'm not sure we should reveal this, from a security perspective.

applications/dashboard/controllers/Api/UsersApiController.php
```diff
@@ -0,0 +1,358 @@
...
+            'hashMethod:s' => 'Hash method for the password.',
+            'email:s' => 'Email address of the user.',
+            'photo:s' => [
```
DaazKu: This field is not properly transformed before being outputted.

screen shot 2017-08-07 at 3 27 45 pm

It should be a full URL.

applications/dashboard/controllers/Api/UsersApiController.php
initvector: I'm not attached to keeping it. We can remove it, altogether, or keep it only when a user has `Garden.Settings.Manage` permissions (although I feel like this makes things a little inconsistent and should probably be avoided).

DaazKu: This is something that should remain private unless you have access to the database itself since it could allow a moderator to bruteforce easy passwords for some users and then gain access to their email account is they used the same password.

linc:
1. `prepareData()` > `massageData()` because it implies both order & specific action.
2. `$userData` > `$data` because it gives needed context. Even `$row` means something to us. `$data` could be any variable in existence.

applications/dashboard/controllers/Api/RolesApiController.php
linc: @tburry

applications/dashboard/controllers/Api/RolesApiController.php
linc: I'm amenable to adding the static variable but I'm not sure it'll be helpful.

I think abstracting this is premature since we don't know if it's going to be on every controller or not, yet. I feel like that's something we can refactor at the end if it makes sense rather than trying to guess at it mid-stream, especially when I'm not sure it'll help clarity at this stage.

initvector: It's been in all the API controllers so far, because they've all been resource controllers. It isn't necessarily a requirement for all API controller endpoints. If there isn't one full, standard schema that we then want to make subsets of, this functionality isn't necessary (e.g. analytics).

applications/dashboard/controllers/Api/RolesApiController.php
initvector: If memory serves, this was done to make the constructor a little less unwieldy. Injecting `CapitalCaseSchema` didn't seem as beneficial.

```php
CommentsApiController::__construct(CommentModel $commentModel, DiscussionModel $discussionModel, UserModel $userModel, CapitalCaseScheme $caseScheme)
```

becomes

```php
CommentsApiController::__construct(CommentModel $commentModel, DiscussionModel $discussionModel, UserModel $userModel)
```

linc:
I tend to agree about the lack of consistency on getter / setter functions but I think Ryan was following @tburry's template here, so I'd wanna know if he did that on purpose or not (and why).

applications/dashboard/controllers/Api/UsersApiController.php
initvector: Fixed with 1ace2d9

applications/dashboard/controllers/Api/UsersApiController.php
initvector: Fixed with 1ace2d9

initvector:
> prepareData() > massageData() because it implies both order & specific action.
> $userData > $data because it gives needed context. Even $row means something to us. $data could be any variable in existence.

Done and done.

applications/dashboard/controllers/Api/RolesApiController.php
DaazKu: We should probably create an abstraction for resource controllers then.
I'm not against waiting a bit as Linc said though.

I will create a an issue about it if this keeps happening.

Issue #5780, opened 2017-07-18
Admins deleting a user account should leave a trace in the logs


DaazKu:
This doesn't need to show in the change logs (since it won't be restorable) but it would be more than handy to have a trace of which administrator deleted the user.
This will help giving better insight on whom deleted a user when we are asked to.

Currently our answer is: We don't know which of the admin did it. Sorry.

Information that should be logged:
- UserID of the admin that deleted the other user.
- UserID, Name and Email of the deleted user (Prior of us changed the name and email value)

PR #5808, opened 2017-07-31
Preserve useful informations when deleting a user

DaazKu:
Closes https://github.com/vanilla/vanilla/issues/5780

To help diagnose deleted users problem the following improvements have been done when a user is deleted:

- Preserve the old name and email of the user
- Log the UserID of the admin/mod that deleted the user. 

Note: I decided to use the user's attribute since activities are cleaned after a while and wanted a permanent trace. It was also unnecessary to put that in the log table since we do not want a deleted user to be restored.


DaazKu:
Do you think we should/could keep the hashed email then?
That way it would still possible to find a user record from remembering the email but we wouldn't have the email per say.

linc:
I think no.

https://www.ftc.gov/news-events/blogs/techftc/2012/04/does-hashing-make-data-anonymous
https://blog.littledata.io/2016/08/03/personally-identifiable-information-pii-hashing-and-google-analytics/
https://adexchanger.com/ad-exchange-news/burning-a-hole-in-the-invincibility-of-email-hashing/

As a matter of privacy, deletion should mean deletion without the ability to do forensics like this.

DaazKu:
These references all speak about hashed email exchanged between third parties.
In this case, the hashed email would be private and not exposed in any part of the forum.

Unless a user specifically request his data from being deleted from the forum (which is rarely the case) conserving the email/username should be fine.

I would like to keep them by default but have a way to specifically state that they should not be kept. As a protection, the latter would require some kind of privileges.
We would also add a warning in the UI stating that the email and username would be kept.

What do you think about that?

tburry:
You would generally have a "purge" option after the delete to do the final delete. This would be analogous to deleting a file to the trash can and then emptying the trash can to finally delete it.

This being said, we have other things to do right now so just deleting the PII for now is fine.

DaazKu:
I'll create a task for the purge thing and update this PR to only add the mod/admin UserID.

kaecyra:
Talking with Todd, it seems like the Log table is actually the best place for this. I think we *do* want to be able to restore deleted users (why would we not?), and the log allows us to purge things from it as well, which satisfies the PII concerns.

tburry:
@kaecyra and I were just talking and if deleted users go into GDN_Log then there shouldn't be an issue.

- The admin that deleted the user is the InsertUserID of the log.
- The user's information is backed up in the log.
- The log can be purged for privacy.

DaazKu:
That's an overhaul of our current system.

screen shot 2017-08-18 at 10 34 29 am
screen shot 2017-08-18 at 10 35 43 am

DaazKu:
I created an issue to answers all the question related to the "user delete" process overhaul and put the username and mod/admin userID in the deleted record for now.

PR #5811, opened 2017-07-31
Apply feedback from Scrutinizer's informational recommendations

initvector:
Scrutinizer gives code feedback. This update applies most of the current "informational" suggestions.

### DIFF
--- a/applications/dashboard/modules/class.pagermodule.php
+++ b/applications/dashboard/modules/class.pagermodule.php
@@ -64,14 +64,14 @@ class PagerModule extends Gdn_Module {
     public $Offset;
 
     /** @var int The last offset of the current page. (ie. Offset to LastOffset of TotalRecords). */
-    private $_LastOffset;
+    protected $_LastOffset;
 
     /**
      * @var bool Certain properties are required to be defined before the pager can build
      * itself. Once they are created, this property is set to true so they are
      * not needlessly recreated.
      */
-    private $_PropertiesDefined;
+    protected $_PropertiesDefined;
 
     /**
      * @var bool A boolean value indicating if the total number of records is known or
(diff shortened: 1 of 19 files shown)


applications/dashboard/modules/class.pagermodule.php
```diff
@@ -64,14 +64,14 @@ class PagerModule extends Gdn_Module {
     public $Offset;
 
     /** @var int The last offset of the current page. (ie. Offset to LastOffset of TotalRecords). */
-    private $_LastOffset;
+    protected $_LastOffset;
```
DaazKu: Why going from private to protected? We want attributes to be private by default.

plugins/Facebook/class.facebook.plugin.php
```diff
@@ -13,7 +13,7 @@ class FacebookPlugin extends Gdn_Plugin {
     const API_VERSION = '2.7';
 
     /** Authentication table key. */
-    const ProviderKey = 'Facebook';
+    const PROVIDER_KEY = 'Facebook';
```
DaazKu: I did not find any reference to ProviderKey so I'm inclined to accept that change but that's a bit risky.

applications/dashboard/modules/class.pagermodule.php
initvector: `MorePagerModule` extends `PagerModule`, but was including its own private `$_LastOffset` property. To ensure child classes can share the property, the access specifier was downgraded from private to protected.

plugins/Facebook/class.facebook.plugin.php
initvector: I've only found references to the `ProviderKey` class constants in older social plugins: Facebook, Google Plus and Twitter. In each case, the constant is only ever used internally within the class. That isn't to say that something, somewhere out there, isn't using these class constants for other purposes. However, it feels like a *relatively* safe change.

applications/dashboard/modules/class.pagermodule.php
DaazKu: I missed that haha. Scrutinizer is great!

PR #5812, opened 2017-08-01
Fix guests' inability to switch to full site from mobile theme

DaazKu:
~~This allows a forum behind Varnish to work properly (Cookies are usually stripped for guests which means that no transient key is generated)~~

Actually guests don't get cookies.

### DIFF
--- a/applications/dashboard/controllers/class.profilecontroller.php
+++ b/applications/dashboard/controllers/class.profilecontroller.php
@@ -541,7 +541,7 @@ class ProfileController extends Gdn_Controller {
      * - 1: Unset the force cookie and use the user agent to determine the theme.
      */
     public function noMobile($type = 'desktop') {
-        // Only validate CSRF for non guests.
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/class.profilecontroller.php
```diff
@@ -541,7 +541,14 @@ public function invitations($userReference = '', $username = '', $userID = '') {
      * - 1: Unset the force cookie and use the user agent to determine the theme.
      */
     public function noMobile($type = 'desktop') {
-        if (!Gdn::request()->isAuthenticatedPostBack(true)) {
+        // Only validate CSRF for non guests.
```
initvector: I'd expand on this, just a little, so we remember *why* we only check authenticated users (e.g. "Only validate CSRF token for authenticated users because guests do not get the transient key cookie.")

DaazKu: That's not true for open source. My hope here is that we can use the blame function of github and track down this PR.

PR #5819, opened 2017-08-02
Add parameter to apply flood control or not to moderators/admin

nervoustwit:
Backport of PR #5698 to 2017-Q2-6


linc:
I missed this in the original, but for future reference: Don't replace instances of `Gdn::session()` with a local variable. We avoid that unless it's gonna be used a ton.

PR #5822, opened 2017-08-03
Increase the complexity of email confirmation tokens

### LINKED ISSUE #5817: Increase complexity of email confirmation token
The email confirmation token is currently an 8-character token from a 36-character pool (upper + digits). At 1000 requests per second, that's an average brute force time of 4 years. Significantly more resources could bring this down to _perhaps_ (if the target can accommodate many multiples of that in requests without triggering a DDoS warning) a week. This is still a poor vector however, because it's much simpler to Mechanical Turk additional user registrations, for which we have no real code-based defense in the product.

Because it's trivially simple to increase the complexity into the "impossible" range, we should do so. I suggest a 12-character token and expanding the pool to include lowercase letters.

initvector:
This update alters email confirmation tokens by increasing them from eight (8) characters to thirty-two (32) and including lowercase letters. This increases possible combinations from 36^8 to 62^32.

Closes #5817

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -72,6 +72,16 @@ class UserModel extends Gdn_Model {
     }
 
     /**
+     * Generate a random code for use in email confirmation.
+     *
+     * @return string
+     */
+    private function confirmationCode() {
+        $result = betterRandomString(32, 'Aa0');
+        return $result;
+    }
+
+    /**
      * Whether or not we are past the user threshold.
      *
      * This is a useful indication that some database operations on the User table will be painfully long.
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -72,6 +72,16 @@ public function __construct() {
     }
 
     /**
+     * Generate a random code for use in email confirmation.
+     *
+     * @return string
+     */
+    public function confirmationCode() {
```
DaazKu: This should probably be private.

linc:
For the record, I don't think setting user-facing tokens to impossible-to-type lengths is a good idea. The 32-character length does not increase security (it was already impossible-to-brute-force at my suggested 12 characters) but what you did accomplish is make this link nigh-impossible for a user to manually type. I know that doesn't seem like an obvious or common scenario, but people have uncommon needs sometimes depending on their device or accessibility requirements. Making changes that have a definable downside and no definable upside (except perhaps scratching some nerd itch to just "make everything 32!") is sub-par design.

DaazKu:
I agree that it is harder to manually type but now it is really bruteforcing proof even if the token would start with "aaa". Since this is a one time thing and that copy pasting exists I'm pretty sure that this is not a problem at all.

Issue #5591, opened 2017-05-16
Reverse engineer /entry/connect functionalities with SSO


DaazKu:
We need documentation on how entry/connect works.
Especially what are the configurations effects and their interaction with the registration types.

Configs:
- `Garden.Registration.AllowConnect`
- `Garden.Registration.AutoConnect`
- `Garden.Registration.ConnectSynchronize`
- `Garden.Registration.EmailUnique`
- `Garden.Registration.NameUnique`
- `Garden.Registration.SendConnectEmail`
- `Garden.SSO.SyncRoles`
- `Garden.SSO.SyncRolesBehavior`

RegistrationMethods:
- Closed
- Basic
- Captcha
- Approval
- Invitation
- Connect

Refer back to https://github.com/vanilla/vanilla/issues/5563#issuecomment-301902989 for more information.

DaazKu:
- `Garden.Registration.EmailUnique`
  - Value: (bool)
  - Default Value: true
  - Description: Specify whether Emails are unique or not.
- `Garden.Registration.NameUnique`
  - Value: (bool)
  - Default Value: true
  - Specify whether Names are unique or not.
- `Garden.Registration.AllowConnect`
  - Value: (bool)
  - Default Value: true
  - Description: Allow SSO connections to link a VanillaUser to a ForeingUser.
  - Note: Setting this to false will force users to create new accounts when connecting through SSO if there is no link already created between the ForeingUser and a VanillaUser.
- `Garden.Registration.AutoConnect` && `Garden.SSO.AutoConnect`
  - Value: (bool)
  - Default Value: true
  - Description: Will automatically try to link users using the provided Email address if the SSO provider is considered to be a "Trusted" authority.
  - Note: 
    - Depends on `Garden.Registration.AllowConnect=true` and `Garden.Registration.EmailUnique=true`
    - `Garden.SSO.AutoConnect` seems to be kind of deprecated.
- `Garden.Registration.ConnectSynchronize`
  - Value: (bool)
  - Default Value: true
  - Description: Update all the user informations (Name, Email, Title, ...)
- `Garden.SSO.SyncRoles`
  - Value: (bool)
  - Default Value: false
  - Description: Tell whether roles needs to be synched when connecting over SSO
- `Garden.SSO.SyncRolesBehavior`
  - Value: (string|bool)
  - Default Value: false
  - Description: Affect `Garden.SSO.SyncRoles` behaviour by only synching roles when you register (first login) if sets to 'register'.
[...]

PR #5831, opened 2017-08-09
Add authentication over API v2

DaazKu:
Depends on ~~#5837~~, ~~#6007~~

# Notes
`Garden.SSO.SyncRolesBehavior === 'register'` has been ditched in favour of the new configuration `Garden.SSO.SyncRolesOnRegistrationOnly`. See [configurations](https://github.com/vanilla/vanilla/issues/5591#issuecomment-317837904).

# TODO

AuthenticateController:
- [x] AutoConnect (if `Garden.Registration.AutoConnect === true`)
    - [x] Allow to auto link existing user by email
- [x] Allow to create a new user automatically if no conflict is detected and no information is missing
- [x] AllowConnect (if `Garden.Registration.AllowConnect === true`)
    - [x] Allow to link to an existing user if the password is correct
         - [x] Preselected list of user
         - [x] User that matches a specific name + email
    - [x] Allow to create a new user manually

AuthenticateAPIController:
- [x] DELETE /authenticate/{authenticator}/{authenticatorID} Delete a link between a user and an authenticator
- [x] DELETE /authenticate/session/{authSessionID} Delete an authentication session
- [x] GET /authenticate/{authenticator}/{authenticatorID} Get a link between a user and an authenticator
- [x] GET /authenticate/session/{authSessionID} Get an authenticate session info. They auto expire after 20 minutes.
- [x] POST /authenticate/{authenticator}/{authenticatorID} Authenticate a user and create a link if there is none already.
- [x] POST /authenticate/link-user/
    -  authSessionID + ((userID || (username + email)) + password

*Note that an authenticator cannot be called "sessionAuthenticator"

# How to test

- Create a custom branch that is a merge of this PR + #6009
- [Install and setup VanillaConnect](https://github.com/vanilla/vanilla/pull/6009).
    - VanillaConnect is the only plugin that has an authenticator.
- From there you want to use VanillaConnect to test that everything is working.
    - I suggest that you create a bunch of users to test the autoconnect feature.

# What to test

- [ ] Log in with a brand new user that has no conflict with Vanilla's user.
- [ ] Log in with an already linked user (log out and log in as the previous user)
- [ ] Log in with a new user that has an email/username clash with a Vanilla's user.
    - [ ] Link to that user using the drop down + password
    - [ ] Link to that user using username + email + password
    - [ ] Create a brand new user.
- Try to sync user roles
    - [ ] With a brand new user
    - [ ] With an existing user

Play with the [configurations](https://github.com/vanilla/vanilla/issues/5591#issuecomment-317837904) to make sure that everything is respected


library/Vanilla/Authenticator/Authenticator.php
```diff
@@ -0,0 +1,74 @@
...
+    public final function getID() {
+        return $this->authenticatorID;
+    }
+
+    /**
+     * Default getName implementation.
+     *
+     * @return string
+     */
+    private function getNameImpl() {
```
initvector: Is this necessary as its own function? Couldn't this just be the body of `getName` and child classes could overwrite it as necessary? If they really needed this exact routine, they could call the parent's method.

library/Vanilla/Authenticator/SSOAuthenticator.php
```diff
@@ -0,0 +1,96 @@
...
+     * @var bool
+     */
+    private $isTrusted = false;
+
+    /**
+     * Authenticator constructor.
+     *
+     * @param string $authenticatorID Currently maps to "UserAuthenticationProvider.AuthenticationKey".
+     */
+    public function __construct($authenticatorID) {
```
initvector: You should be able to ditch this, since all your doing is calling the parent constructor.

library/Vanilla/Authenticator/SSOAuthenticator.php
```diff
@@ -0,0 +1,96 @@
...
+        return $this->isTrusted;
+    }
+
+    /**
+     * Setter of isTrusted.
+     *
+     * @param bool $isTrusted
+     */
+    protected function setTrusted($isTrusted) {
+        $this->isTrusted = $isTrusted;
```
initvector: It might not be a bad idea to force this to a boolean value or test it and throw an exception if it isn't.

library/Vanilla/Authenticator/Authenticator.php
```diff
@@ -0,0 +1,74 @@
...
+
+        if (array_pop(explode('\\', static::class)) !== $this->getNameImpl(static::class).'Authenticator') {
+            throw new \Exception('Authenticator class name must end with Authenticator');
+        }
+    }
+
+    /**
+     * Validate an authentication by using the request's data.
+     *
+     * @throw Exception Reason why the authentication failed.
```
initvector: Typo in the tag:`@throws`. There's a few more like this too.

library/Vanilla/Models/SSOInfo.php
```diff
@@ -0,0 +1,105 @@
...
+     * @param array $associativeArray
+     * @throws \Exception If a non associative array is passed to the constructor.
+     */
+    public function __construct(array $associativeArray = []) {
+        if (array_key_exists(0, $associativeArray)) {
+            throw new \Exception(__CLASS__.' can only be initialized with an associative array.');
+        }
+
+        foreach($associativeArray as $name => $value) {
+            $this->$name = $value;
```
initvector: Instead of dynamically declaring a bunch of properties on your object, how about keeping them in a dedicated data array that you define on the class? I know it probably won't mesh well with your intentions for using `ArrayAccessTrait`, but I think we should stay away from dynamically-declared properties, especially when they're arbitrary.

library/Vanilla/Models/SSOInfo.php
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Getter with default value.
+     * Deprecate when we can use the coalesce operator.
+     *
+     * @param $name
+     * @param $default
+     * @return mixed The extra information or the default if not found.
+     */
+    public function coalesce($name, $default = null) {
```
initvector: Can you use `get` here, instead? That's pretty much the standard in Vanilla.

library/Vanilla/Models/SSOInfo.php
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @throw \Exception If the validation fails.
+     */
+    public function validate() {
+        $required = ['authenticatorName', 'authenticatorID', 'authenticatorIsTrusted', 'uniqueID'];
+
+        foreach ($required as $name) {
+            // Empty is used on purpose here. Nothing should be of empty value in there.
+            if (empty($this->$name)) {
+                throw new \Exception("SSOInfo's '$name' is required.");
```
initvector: If there's any way to collect these error messages and then throw the exception with all of them, I recommend doing that. It's always a pain trying to something and getting one-at-a-time errors until it actually goes through. In my experience, it's always been better to get a full list of errors and then report them, all at once.

The `Validation` class and `ValidationException` exception in the garden-schema library might come in handy here. They're used in a similar fashion in `AddonModel` and `InstallModel`.

library/Vanilla/Models/SSOInfo.php
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @package Vanilla\Models
+ */
+class SSOInfo implements \ArrayAccess {
+    use ArrayAccessTrait;
+
+    /**
+     * Maps to "GDN_UserAuthenticationProvider.AuthenticationSchemeAlias"
+     *
+     * @var string
```
initvector: If it's possible to fit a variable's doc on a single line, you should stick with the one-line syntax.

```php
/** @var string Maps to "GDN_UserAuthenticationProvider.AuthenticationSchemeAlias" */
```

library/Vanilla/Models/SSOModel.php
```diff
@@ -0,0 +1,319 @@
...
+        $name = $ssoInfo->coalesce('name');
+
+        if (!$email && !$name) {
+            return false;
+        }
+
+        $userInfo = [
+            'Name' => $name,
+            'Email' => $email,
+            'Password' => randomString('32', 'Aa0!'),
```
initvector: It'd be better if the length was a class constant.

initvector: `randomString` just passes things through to `betterRandomString`, so let's go with that.

library/core/class.authenticationprovidermodel.php
```diff
@@ -163,6 +163,28 @@ public static function getProviderByScheme($authenticationSchemeAlias, $userID =
...
     }
 
+
+    /**
+     * Get a list of providers by (read type) scheme.
+     *
+     * @param $authenticationSchemeAlias
+     * @return array
+     */
+    public function getProvidersByScheme($authenticationSchemeAlias) {
```
initvector: Since we're already in the "Authentication Provider" model, this should probably be shortened to `getByScheme`.

library/Vanilla/Models/SSOInfo.php
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Class SSOInfo
+ *
+ * This is a data object containing information returned by an SSOAuthenticator.
+ * Adding methods to this object probably means that you are doing something wrong.
+ *
+ * @package Vanilla\Models
+ */
+class SSOInfo implements \ArrayAccess {
```
initvector: This is probably me nitpicking, but the name of this class bugs me. I believe "data" is more commonly used in Vanilla. References to this class (and even it's description) refer to its contents as "data" too. I recommend either renaming it to "SSOData" or possibly even renaming it to "Data" and putting it under the "Vanilla\Models\SSO" namespace. The latter is a similar practice to what we see done with `Garden\Web\Data`.

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+use Vanilla\AddonManager;
+use Vanilla\Authenticator\Authenticator;
+use Vanilla\Authenticator\SSOAuthenticator;
+use Vanilla\Models\SSOModel;
+use Vanilla\Utility\CamelCaseScheme;
+
+/**
+ * API Controller for the `/authenticate` resource.
+ */
+class AuthenticateApiController extends AbstractApiController {
```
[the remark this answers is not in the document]
DaazKu: I thought about it too but they are "authentication sessions".
There's also no point in allowing to create an authentication session manually and in fact that would be dangerous.

That's why they are on the controller. /authenticate/session really makes sense here.

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+     * Unlink a user from the specified authenticator.
+     * If no user is specified it will unlink the current user.
+     *
+     *
+     * @param string $authenticator
+     * @param string $authenticatorID
+     * @param array $query The query string as an array.
+     * @throws Exception
+     */
+    public function delete($authenticator, $authenticatorID = '', array $query) {
```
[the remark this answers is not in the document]
DaazKu: I don't know. $query is injected by the container using reflection so it is possible to have the seconds parameter being optional in that case.

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+            $userID = $query['UserID'];
+        } else {
+            $this->permission('Garden.SignIn.Allow');
+            $userID = $this->getSession()->UserID;
+        }
+
+        $authenticatorInstance = $this->getAuthenticator($authenticator, $authenticatorID);
+
+        $data = [];
+        $this->userModel->getDelete(
```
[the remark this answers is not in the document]
DaazKu: That function is weird. I'm just using it.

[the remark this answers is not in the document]
DaazKu: [3:23 PM] Alexandre Chouinard: Long story short, that's the only way I found to delete a record from Gdn_UserAuthentication by using a model
[3:24 PM] Alexandre Chouinard: Someone had the great idea to not create a Gdn_UserAuthentication model and put some utility function in Gdn_User instead

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+        );
+    }
+
+    /**
+     * Delete a session.
+     *
+     * @param string $authSessionID
+     */
+    public function delete_session($authSessionID) {
+        $this->permission();
```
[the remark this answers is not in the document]
DaazKu: I don't think so. Sessions are created by guest and this is a convenience method that let you kill the session right away if you are not going to use it. (before it expires)

Knowing the sessionID is the validation here. At worst you are just gonna expire an authentication session if you can guess a sessionID.

I'll put `Garden.SignIn.Allow`

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+            }
+
+            $userID = $userResults[0]['UserID'];
+        }
+
+        $tmp = ['UserID' => $userID];
+        $this->userModel->expandUsers($tmp, ['UserID']);
+        $user = $tmp['User'];
+
+        $passwordHash = new Gdn_PasswordHash();
```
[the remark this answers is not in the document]
DaazKu: Same as the `caseScheme()`.

If we every need to use it in multiple functions we can make it a class attribute.
This is also an endpoint that won't be hit often.

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+                'authenticatorIsTrusted:b' => 'If the authenticator is trusted to sync user\'s information.',
+                'uniqueID:s' => 'Unique ID of the user supplied by the provider.',
+                'email:s?' => 'Email of the user.',
+                'name:s?' => 'Name of the user.',
+                'roles:a?' => [
+                    'description' => 'One or more role name.',
+                    'items' => ['type' => 'string'],
+                    'style' => 'form',
+                ],
+                '...:s?' => 'Any other information.',
```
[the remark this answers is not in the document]
DaazKu: This is just a placeholder for the documentation.
I could remove the type I guess.

library/core/class.authenticationprovidermodel.php
DaazKu: I'm following the current convention. Check the method above.

library/Vanilla/Models/SSOModel.php
DaazKu: It is used in one place and is not relevant (the hashmethod is random so any arbitrary number will do). No need to expose this as a class constant.

library/Vanilla/Models/SSOInfo.php
DaazKu: This object is meant to be used like an array and this function name does exactly what it says. (It is clearer than "get")
It will be removed when we stop supporting php 5.6 (early next year)

library/Vanilla/Models/SSOInfo.php
DaazKu: That's actually how I did it first but that's a no go. This is an array like object and the fact that it is an object should be transparent.

Due to the fact that this is an array like object, dynamically declared properties is perfectly fine.

library/Vanilla/Authenticator/SSOAuthenticator.php
DaazKu: Everything can be casted to a boolean so it is not a problem. 

The doc says that it wants a boolean so if you pass it something else you are playing with fire.

library/Vanilla/Authenticator/SSOAuthenticator.php
DaazKu: The docblock is important so it has to stay.

library/Vanilla/Authenticator/Authenticator.php
DaazKu: `getNameImpl()` is also used in the constructor ;)

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -0,0 +1,492 @@
...
+        $this->permission();
+
+        $this->schema([
+            'authenticator:s' => 'The authenticator that will be used.',
+            'authenticatorID:s?' => 'Authenticator instance\'s identifier.',
+            'userID:i?' => 'UserID to unlink authenticator from.',
+        ], 'in')->setDescription('Delete the link between an authenticator and a user.');
+        $this->schema([], 'out');
+
+        $in->validate($query);
```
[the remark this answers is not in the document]
DaazKu: Fixed.

library/Vanilla/Models/SSOInfo.php
DaazKu: I can do SSOData.

library/Vanilla/AddonManager.php
```diff
@@ -218,7 +218,7 @@ public function findClasses($pattern, $searchAll = false) {
         if ($searchAll === false) {
             foreach ($this->autoloadClasses as $classKey => $classesEntry) {
                 foreach($classesEntry as $namespace => $classData) {
-                    if ($fn($namespace.$classKey)) {
+                    if ($fn($namespace.$classData['className'])) {
```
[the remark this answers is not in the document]
DaazKu: Yeah I'll do that.

library/Vanilla/Models/SSOModel.php
```diff
@@ -0,0 +1,319 @@
...
+
+                // This coupling sucks but I feel like that's the best way to accommodate the config!
+                if ($syncRolesOnlyRegistration && val('connectOption', $ssoInfo) !== 'createuser') {
+                    $syncRoles = false;
+                } else {
+                    // Synchronize user's roles.
+                    $syncRoles = $this->config->get('Garden.SSO.SyncRoles', false);
+                }
+
+                if (!$this->syncUser($ssoInfo, $user, $syncInfo, $syncRoles)) {
```
initvector: Syncing user info with a trusted provider does not appear to sync roles. Using Vanilla Connect and this branch, `$syncRoles` is always false, even though the connection is trusted.

DaazKu: Is Garden.SSO.SyncRoles  set to true?

library/Vanilla/Models/SSOInfo.php
DaazKu: Done

library/Vanilla/Models/SSOModel.php
DaazKu: Fixed

DaazKu:
The build is failing because of the conversation endpoint and I need to change [this](https://github.com/vanilla/vanilla/blob/812258d37fc361c629d9f10659da741bbdd37727/applications/dashboard/controllers/api/AuthenticateApiController.php#L410) to a ConfigurationException.

Everything is done otherwise.

DaazKu:
I created the issue for the profile extender: https://github.com/vanilla/vanilla/issues/6074

DaazKu:
I've also created https://github.com/vanilla/vanilla/issues/6075 to handle that [TODO](https://github.com/vanilla/vanilla/pull/5831/files#diff-8ffe72f3aa055e97957fd1b12735810bR320)

PR #5835, opened 2017-08-11
Deprecate & null the API configuration endpoint

linc:
`/configuration` under API v1, which sits at /settings/configuration in the Dashboard.

* I cannot figure out what this was ever for.
* It's a bad security idea to leave anything like this laying around.
* It's been broken for a long time due to manually requiring a `transientkey` in a `GET`.

I've elected to leave a deprecated stub despite it consistently returning a 403 to all API request as-is, meaning it's extremely likely no one's using it.

### DIFF
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -697,44 +697,11 @@ class SettingsController extends DashboardController {
 
     /**
      *
-     *
+     * @deprecated
      * @throws Exception
      */
     public function configuration() {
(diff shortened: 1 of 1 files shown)


initvector:
This endpoint is available as a part of API v1: [Forum Administration](http://docs.vanillaforums.com/api/forum-administration). API calls "skip" TK checks.

linc:
This is a manual `GET` check so it is broken for all API calls. A customer reported perpetual 403 errors when attempting to use it. I've removed the endpoint from the old blog docs and cross-referenced their removal from the newer API docs now as well.

PR #5844, opened 2017-08-14
Backport: Fix max recipient check

DaazKu:
Backport #5843 

I had to change the case of the variables from the origin commit.


linc:
Please use proper PR titles and the cross-linking feature for backports.

DaazKu:
Gotcha

PR #5847, opened 2017-08-14
Add param to comments index endpoint to allow lookup by author

### LINKED ISSUE #5800: Add filter by InserUserID to comments API v2 endpoint
There needs to be a way to filter comments by a particular user in the /api/v2/comments index.

See [`DiscussionsApiController::index`](https://github.com/vanilla/vanilla/blob/78605ad92f591c3a68a5324b7eba478e96e67021/applications/vanilla/controllers/api/DiscussionsApiController.php#L240)

initvector:
This update adds the insertUserID parameter to the comments index endpoint under API v2 (/api/v2/comments). Performing a GET request to this endpoint using this parameter will allow you to filter the results based on who authored the comment. This can be used on its own to retrieve all comments by a user, or it can be used with the discussionID parameter to filter comments on a particular discussion by a specific user. The index endpoint now requires insertUserID *or* discussionID.

This new functionality uses the `CommentModel::getByUser2` method. The ability to control record sorting and filtering by date has been added to this function.

Closes #5800

### DIFF
--- a/applications/vanilla/controllers/api/CommentsApiController.php
+++ b/applications/vanilla/controllers/api/CommentsApiController.php
@@ -220,7 +220,7 @@ class CommentsApiController extends AbstractApiController {
         $this->permission();
 
         $in = $this->schema([
-            'discussionID:i' => 'The discussion ID.',
+            'discussionID:i?' => 'The discussion ID.',
             'page:i?' => [
                 'description' => 'Page number.',
                 'default' => 1,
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/api/CommentsApiController.php
```diff
@@ -233,34 +233,50 @@ public function index(array $query) {
...
+        // Lookup by discussion or by user?
+        if (array_key_exists('discussionID', $query)) {
+            $discussion = $this->discussionByID($query['discussionID']);
+            list($offset, $limit) = offsetLimit("p{$query['page']}", $query['limit']);
+            $this->discussionModel->categoryPermission('Vanilla.Discussions.View', $discussion['CategoryID']);
 
-        if (isset($query['after'])) {
-            $where['dateInserted >'] = $query['after'];
+            // Build up the where clause.
+            $where = ['discussionID' => $query['discussionID'], 'joinUsers' => false];
```
DaazKu: Are you sure that thi where will work on a case sensitive system?

DaazKu: Even if it does I think we should use the properly named columns

initvector: I agree. The columns used in the where clause should use the same casing established in the applications structure.

DaazKu: I did a double check as an FYI and only table names are case sensitive in MySQL.
https://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html

applications/vanilla/controllers/api/CommentsApiController.php
```diff
@@ -233,34 +233,50 @@ public function index(array $query) {
...
             ],
+            'insertUserID:i?' => 'Filter by author.',
             'after:dt?' => 'Get only comments after this date.',
             'expand:b?' => [
                 'description' => 'Expand associated records.',
                 'default' => false
-            ]
-        ], 'in')->setDescription('List comments.');
+            ],
+            'lookup:s?' => 'The field used to lookup comments.'
```
DaazKu: I think you forgot to strip this field.

applications/vanilla/controllers/api/CommentsApiController.php
```diff
@@ -233,34 +233,50 @@ public function index(array $query) {
...
+
+            $rows = $this->commentModel->getWhere(
+                $where,
+                'DateInserted',
+                'asc',
+                $limit,
+                $offset
+            )->resultArray();
+        } else {
+            $rows = $this->commentModel->getByUser2(
```
DaazKu: This does not honour the after parameter.

DaazKu:
My last concern here is that if you request user's comments by discussion the results are not filtered by other plugins but if you request user's comments by InsertUserID they are.

This can be seen in action with the Troll plugin. This causes some discrepancy the results.

DaazKu:
So we talked about this issue and we "should" be using the models so that plugins can filter the data.
The model should be updated/refactored if it can't be used as is.

initvector:
I've filed a separate issue for looking up discussions: #5854

PR #5856, opened 2017-08-14
Ignore .vscode folder

Charrondev:
This makes it easier for people that use VSCode to make commits without accidently commiting their editor's project specific configuration (Such as XDebug config)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ composer-local.json
 .settings
 .buildpath
 .idea/*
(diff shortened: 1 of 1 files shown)


DaazKu:
Should we update https://github.com/vanilla/vanilla-cli then?

https://github.com/vanilla/vanilla-cli/tree/eed768baad8b57b83d0cef4c9127901339192555/.vscode

PR #5859, opened 2017-08-16
Use the DI container to instantiate plugins

DaazKu:
The plugin manager was not using the DI Container to instantiate plugins,
Now it does :)

### DIFF
--- a/library/core/class.pluginmanager.php
+++ b/library/core/class.pluginmanager.php
@@ -67,6 +67,11 @@ class Gdn_PluginManager extends Gdn_Pluggable implements ContainerInterface {
     private $addonManager;
 
     /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    /**
      * @var EventManager
      */
     private $eventManager;
@@ -75,12 +80,13 @@ class Gdn_PluginManager extends Gdn_Pluggable implements ContainerInterface {
      * Initialize a new instance of the {@link Gdn_PluginManager} class.
      *
      * @param AddonManager $addonManager The addon manager that manages all of the addons.
+     * @param ContainerInterface $container The DI container.
      * @param EventManager $eventManager The event manager that handles all plugin events.
      */
-    public function __construct(AddonManager $addonManager = null, EventManager $eventManager = null) {
+    public function __construct(AddonManager $addonManager = null, ContainerInterface $container, EventManager $eventManager = null) {
         parent::__construct();
         $this->addonManager = $addonManager;
-
+        $this->container = $container;
         $this->eventManager = $eventManager ?: new EventManager($this);
     }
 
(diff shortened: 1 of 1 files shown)


library/core/class.pluginmanager.php
```diff
@@ -75,12 +80,13 @@ class Gdn_PluginManager extends Gdn_Pluggable implements ContainerInterface {
      * Initialize a new instance of the {@link Gdn_PluginManager} class.
      *
      * @param AddonManager $addonManager The addon manager that manages all of the addons.
+     * @param ContainerInterface $container The DI container.
      * @param EventManager $eventManager The event manager that handles all plugin events.
      */
-    public function __construct(AddonManager $addonManager = null, EventManager $eventManager = null) {
+    public function __construct(AddonManager $addonManager = null, ContainerInterface $container, EventManager $eventManager = null) {
```
tburry: You've sandwiched a required parameter between two optional ones. I'd just use the container from the event manager.

DaazKu:
Welp. Bause we hacked Gdn_PluginManager to be instantiable with no argument to be able to unit test it the changes required to be able to properly use the container are way greater than I initially thought.

The plugin manager use itself to [instantiate plugins](https://github.com/vanilla/vanilla/blob/9ec24f24c9e76eebd5dd4512aa60b27844c6bad4/library/core/class.pluginmanager.php#L1295).
This prevent us from using the container to instantiate plugins when doing unit tests because it will cause an infinite call stack between `Gdn_PluginManager::getPluginInstance()` and `Gdn_PluginManager::get()`.

To fix the issue we would have to refactor all unit tests that use the PluginManager to properly use the DI container. This is something that I do not have the time to do now.

I'll create an issue about it and close this PR.

tburry:
You cant just do $this->eventManager->getContainer()->get(...')?

DaazKu:
No. Like I said if you instantiate the PluginManager like so: `new Gdn_PluginManager()` 

```
    public function __construct(AddonManager $addonManager = null, EventManager $eventManager = null) {
        parent::__construct();
        $this->addonManager = $addonManager;
        $this->eventManager = $eventManager ?: new EventManager($this); // Assign itself as the container.
    }
```

You end up with an infinite call stack generated by `Gdn_PluginManager::get()` calling `Gdn_PluginManager::getPluginInstance()` which would call `$this->eventManager->getContainer()->get(...')` which is the same as calling `Gdn_PluginManager::get()`.

DaazKu:
After speaking with Todd we concluded that we will detect if the proper DI container was passed in `Gdn_PluginManager`. If so, we will use it to instantiate the class, otherwise, we will use new `$class()`.

The latter won't happen when running Vanilla and will only happen in unit tests. Fortunately the methods using the container are not unit tested and will never be.

DaazKu:
A proper fix will be made when https://github.com/vanilla/vanilla/issues/5863 is done.

PR #5862, opened 2017-08-16
Use the DI container to instantiate plugins

DaazKu:
See https://github.com/vanilla/vanilla/pull/5859 for details.

### DIFF
--- a/library/core/class.pluginmanager.php
+++ b/library/core/class.pluginmanager.php
@@ -72,6 +72,14 @@ class Gdn_PluginManager extends Gdn_Pluggable implements ContainerInterface {
     private $eventManager;
 
     /**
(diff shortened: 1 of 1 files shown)


library/core/class.pluginmanager.php
```diff
@@ -614,10 +622,14 @@ public function getPluginInstance($name, $accessType = self::ACCESS_CLASSNAME, $
         }
 
         if (!isset($this->instances[$className])) {
-            if ($sender === null) {
-                $object = new $className();
+            if ($this->hasProperContainer) {
+                $object = $this->eventManager->getContainer()->get($className);
             } else {
-                $object = new $className($sender);
+                if ($sender === null) {
```
tburry: Clever catch. Can you hoist this nest to the parent if tho?

PR #5864, opened 2017-08-16
Allow plugins to set a dependency on Gdn_Controller

DaazKu:
Allows plugins, or any class really, to have a dependency on the current controller without calling Gdn::controller().

In the plugins it is useful if the event was not fire from a controller but you still need to reference it.

```php
    /**
     * @param Gdn_Controller $controller Current controller.
     */
    public function __construct(Gdn_Controller $controller) {
        $this->controller = $controller;
    }
   
    public function model_event_handler() {
        $this->controller->jsonTarget({TARGET}, {DATA}, {TYPE});
    }
```

I tried adding
```
    ->rule('Gdn_Controller')
    ->setInherit(true)
    ->setShared(true)
```
in `bootstrap.php` but the problem is that when you instantiate a new Controller you end up having a `Gdn_Controller` instance instead of a `SpecificController` instance in the container.

I was also playing with the idea of creating a `SharedInstanceInterface` in `Garden\Container` but wasn't sure it was the way to go.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -842,8 +842,12 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
     private function createController($controllerName, $request, &$routeArgs) {
         /* @var Gdn_Controller $controller */
         $controller = $this->container->get($controllerName);
+
         // Allow classes to have a dependency on Gdn_Controller.
-        $this->container->setInstance('Gdn_Controller', $controller);
+        // It is possible that the controller does not inherit Gdn_Controller :(
+        if (is_a($controller, Gdn_Controller::class)) {
+            $this->container->setInstance(Gdn_Controller::class, $controller);
+        }
         Gdn::controller($controller);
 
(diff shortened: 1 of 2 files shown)


DaazKu:
Note that I changed the dispatcher constructor. We'll have to test on Infra to make sure that this change does break something.

library/core/class.dispatcher.php
```diff
@@ -830,7 +841,9 @@ private function dispatchController($request, $routeArgs) {
      */
     private function createController($controllerName, $request, &$routeArgs) {
         /* @var Gdn_Controller $controller */
-        $controller = new $controllerName();
+        $controller = $this->container->get($controllerName);
+        // Allow classes to have a dependency on Gdn_Controller.
+        $this->container->setInstance('Gdn_Controller', $controller);
```
tburry: I'd add a check to make sure that $controller is an instance of Gdn_Controller since that isn't a hard requirement.

PR #5874, opened 2017-08-18
Add ability to translate Activity HeadlineFormat on sub-communities

nervoustwit:
Save the raw, untranslated format of an activity headline in the data column of the activity table. If the activity is displayed in another language it can be translated from the original.


DaazKu:
Not good enough: https://github.com/vanilla/vanilla/issues/5881

nervoustwit:
This is going to be handled some other way.

PR #5877, opened 2017-08-22
Add /tokens API v2 endpoint and UI

### LINKED ISSUE #5876: /tokens
This endpoint requires the modifying GDN_AccessToken to add the AccessTokenID auto-increment column.

The /tokens endpoints will initially be for granting/revoking personal access tokens.

## POST /tokens

- name: A user-defined name indicating what the access token will be used for.
- transientKey: The users's CSRF token. This limits the endpoint to the web UI (more or less).

Create an access token. The type of access token will be "personal". The name doesn't have a column so add it to the attributes.

### Output

- accessTokenID
- name
- accessToken: This is the signed access token suitable for making requests.
- dateInserted

## GET /tokens

Get the active personal access tokens for a user. This does not include expired or revoked tokens.

### Output

- accessTokenID: An ID used to revoke the token.
- name
- dateInserted

## DELETE /tokens/:id

Revoke the access token with the given ID.

## GET /tokens/:id

- transientKey: The users's CSRF token. This limits the endpoint to the web UI (more or less).

Reveal the access token with the given ID.

### Output

- accessTokenID
- name
- accessToken
- dateInserted

This is the same as the output for POST /tokens

initvector:
First and foremost, this update adds the /tokens endpoint for API v2. To that end, it does several things:

1. Adds index, GET id and DELETE methods to the endpoint. The index will list all of a user's active tokens. Getting a specific token by its AccessTokenID will return a usable `accessToken`. Deleting a token retains the record, but revokes and expires it.
1. To limit the usage of this endpoint to UI operations, some actions require a transient key. By design, API calls dodge transient key checks, so more direct checks have been added to `TokensApiController`. This has necessitated modifications to `Gdn_Session` to allow better access to its transient key generation/verification methods.
1. The AccessToken table now has an auto-increment primary key: AccessTokenID. InsertUserID has also been added to this table.
1. A new permission, Garden.Tokens.Add, has been added. Users with this permission have access to managing their own access tokens for a site.
1. `AccessTokenModel` has been updated alongside the AccessToken table. `AccessTokenModel::getID` looks up an access token by the AccessTokenID field. `AccessTokenModel::getToken` looks up an access token by the Token field.
1. `InternalClient`, used in automated testing, has been updated to allow GET and POST calls with a transient key field.
1. `InternalRequest`, also used in automated testing, has been updated to support cookies.
1. Unit tests have been added for the /tokens endpoint.

Closes #5876

### DIFF
--- a/applications/dashboard/controllers/Api/TokensApiController.php
+++ b/applications/dashboard/controllers/Api/TokensApiController.php
@@ -5,6 +5,7 @@
  */
 
 use Garden\Schema\Schema;
+use Garden\Web\Data;
 use Garden\Web\Exception\ClientException;
 use Garden\Web\Exception\NotFoundException;
 
@@ -50,11 +51,11 @@ class TokensApiController extends AbstractApiController {
      * @throws ClientException if current user isn't authorized to delete the token.
      */
     public function delete($id) {
-        $this->schema($this->idSchema(),'in')->setDescription('Revoke an authentication token.');
+        $this->schema($this->idSchema(),'in')->setDescription('Revoke an access token.');
         $out = $this->schema([], 'out');
 
         $row = $this->token($id);
-        if ($row['UserID'] != $this->session->UserID) {
+        if ($row['UserID'] != $this->getSession()->UserID) {
             $this->permission('Garden.Settings.Manage');
         }
 
@@ -142,7 +143,7 @@ class TokensApiController extends AbstractApiController {
         $in = $this->schema([
             'id',
             'transientKey:s' => 'A valid CSRF token for the current user.'
-        ], 'in')->add($this->idSchema())->setDescription('Reveal a usable authentication token.');
+        ], 'in')->add($this->idSchema())->setDescription('Reveal a usable access token.');
         $out = $this->schema($this->sensitiveSchema(), 'out');
 
         $query['id'] = $id;
@@ -150,8 +151,8 @@ class TokensApiController extends AbstractApiController {
         $this->validateTransientKey($query['transientKey']);
 
         $row = $this->token($id);
-        if ($row['UserID'] != $this->session->UserID) {
-            if ($this->session->checkPermission('Garden.Settings.Manage') === false) {
+        if ($row['UserID'] != $this->getSession()->UserID) {
+            if ($this->getSession()->checkPermission('Garden.Settings.Manage') === false) {
                 throw new NotFoundException('Access Token');
             }
         }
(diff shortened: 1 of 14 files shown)


applications/dashboard/controllers/Api/TokensApiController.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /**
+     * Validate the transient key for the current request.
+     *
+     * @param $transientKey
+     * @throws ClientException
+     */
+    public function validateTK($transientKey) {
```
DaazKu: This function's name is poorly chosen. I thought it was validateToken.

initvector: A "token" in this context is an access token. This function will be responsible for validating the transient key ("TK") for some actions that require it.

DaazKu: `validateTransientKey()`

library/Vanilla/Web/Controller.php
```diff
@@ -27,7 +27,7 @@
     /**
      * @var SessionInterface
      */
-    private $session;
+    protected $session;
```
DaazKu: This should remain private I think,
`getSession()` should be used to access it. We probably do not want child classes to modify the variable.

applications/dashboard/controllers/Api/TokensApiController.php
```diff
@@ -0,0 +1,316 @@
...
+use Garden\Web\Exception\ClientException;
+use Garden\Web\Exception\NotFoundException;
+
+/**
+ * API Controller for the `/tokens` resource.
+ */
+class TokensApiController extends AbstractApiController {
+
+    /** Default expiry for issued tokens. */
+    const DEFAULT_EXPIRY = '10 years';
```
DaazKu: From a security perspective, I'm concerned about tokens virtually not expiring on their own.

What will they be used for?

applications/dashboard/controllers/Api/TokensApiController.php
```diff
@@ -0,0 +1,316 @@
...
+        }
+        unset($token);
+        array_walk($activeTokens, [$this, 'prepareRow']);
+
+        $result = $out->validate($activeTokens);
+        return $result;
+    }
+
+    /**
+     * Issue a new transient key for the current user.
```
DaazKu: access token?

We need to make sure to only use one terminology when speaking about something.

`->setDescription('Issue a new authentication token for the current user.');`
should probably by
`->setDescription('Issue a new access token for the current user.');`

applications/dashboard/controllers/Api/TokensApiController.php
```diff
@@ -0,0 +1,316 @@
...
+/**
+ * API Controller for the `/tokens` resource.
+ */
+class TokensApiController extends AbstractApiController {
+
+    /** Default expiry for issued tokens. */
+    const DEFAULT_EXPIRY = '10 years';
+
+    /** The maximum number of tokens in a response. */
+    const RESPONSE_LIMIT = 200;
```
DaazKu: If we limit a response we need pagination.

Not having pagination make the possibility of having an issue not being able to get limit + 1 result even if the limit is stupidly high.
Not having ensure that we will never have that problem.
If we want to impose a limit for scaling reason we should do so on the create endpoint, not the get.

tests/InternalClient.php
```diff
@@ -35,6 +46,21 @@ public function __construct(Container $container, $baseUrl = '/api/v2') {
     }
 
     /**
+     * Add the transient key cookie header to an array of headers.
+     *
+     * @param array $headers
+     */
+    public function addTransientKeyHeader(array &$headers) {
```
DaazKu: I'm really not a big fan of all these functions modifying parameters by references instead of returning the modified object.
We are gonna end up with 2 different ways to do things and I that's not great.
Not having to think twice and always assign back the variable to itself when calling the functions is simpler.

Also, passing by reference does is not an optimization at all.

PR #5880, opened 2017-08-23
Add missing throws to Controller::permission()

DaazKu:
Seems like we failed to properly do the Controller::permission() docblock.


hgtonight:
+1

DaazKu:
Thanks for that extra commit Ryan. Should have done that!

Issue #5881, opened 2017-08-24
[Translation] Literals should not be combined because it causes issue in some languages


DaazKu:
This can be seen in french on the activity page:

*UP = User Perspective*
*GP = Guest Perspective*
<table>
  <tr>
    <th></th>
    <th>English UP</th>
    <th>French UP</th>
    <th>English GP</th>
    <th>French GP</th>
  </tr>
  <tr>
    <td>English</td>
    <td colspan="4" rowspan="2">
<img src="https://user-images.githubusercontent.com/2412909/29672657-e83b0adc-88ba-11e7-85f8-d73986b36fc2.png"/>
</td>
  </tr>
  <tr>
    <td>French</td>
  </tr>
</table>

#### First error:
On the english site: `You changed your profile picture`
Everything it properly translated "words for words" but the problem is that the verb needs to be conjugated with the pronoun. In this case the correct translation would be: `Vous avez changé votre photo de profil`

#### Second error:
On the english site: `DaazKu changed their profile picture`
The translation from english to french of their is leur.

You end up with `DaazKu a changé leur photo de profil` but leur is plural and cannot be used to reference something singular. The correct translation should be `DaazKu a changé sa photo de profil` which is the same as using his.

All these translations are done with a single literal: `{RegardingUserID,You} changed {ActivityUserID,your} profile picture.` which is the default for `HeadlineFormat.PictureChange` but we need 2 distinct literals for the french language. One for `[link](You) changed [link](your) profile picture` and one for `[link](User) changed [link](their) profile picture`.

We have the same kind of problem with `All %s`. In french `All` is translated word for word to `Tous` but this changes if the subject is feminin. In that case the translation would be `Toutes`. This can be seen on the notifications vs messages:
`All messages -> Tous les messages` Correct
`All notifications -> Tous les notifications` Incorrect! This should be `Toutes les notifications`

I think we should drop all literals that uses compositions of other literals and have a talk about the activity page.

charrondev:
We're essentially never going to being something like this, but in this case I don't think this will be fix until an overhaul of activity is done.

PR #5883, opened 2017-08-24
Fix "All Notifications" and "All Bookmarks" strings for translation.

nervoustwit:
Back porting #5834


DaazKu:
Relagted https://github.com/vanilla/vanilla/issues/5881

PR #5884, opened 2017-08-24
Fix “All messages” string for translations

nervoustwit:
This is to un-couple "All" and "messages" for more accurate translations in certain languages.


DaazKu:
Related: https://github.com/vanilla/vanilla/issues/5881

PR #5887, opened 2017-08-24
Fix All Messages so that it translates better.

nervoustwit:
Backport/5884

### DIFF
--- a/applications/conversations/views/messages/popin.php
+++ b/applications/conversations/views/messages/popin.php
@@ -45,7 +45,7 @@
         <?php endforeach; ?>
         <li class="Item Center">
             <?php
-            echo anchor(sprintf(t('All %s'), t('Messages')), '/messages/inbox');
+            echo anchor(t('All Messages'), '/messages/inbox');
             ?>
         </li>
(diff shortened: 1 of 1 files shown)


applications/conversations/views/messages/popin.php
```diff
@@ -45,7 +45,7 @@
         <?php endforeach; ?>
         <li class="Item Center">
             <?php
-            echo anchor(sprintf(t('All %s'), t('Messages')), '/messages/inbox');
+            echo anchor(t('All Messages'), '/messages/inbox');
```
initvector: Is there already a translation for "All Messages"? If not, this is going to break this translation everywhere.

initvector: I've added the source string with https://github.com/vanilla/locales/commit/d95e4ff74a585bbd721d678837eda47a5c4c86fe but we'll need to wait until it's been translated before this backport is merged.

linc: Make the change backwards compatible instead by using the existing version as the fallback.

PR #5892, opened 2017-08-29
Fix configuring cookie persistence 

jamesinc:
This appears to be a longstanding typo. I found it by accident while tracing the code path for setting the long-term login persistence cookie.


initvector:
Great find! Thanks!

PR #5894, opened 2017-08-29
Use Session->hourOffset() in class Format

R-J:
Class Session contains a helper method hourOffset() which returns the correct $hourOffset for current user/guest.
In class format that same offset calculation was done in two methods. I replaced that calculation directly with the sessions method.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -512,8 +512,6 @@ class Gdn_Format {
      * @return string
      */
     public static function date($timestamp = '', $format = '') {
-        static $guestHourOffset;
-
         // Was a mysqldatetime passed?
         if ($timestamp !== null && !is_numeric($timestamp)) {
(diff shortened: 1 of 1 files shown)


R-J:
Instead of
~~~
    public static function toTimezone($timestamp) {
        // Alter the timestamp based on the user's hour offset.
        $hourOffset = Gdn::session()->hourOffset();

        if ($hourOffset <> 0) {
            $timestamp += $hourOffset * 3600;
        }

        return $timestamp;
    }
~~~

I would suggest to use the much shorter but still readable

~~~
    public static function toTimezone($timestamp) {
        return $timestamp + Gdn::session()->hourOffset() * 3600;
    }
~~~

If you like me to, I would change this PR accordingly

DaazKu:
Thanks!

PR #5903, opened 2017-08-30
Use photoUrl in users API v2 endpoint

### LINKED ISSUE #5900: [API v2] UserApiController use photo instead of photoURL
Everywhere else photoURL is used. See `getUserFragment()`.

initvector:
The standard for outputting a user's photo as part of a record in API endpoints is in the `photoUrl` field. The output schema used under the /users API v2 isn't adopting this standard and only uses the normal "photo" field.

This update alters the output schemas under the /users endpoint to put a URL to a user's photo under the `photoUrl` field.

"Photo" is going to remain a part of the full schema of the user record, because it is beneficial for setting a photo or reading a raw photo value (e.g. s3://path/to/file.jpg), whereas photoUrl should always be a client-consumable URL to the file.

Closes #5900

### DIFF
--- a/applications/dashboard/controllers/Api/UsersApiController.php
+++ b/applications/dashboard/controllers/Api/UsersApiController.php
@@ -71,12 +71,12 @@ class UsersApiController extends AbstractApiController {
             'photo:s' => [
                 'allowNull' => true,
                 'minLength' => 0,
-                'description' => 'Photo of the user.'
+                'description' => 'Raw photo field value from the user record.'
             ],
             'photoUrl:s' => [
                 'allowNull' => true,
                 'minLength' => 0,
-                'description' => 'Url to the user photo.'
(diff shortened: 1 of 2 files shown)


DaazKu:
It is not clear what the difference between Photo and PhotoURL is.
Why do we need Photo is we have Photo URL? Can you actually upload an image via API?

applications/dashboard/controllers/Api/UsersApiController.php
```diff
@@ -73,6 +73,11 @@ protected function fullSchema() {
                 'minLength' => 0,
                 'description' => 'Photo of the user.'
             ],
+            'photoUrl:s' => [
+                'allowNull' => true,
+                'minLength' => 0,
+                'description' => 'Url to the user photo.'
```
DaazKu: It should be URL in the description. Not Url.

PR #5904, opened 2017-08-30
Add ability to move categories between parents with API v2

### LINKED ISSUE #5799: Add action to move categories between parents in API v2 categories endpoint
Categories cannot be deleted when they still have children. We need an easy way to move categories from one parent to another.

initvector:
This update adds a PUT action on the /categories API v2 endpoint for moving categories between parents.

To use, perform a PUT request to /api/v2/categories/[categoryID]/parentCategoryID where the body is a JSON of the new value:

```json
{ "parentCategoryID": 123 }
```

Closes #5799

### DIFF
--- a/applications/vanilla/controllers/api/CategoriesApiController.php
+++ b/applications/vanilla/controllers/api/CategoriesApiController.php
@@ -76,13 +76,12 @@ class CategoriesApiController extends AbstractApiController {
      * Lookup a single category by its numeric ID or its URL code.
      *
      * @param int|string $id The category ID or URL code.
-     * @param bool $throw Throw an exception if the category cannot be found.
      * @throws NotFoundException if the category cannot be found.
      * @return array
      */
-    public function category($id, $throw = true) {
+    public function category($id) {
         $category = CategoryModel::categories($id);
-        if (empty($category) && $throw) {
+        if (empty($category)) {
             throw new NotFoundException('Category');
         }
         return $category;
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/api/CategoriesApiController.php
```diff
@@ -353,6 +343,70 @@ public function post(array $body) {
...
+        if ($id == $parentCategoryID) {
+            throw new ClientException('A category cannot be the parent of itself.');
+        }
+
+        if ($parentCategoryID < 1 && $parentCategoryID !== -1) {
+            throw new ClientException('parentCategoryID must be -1 or greater than zero.');
+        }
+
+        $parentCategory = $this->category($parentCategoryID, false);
+        if (!is_array($parentCategory)) {
```
DaazKu: Instead of opting out of category throwing an error (there should not be an option because may be confusing down the line) you should try {} catch () {} and throw your client exception instead of testing if parentCategory is not an array.

applications/vanilla/controllers/api/CategoriesApiController.php
```diff
@@ -353,6 +343,70 @@ public function post(array $body) {
...
+     * @param int $id The ID of the category.
+     * @param array $body The request body.
+     * @throws NotFoundException if unable to find the category.
+     * @throws ClientException if the parent and category are the same.
+     * @throws ClientException if the parent category ID is invalid.
+     * @throws ClientException if the target parent category does not exist.
+     * @throws ClientException if trying to move a category under one of its own children.
+     * @return array
+     */
+    public function put_parentCategoryID($id, array $body) {
```
DaazKu: This should be done in patch. We are affectively updating parentCategoryID of a category. 

PUT is for replacing a whole record.

DaazKu:
Seems good to me but I think this is the kind of feature that will really benefit some unit tests.

- Testing that the category really moved.
- Testing that the parent counts where properly updated.

initvector:
I've added the following tests:

* The category was successfully moved.
* Verify the parent exists before moving.
* Verify the category won't be a parent of itself.
* Verify the category won't be under one of its own children.

PR #5908, opened 2017-08-31
Allow conversation to be created without message

DaazKu:
Add an option that allows to create a conversation without a message.
This is required by the ConversationAPIController.

Once we have time for it we can refactor both models to be separated instead of being mixed together.

What to check when testing:
- Users notifications (Activities)
- Users unread message count
- Conversation's message related fields (FirstMessageID, LastMessageID, CountMessages...)
- UserConversation message related fields (CountReadMessages, LastMessageID...)
- All the above when adding a user to a conversation that has no messages.

Easy way to clean your DB for testing purpose:
```sql
TRUNCATE GDN_Conversation;
TRUNCATE GDN_ConversationMessage;
TRUNCATE GDN_UserConversation;
UPDATE GDN_User SET CountUnreadConversations = NULL;
DELETE FROM GDN_Activity WHERE ActivityTypeID IN (20, 21);
```

To test the option really easily from the UI you can switch 
`$createMessage = empty($settings['ConversationOnly']);`
to
`$createMessage = !empty($settings['ConversationOnly']);`
in `ConversationModel::save()`

### DIFF
--- a/applications/conversations/models/class.conversationsmodel.php
+++ b/applications/conversations/models/class.conversationsmodel.php
@@ -94,4 +94,44 @@ abstract class ConversationsModel extends Gdn_Model {
         $conversationMembers = $this->getConversationMembers($conversationID);
         return (in_array($userID, $conversationMembers));
     }
+
+    /**
+     * Notify users when a new message is created.
+     *
+     * @param array|object $conversation
+     * @param array|object $message
+     * @param array $notifyUserIDs
+     */
+    protected function notifyUsers($conversation, $message, $notifyUserIDs) {
+        $conversation = (array)$conversation;
+        $message = (array)$message;
+
+        $activity = [
+            'ActivityType' => 'ConversationMessage',
+            'ActivityUserID' => $conversation['InsertUserID'],
+            'HeadlineFormat' => t('HeadlineFormat.ConversationMessage', '{ActivityUserID,User} sent you a <a href="{Url,html}">message</a>'),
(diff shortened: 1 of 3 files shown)


applications/conversations/models/class.conversationsmodel.php
```diff
@@ -94,4 +94,44 @@ public function validConversationMember($conversationID, $userID) {
...
+     * @param array $notifyUserIDs
+     */
+    protected function notifyUsers($conversation, $message, $notifyUserIDs) {
+        $conversation = (array)$conversation;
+        $message = (array)$message;
+
+        $activity = [
+            'ActivityType' => 'ConversationMessage',
+            'ActivityUserID' => $conversation['InsertUserID'],
+            'HeadlineFormat' => t('HeadlineFormat.ConversationMessage', '{ActivityUserID,User} sent you a <a href="{Url,html}">message</a>'),
```
initvector: Since `ActivityUserID` is being set to the user who created the conversation, not the message, notifications are going to say the conversation creator is the one sending everything.

You can test this out by creating a conversation as one user A and sending a message to it as user B. You will likely get a notification on user B that user A sent the message.

initvector:
- [x] Users notifications (Activities)
- [x] Users unread message count
- [x] Conversation's message related fields (FirstMessageID, LastMessageID, CountMessages...)
- [x] UserConversation message related fields (CountReadMessages, LastMessageID...)
- [x] All the above when adding a user to a conversation that has no messages.

PR #5910, opened 2017-08-31
Allow table-related HTML elements to have classes in posts

nervoustwit:
Presently the VanillaHtmlFormatter will allow certain css classes to be added to certain DOM elements. This PR will allow table, tr and td elements to have css classes added to them.

### DIFF
--- a/library/core/class.vanillahtmlformatter.php
+++ b/library/core/class.vanillahtmlformatter.php
@@ -123,7 +123,10 @@ class VanillaHtmlFormatter {
         'pre',
         'span',
         'strong',
-        'ul'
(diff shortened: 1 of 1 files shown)


linc:
My thoughts on this:

* It appears to be safe, because we separately whitelist what classes are allowed to be added in UGC.
* It appears necessary, because some sites do use tables in UGC and wish to format them in specific ways when publishing content as a discussion. I foresee more use cases for this in the future.

DaazKu:
How about `caption`?

https://www.w3.org/TR/html5/tabular-data.html#the-table-element

PR #5917, opened 2017-09-01
Add ability to filter and sort discussions endpoint index by pinned status

### LINKED ISSUE #5534: Add "pinned" support to /discussions endpoint index
We'll need to add support for adding pinned discussions (announcements) to their proper position when pulling posts from the /discussions endpoint index.

Vanilla will be moving towards using "pin"/"pinned" instead of "announce"/"announced". These endpoint additions will implement the updated nomenclature.

1. `DiscussionApiController`
    1. Add a "pinned" option to the endpoint. This should be an optional boolean field.
    1. Add a "pinOrder" option to the endpoint. This should be an enum field, consisting of two options: "mixed" and "first".
    1. If the "pinned" parameter is `false`, do not include discussions with a truthy Announce field.
    1. If the "pinned" parameter is `true`, include ONLY discussions with a truthy Announce field.
    1. If the "pinned" parameter is `null` and "pinOrder" is "mixed", include discussions with a truthy Announce field in their natural sort order.
    1. If the "pinned" parameter is `null` and "pinOrder" is "first", include discussions with a truthy Announce field at the top of the first page of results.
1. `DiscussionModel`
    1. Update `DiscussionModel::calculate` to add a "Pinned" field, based on the value of "Announce". It will be cast to a boolean.
    1. Add a pinLocation enum that can either be "category" or "discussions". Announce of 1 means "discusisons" and announce of 2 means "category".

initvector:
This update adds the ability to sort and filter the discussions endpoint index by pinned/announced status. The implementation follows what is seen on Vanilla's frontend. If you have 30 normal discussions in a result and there are two pinned discussions, your result is going to be 32 total discussions.

Specifics on usage of the new parameters can be found in the associated issue.

Closes #5534

### DIFF
--- a/applications/vanilla/controllers/api/DiscussionsApiController.php
+++ b/applications/vanilla/controllers/api/DiscussionsApiController.php
@@ -230,7 +230,7 @@ class DiscussionsApiController extends AbstractApiController {
             'pinned:b?' => 'Whether or not to include pinned discussions. If true, only return pinned discussions. Cannot be used with the pinOrder parameter.',
             'pinOrder:s?' => [
                 'default' => 'first',
-                'description' => 'If including pinned posts, in what order should they be integrated? Cannot be used with the pinned parameter.',
+                'description' => 'If including pinned posts, in what order should they be integrated? When "first", discussions pinned to a specific category will only be affected if the discussion\'s category is passed as the categoryID parameter. Cannot be used with the pinned parameter.',
                 'enum' => ['first', 'mixed'],
             ],
             'page:i?' => [
@@ -252,7 +252,11 @@ class DiscussionsApiController extends AbstractApiController {
 
         $query = $this->filterValues($query);
         $query = $in->validate($query);
+
         $where = array_intersect_key($query, array_flip(['categoryID', 'insertUserID']));
+        if (array_key_exists('categoryID', $where)) {
+            $where['d.CategoryID'] = $where['categoryID'];
+        }
 
         list($offset, $limit) = offsetLimit("p{$query['page']}", $query['limit']);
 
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -125,6 +125,11 @@ protected function fullSchema() {
             'insertUser?' => $this->getUserFragmentSchema(),
             'bookmarked:b' => 'Whether or no the discussion is bookmarked by the current user.',
             'announce:b' => 'Whether or not the discussion has been announced (pinned).',
+            'pinned:b|n' => 'Whether or not the discussion has been pinned.',
```
DaazKu: `Boolean | null` does not make sense.

A discussion is either pinned or not.

DaazKu: `Boolean | null` does not make sense.

A discussion is either pinned or not.

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -222,6 +227,14 @@ public function index(array $query) {
 
         $in = $this->schema([
             'categoryID:i?' => 'Filter by a category.',
+            'pinned:b|n' => [
```
DaazKu: As above. `b|n` does not make sense. This is a filter so `isPinned:b?` would probably make more sense.

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -222,6 +227,11 @@ public function index(array $query) {
 
         $in = $this->schema([
             'categoryID:i?' => 'Filter by a category.',
+            'pinned:b?' => 'Whether or not to include pinned discussions. If true, only return pinned discussions.',
+            'pinOrder:s?' => [
+                'enum' => ['first', 'mixed'],
+                'description' => 'If including pinned posts, in what order should they be integrated?'
```
DaazKu: The description needs an update. It should state that this filter will only work if `pinned` is not used.
There should also probably be a default value set to 'mixed' since that's how discussions are returned if no filter is used.

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -125,6 +125,11 @@ protected function fullSchema() {
             'insertUser?' => $this->getUserFragmentSchema(),
             'bookmarked:b' => 'Whether or no the discussion is bookmarked by the current user.',
             'announce:b' => 'Whether or not the discussion has been announced (pinned).',
```
DaazKu: Shouldn't we remove announce?

initvector: This'll be nuked as part of #5937. Right now, it's needed when using PUT announce.

PR #5923, opened 2017-09-05
Remove Database.CharacterEncoding from inf configs

linc:
This is a pure infrastructure change, so it should never go to `master`.

When we added utf8mb4 compatibility as the default, we manually set every existing forum to use utf8 via the config so we wouldn't get an impromptu change. We're now ready to do that switchover, and ops has already made the necessary changes to the my.cnf. So now it's time to undo the manual setting of this config globally.

Because we've already deployed a change effecting the collation, we're currently blocked from further database changes until this is deployed and activated.

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -892,7 +892,7 @@ if (c('Garden.Registration.CaptchaPublicKey')) {
 }
 
 // Remove Charset from inf.
-if (c('Database.CharacterEncoding')) {
+if (c('Database.CharacterEncoding') !== 'utf8mb4') {
     removeFromConfig('Database.CharacterEncoding');
     removeFromConfig('Database.ExtendedProperties.Collate');
 }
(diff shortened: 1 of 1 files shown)


linc:
Why do we want to remove the collate? I believe that is the correct one to set.

DaazKu:
No the collate would be `utf8mb4_unicode_ci`

DaazKu:
Long story short if we remove hardcoded `CharacterEncoding` from the configs we need to remove the hardcoded collates too.

linc:
Oh, I was looking in the config-default which is set to the correct one, than misreading your copy/paste.

What's further weird is that I'm seeing this in the open.vf config right now:

        "CharacterEncoding": "utf8",
        "ExtendedProperties": {
            "Collate": "utf8mb4_unicode_ci"
        }

which doesn't make a ton of sense. But anyway, yes, that makes sense we should eliminate both if we're eliminating one. I've amended the PR.

Still unclear on the Travis fail root cause but I assume it's dying on utility/update for some reason.

DaazKu:
I'm gonna approve the changes but we should probably figure out what is breaking before merging.

linc:
I placated Travis by amending the logic to keep him from executing that block on a new install.

PR #5925, opened 2017-09-05
Require PHPUnit ~5.0 as a dev install to allow unit testing

DaazKu:
Install PHPUnit as a development dependency.
This allows for unit testing locally without installing phpunit elsewhere.

Especially useful for https://github.com/vanilla/vanilla-docker


linc:
I have no objection to this.

PR #5926, opened 2017-09-05
Move session initiation in SiteTestTrait

DaazKu:
This allows APIController calls in `setupBeforeClass` methods.
Otherwise you get permissions exceptions.

Example: https://github.com/vanilla/vanilla/blob/fa2b4ea8429b9c052f57edf087d784f0f8aa7939/tests/APIv2/MessagesTest.php#L31

### DIFF
--- a/tests/InternalClient.php
+++ b/tests/InternalClient.php
@@ -84,6 +84,10 @@ class InternalClient extends HttpClient {
     public function setUserID($userID) {
         $this->userID = $userID;
 
+        /* @var \Gdn_Session $session */
+        $session = $this->container->get(\Gdn_Session::class);
+        $session->start($userID, false, false);
+        
         return $this;
     }
 }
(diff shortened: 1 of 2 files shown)


tests/InternalClient.php
```diff
@@ -84,10 +84,6 @@ public function getUserID() {
     public function setUserID($userID) {
         $this->userID = $userID;
 
-        /* @var \Gdn_Session $session */
-        $session = $this->container->get(\Gdn_Session::class);
-        $session->start($userID, false, false);
-
```
initvector: I recommend keeping this part in here. The function allowed you to change users on the current session. Now, it's just setting a property on the object and you're stuck with always being the admin user.

DaazKu: This may be a premature removal but I can already tell that handling the session management there won't be enough. I'll close for now and apply the session start in setupBeforeClass and we can address that issue later then.

PR #5931, opened 2017-09-06
Improves dispatcher errors while unit testing

DaazKu:
I had a really bad time trying to find out what was not working while unit testing the API v2.

This change add the stack trace to reported errors from the new Dispatcher.
With the source of the error the debugging is waaayyy faster.

~The stack trace is only added when we are in the UnitTest environment so that's a good thing!~

The stack is set as part of the Data meta informations which are never displayed on the client side. 
We extract that information in VanillaTests\InternalRequest and set it as a custom header on the `HttpResponse`.
We then check, in `VanillaTests\InternalClient` if the custom header contains an error stack trace.

### DIFF
--- a/library/Garden/Web/Dispatcher.php
+++ b/library/Garden/Web/Dispatcher.php
@@ -139,6 +139,8 @@ class Dispatcher {
         } elseif ($raw instanceof \Exception) {
             $data = $raw instanceof \JsonSerializable ? $raw->jsonSerialize() : ['message' => $raw->getMessage(), 'status' => $raw->getCode()];
             $result = new Data($data, $raw->getCode());
+            // Provide stack trace as meta information.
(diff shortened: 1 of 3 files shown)


DaazKu:
You can see the difference here: 

https://travis-ci.org/vanilla/vanilla/jobs/272547397
vs 
https://travis-ci.org/vanilla/vanilla/jobs/272571249

DaazKu:
We don't want the traces to be returned by the API.
Using Debug is a no go because it causes other problems and unit tests do not run with it turned on.
This is the easiest/cleaned solution I've found.

DaazKu:
@Todd we can't propagate `Data`'s meta over an `HttpResponse`. We have to put the trace in the body which is why I've tested that we are in the unit test environment before doing so.

PR #5932, opened 2017-09-06
Refactor ConversationsModel to accommodate the API Controller

DaazKu:
Allow to get conversations members full data set and add pagination to it!


applications/conversations/models/class.conversationsmodel.php
```diff
@@ -64,19 +64,26 @@ public function checkForSpam($type, $skipSpamCheck = false) {
      * Get all the members of a conversation from the $conversationID.
      *
      * @param int $conversationID The conversation ID.
+     * @param bool $idsOnly The returns only the userIDs or everything from UserConversation.
+     * @param bool $limit
+     * @param bool $offset
      *
      * @return array Array of user IDs.
```
initvector: Now that more fields can be returned, the description for the return tag isn't accurate.

PR #5944, opened 2017-09-11
add pptx for AllowedFileExtensions variable.

torut:
word(.doc) and excel(.xls) are includes .docx and .xlsx extension for AllowedFileExtensions.
but powerpoint(ppt) is not include .pptx extension in AllowedFileExtensions.


linc:
Good observation. Thanks for contributing.

PR #5957, opened 2017-09-13
Add option to hide promoted content module if no results are found

slafleche:
We've had a request to hide the promoted content module when it's empty. I've added an option to do that, but by default we display it. I think this is better, because it'll avoid the situation where someone thinks the module doesn't work, but it's because there's nothing to show.


applications/vanilla/modules/class.promotedcontentmodule.php
```diff
@@ -134,6 +142,7 @@ public function validate($parameters = []) {
      */
     public function getData() {
         $this->setData('Content', false);
+        $this->setData('ShowIfNoResults', $this->ShowIfNoResults);
```
initvector: You shouldn't need this. This is already a property on the object. You don't need to set it in the data array.

applications/vanilla/views/modules/promotedcontent.php
```diff
@@ -1,33 +1,36 @@
...
-        if ($Content):
+$ShowIfNoResults = $this->data('ShowIfNoResults');
 
-            if ($this->Group):
-                $Content = array_chunk($Content, $this->Group);
-            endif;
+$ContentItems = is_array($Content) ? sizeof($Content) : 0;
 
-            foreach ($Content as $ContentChunk):
+if ($ContentItems != 0 || $ShowIfNoResults):
```
initvector: You can condense this down to:

```php
if (!empty($Content) || $this->ShowIfNoResults):
```

`empty` will return `true` for an empty array. `ShowIfNoResults` is already a property on `$this`. You don't need to set `$ContentItems` or `$ShowIfNoResults`.

applications/vanilla/modules/class.promotedcontentmodule.php
```diff
@@ -93,6 +96,10 @@ public function load($parameters = []) {
         }
     }
 
+    public function setShowIfNoResults($val) {
```
initvector: Need a doc block here

applications/vanilla/views/modules/promotedcontent.php
```diff
@@ -1,33 +1,36 @@
...
-?>
-<div class="Box BoxPromoted">
-    <?php echo panelHeading(t('Promoted Content')); ?>
-    <div class="PanelInfo DataList">
-        <?php
-        $Content = $this->data('Content');
-        $ContentItems = sizeof($Content);
-
-        if ($Content):
+$ShowIfNoResults = $this->data('ShowIfNoResults');
```
initvector: This isn't used any more, is it?

applications/vanilla/views/modules/promotedcontent.php
```diff
@@ -1,33 +1,36 @@
...
 
-            if ($this->Group):
-                $Content = array_chunk($Content, $this->Group);
-            endif;
+$ContentItems = is_array($Content) ? sizeof($Content) : 0;
```
initvector: This isn't used any more, either, right?

applications/vanilla/modules/class.promotedcontentmodule.php
initvector: `ShowIfNoResults` is a public property in this class. A setter is kinda pointless. If you want to force a boolean (which is a good idea), I recommend lowering the visibility of this property down to private and making a getter method to fetch the value.

PR #5974, opened 2017-09-22
Allow dismissal of locally announced categories

### LINKED ISSUE #5969: Unable to dismiss announced discussions that are only announced in their own category
Discussions can only be dismissed if they announce in recent discussions and their own categories. 

image

This seems to be due to [this line])(https://github.com/vanilla/vanilla/blob/32fe13d05dee9dc2bf731b942a68eba5bd466c5f/applications/vanilla/views/discussion/helper_functions.php#L340). It only checks for `$discussion->Announce === 1` when  `$discussion->Announce === 2` is also an announced discussion. We should check for both unless this is an intentional feature.

Charrondev:
Closes #5969

Posts announced only in their own category were not dismissible (most likely an oversite). This PR fixes this by checking for both types of announced categories instead of just 1.

Would it be better to check for all >0 values or all non-zero values instead of just additionally adding  the check for 2?


Charrondev:
I've updated my commit and had both `->Announce` and `->Dismissed` evaluate as booleans.

PR #5975, opened 2017-09-23
Add ability to register users with the API

### LINKED ISSUE #5528: Add registration to UsersApiController
### Proposed Solution

1. `public function post_registrationApply(array $body): array`
    * Will require the following fields: Name, Password, Email and DiscoveryText.
    * Use `UserModel::insertForApproval` to save registration.
1. `public function post_registrationBasic(array $body): array`
    * Will require the following fields: Name, Password and Email.
    * Use `UserModel::insertForBasic` to save registration.
1. `public function post_registrationInvitation(array $body): array`
    * Will require the following fields: Name, Password, Email and InvitationCode.
    * Use `UserModel::insertForInvitation` to save registration.

### Notes

1. Only the configured registration method's associated endpoint is active. If registration is set to "basic", then a POST to /api/v2/users/registrationApply is going to respond with an error.
1. Should acceptance of the TOS be a field that gets submitted?
1. CAPTCHA addons may need to be refactored to work with API requests.
1. `UserModel::register` currently only allows the configured registration method for the site. Since we're disabling endpoints that aren't configured, anyway, this probably isn't a big deal. However, adding the flexibility of passing a method override parameter to the function would be easy way to get around this limitation.
1. Extended fields (e.g. Profile Extender) are not supported and will be added in a future revision.
1. Required CAPTCHA values will need to be generated on the client, using the site's CAPTCHA config (e.g. reCAPTCHA keys). The client will be responsible for inserting the script to generate the widget and passing along the user-supplied answer with the rest of the request.

initvector:
This update adds the ability to register users with the API, using the following registration methods: Invitation, Approval and Basic.

1. Invitations can be managed with the /invitations endpoint. Currently, you can create an invitation code, list your invitation codes or delete a code.
1. Applications can be managed with the /applications endpoint. This is where users can apply to the site and site managers can review applications. Applications are essentially just users in the applicant role, but this endpoint attempts to make them appear to be separate entities. The underlying data is still stored in the users table.
1. Basic and invitation-based registrations are now handled with POST requests to the /users/register endpoint.
1. Registration fields have not been changed. Use the same field names you'd use to register with Vanilla's UI.
1. Several underlying functions used by these new controllers have had to be refactored for the sake of addressing errors or making them usable outside of the existing entry and user controllers.
1. There's currently a [kludge in `Garden\Web\Dispatcher`](https://github.com/vanilla/vanilla/blob/df9b7536d3bf12d8a56d3db9f55501107568cefc/library/Garden/Web/Dispatcher.php#L83) to require an access token for API POST requests. This doesn't make a lot of sense for user registrations, so I've added the ability to define exceptions for endpoints.
1. Relevant tests have been added.

### Notes

1. Extended fields (e.g. Profile Extender) are not currently supported.
1. CAPTCHA is supported.
1. To test registrations with CAPTCHA, I recommend creating a simple HTML page with the reCAPTCHA widget included. Make the form action post to the desired endpoint (e.g. /api/v2/applications, /api/v2/users/register)

Closes #5528


applications/dashboard/controllers/Api/ApplicationsApiController.php
```diff
@@ -0,0 +1,221 @@
...
+     *
+     * @param int $id
+     * @param array $body
+     * @throws ClientException if the user record is not an applicant.
+     * @return array
+     */
+    public function patch($id, array $body) {
+        $this->permission('Garden.Users.Approve');
+
+        $in = $this->schema([
```
DaazKu: You should add an extra in schema so that we know that $id is the userID

applications/dashboard/controllers/Api/ApplicationsApiController.php
```diff
@@ -0,0 +1,221 @@
...
+     */
+    public function post(array $body) {
+        $this->permission();
+
+        $in = $this->schema([
+            'email:s' => 'The email address for the user.',
+            'name:s' => 'A username for the user.',
+            'password:s' => 'A password for the user.',
+            'discoveryText:s' => 'Why does the user wish to join?',
+            'termsOfService:b' => 'Were the terms of use accepted?'
```
DaazKu: Same as below: https://github.com/vanilla/vanilla/pull/5975/files#diff-69178cf6662982178befa1af6e5022edR314

applications/dashboard/controllers/Api/UsersApiController.php
```diff
@@ -285,6 +291,72 @@ public function post(array $body) {
...
+
+        $inputProperties = [
+            'email:s' => 'An email address for this user.',
+            'name:s' => 'The username.',
+            'password:s' => 'A password for this user.',
+            'termsOfService:b' => 'Were the terms of use accepted?'
+        ];
+        if ($registrationMethod === 'invitation') {
+            $inputProperties['invitationCode:s'] = 'An invitation code for registering on the site.';
+        } elseif ($this->userModel->isRegistrationSpam($userData)) {
```
DaazKu: We can't change the API spec depending on what is sent to it because we will never be able to generate the documentation properly if we do this.

We should put `discoveryText:s?` in the inputProperties and incidate that this field will be required if the response is a 202. THEN we can remove the ? if the $userData is flagged as spam which will make the validation work as intended.

applications/dashboard/models/class.invitationmodel.php
```diff
@@ -229,14 +249,14 @@ public function delete($invitationID) {
...
+        $result = $this->SQL->delete($this->Name, ['InvitationID' => $id]);
 
         // Add the invitation back onto the user's account if the invitation has not been accepted.
-        if (!$invitation->AcceptedUserID) {
+        if ($result && !$invitation['AcceptedUserID']) {
             Gdn::userModel()->increaseInviteCount($userID);
         }
 
-        return true;
+        return $result;
```
DaazKu: I'm not sure that $result is actually a boolean. Can we test this?

applications/dashboard/models/class.usermodel.php
```diff
@@ -1327,21 +1327,29 @@ public function getApplicantCount() {
     /**
      * Returns all users in the applicant role.
      *
+     * @param int|bool $limit
```
DaazKu: Just int|false here would be better.
You can actually pass `true` but that would be dumb.

applications/dashboard/models/class.usermodel.php
```diff
@@ -3416,26 +3424,11 @@ public function validateUniqueFields($username, $email, $userID = '', $return =
      * Approve a membership applicant.
      *
      * @param int $userID
-     * @param string $email
      * @return bool
      * @throws Exception
      */
-    public function approve($userID, $email) {
```
DaazKu: Deprecate the extra parameter. We'll delete it later.

applications/dashboard/models/class.usermodel.php
```diff
@@ -4921,4 +4914,61 @@ public function clearSectionNavigationPreference($url = '', $userID = '', $reset
...
+
+        $applicantRoleIDs = RoleModel::getDefaultRoles(RoleModel::TYPE_APPLICANT);
+
+        // Make sure the user is an applicant.
+        $roleData = $this->getRoles($userID);
+        if ($roleData->numRows() == 0) {
+            throw new Exception(t('ErrorRecordNotFound'));
+        } else {
+            $appRoles = $roleData->result(DATASET_TYPE_ARRAY);
+            foreach ($appRoles as $appRole) {
```
DaazKu: You can also foreach directly on it. Long story short. Let's use it as an array!

applications/dashboard/models/class.usermodel.php
```diff
@@ -4921,4 +4914,61 @@ public function clearSectionNavigationPreference($url = '', $userID = '', $reset
...
+     * @return bool
+     */
+    public function isApplicant($userID) {
+        $result = false;
+
+        $applicantRoleIDs = RoleModel::getDefaultRoles(RoleModel::TYPE_APPLICANT);
+
+        // Make sure the user is an applicant.
+        $roleData = $this->getRoles($userID);
+        if ($roleData->numRows() == 0) {
```
DaazKu: You can call `count()` on a Gdn_DataSet.
`class Gdn_DataSet implements IteratorAggregate, Countable, JsonSerializable`

applications/dashboard/settings/class.hooks.php
```diff
@@ -654,6 +654,22 @@ public function gdn_dispatcher_appStartup_handler($sender) {
             }
         }
         $this->checkAccessToken();
+
```
DaazKu: @tburry You are gonna have to review this part.

applications/dashboard/controllers/Api/ApplicationsApiController.php
```diff
@@ -0,0 +1,221 @@
...
+     * @param int $id
+     * @param array $body
+     * @throws ClientException if the user record is not an applicant.
+     * @return array
+     */
+    public function patch($id, array $body) {
+        $this->permission('Garden.Users.Approve');
+
+        $in = $this->schema([
+            'status:s' => [
```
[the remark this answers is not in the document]
initvector: I'd like to avoid "user" references under the applications endpoint as much as we can. I know they are technically the same resource, but this endpoint aims to treat them as separate entities.

applications/dashboard/controllers/Api/ApplicationsApiController.php
```diff
@@ -0,0 +1,221 @@
...
+        $out = $this->schema($this->fullSchema(), 'out');
+
+        $row = $this->userByID($id);
+        $this->prepareRow($row);
+
+        if ($this->userModel->isApplicant($id) === false) {
+            throw new ClientException('The application specified is already an active user.');
+        }
+
+        $body = $in->validate($body, true);
```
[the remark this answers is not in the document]
initvector: Sparse validation is standard for PATCH endpoints. I know it doesn't seem necessary here, now, but more fields may be added in the future.

applications/dashboard/controllers/Api/InvitationsApiController.php
```diff
@@ -0,0 +1,203 @@
...
+                'email:s' => 'The email address associated with an invitation.',
+                'code:s' => 'An invitation code.',
+                'status' => [
+                    'description' => 'Current status for the invitation.',
+                    'enum' => ['accepted', 'pending']
+                ],
+                'insertUserID:i' => 'User who created the invitation.',
+                'dateInserted:dt' => 'When the invitation was created.',
+                'acceptedUser:o|n?' => $this->getUserFragmentSchema(),
+                'acceptedUserID:i|n' => 'User who accepted the invitation.',
```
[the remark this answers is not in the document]
initvector: That's modifying the underlying structure of this resource. I agree it would be a nice bit of information to have, but I believe it's outside the scope of this update.

applications/dashboard/controllers/Api/InvitationsApiController.php
```diff
@@ -0,0 +1,203 @@
...
+                'description' => 'The number of items per page.',
+                'default' => 30,
+                'minimum' => 1,
+                'maximum' => 100
+            ],
+            'expand:b?' => 'Expand associated records.'
+        ], 'in')->setDescription('Get a list of invitations for the current user.');
+        $out = $this->schema([
+            ':a' => $this->fullSchema()
+        ]);
```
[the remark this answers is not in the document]
initvector: Good eye!

applications/dashboard/models/class.usermodel.php
initvector: My IDE complains when I give it a specific value, instead of a type. :frowning_face:

applications/dashboard/controllers/Api/ApplicationsApiController.php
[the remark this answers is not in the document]
DaazKu: Yeah that makes sense. I wrote this before realizing that the userID is actually the applicationID.

Still:
```
        $this->schema(['id' => 'Application ID'], 'in');
        $in = $this->schema([
            'status:s' => [
                'description' => 'Current status of the application.',
                'enum' => ['approved', 'declined', 'pending']
            ]
        ], 'in')->setDescription('Modify a user application.');
        $out = $this->schema($this->fullSchema(), 'out');
```

applications/dashboard/controllers/Api/ApplicationsApiController.php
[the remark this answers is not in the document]
DaazKu: Good point.

applications/dashboard/settings/class.hooks.php
tburry: This is how you [bypass a CSRF ban](https://github.com/vanilla/hootsuite/blob/master/addons/hootsuite/controllers/hootsuiteapicontroller.php#L156) for API controllers.

applications/dashboard/models/class.usermodel.php
DaazKu: Welp. Mine doesn't. No big deal..

applications/dashboard/settings/class.hooks.php
initvector: I needed to merge in the update from #5990. Works, now!

applications/dashboard/controllers/Api/InvitationsApiController.php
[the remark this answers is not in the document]
initvector: Structural updates added in a separate PR: #6008

applications/dashboard/controllers/Api/InvitationsApiController.php
```diff
@@ -0,0 +1,229 @@
...
+        $invitationData = $this->caseScheme->convertArrayKeys($body);
+        $row = $this->invitationModel->save($invitationData, ['ReturnRow' => true]);
+        $this->validateModel($this->invitationModel);
+        $this->prepareRow($row);
+
+        $result = $out->validate($row);
+        return new Data($result, 201);
+    }
+
+    public function prepareRow(array &$row) {
```
DaazKu: Missing docblock

PR #5980, opened 2017-09-26
Don't assume we get an array in convo save

linc:
A few lines we account for the deprecated condition of `$settings` being an object. Unfortunately, that fatals this first line.

Closes https://github.com/vanilla/internal/issues/1164


DaazKu:
The unit tests fails and I took care of it: https://github.com/vanilla/vanilla/pull/5981

PR #5989, opened 2017-09-27
[Backport] Allow dismissal of locally announced categories

Charrondev:
Backport of #5974

Do we normally deploy backports like this everyone where or just on an as-needed basis?


linc:
It can be deployed everywhere.

PR #6004, opened 2017-10-02
Fix losing query parameters on homepage

initvector:
[The dispatcher rewrites the full request to the default controller when visiting the homepage.](https://github.com/vanilla/vanilla/blob/8799b3ae7c468649afbf6e28c2dfa392aa0678af/library/core/class.dispatcher.php#L723) This means query parameters used by some addons (e.g. Multilingual's "locale" parameter) can be inaccessible on the homepage.

This update takes a snapshot of the query before the homepage rewrite, then merges in the original query parameters, if there were any.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -722,7 +722,7 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
             $defaultController = Gdn::router()->getRoute('DefaultController');
             $originalGet = $request->get();
             $request->pathAndQuery($defaultController['Destination']);
-            if (is_array($originalGet) && !empty($originalGet)) {
+            if (is_array($originalGet) && count($originalGet) > 0) {
                 $request->setQuery(array_merge($request->get(), $originalGet));
             }
         }
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -720,7 +720,11 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
         } elseif (in_array($request->path(), ['', '/'])) {
             $this->isHomepage = true;
             $defaultController = Gdn::router()->getRoute('DefaultController');
+            $originalGet = $request->get();
             $request->pathAndQuery($defaultController['Destination']);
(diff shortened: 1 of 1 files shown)


library/core/class.dispatcher.php
```diff
@@ -720,7 +720,11 @@ private function rewriteRequest($request) {
         } elseif (in_array($request->path(), ['', '/'])) {
             $this->isHomepage = true;
             $defaultController = Gdn::router()->getRoute('DefaultController');
+            $originalGet = $request->get();
             $request->pathAndQuery($defaultController['Destination']);
+            if (is_array($originalGet) && !empty($originalGet)) {
```
DaazKu: Nitpicking: Since we know that the variable is an array we should probably use `count()`

library/core/class.dispatcher.php
```diff
@@ -720,7 +720,11 @@ private function rewriteRequest($request) {
...
+                $request->setQuery(array_merge($request->get(), $originalGet));
```
DaazKu: We should be using `array_merge($originalGet, $request->get())` to preserve the DefaultController query string.

initvector: That's going to lead to unexpected behavior. If I'm supplying something that's a "default", I wouldn't expect it to overwrite an explicit parameter being passed in. To a user, it would appear as if their query string was being ignored.

DaazKu: The inverse is also true.
If you set a query string parameter in the default route it can be important. Letting someone overwrite it might not be a desired behaviour. It would be safer to not overwrite them. 

If there is a need to overwrite it I would argue that the caller should request the final path with the desired query string parameter.

DaazKu: A non optimal setup could redirect the default page to a page that use a query parameter to do a subsequent redirect. Letting someone overwrite that parameter might lead to an unexpected behaviour.

DaazKu: After discussing that the consensus is that we should not restrict anything because it would be weirdly used.

PR #6018, opened 2017-10-06
Add drafts API endpoint

### LINKED ISSUE #5535: Add the DraftsApiController
### Proposed Solution

1. Draft records returned by this endpoint should include the following fields:
    * `draftID`
    * `recordType` (e.g. "discussion", "comment")
    * `name` (e.g. the discussion title or `null` for a comment)
    * `body`
    * `insertUserID`
    * `dateInserted`
    * `updateUserID`
    * `dateUpdated`
    * `attributes` - This is where most of the current comment- and discussion-specific fields will go. This includes: `CategoryID`, `DiscussionID`, `Closed`, `Announce`, `Sink` and `Format`. Ideally, this will allow us to move to a more generic table format in the future. Simulating it now will make the transition easier, later.
1. `public function index($query): array`
    * Query the `Drafts` table, selecting open drafts for the current user, ordering by `DateInserted`, descending.
    * Allow filtering by `recordType`. Right now, "discussion" is a draft with no valid `DiscussionID` (null or less than 1) and "comment" is a draft with a discussion ID.
    * Paginate results.
1. `public function get($id): array`
    * Get a single draft by its unique ID in the format specified above.
1. `public function post(array $body): array`
    * Use the following fields to create a new draft:
        * `recordType` - enum - Currently, this is just "comment" and "discussion".
        * `name` - string - optional
        * `body` - string
        * `attributes` - object - All the resource-specific fields that may be needed for this draft.
1. `public function patch($id, array $body): array`
    * Edit an existing draft.
    * Parameters same as POST.
1. `public function delete($id): null`
    * Given a draft's unique ID, delete it.

### Notes

Querying a comment draft for a particular discussion isn't very easy with this interface. As it stands, I would need to query `recordType=comment` and then look through the attributes of all the records until I found (or didn't find) a `DiscussionID` value to match for a particular discussion. It might be worth adding `ForeignType` and `ForeignID` fields to proposed structure, so we could "attach" a draft to an existing record, potentially of a different type. I could then fetch a comment draft for discussion \#123 by querying `recordType=comment&foreignType=discussion&foreignID=123`.

initvector:
This update adds support for creating drafts with the API. The specifics of the actual implementation are detailed in the associated issue. The API implementation is what will become of drafts: all resource-specific fields going into the `Attributes` column. Right now, the API is serving as a translation layer. Once the necessary refactoring occurs to make drafts a more universally-usable resource, the translations can be dropped.

`DraftsModel` needed some help, so I've also performed the following updates:

1. Do not always require the `Body` field. This was blocking sparse updating with PATCH. The implemented fix mirrors similar `Body` requirement conditions from `DiscussionModel`.
1. Do not unset the `DiscussionID` if it is a `null` value. The condition appeared to be looking for an empty string, but `null` values were also tripping its loose comparison.
1. `DraftModel::updateUser` was using `DraftModel::getCount`, which is deprecated. It has been updated to use `DraftModel::getCountByUser`, instead.

Closes #5535

### DIFF
--- a/applications/vanilla/controllers/api/DraftsApiController.php
+++ b/applications/vanilla/controllers/api/DraftsApiController.php
@@ -64,7 +64,7 @@ class DraftsApiController extends AbstractApiController {
      */
     public function draftByID($id) {
         $row = $this->draftModel->getID($id, DATASET_TYPE_ARRAY);
-        if (!$row || $row['Deleted'] > 0) {
+        if (!$row) {
             throw new NotFoundException('Draft');
         }
         return $row;
@@ -84,7 +84,7 @@ class DraftsApiController extends AbstractApiController {
                 Schema::parse(
                     ['recordType', 'parentRecordID?', 'attributes']
                 )->add($this->fullSchema()),
-                'DiscussionPost'
+                'DraftPost'
             );
         }
 
@@ -145,6 +145,28 @@ class DraftsApiController extends AbstractApiController {
 
         $in = $this->idParamSchema('in')->setDescription('Get a draft.');
         $out = $this->schema($this->fullSchema(), 'out');
+
+        $row = $this->draftByID($id);
+        if ($row['InsertUserID'] !== $this->getSession()->UserID) {
+            $this->permission('Garden.Moderation.Manage');
+        }
+        $this->prepareRow($row);
+
+        $result = $out->validate($row);
+        return $result;
+    }
+
+    /**
+     * Get a draft for editing.
+     *
+     * @param int $id The unique ID of the draft.
+     * @return array
+     */
+    public function get_edit($id) {
+        $this->permission('Garden.SignIn.Allow');
+
+        $in = $this->idParamSchema('in')->setDescription('Get a draft for editing.');
+        $out = $this->schema(['draftID', 'parentRecordID', 'attributes'], 'out')->add($this->fullSchema());
 
         $row = $this->draftByID($id);
         if ($row['InsertUserID'] !== $this->getSession()->UserID) {
@@ -241,7 +263,7 @@ class DraftsApiController extends AbstractApiController {
         $this->permission('Garden.SignIn.Allow');
 
         $in = $this->draftPostSchema('in')->setDescription('Update a draft.');
-        $out = $this->schema($this->fullSchema(), 'out');
+        $out = $this->schema(['draftID', 'parentRecordID', 'attributes'], 'out')->add($this->fullSchema());
 
         $row = $this->draftByID($id);
         if ($row['InsertUserID'] !== $this->getSession()->UserID) {
@@ -249,15 +271,16 @@ class DraftsApiController extends AbstractApiController {
         }
 
         $body = $in->validate($body, true);
-        $draftData = $this->translateRequest($body);
+        $recordType = !empty('DiscussionID') ? 'comment' : 'discussion';
+        $draftData = $this->translateRequest($body, $recordType);
         $draftData['DraftID'] = $id;
         $this->draftModel->save($draftData);
         $this->validateModel($this->draftModel);
 
-        $row = $this->draftByID($id);
-        $this->prepareRow($row);
+        $updatedRow = $this->draftByID($id);
+        $this->prepareRow($updatedRow);
 
-        $result = $out->validate($row);
+        $result = $out->validate($updatedRow);
         return $result;
     }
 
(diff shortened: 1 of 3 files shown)


applications/vanilla/controllers/api/DraftsApiController.php
```diff
@@ -0,0 +1,394 @@
...
+     * @return array
+     */
+    private function translateRequest(array $body, $recordType = null) {
+        // If the record type is not explicitly defined by the parameters, try to extract it from $body.
+        if ($recordType === null && array_key_exists('recordType', $body)) {
+            $recordType = $body['recordType'];
+        }
+
+        if (array_key_exists('attributes', $body)) {
+            $columns = ['announce', 'body', 'categoryID', 'closed', 'format', 'name', 'sink', 'tags'];
```
DaazKu: I think that these should be fetch from the relevant record's schema.

applications/vanilla/controllers/api/DraftsApiController.php
```diff
@@ -0,0 +1,394 @@
...
+     * List drafts created by the current user.
+     *
+     * @param array $query The query string.
+     * @return array
+     */
+    public function index(array $query) {
+        $this->permission('Garden.SignIn.Allow');
+
+        $in = $this->schema([
+            'recordType:s?' => [
```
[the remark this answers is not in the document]
initvector: Yeah, but recordType is a parameter here. It's used for filtering the type of drafts to be displayed. It's not required. If you don't give it a recordType, it should show you all drafts, regardless of type.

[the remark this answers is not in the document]
DaazKu: Ho yeah my bad. Btw, for consistency we should use "filter" instead of "limit" in the description.

applications/vanilla/controllers/api/DraftsApiController.php
initvector: These are actually just coming off the drafts table. They wouldn't change, based on the resource type. It's pulling whitelisted "drafts" columns and extracting them from submitted `attributes` property values, before saving.

DaazKu: Most of them are bound to the discussion record.
The draft table should not impact the API endpoint.

DaazKu: body, format, name are the only fields that SHOULD be common among all draft types.

initvector: This is for translating a request into a format that can be passed to `DraftModel::save`. It has to follow the drafts table format. When things are refactored and the drafts table is structured like the API schema, then we can ditch this translation layer.

DaazKu: Sorry I got confused.. I was referring to https://github.com/vanilla/vanilla/pull/6018/files/67bb66c03b353c2cebd8295d69ced5e210558003#r143755372

DaazKu: The description and the validation should be done using the recordType schema ^

applications/vanilla/controllers/api/DraftsApiController.php
```diff
@@ -0,0 +1,394 @@
...
+            $schema = Schema::parse([
+                'draftID:i' => 'The unique ID of the draft.',
+                'recordType:s' => [
+                    'description' => 'The type of record associated with this draft.',
+                    'enum' => ['comment', 'discussion']
+                ],
+                'parentRecordID:i|n' => 'The unique ID of the intended parent to this record.',
+                'attributes:o' => [
+                    'description' => 'A free-form object containing all custom data for this draft.',
+                    'properties' => [
```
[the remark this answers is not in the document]
initvector: > The description and the validation should be done using the recordType schema

I don't think that's the route to go either. We can't dynamically change the schema based on the parameters it receives. It's possible a drafts UI might want to stash some information that isn't necessarily a column in the resource's table (a temporary selection of some UI element, etc.). Maybe there are fields we don't want to endorse the usage of in drafts by making them valid attributes for a resource. It's going to get increasingly trickier to make this flexible enough to be a valid for anything, but still including resource-specific validation. I think the simplest route is letting the UI stash whatever it wants.

I agree that the attributes here were too specific. They work great for now, but documenting these doesn't make much sense if the drafts API is going to be more open-ended.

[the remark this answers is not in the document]
DaazKu: The description should say to look for the recordType model.
The validation should be done using the recordType schema but any value could be null.
Wouldn't that be the best solution?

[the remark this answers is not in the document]
initvector: [One of the requirements mentioned on the original task](https://github.com/vanilla/vanilla/issues/5535#issuecomment-330909057) was for the attributes field to "load/save anything given to the draft". I interpreted this to mean *anything*. I am concerned that enforcing additional (and dynamic) validation requirements might work well enough for Vanilla's current capabilities with discussions and comments, but may limit the usefulness for drafts of other resource types in the future.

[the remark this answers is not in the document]
DaazKu: If a recordType is specified we should validate that each field is either null or matching the schema field's type and any extra fields should be left untouched. This would meet all the requirements and give some insight of what should be excepted by the API user.

This could be done in a future iteration though. We could just not validate the attribute data at all for now.

Issue #6015, opened 2017-10-06
AuthenticateAPIController unit tests - Part 1


DaazKu:
Update the test suite to work properly.

- Fix sessions
- Fix bootstraping
...

DaazKu:
Closing in favour or tracking on the PR

PR #6019, opened 2017-10-06
Miscellaneous improvements to testing

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/6015

- Allows to unit tests without a session being started.
- Add missing alias "RequestInterface" for Gdn_Request.
- Call `startAuthenticator()` after the site was installed (otherwise you get an error when calling `$session->end()` because the password authenticator is not loaded.

### DIFF
--- a/tests/APIv2/AbstractAPIv2Test.php
+++ b/tests/APIv2/AbstractAPIv2Test.php
@@ -20,14 +20,32 @@ abstract class AbstractAPIv2Test extends \PHPUnit_Framework_TestCase {
     private $api;
 
     /**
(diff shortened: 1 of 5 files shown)


tests/APIv2/AbstractAPIv2Test.php
```diff
@@ -20,14 +20,32 @@
...
 
     /**
+     * @var bool Set to false before setUp() to skip the session->start();
+     */
+    private $startSessionOnSetup = true;
+
+    /**
+     * Whether to start a session on setUp() or not.
+     *
+     * @param $enabled
```
initvector: Should add a type here.

PR #6025, opened 2017-10-11
Add bearer token in HTTP header for profile requests in OAuth2

nervoustwit:
Vanilla's OAuth2 passes the access token in the body of the request for a user's protected resources from an authenticating server. According to [rfc6750](https://tools.ietf.org/html/rfc6750) this is acceptable, but the preferred way is by passing it in the "Authorization" http request header field, as in:
>    GET /oauth/me HTTP/1.1
     Host: server.example.com
     Authorization: Bearer mF_9.B5f-4.1JqM

This PR will:

- Add a checkbox to the Oauth2 settings in the dashboard to opt into this workflow.
- If the provider is set to BearerToken=true, it will send not send the access_token in the body but send it in the HTTP request headers.
- These changes are fully backwards compatible.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -725,23 +725,30 @@ class Gdn_OAuth2 extends Gdn_Plugin {
     public function getProfile() {
         $provider = $this->provider();
         $uri = $this->requireVal('ProfileUrl', $provider, 'provider');
-        $defaultParams = [
-            'access_token' => $this->accessToken()
-        ];
+        $defaultParams = [];
+        $defaultOptions = [];
 
-        // Either send the Access Token in the GET request or as an Authorization header, depending on the client workflow.
+        // Send the Access Token as an Authorization header, depending on the client workflow.
         if (val('BearerToken', $provider)) {
-            $defaultParams = [];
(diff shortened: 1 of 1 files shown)


derrickeason:
Can this also get backported to production in relatively short order after it's been merged? The most immediate need is on cl412

library/core/class.oauth2.php
```diff
@@ -728,11 +729,19 @@ public function getProfile() {
...
         ];
 
+        // Either send the Access Token in the GET request or as an Authorization header, depending on the client workflow.
+        if (val('BearerToken', $provider)) {
+            $defaultParams = [];
+            $profileRequestOptions = array_merge($this->getProfileRequestOptions(), ['Authorization-Header-Message' => 'Bearer '.$this->accessToken()]);
+        } else {
+            $profileRequestOptions = $this->getProfileRequestOptions();
+        }
+
```
linc: This all reads very confusedly for what it's doing.

Why let `$defaultParams` be set then unset it? 

Why put `$profileRequestOptions = $this->getProfileRequestOptions()` in an else block rather than just doing that first either way?

You're twisting together logic here that should just be 1 nice if block.

PR #6028, opened 2017-10-11
Revise Travis build process

### LINKED ISSUE #5815: Travis CI build failing on Trusty
Travis CI is gradually making Trusty the default distro for builds. Vanilla's Travis build is not currently compatible with the Trusty distro. The following error is encountered:

> Syntax error or access violation: 1044 Access denied for user 'travis'@'%' to database 'vanilla_test'

Related: #5637

initvector:
This update performs the following updates to the Travis build process:

1. Stop forcing Ubuntu Precise. This LTS release is no longer supported and Travis is phasing it out in favor of the next LTS release: Ubuntu Trusty.
1. Comment the build file.
1. Reorganize the build file (mainly alphabetizing).
1. Move Composer installation/updating from the install step to the before_script step. The install step is typically used for system dependencies and the before_script setup is used to prepare for the build script to be executed.
1. Cache Composer package files.
1. Stop caching apt (Debian) packages. [Travis's caching documentation](https://docs.travis-ci.com/user/caching/#Things-not-to-cache) recommends these types of files should be avoided, because there's no real benefit.
1. Move linting, code coverage generation and phpUnit into a separate script.
1. Do not perform linting and code coverage generation unless the current branch is "master".
1. Remove testing against HHVM.

Closes #5815

### DIFF
--- a/tests/travis/main.sh
+++ b/tests/travis/main.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 cd $TRAVIS_BUILD_DIR
 
--- a/tests/travis/main.sh
+++ b/tests/travis/main.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+cd $TRAVIS_BUILD_DIR
+
+DO_COVERAGE=false
+DO_LINT=false
+
+if [ "$TRAVIS_BRANCH" = "master" ]; then
+    DO_COVERAGE=true
+    DO_LINT=true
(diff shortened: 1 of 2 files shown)


tests/travis/main.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
DaazKu: https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

PR #6030, opened 2017-10-11
Add filtering by status in conversations participants endpoint

### LINKED ISSUE #5950: Address Conversation's participants "deleted" flag issue
This "deleted" value means that a participant "left the conversation".
The flag's name does not reflect its current meaning.

Also note that get_participants returns every participant even if they left. Which is bizarre at first sigh. (The conversation's countParticipants does not count deleted participants)

@tburry has his view on how he wants us to address that.

initvector:
This update makes a couple modifications to retrieving conversation participants via the API:

1. Allow filtering by a "status" parameter:
    * participating: active users in the conversation
    * deleted: users who have left the conversation
    * all (default): all users who are, or were, part of the conversation
1. Change the "deleted" attribute to "status" when getting a list of participants for a conversation. The value of this new field can be either "participating" or "deleted", where "deleted" is someone who has left the conversation.

Closes #5950


applications/conversations/controllers/api/ConversationsApiController.php
```diff
@@ -207,6 +207,11 @@ public function get_participants($id, array $query) {
         $this->idParamSchema();
 
         $in = $this->schema([
+            'status:s' => [
```
DaazKu: If there's a default the parameter should be optional.

applications/conversations/models/class.conversationsmodel.php
```diff
@@ -67,16 +67,21 @@ public function checkForSpam($type, $skipSpamCheck = false) {
...
 
         $userConversation = new Gdn_Model('UserConversation');
-        $userMembers = $userConversation->getWhere([
-            'ConversationID' => $conversationID
-        ], 'UserID', 'asc', $limit, $offset)->resultArray();
+        $where = ['ConversationID' => $conversationID];
+        if ($active === true) {
+            $where['Deleted'] = 0;
+        } elseif ($active === false) {
+            $where['Deleted >'] = 0;
```
DaazKu: I know it works but this is implying that Deleted **COULD BE** different than 1 or 0 which would be anormal.
We should put a note highlighting that or use `$where['Deleted'] = 1;`.
The latter is also more resilient to changes, if we decide to make Deleted = 2 a thing. (Doesn't make sense to present me though.)

FYI: In general GT is also bad for performance but since we are already filtering by conversation it is not a real concern in this case.

DaazKu:
There's a conflict

PR #6039, opened 2017-10-16
Improve flyout accessibility

slafleche:
# This PR is to improve flyout accessibility in Vanilla. 

The changes in this PR should not affect the UI visually. The states are set after the JS has opened or closed. I made it this way because there's multiple ways we handle flyouts in Vanilla and I wanted to minimize the risk of breaking existing JS and allowed it to easily be used elsewhere. The functions used are added to the "gdn" object for reuse.

### Flyout toggles:
* get "aria-haspopup" set to true
*  tabindex set to "0" -This will allow you to tab through spans (like in the editor) and through links (some FF and Safari don't allow you by default to tab through links).
* "role" set to "button" (they are not links)
* "aria-expanded" set to true or false, depending on the state of the flyout

### Flyouts get:
* "aria-hidden" set to true or false, depending on the state of the flyout

### Links (inside Flyouts get):
* Tabindex set to 0 to allow proper tabbing

____
Also note, most of our themes don't have proper focus styles. For testing, I suggest you throw in this bit of CSS:
```
*:focus {
  outline: solid orange 2px;
}
```


js/global.js
```diff
@@ -774,21 +778,32 @@ jQuery(document).ready(function($) {
...
     $(document).delegate('.ButtonGroup > .Handle', 'click', function() {
-        var buttonGroup = $(this).closest('.ButtonGroup');
-        if (buttonGroup.hasClass('Open')) {
+        var $buttonGroup = $(this).closest('.ButtonGroup');
+        if ($buttonGroup.hasClass('Open')) {
             // Close
-            $('.ButtonGroup').removeClass('Open');
+            $('.ButtonGroup').each(function(){
+                $(this).removeClass('Open');
+                $(this).setFlyoutAttributes();
```
tburry: Shouldn't both these calls work on all matched elements and thus not need the `each()`?

js/global.js
```diff
@@ -774,21 +778,32 @@ jQuery(document).ready(function($) {
...
+            });
+
         } else {
             // Close all other open button groups
-            $('.ButtonGroup').removeClass('Open');
+            $('.ButtonGroup').each(function(){
```
tburry: Shouldn't both these calls work on all matched elements and thus not need the `each()`?

js/global.js
```diff
@@ -844,14 +862,19 @@ jQuery(document).ready(function($) {
...
     });
 
     $(document).delegate(document, 'click', function() {
         if (lastOpen) {
             $('.Flyout', lastOpen).hide();
             $(lastOpen).removeClass('Open').closest('.Item').removeClass('Open');
         }
-        $('.ButtonGroup').removeClass('Open');
+
+        $('.ButtonGroup').each(function(){
```
tburry: Shouldn't both these calls work on all matched elements and thus not need the `each()`?

js/global.js
```diff
@@ -2102,20 +2125,65 @@ if (typeof String.prototype.trim !== 'function') {
...
+    accessibleFlyoutHandle: function (isOpen) {
+        $(this).attr('aria-expanded', isOpen.toString());
+    },
+
+    accessibleFlyout: function (isOpen) {
+        $(this).attr('aria-hidden', (!isOpen).toString());
+    },
+
+    setFlyoutAttributes: function () {
+        var $wrapper = $(this);
```
tburry: This should probably have an `each()` to handle multiple flyouts being selected.

library/core/functions.general.php
```diff
@@ -309,7 +309,7 @@ function attribute($name, $valueOrExclude = '') {
             $exclude = $valueOrExclude;
         }
         foreach ($name as $attribute => $val) {
-            if (empty($val) || ($exclude && stringBeginsWith($attribute, $exclude))) {
+            if ((empty($val) && $val !== '0') || ($exclude && stringBeginsWith($attribute, $exclude))) {
```
tburry: We want the numeric zero to also render, so:

```php
!in_array($val, [0, '0'], true)
```

plugins/editor/views/editor.php
```diff
@@ -40,11 +54,13 @@
...
       }
 
+
+
       switch ($button['action']) {
 
          case 'link':
             $html_toolbar .= wrap(
-               wrap($html_arrow_down, 'span', $button['attr']).''.
+               wrap($html_arrow_down . $screenReaderMarkup, 'span', $button['attr']).''.
```
tburry: No spaces around the `.` operator is our coding standard.

plugins/editor/views/editor.php
```diff
@@ -57,7 +73,7 @@
 
          case 'image':
             $html_toolbar .= wrap(
-               wrap($html_arrow_down, 'span', $button['attr']).''.
+               wrap($html_arrow_down . $screenReaderMarkup, 'span', $button['attr']).''.
```
tburry: No spaces around the `.` operator is our coding standard.

plugins/editor/views/editor.php
```diff
@@ -8,14 +8,28 @@
    : 'editor-desktop';
 
 $html_toolbar = '<div class="editor editor-format-'.$format.' '.$css_ismobile_class.'">';
-$html_arrow_down = '<span class="icon icon-caret-down"></span>';
+$html_arrow_down = '<span class="icon icon-caret-down" aria-hidden="true"></span>';
 $editor_file_input_name = $this->data('_editorFileInputName');
 
 foreach ($this->data('_EditorToolbar') as $button) {
+   $screenReaderMarkup = '<span class="sr-only">' . $button['attr']['title'] . '</span>';
```
tburry: We only put a `$` in front of jQuery variables in javascript.

plugins/editor/views/editor.php
```diff
@@ -69,7 +85,7 @@
          case 'fileupload':
             $accept = $this->data('Accept', '');
             $html_toolbar .= wrap(
-                wrap($html_arrow_down, 'span', $button['attr']).''.
+                wrap($html_arrow_down . $screenReaderMarkup, 'span', $button['attr']).''.
```
tburry: No spaces around the `.` operator is our coding standard.

plugins/editor/views/editor.php
```diff
@@ -85,7 +101,7 @@
          case 'imageupload':
             $accept = $this->data('AcceptImage', '');
             $html_toolbar .= wrap(
-                wrap($html_arrow_down, 'span', $button['attr']).''.
+                wrap($html_arrow_down . $screenReaderMarkup, 'span', $button['attr']).''.
```
tburry: No spaces around the `.` operator is our coding standard.

plugins/editor/views/editor.php
```diff
@@ -135,7 +151,7 @@
             $colorOptions = $textColorOptions.$highlightColorOptions;
 
             $html_toolbar .= wrap(
-               wrap($html_arrow_down, 'span', $button['attr']).''.
+               wrap($html_arrow_down . $screenReaderMarkup, 'span', $button['attr']).''.
```
tburry: No spaces around the `.` operator is our coding standard.

plugins/editor/views/editor.php
```diff
@@ -144,7 +160,7 @@
          // All other dropdowns (color, format, emoji)
          default:
             $html_toolbar .= wrap(
-               wrap($html_arrow_down, 'span', $button['attr']).''.
+               wrap($html_arrow_down . $screenReaderMarkup, 'span', $button['attr']).''.
```
tburry: No spaces around the `.` operator is our coding standard.

plugins/editor/views/editor.php
tburry: Oh whoops, sorry!

PR #6044, opened 2017-10-16
Use fallback for translations of approval messages

Charrondev:
Add fallback translations for approval messages. The existing definitions are the same and we need to order translations for them, so instead of ordering multiple of the same translation, we fall back to the default, but still allow it to be overridden on a granular level in the future.

Goes with https://github.com/vanilla/locales/pull/47


Charrondev:
I've updated this PR to just directly use what used to be the fallback strings. We already have them in transifex so it saves some time/money there too.

https://github.com/vanilla/locales/blob/master/tx-source/site_core.php#L1618-L1620

PR #6055, opened 2017-10-18
Fix issue with namespacing classes during lookup in AddonManager::lookupByClassname

tburry:
Fixes https://github.com/vanilla/vanilla/issues/6065

Return the correct Addon if multiple addons have the same name in different namespaces.


tburry:
@DaazKu I added a failing unit test to illustrate the bug.

PR #6058, opened 2017-10-19
Fix tests for conversation and message API v2 endpoints

### LINKED ISSUE #6061: Fix conversation and message unit tests
After a security patch was applied the unit test for the conversation and message endpoint fails.

Unit tests should be rewritten to test the endpoints with the Conversations.Moderation.Allow configuration turned ON and OFF.

DaazKu:
Fixes #6061 

- Add the new ConfigurationException
   - `permissionException()` should not have been used in APIControllerr
- Fix ConversationModel::inConversation function.
  - This was a result of bad merges and the patch repo not being in sync with the master repo.
- Fix inability to disable floodControl check on ConversationMessageModel
  - If the user was not an administrator `setFloodControlEnabled()` was called which was overriding any prior call to it. This was preventing us from disabling the floodControl check on the model for API tests.
- Update the Conversation/Message test suites to test with and without AllowModeration enabled.

Backport to `release/2.5a`

### DIFF
--- a/library/Vanilla/Exception/ConfigurationException.php
+++ b/library/Vanilla/Exception/ConfigurationException.php
@@ -16,14 +16,60 @@ class ConfigurationException extends ForbiddenException {
     /**
      * Construct a {@link ConfigurationException} object.
      *
-     * @param string The configuration that failed the test.
+     * @param array The configurations that failed the test.
      * @param array $context Additional information for the error.
+     *   - You can set $context['configurationsRequiredValues'][{name}] = value|array
      */
-    public function __construct($configuration, array $context = []) {
-        $context['configuration'] = $configuration;
+    public function __construct($configurations, array $context = []) {
+        if (!is_array($configurations)) {
+            $configurations = [$configurations];
+        }
 
-        $msg = sprintf('The %s configuration value disallow you to do that.', $configuration);
+        $context['configurations'] = $configurations;
 
-        parent::__construct($msg, $context);
+        $messageParts = [];
+        $messageParts[] = t('The site is not configured to support the current action.');
+
+        $or = t('or');
+        foreach ($configurations as $configurationName) {
+            if (isset($context['requiredValues']) && array_key_exists($configurationName, $context['requiredValues'])) {
+                $requiredValues = $context['requiredValues'][$configurationName];
+                if (!is_array($requiredValues)) {
+                    $requiredValues = [$requiredValues];
+                }
+                array_walk($requiredValues, function(&$value) { $value = $this->translateValue($value); });
+
+                $messageParts[] = sprintft("The $configurationName config must be set to %s to support the current action.", implode(" $or ", $requiredValues));
+            } else {
+                $messageParts[] = t("The $configurationName config is required to access this.");
+            }
+        }
+
+        parent::__construct(implode(' ', $messageParts), $context);
+    }
+
+    /**
+     * Translate values to human readable format.
+     *
+     * @param mixed $value
+     * @return string
+     */
+    protected function translateValue($value) {
+        if (is_string($value)) {
+            $value = "\"$value\"";
+        } elseif ($value === true) {
+            $value = 'true';
+        } elseif ($value === false) {
+            $value = 'false';
+        } elseif ($value === null) {
+            $value = 'null';
+        } elseif (is_array($value)) {
+            foreach ($value as &$content) {
+                $content = $this->translateValue($content);
+            }
+            $value = str_replace(["\n    ", "\n", 'Array(, '], [', ', '', 'Array('], print_r($value, true));
+        }
+
+        return (string)$value;
     }
 }
--- a/library/Vanilla/Exception/ConfigurationException.php
+++ b/library/Vanilla/Exception/ConfigurationException.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * @author Alexandre (DaazKu) Chouinard <alexandre.c@vanillaforums.com>
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license http://www.opensource.org/licenses/gpl-2.0.php GNU GPL v2
+ */
+
+namespace Vanilla\Exception;
+
+use Garden\Web\Exception\ForbiddenException;
+
+/**
+ * An exception tha represents a configuration test failing.
+ */
+class ConfigurationException extends ForbiddenException {
+    /**
+     * Construct a {@link ConfigurationException} object.
+     *
+     * @param string The configuration that failed the test.
+     * @param array $context Additional information for the error.
+     */
+    public function __construct($configuration, array $context = []) {
+        $context['configuration'] = $configuration;
+
+        $msg = sprintf('The %s configuration value disallow you to do that.', $configuration);
+
+        parent::__construct($msg, $context);
+    }
+}
--- a/library/Vanilla/Exception/ConfigurationException.php
+++ b/library/Vanilla/Exception/ConfigurationException.php
@@ -16,59 +16,23 @@ class ConfigurationException extends ForbiddenException {
     /**
      * Construct a {@link ConfigurationException} object.
      *
-     * @param array The configurations that failed the test.
+     * @param string The configurations that failed the test.
      * @param array $context Additional information for the error.
-     *   - You can set $context['configurationsRequiredValues'][{name}] = value|array
+     *   - You can set $context['configurationValue'] = value; to specify that the configuration requires a specific value to be set.
      */
-    public function __construct($configurations, array $context = []) {
-        if (!is_array($configurations)) {
-            $configurations = [$configurations];
+    public function __construct($configuration, array $context = []) {
+        $context['configuration'] = $configuration;
+
+        if (array_key_exists('configurationValue', $context)) {
+            $message = sprintft(
+                "The %s config must be set to %s to support the current action.",
+                $configuration,
+                json_encode($context['configurationValue'], JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE)
+            );
+        } else {
+            $message = sprintft("The %s config is required to support the current action.", $configurationName);
         }
 
-        $context['configurations'] = $configurations;
-
-        $messageParts = [];
-
-        $or = t('or');
-        foreach ($configurations as $configurationName) {
-            if (isset($context['requiredValues']) && array_key_exists($configurationName, $context['requiredValues'])) {
-                $requiredValues = $context['requiredValues'][$configurationName];
-                if (!is_array($requiredValues)) {
-                    $requiredValues = [$requiredValues];
-                }
-                array_walk($requiredValues, function(&$value) { $value = $this->translateValue($value); });
-
-                $messageParts[] = sprintft("The $configurationName config must be set to %s to support the current action.", implode(" $or ", $requiredValues));
-            } else {
-                $messageParts[] = t("The $configurationName config is required to support the current action.");
-            }
-        }
-
-        parent::__construct(implode(' ', $messageParts), $context);
-    }
-
-    /**
-     * Translate values to human readable format.
-     *
-     * @param mixed $value
-     * @return string
-     */
-    protected function translateValue($value) {
-        if (is_string($value)) {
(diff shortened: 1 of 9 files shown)


DaazKu:
I concur. A user encountering the error can ask the site owner why that specific error is happening and it will be relevant since the actual configuration is part of the message.
It follows the same principle that was applied to the permissions.

initvector:
The checks/exceptions are used so often that it isn't as feasible to make a custom error message for all of them. If we could potentially throw a config check on every page load, I'd agree it should follow the same basic messaging as permissions. However, in the case of a config check, it's pretty straightforward to give a friendlier error message. At this point, a config exception seems unnecessary and not as flexible as using an existing exception class with better messaging.

DaazKu:
As a non developer a "friendlier" message will tell you nothing relevant about why you are getting an error. 

Stating the configuration that causes the request to fails helps to actually understand the cause of the error. Same goes for permissions.

- As a user, I usually do not care about the meaning of the error if copy/pasting it to the support team/administrator gives me insight about what is the problem.
- As a support team member/administrator I want the error message to be as precise as possible so that I have a good idea of what the cause it. Having a configuration exception stating what configuration was the cause of the error is golden.
   - This is especially relevant for our cloud hosted customers that will likely pass the error to the support team.

I would also point out that the API is NOT a frontend component any users will hit in the wild so I do not see any problem with a ConfigurationException being thrown.

initvector:
> As a non developer a "friendlier" message will tell you nothing relevant about why you are getting an error.

As a non-developer, if I saw "the site isn't configured to allow conversation moderation", it would be a lot easier for me to digest than "The Conversations.Moderation.Allow configuration value disallow you to do that".

> Stating the configuration that causes the request to fails helps to actually understand the cause of the error. Same goes for permissions.

I feel like this is backwards. With the config exception error messages, I need to glean what a config does from its name and then figure out where I can change that in the dashboard. If a site admin doesn't have direct access to the config, a configuration name isn't going to tell them as much or be as helpful as stating what that config does.

Permission names map directly to the UI. It's a lot easier to change these in the dashboard, if you have a config name. Permissions are also boolean. Config values can be anything. Any usefulness of the config exception is almost entirely limited to `true`/`false` configs.

> As a user, I usually do not care about the meaning of the error if copy/pasting it to the support team/administrator gives me insight about what is the problem.

As a user, I care about the meaning of the error. I prefer knowing what's going on, instead of getting an obscure error message, even if I'm just passing the error on to support.

> As a support team member/administrator I want the error message to be as precise as possible so that I have a good idea of what the cause it. Having a configuration exception stating what configuration was the cause of the error is golden.

If I'm managing a site, but don't have direct access to the config, a config name isn't going to be that helpful. Telling me a specific setting in an application is limiting my request and why will be easier to follow when trying to locate it in the dashboard.

> This is especially relevant for our cloud hosted customers that will likely pass the error to the support team.

I agree this is especially relevant to cloud-hosted customers. Instead of passing every config-name error to support, they have a chance at fixing it themselves in the dashboard if their users give them an easy-to-understand, informative error message when the config is blocking an action.

> I would also point out that the API is NOT a frontend component any users will hit in the wild so I do not see any problem with a ConfigurationException being thrown.

Never say never. Vanilla is used by an awful lot of people in the open source community. If someone builds a custom frontend using the API (as I believe there are plans to do), they're likely to encounter these configuration exceptions.

Right now, there's only two: `ServerException` (something unexpected happen on the server) and `ClientException` (there was a problem with contents of the request). A `ClientException` explaining the request was invalid due to a setting on the site (e.g. "the site is is not configured to support X") seems like a better idea than a new exception specifically for config "errors".

DaazKu:
> As a non-developer, if I saw "the site isn't configured to allow conversation moderation", it would be a lot easier for me to digest than "The Conversations.Moderation.Allow configuration value disallow you to do that".

I concede that the sentence could be improved.

> I feel like this is backwards. With the config exception error messages, I need to glean what a config does from its name and then figure out where I can change that in the dashboard. If a site admin doesn't have direct access to the config, a configuration name isn't going to tell them as much or be as helpful as stating what that config does.

Some configuration are not accessible via the dashboard and this is one of them.
In that case it makes more sense to put the name of the configuration in the message for multiple reasons:
- If you don't understand what the error is about you are not the intended target. Pass the error up the chain.
- If you are an admin you can guess pretty easily where the configuration can be changed in the dashboard. (more on that below) 
- If you are an admin of your OSS version and you do not know how to play with your configuration file you will learn a new thing!
- If you are an admin on our cloud-hosted platform, the support team will be able to help you pretty fast with the problem since the relevant information is part of the message.
  - Ideally the endpoint documentation should state that some actions are configurable and can throw an exception.

> Permission names map directly to the UI. It's a lot easier to change these in the dashboard, if you have a config name. Permissions are also boolean. Config values can be anything. Any usefulness of the config exception is almost entirely limited to true/false configs.

Good point.

The permission error message is irrelevant to anyone but administrators and maybe moderators though.
The intended target for a permission errors are these 2 groups.
It is the same for a ConfigurationException.

> As a user, I care about the meaning of the error. I prefer knowing what's going on, instead of getting an obscure error message, even if I'm just passing the error on to support.

As said above, the sentence could be improved but if you do not understand an error message it's because you probaly cannot do anything about it. Pass it up the chain.
Not all errors should be aimed to be understood by an end-user.

> If I'm managing a site, but don't have direct access to the config, a config name isn't going to be that helpful. Telling me a specific setting in an application is limiting my request and why will be easier to follow when trying to locate it in the dashboard.

That's not true. Normally, configurations are specifically named so that it is easy to guess where they can be changed.
Conversations.Moderation.Allow would probably be a moderation option in the conversation section of the dashboard.

> I agree this is especially relevant to cloud-hosted customers. Instead of passing every config-name error to support, they have a chance at fixing it themselves in the dashboard if their users give them an easy-to-understand, informative error message when the config is blocking an action.

That's not true either. You do have a chance to fix the error yourself since you can easily guess where in the dashboard that configuration should be located.

Anyway... It is not applicable in this case since you can't fix this unless you have access to the configuration.

> Never say never. Vanilla is used by an awful lot of people in the open source community. If someone builds a custom frontend using the API (as I believe there are plans to do), they're likely to encounter these configuration exceptions.
>
> Right now, there's only two: ServerException (something unexpected happen on the server) and ClientException (there was a problem with contents of the request). A ClientException explaining the request was invalid due to a setting on the site (e.g. "the site is is not configured to support X") seems like a better idea than a new exception specifically for config "errors".

ConfigurationException is actually a ClientException.

The intended audience of the API are **developpers** so the error message you be crafted to be meaningful to them. More on that below.

There is two way to go about error reporting when you use an API. 
1. You print the API errors directly to your users.
1. You are nice and you handle the API errors and print them a more friendly version of it. (That's pretty much what we do for every API that we use in plugins)
For the first option to be end-user friendly

-----

I think that having a configuration error is correct but the actual message could use some improvements.

I don't think that we should always have end-users friendly exception reported by the API since the end-user is not the direct consumer of the API.
A good error should be self explainatory to the **intended audience** and the thing is that the API audience are **developers** and trying to make a configuration error friendly to the end-user, IMO, will actually causes more problem to the intended audience than anything.

That being said, some errors are pretty self explanatory to everyone and that's a good thing. Example: Discussion not found.

I don't think we have a specific way of reporting errors yet.
Maybe we will decide to have all errors be friendly to the end-users but for that to happen we need to have way better documentation and have error logging that is accessible to the **developers** which are the real user of the API. That's not something that is available right now.


We could also endup having the API report errors with 2 messages. One low level message and one end-user friendly.
This would allow **developpers** to have more insight on what is going on and they can forward the end-user friendly message to the UI.

TLDR: The main consummer of the API will be **developpers** and the errors should as relevant as possible to them. If the error message is not relevant to the consummer it indicates that it should be passed up the chain.

tburry:
My preference is generally for friendlier error messages. The API will be surfaced to users and I think that error message shouldn't be too scary. We can add extra structured information that comes out in the JSON if necessary. Also remember that we can add documentation with specific friendly error messages and then Google will find them for more context. That's how we troubleshoot tools all the time.

WRT implementation, we need some way to bypass these things for the OpenAPI. This could be helped by a structured exception that has an array of keys/values required where the OpenAPI could then temporarily set those and try the endpoint again. I'm also okay with just kludging those few endpoints in the addon to make things easier. We don't have too many endpoints that get turned off, nor do we want that.

library/Vanilla/Exception/ConfigurationException.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Construct a {@link ConfigurationException} object.
+     *
+     * @param string The configuration that failed the test.
+     * @param array $context Additional information for the error.
+     */
+    public function __construct($configuration, array $context = []) {
+        $context['configuration'] = $configuration;
+
+        $msg = sprintf('The %s configuration value disallow you to do that.', $configuration);
```
tburry: Some configurations must have a specific value such as registration method. I imagine this exception would be nice if you could

1. Specify a required value which defaults to true.
2. Display an error message that nicely works for both cases.
3. Add a t() wrap so that we can make adjustments to non-friendly error messages if we want to.

```php
if ($value === true) {
    $msg = t("The $configuration config is required to access this.");
} else {
   $message = sprintf(t("The $configuration config must be \"%s\" to access this."), $this->translateValue($value));
}
```

tburry:
Just want to point out my notes here: https://github.com/vanilla/vanilla/pull/6058/files/a390a2e332ef3973e0caabd78c5fc02714e54a88#r145999258

DaazKu:
Without context:
```php
            throw new ConfigurationException('Conversations.Moderation.Allow');
```
> The site is not configured to support the current action. The Conversations.Moderation.Allow config is required to access this.

With context:
```php
            throw new ConfigurationException(
                'Conversations.Moderation.Allow',
                ['requiredValues' => ['Conversations.Moderation.Allow' => true]]
            );
```
> The site is not configured to support the current action. The Conversations.Moderation.Allow config must be set to true to support the current action.

Complexe usage:
```php
            throw new ConfigurationException(
                [
                    'Conversations.Moderation.Allow',
                    'Conversations.Moderation.Allow2',
                    'Conversations.Moderation.Allow3',
                    'Conversations.Moderation.Allow4',
                ],
                [
                    'requiredValues' => [
                        'Conversations.Moderation.Allow' => true,
                        'Conversations.Moderation.Allow2' => ['string', 1.9, true, false],
                        'Conversations.Moderation.Allow3' => ['array' => ['blabla' => false], null],
                    ],
                ]
            );
```
> The site is not configured to support the current action. The Conversations.Moderation.Allow config must be set to true to support the current action. The Conversations.Moderation.Allow2 config must be set to "string" or 1.9 or true or false to support the current action. The Conversations.Moderation.Allow3 config must be set to Array([blabla] => false) or null to support the current action. The Conversations.Moderation.Allow4 config is required to access this.

I'll create a task to maybe move the translate function somewhere more accessible so that we can reuse it and unit test it.

DaazKu:
@tburry If that's good for you we can merge.

library/Vanilla/Exception/ConfigurationException.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Construct a {@link ConfigurationException} object.
+     *
+     * @param array The configurations that failed the test.
+     * @param array $context Additional information for the error.
+     *   - You can set $context['configurationsRequiredValues'][{name}] = value|array
+     */
+    public function __construct($configurations, array $context = []) {
+        if (!is_array($configurations)) {
+            $configurations = [$configurations];
```
tburry: FYI: This is the same as:

```php
$configurations = (array)$configurations;
```

DaazKu:
I update the code to simplify the ConfigurationException.
- Use json_encode on the value
- Only allow one configuration to be set. (If there is a case where multiple configurations are incorrect we will need to throw multiple errors which will imply that someone will have to fix the first one before seeing the other)
- Make the strings properly translatable.

tburry:
I decided to [simplify the configuration exception](https://github.com/vanilla/vanilla/pull/6058/commits/667ec03f2b74f61bf5edd6af02d668c5336ca865). My reasoning is on the commit.

Issue #5539, opened 2017-05-08
API v2 Permissions


tburry:
Permissions are due for an overhaul in Vanilla and a good starting point is implementing the desired changes with the API.  Here's an overview of the proposed changes:

### Overview

1. A standard naming convention is needed. Right now, permissions are three levels deep and there isn't much in the way of standardization. The proposed format is: [resource].[action] (e.g. discussions.add). Addon names should not be included in the permission.
1. Permission names should be camel-case. There aren't a lot of two-word permission segments, but where there are, they need to be properly cased (e.g. resources.longAction).
1. The pluralized version of a resource should be used in the permission (e.g. "discussions", not "discussion")
1. The goal will be to try to match up permission actions with HTTP verbs, as closely as possible. If a POST is required, the permission would use the "add" action (e.g. discussions.add).
    * CRUD verbs
        * view: GET
        * add: POST
        * edit: PATCH
        * delete: DELETE
    * Standard combined verbs. These verbs can encompass multiple operations.
        * allow: Regular member operations
        * moderate: Moderator operations
        * manage: Administrator operations
    * If an action doesn't fit with a standard verb then creating a verb for a permission is fine, but should be done with caution.
1. Default category permissions should be represented by 0, not -1, as they are in the database.
1. **The initial version of the permission renaming will apply to the API and permission administration only.** The controllers will still request their permissions using the old permission names at this time. If/when we update all permissions then calls to the permissions should be backwards compatible which will require some optimization work seperate from this process.

### Format

The structure of permissions data used with the API should adhere to a standard format. The current proposed format is:

```json
{
  "permissions": {
    "global": {
      "reactions.negative.add": false,
      "discussions.add": false,
      "comments.add": true
  },
  "categories": [
    {
      "id": 123,
      "permissions": {
        "discussions.add": true,
        "comments.add": true
      }
    }
  ]
}
```

Global permissions go under the "global" key. Resource-specific permissions go under a key with the name of the resource.

### Deprecated Permissions

Some current permissions will be deprecated or at least changed enough to not be included in the API at this time. The API should not allow these permissions to be set, nor should it return them in any responses. In essence, deprecated permissions are invisible to the API.

The current list of proposed deprecated permissions:

- Garden.Activity.Delete
- Garden.Activity.View
- Garden.SignIn.Allow
- Garden.Curation.Manage
- Vanilla.Approval.Require
- Vanilla.Comments.Me

### Updated Permission Names

[...]

PR #6063, opened 2017-10-20
Add simplified permissions roles API output

### LINKED ISSUE #6062: Implement permission translation in roles API output
A new, simplified format for permissions was proposed in #5539. To use this format, a method of translating old-to-new needs to be added to the output of the roles API controller.

initvector:
An overhaul of the structure of permissions was proposed in #5539. This update implements that permission structure in roles API responses.

Closes #6062

### DIFF
--- a/applications/dashboard/controllers/api/RolesApiController.php
+++ b/applications/dashboard/controllers/api/RolesApiController.php
@@ -121,9 +121,8 @@ class RolesApiController extends AbstractApiController {
             }
 
             $result[$name] = $val;
-            unset($currentPerm, $pass, $fail);
+            unset($currentPerm, $pass);
         }
-        unset($perms);
 
         return $result;
     }
--- a/applications/dashboard/controllers/api/RolesApiController.php
+++ b/applications/dashboard/controllers/api/RolesApiController.php
@@ -9,14 +9,34 @@ use Garden\Web\Data;
 use Garden\Web\Exception\NotFoundException;
 use Garden\Web\Exception\ServerException;
 use Vanilla\Utility\CapitalCaseScheme;
+use Vanilla\Utility\CamelCaseScheme;
 
 /**
  * API Controller for the `/roles` resource.
  */
 class RolesApiController extends AbstractApiController {
 
+    /** @var CamelCaseScheme */
+    private $camelCaseScheme;
+
     /** @var CapitalCaseScheme */
     private $caseScheme;
+
+    /** @var array Groups of permissions that can be consolidated into one. */
+    private $consolidatedPermissions = [
+        'discussions.moderate' => ['discussions.announce', 'discussions.close', 'discussions.sink'],
+        'discussions.manage' => ['discussions.delete', 'discussions.edit']
+    ];
+
+    /** @var array Permissions that have been deprecated and should no longer be used. */
+    private $deprecatedPermissions = [
+        'Garden.Activity.Delete',
+        'Garden.Activity.View',
+        'Garden.SignIn.Allow',
+        'Garden.Curation.Manage',
+        'Vanilla.Approval.Require',
+        'Vanilla.Comments.Me'
+    ];
 
     /** @var Schema */
     private $idParamSchema;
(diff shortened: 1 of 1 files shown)


initvector:
Will need to rebase, once the build on `master` is fixed.

applications/dashboard/controllers/api/RolesApiController.php
```diff
@@ -243,6 +387,39 @@ public function post(array $body) {
     }
 
     /**
+     * Rename a legacy Vanilla permission slug.
+     *
+     * @param string $permission
+     * @return string
+     */
+    private function renamePermission($permission) {
+        $result = $permission;
```
DaazKu: That line does nothing.

applications/dashboard/controllers/api/RolesApiController.php
```diff
@@ -42,6 +83,48 @@ public function __construct(RoleModel $roleModel, PermissionModel $permissionMod
         $this->roleModel = $roleModel;
         $this->permissionModel = $permissionModel;
```
DaazKu: $permissionModel is missing from the DOC block

applications/dashboard/controllers/api/RolesApiController.php
```diff
@@ -9,21 +9,62 @@
...
+        'Garden.Settings.Manage' => 'site.manage',
+        'Garden.Users.Approve' => 'applicants.manage',
+        'Groups.Group.Add' => 'groups.add',
+        'Groups.Moderation.Manage' => 'groups.moderate',
+        'Reputation.Badges.Give' => 'badges.moderate',
+        'Vanilla.Tagging.Add' => 'tags.add'
+    ];
+
+    /** @var array Permissions with a fixed name. Do not change them. */
+    private $fixedPermissions = [
```
DaazKu: I don't understand why we have this array of permissions.

applications/dashboard/controllers/api/RolesApiController.php
initvector: This was done to help debug this method, now and in the future. It's easier to compare the original value with the result.

applications/dashboard/controllers/api/RolesApiController.php
initvector: There are some permissions that should be excluded from being renamed at this time. Currently, it's just these two: Reactions.Negative.Add and Reactions.Positive.Add. It's these two, in particular, because they don't really match the naming scheme of "resource.action". If we clip the addon namespace off, they become "negative.add" and "positive.add", which doesn't do much to convey what the permission actually does. We're also trying to avoid two-word sections (e.g. reactions.positiveAdd). These permission names will probably be revisited in the near future, but for the time being, they're edge cases that we shouldn't rename.

applications/dashboard/controllers/api/RolesApiController.php
DaazKu: I don't think we should let 'debug' code lie around. We can easily add that line back if we need to actually debug the thing. Removing it will also shut up the editors about unused variable.

applications/dashboard/controllers/api/RolesApiController.php
DaazKu: What about all the other plugins permissions?

- Plugins.Debugger.View
- Plugins.Pockets.Manage
- Plugins.Flagging.Notify
- AvatarPool.CustomUpload.Allow
- Plugins.Attachments.Upload.Allow
- ....

initvector: * Plugins.Debugger.View - becomes "debugger.view"
* Plugins.Pockets.Manage - becomes "pockets.manage"
* Plugins.Flagging.Notify - becomes "flagging.notify"
* AvatarPool.CustomUpload.Allow - "customupload.allow"
* Plugins.Attachments.Upload.Allow - Not renamed, because it's four parts.

Some of these aren't ideal, but they're much more indicative of what the permission actually is.

DaazKu:
I see multiple problems that are more or less important to fix here.

- We are polluting the permissions list with permissions defined in plugins that are not part of vanilla/vanilla.
- OS plugins permission will be affected by this change with no way to opt out of the automatic renaming.
- We are gonna have to do a PR against the "fixedPermissions" array every time we find a a new permission that doesn't work well with our automatic renaming.

PR #6068, opened 2017-10-23
Move SQL building after categoryPermission call

DaazKu:
Fixes https://github.com/vanilla/internal/issues/1231

I moved the query building part of `DiscussionModel::getWhere()` after the call to `DiscussionModel::categoryPermissions()` which ensure that there is no call to something that can go a query before `$this->sql->get()` is called.


DaazKu:
The PR is failed unit tests because https://github.com/vanilla/vanilla/pull/6058 hasn't been merged yet.

PR #6077, opened 2017-10-27
Allow quoting usernames with dots

dannimajdalani:
Added the ability to Quote Usernames with dots in their usernames

### DIFF
--- a/plugins/Quotes/class.quotes.plugin.php
+++ b/plugins/Quotes/class.quotes.plugin.php
@@ -481,7 +481,7 @@ BQ;
                 case 'TextEx':
                     $quoteBody = $data->Body;
                     $insertName = $data->InsertName;
-                    if (strpos($insertName, ' ') !== false) {
+                    if ((strpos($insertName, ' ') OR strpos($insertName, '.')) !== false) {
                         $insertName = '"'.$insertName.'"';
                     }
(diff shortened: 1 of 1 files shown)


plugins/Quotes/class.quotes.plugin.php
```diff
@@ -481,7 +481,7 @@ protected function formatQuote($type, $iD, &$quoteData, $format = false) {
                 case 'TextEx':
                     $quoteBody = $data->Body;
                     $insertName = $data->InsertName;
-                    if (strpos($insertName, ' ') !== false) {
+                    if ((strpos($insertName, ' ') OR strpos($insertName, '.')) !== false) {
```
Charrondev: This doesn't seem like it works the way your expecting. It should be
```php
if (strpos($insertName, ' ') !== false ||  strpos($insertName, '.') !== false) {
```

Charrondev: A little bit more explanation on this. The first thing that would evaluate would be `(bool)strpos($insertName, ' ')`. PHP reads the first expression before your OR and evaluates if it truthy or falsy by converting it to a boolean. `strpos` could return 0 (first position) in which case it would evaluate to false even though it found the character ` `

Issue #6021, opened 2017-10-10
Dashboard: Fluid fixed navigation panel is broken


charrondev:
The dashboard left navigation panel used it fix itself so that it couldn't be scrolled out of view. This no longer works.

This seems to originate from  this commit: https://github.com/vanilla/vanilla/commit/56b380f31c6f9cd12e9367aac64d3e5e24777ee3#diff-bcd04ed30cdf4a89c731b25091d61891

Charrondev:
While reverting that commit seems to restore the old behaviour, it still leaves this bug: https://github.com/vanilla/vanilla/issues/4868, which seems to be caused by the height not being recalculated correctly when the side panels are collapsed and expanded.

https://github.com/vanilla/vanilla/commit/56b380f31c6f9cd12e9367aac64d3e5e24777ee3 Seems to have resolved the issue by disabled the fluid part of fixing the panel altogether. A more granular solution will be required.

Charrondev:
I've determined that this will require significantly more work than I had initially imagined. I'm going to table this until we have to time to rip apart some of the dashboard's javascript as even if I fixed this edge case the actual performance of the fluid fixed functionality it terrible.

Issue #4868, opened 2016-11-27
Dashboard v3 scrolling issue on short pages


linc:
When I load `/dashboard/settings/registration`, on my 13" Macbook in Firefox, the footer is generated not too far from the end of the brief settings. I cannot scroll past it, making everything below "Reputation" inaccessible in the sidebar navigation. Toggling a sidebar section closed/open causes the footer to move down the page below the sidebar, fixing the issue for that pageload.

slafleche:
In PR https://github.com/vanilla/vanilla/pull/5506

PR #6083, opened 2017-10-31
Revert "Fix initial logic for having a “fixed” panel"

Charrondev:
This reverts commit 56b380f31c6f9cd12e9367aac64d3e5e24777ee3.

Due to that commit we now never handle scrolling with the side panel. 

For `handleScroll` to be true both `containerOuterHeight < vars.windowHeight` and `containerOuterHeight > vars.windowHeight` would be true.

The initial PR was here: https://github.com/vanilla/vanilla/pull/5506.

Fulfils: https://github.com/vanilla/vanilla/issues/6021

### DIFF
--- a/applications/dashboard/js/jquery.fluidfixed.js
+++ b/applications/dashboard/js/jquery.fluidfixed.js
@@ -97,7 +97,7 @@
              */
 
             var containerOuterHeight = vars.containerHeight + vars.offsetTop + vars.offsetBottom;
-            var fixObject = containerOuterHeight < vars.documentHeight && containerOuterHeight < vars.windowHeight;
+            var fixObject = containerOuterHeight < vars.documentHeight;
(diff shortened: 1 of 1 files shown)


slafleche:
If there's a bug with https://github.com/vanilla/vanilla/commit/56b380f31c6f9cd12e9367aac64d3e5e24777ee3, you still need to address https://github.com/vanilla/vanilla/issues/4868.

Charrondev:
Hmm, that bug seems to be caused by collapsing/uncollapsing parts of the panel and the real height not being properly calculated. This is probably going to require a much more in depth fix. I've updated #6021 and will revisit this later.

Charrondev:
I'm going to close out this PR. This is probably going to a require a lot more work to fix than it initially seemed.

PR #6094, opened 2017-11-02
Fire event before displaying per-category notification preferences

Charrondev:
I'm trying to fulfill a service that requires hiding the email preferences for particular users with criteria to be determined in a separate plugin. The email preferences could just be hidden with CSS except that their is a double header with column spans set. This is not changeable through a stylesheet. So I've fired an event that allows a Plugin to set its own criteria for if these email preferences show up or not. This is a fallback and will not change any existing behaviour.

Needs backport to `release/2017-Q2-6` and `release/2.5a`.


Charrondev:
TIL you can save a config value only in memory. This is not necessary in that case.

PR #6097, opened 2017-11-02
Allow POST requests to dispatch to Controller::patch() in some circumstances

tburry:
This is a bit of a kludge to allow POST to be used against the usual PATCH method to allow for multipart/form-data on PATCH (edit) endpoints.

The initial implementation of this is narrow so POST cannot dispatch to `patch_action` style methods, just the default `patch` method.


tburry:
I think this PR is safe as long as our tests pass since it affects only tested code.

PR #6101, opened 2017-11-02
Improve file upload support for API uploads

### LINKED ISSUE #6084: Update Gdn_Request to normalize file uploads into POST data
The `Gdn_Request` class needs to be modified to normalize file uploads into the post data. This will be done to make it easier to handle file uploads with API v2.

A new `Vanilla\UploadedFile` class should be created to contain an uploaded file's information. It should adhere to `UploadedFileInterface` from [PSR-7](http://www.php-fig.org/psr/psr-7), but without the `getStream` method.

The superglobal `$_FILES` array should be processed and added to the POST data in `Gdn_Request`. The recommendation implementation for normalizing uploads can be found on the PSR-7 page: [Uploaded Files](http://www.php-fig.org/psr/psr-7/#16-uploaded-files)

If an element in `$_FILES` shares a name with an element in `$_POST`, the POST element takes precedence.

### LINKED ISSUE #6085: Allow schemas with file validation
[Garden Schema](https://github.com/vanilla/garden-schema) is used extensively throughout API v2 controllers. However, the project only supports JSON data and Vanilla will need a schema that supports validating file fields. The `Garden\Schema\Schema` class should be extended in Vanilla as 
`UploadedFileSchema`. Since Garden Schema supports nested schemas, this custom Vanilla schema class can then be used in a normal schema definition.

1. `UploadedFileSchema` should validate if its value is an instance of `Vanilla\UploadedFile`.
1. The schema array in `UploadedFileSchema` can be hardcoded with the following information, [as defined in the OpenAPI v3 spec](https://swagger.io/specification/#considerations-for-file-uploads-61):
    ```json
    {
        "id": "UploadedFile",
        "type": "string",
        "format": "binary"
    }
    ```
1. Tweaking the `Garden\Schema\Schema` class to support this functionality may be necessary. Specifically, some methods may need to have their visibility altered from `private` to `protected`.

initvector:
In preparation for using file uploads with the API, this update does a few things:

1. Adds the new `Vanilla\UploadedFile` class, which attempts to be adherent to the `UploadedFileInterface` from [PSR-7](http://www.php-fig.org/psr/psr-7), with the exception that streams are not supported.
1. Merges entries from the `$_FILES` array into POST data in `Gdn_Request` as instances of `Vanilla\UploadedFile`. The structure of file attributes is normalized to something a little more sane in the process.
1. Adds a new `Vanilla\UploadedFileSchema` class, which extends `Garden\Schema\Schema`. This class will be used to verify a field in an API request is populated with a file (instance of `Vanilla\UploadedFile`).
1. Adds tests for `Vanilla\UploadedFile`.

To test out `Vanilla\UploadedFileSchema`, all you need to do is add it into an input schema.

```php
$in = $this->schema([
    'myFile' => new Vanilla\UploadedFileSchema()
]);
```

Closes #6084
Closes #6085

### DIFF
--- a/library/Vanilla/UploadedFile.php
+++ b/library/Vanilla/UploadedFile.php
@@ -1,0 +1,207 @@
+<?php
+namespace Vanilla;
+
+use Gdn;
+use InvalidArgumentException;
+use RuntimeException;
+
+/**
+ * Value object representing a file uploaded through an HTTP request.
+ */
+class UploadedFile {
+
+    /** @var string */
+    private $clientFileName;
+
+    /** @var string */
+    private $clientMediaType;
+
+    /** @var int */
+    private $error;
+
+    /** @var string */
+    private $file;
+
+    /** @var string */
+    private $moved = false;
+
+    /** @var int */
+    private $size;
+
+    /**
+     * UploadedFile constructor.
+     *
+     * @param string $file
+     * @param int $size
+     * @param int $error
+     * @param string|null $clientFileName
+     * @param string|null $clientMediaType
+     */
+    public function __construct($file, $size, $error, $clientFileName = null, $clientMediaType = null) {
+        $this->setSize($size);
+        $this->setError($error);
+        $this->setClientFileName($clientFileName);
+        $this->setClientMediaType($clientMediaType);
+
+        if ($this->getError() === UPLOAD_ERR_OK) {
+            $this->setFile($file);
+        }
+    }
+
+    /**
+     * Retrieve a stream representing the uploaded file.
+     *
+     * @throws RuntimeException
+     */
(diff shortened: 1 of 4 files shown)


library/Vanilla/UploadedFile.php
```diff
@@ -0,0 +1,212 @@
+<?php
+/**
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license GNU GPLv2
+ */
+
+namespace Vanilla;
+
+use Gdn;
```
DaazKu: You can't have a dependency on Gdn in the Vanilla namespace.

Receive the container or Gdn_Upload in the constructor.

PR #6112, opened 2017-11-06
fixed non-existing permission

dannimajdalani:
### Issue
Users with admin rights did not have access to /dashboard/settings/bans/find/["userid"]

### DIFF
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -2049,7 +2049,6 @@ class SettingsController extends DashboardController {
      * @param int|string $userIdentifier Either the username or user ID.
      */
     private function findBanRule($userIdentifier) {
-        $this->permission('Garden.Settings.Manage');
 
         $userModel = new UserModel();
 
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/class.settingscontroller.php
```diff
@@ -2049,7 +2049,7 @@ public function mobileThemes($ThemeName = '', $TransientKey = '') {
      * @param int|string $userIdentifier Either the username or user ID.
      */
     private function findBanRule($userIdentifier) {
-        $this->permission('Moderation.Bans.Manage');
+        $this->permission('Garden.Settings.Manage');
```
initvector: This permission check seems redundant. `SettingsController::findBanRule` is a private method. The only place it's called is in `SettingsController::bans`, which [already has a permission check at the top of it](https://github.com/vanilla/vanilla/blob/cbb18ce9e5be5eb85126d723ba0fd2d3a98dacff/applications/dashboard/controllers/class.settingscontroller.php#L568).

DaazKu:
You need to sign the CLA

dannimajdalani:
signed

PR #6118, opened 2017-11-07
Use valr() to translate multidimensional arrays in arrayTranslate()

nervoustwit:


nervoustwit:
The use case of this PR is that in SSO plugins we use the function arrayTranslate() to translate the profile  in the JSON response that is returned by the authenticating server. We allow admins to configure the names of array keys in the dashboard to map to the the JSON response. Presently any nested arrays in the JSON response are not accessible.

In this PR we user `valr()` to get the values from the JSON response so that the keys can use Vanilla's dot notation. For example by setting "Photo" to "picture.thumbnail" in the dashboard we could get the thumbnail from the response below.

```
$arrayMap = [
  'email' => 'Email',
  'user_name' => 'Name',
  'picture.thumbnail' => 'Photo
]

$JSONResponse = {
  "email": "patrick.k@vanillaforums.com",
  "user_name": "patrick",
  "picture": {
     "thumbnail": "https://somecdn.com/patricksthumbnail.png",
     "photo": "https://somecdn.com/patricksphoto.png"
     }
}
```

tburry:
Beyond the broken build I think this is too much of a change to arrayTranslate. Can you instead make an arrayTranslateR function and we can use that in SSO? This will also need some unit testing.

nervoustwit:
Agreed. arrayTranslate() touches a lot of code.

PR #6119, opened 2017-11-08
Add the Attributes class

tburry:
This class is an array wrapper for the "attributes" style columns of API call results. We need this class because attributes can be empty which then JSON encode to `[]` instead of `{}` which can throw non-PHP clients off.


tburry:
Closed in favor of #7049.

PR #6123, opened 2017-11-08
Add date filter parsing with schemas

### LINKED ISSUE #6037: Add date parser for API date parameters
The API will need parsing for date field filters. Current requirements for the parser are:

1. Date fields should be strings.
1. The time portion is optional.
1. The date strings should provide a comparison operators.
    * Supported operator: >, >=, <, <=
    * Example: `>=2017-09-21` would be interpreted as "greater than or equal to 00:00:00 on September 21st, 2017"
    * Example: `<2017-09-21` would be interpreted as "less than 00:00:00 on September 21st, 2017"
1. Limited support for [interval notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals) needs to be implemented to support date ranges.
    * Supported operators:
        * Inclusive: [, ]
        * Exclusive: (, )
    * Example: `[2017-09-21,2017-09-22)` would be interpreted as "between 00:00:00 on September 21st, 2017 and 23:59:59 on September 21nd, 2017"

initvector:
This update adds the `DateFilterSchema` class, which can be used to parse date filter strings into an associative array that can be used for building date filters in API endpoints. Examples of the supported date filter format can be found on the original issue.

```php
$data = ['myDate' => '[2017-10-01,2017-10-30]']
$in = $this->schema(['myDate' => new DateFilterSchema()]);
$data = $in->validate($data);
/**
 * $data = [
 *     'myDate' => [
 *         'op' => '[]',
 *         'value' => [
 *             DateTimeImmutable,
 *             DateTimeImmutable
 *         ]
 *     ]
 * ]
 */
```

Closes #6037 

*Note: The original issue mentions building out a subschema array for these fields and adding a filter function to be called on the data, but this idea was scrapped in favor of subclassing `Garden\Schema\Schema` and using that as a nested schema object, similar to the method used for the [file upload schema](https://github.com/vanilla/vanilla/pull/6101).*

### DIFF
--- a/library/Vanilla/DateFilterSchema.php
+++ b/library/Vanilla/DateFilterSchema.php
@@ -1,0 +1,210 @@
+<?php
+/**
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license GNU GPLv2
+ */
+
+namespace Vanilla;
+
+use DateTimeImmutable;
+use Garden\Schema\Invalid;
+use Garden\Schema\Schema;
+use Garden\Schema\ValidationField;
+use Garden\Schema\ValidationException;
+
+class DateFilterSchema extends Schema {
+
+    /** @var string Regular expression for matching a datetime. */
+    private $dateRegEx = '(?<date>\d{4}-\d{2}-\d{2})(?<time>[ T]\d{2}:\d{2}(:\d{2})?)?';
+
+    /** @var array Valid characters for opening an interval-notation range. */
+    private $intervalOpen = ['(', '['];
+
+    /** @var array Valid characters for closing an interval-notation range. */
+    private $intervalClose = [')', ']'];
+
+    /** @var array Valid operators for simple date comparisons. */
+    private $simpleOperators = ['=', '>', '<', '>=', '<='];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(array $schema = []) {
+        if (!empty($schema)) {
+            throw new \InvalidArgumentException(self::class.' does not support custom schemas.');
+        }
+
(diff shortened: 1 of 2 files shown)


library/Vanilla/DateFilterSchema.php
```diff
@@ -0,0 +1,200 @@
...
+
+namespace Vanilla;
+
+use DateTimeImmutable;
+use Garden\Schema\Invalid;
+use Garden\Schema\Schema;
+use Garden\Schema\ValidationField;
+use Garden\Schema\ValidationException;
+
+class DateFilterSchema extends Schema {
```
DaazKu: Docblock

library/Vanilla/DateFilterSchema.php
```diff
@@ -0,0 +1,200 @@
...
+
+    /** @var array Valid operators for simple date comparisons. */
+    private $simpleOperators = ['=', '>', '<', '>=', '<='];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(array $schema = []) {
+        if (!empty($schema)) {
+            throw new \InvalidArgumentException(self::class.' does not support custom schemas.');
```
DaazKu: Just declare the constructor with no parameter.
```php
/**
 * DateFilterSchema constructor.
 */
    parent::__construct();
}
```

initvector: My intention here is to make sure that anyone attempting to pass a schema spec to this subclass immediately knows it isn't going to work, instead of having it silently fail to do what they were attempting.

Issue #6127, opened 2017-11-09
Default to current user on /api/v2/conversations


tburry:
If i call /api/v2/conversations with no parameters it should use the authenticated user's ID. Currently I get the following error:

```json
{
    "message": "One of insertUserID, participantUserID are required.",
    "status": 400,
    "errors": [
        {
            "field": "",
            "code": "missingField",
            "message": "One of insertUserID, participantUserID are required."
        }
    ]
}
```

initvector:
Closing in favor of tracking on the associated PR: #6265

PR #6128, opened 2017-11-09
Conversations endpoint tweaks for theme development

tburry:
When developing the theme against the conversations endpoint we found
we needed the following:

- [x] The GET /conversations endpoint should return my inbox without any
parameters.
- [x] The conversations that are returned should always have a name field
for easier display in themes/apps.
- [x] This lead to changing the POST /conversations/:id/participants
endpoint to return a similar result to GET
/conversations/:id/participants.
- [x] Need lastMessage information.
- [x] Need abbreviated participants for GET /conversations
- [x] body, url for GET /conversations. The body should be either the last message for the viewing user or the first message in the conversation.


tburry:
Closes #6127.

applications/conversations/controllers/api/ConversationsApiController.php
```diff
@@ -67,7 +67,7 @@ private function checkModerationPermission() {
      * @throws NotFoundException if the conversation could not be found.
      * @return array
      */
-    private function conversationByID($id, $viewingUserID = null) {
+    private function conversationByID($id, $viewingUserID = null, $participants = 0) {
```
DaazKu: The docblock needs to be updated.

PR #6133, opened 2017-11-13
Update expand API parameter to be a CSV of specific fields

### LINKED ISSUE #6106: Update expand parameter on comments and discussions endpoint to take a CSV
`CommentsApiController` and `DiscussionsApiController` each have an "expand" parameter. This parameter is intended to automatically expands associated records, like the author. The only accepted values for this parameter are boolean, however more control will need to be implemented. Handling of the "expand" parameter should be updated to allow for a CSV of values. For example:

```
GET /discussions/123?expand=reactions,insertUser
```

initvector:
This update alters the "expand" parameter on several API endpoints, changing it from a boolean to a CSV of field names. The intended effect is to allow more control over what data is expanded.

Addons should be able to modify the allowed "expand" values in these schemas by hooking into relevant events (e.g. `controller_schema`).

Closes #6106

### DIFF
--- a/applications/dashboard/controllers/api/AbstractApiController.php
+++ b/applications/dashboard/controllers/api/AbstractApiController.php
@@ -44,6 +44,47 @@ abstract class AbstractApiController extends \Vanilla\Web\Controller {
     }
 
     /**
+     * Determine which fields should be expanded, using a request and a field map.
+     *
+     * @param array $data An array representing request data.
+     * @param array $map An array of short-to-full field names (e.g. insertUser => InsertUserID).
+     * @param string $field The name of the field where the expand fields can be found.
+     * @return array
+     */
+    protected function getExpandFields(array $data, array $map, $field = 'expand') {
+        $result = [];
+        if (array_key_exists($field, $data)) {
+            $expand = $data[$field];
+            foreach ($map as $short => $full) {
+                if (in_array($short, $expand)) {
+                    $result[] = $full;
+                }
+            }
+        }
+        return $result;
+    }
+
+    /**
+     * Get a simple schema for nesting as an "expand" parameter.
+     *
+     * @param array $fields Valid values for the expand parameter.
+     * @return Schema
+     */
+    protected function getExpandFragment(array $fields) {
+        $result = $this->schema([
+            'description' => 'Expand associated records.',
+            'items' => [
+                'enum' => $fields,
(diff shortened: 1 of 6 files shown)


applications/dashboard/controllers/api/AbstractApiController.php
```diff
@@ -44,6 +44,47 @@ public function formatField(array &$row, $field, $format) {
...
 
     /**
+     * Determine which fields should be expanded, using a request and a field map.
+     *
+     * @param array $data An array representing request data.
+     * @param array $map An array of short-to-full field names (e.g. insertUser => InsertUserID).
+     * @param string $field The name of the field where the expand fields can be found.
+     * @return array
+     */
+    protected function getExpandFields(array $data, array $map, $field = 'expand') {
```
DaazKu: Can we give a more meaningful name to the `$data` parameter? `$requestData` would be perfect.

DaazKu: The function could also use a name change. It does not "get" expand fields. It gets, filters, and converts them.
Something like "resolveExpandFields" would be more explanatory. (I'm not sold on the word resolve but that's all I can think of right now.)

```
function resolveExpandFields(&$expandFields, array $filterMap) {
    if (!is_array($expandFields) {
        return [];
    }
    ... 
}
----
$resolvedFields = resolveExpandFields($query['expand'], ['insertUser' => 'InsertUserID']);
```

Having &expandFields being a reference allow us to pass $query['expand'] even if it is undefined which is pretty neat.

applications/dashboard/controllers/api/AbstractApiController.php
```diff
@@ -44,6 +44,47 @@ public function formatField(array &$row, $field, $format) {
...
+    }
+
+    /**
+     * Get a simple schema for nesting as an "expand" parameter.
+     *
+     * @param array $fields Valid values for the expand parameter.
+     * @return Schema
+     */
+    protected function getExpandFragment(array $fields) {
+        $result = $this->schema([
```
DaazKu: This should not be a model of it's own. Use `Schema::parse();`
Or if we want to expose it as it's own model the enum part should be set to '...' on the first declaration and set properly before being returned otherwise the first call to it will set the items to something that will probably not be relevant for many endpoints.

PR #6141, opened 2017-11-15
Add ability to update permissions to RolesAPIController

### LINKED ISSUE #5775: Add ability to manage permissions to /roles endpoints in API v2
## PATCH /roles/:id/permissions

Update permissions for the given role.

### Eample: Global Permissions

```
PATCH /roles/123/permissions

[
  {
    type: "global",
    permissions: {
      "reactions.negative.add": true,
      "discussions.add": false,
      "comments.add": true
    }
  }
]
```

This would set the given global permissions of the role. If the role has other global permissions then they will not be overridden.

### Example: Category Permissions

```
PATCH /roles/123/permissions

[
  {
    type: "category",
    id: 123,
    permissions: {
      "discussions.add": true,
      "comments.add": true
    }
  }
]
```

This would add the given category permissions to the role. If the cateogy's permissionCategoryID is set to another category then it
would get set to itself at this point and all of its children that do not have custom perissions would have to get updated.

## Add a field to the categories endpoint

We need to add a `customPermissions` boolean to the categories endpoint that controls the permissionCategoryID behavior for the category.

## GET /roles/:id changes

- Add an `expand` parameter that will expand permissions on the roles.
- This should display a max of 100 permission rows. Select 101 and if there are that many throw a 416 client error saying there are too many permissions to display.

## POST /roles & PATCH /roles/:id changes

- Ad an optional permissions object that will add the permissions on the new role.

### Example: Overwriting the global and a category permissions

```
PUT /roles/:id/permissions

[
  {
    "type": "global",
    "permissions": {
      "reactions.negative.add": true
    }
  },
  {
    "type": "category",
    "id": 123,
    "permissions": {
        "discussions.add": true,
        "comments.add": true
    }
  }
]
```

- So the role will end up with only "reactions.negative.add" permissions.
- So the role will end up with only "discussions.add" and "comments.add" permissions for category 123 (ex. "discussions.view" will be false).

initvector:
This update allows setting permissions on a role with the /roles endpoint. Specifics are in the associated issue, including the permission structure, but here's a rundown:

- [x] Update (not overwrite) global permissions with PATCH /roles/:id/permissions
- [x] Overwrite global permissions with PUT /roles/:id/permissions
- [x] Update (not overwrite) per-category permissions with PATCH /roles/:id/permissions
- [x] Overwrite per-category permissions with PUT /roles/:id/permissions
- [x] Add `expand` parameter to /roles GET and index, allow users to expand permission rows
- [x] Add limitation of 100 permission rows when expanded. Exceeding this limit throws an [HTTP 416](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416)
- [x] Add `permissions` parameter to POST /roles
- [x] Add `permissions` parameter to  PATCH /roles/:id
- [x] Add `customPermissions` field to /categories to indicate if custom permissions are set on this category and allow toggling them off with the same parameter.

Closes #5775

### DIFF
--- a/applications/dashboard/controllers/api/RolesApiController.php
+++ b/applications/dashboard/controllers/api/RolesApiController.php
@@ -89,6 +89,7 @@ class RolesApiController extends AbstractApiController {
      *
      * @param RoleModel $roleModel
      * @param PermissionModel $permissionModel
+     * @param CategoryModel $categoryModel
      */
     public function __construct(RoleModel $roleModel, PermissionModel $permissionModel, CategoryModel $categoryModel) {
         $this->roleModel = $roleModel;
@@ -242,6 +243,7 @@ class RolesApiController extends AbstractApiController {
      * Get a single role.
      *
      * @param int $id The ID of the role.
+     * @param array $query
      * @throws NotFoundException if unable to find the role.
      * @return array
      */
@@ -313,6 +315,11 @@ class RolesApiController extends AbstractApiController {
         return $result;
     }
 
+    /**
+     * Return a schema to represent a collection of permission rows.
+     *
+     * @return static
+     */
     private function getPermissionsFragment() {
         static $permissionsFragment;
 
@@ -348,6 +355,7 @@ class RolesApiController extends AbstractApiController {
     /**
      * List roles.
      *
+     * @param array $query
      * @return array
      */
     public function index(array $query) {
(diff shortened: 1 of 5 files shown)


DaazKu:
This endpoint has the same issue than the TOKEN endpoint.
We cannot fetch the permissions of a role if it has 100+ permissions instead of limiting the number of permissions that can be created or giving an alternate mean of getting the permission of a role in a paginated manner.

I've been told that this design flaw was known sooooo I'm gonna approve the PR since everything else is good.

PR #6147, opened 2017-11-16
Fix bitwise math to validate if an addon exists or is disabled

### LINKED ISSUE #6146: Addon requirements don't properly verify names or version of required addons
## Steps to reproduce
- Create a new addon and symlink it into the plugins folder
- Create an `addon.json` file it with the following require key
```
"require": {
    "alskdfjl;askdjfl;aksdfj": ">1.0",
    "vanilla": ">= 4320"
}
```
- Attempt to enable the addon

## Expected Behaviour
The addon will not be enabled and an error will be displayed. The first addon does not exist and the second addon does not have a compatible version.

## Actual Behavior
The addon is successfully enabled with no visible warnings.

Charrondev:
Closes #6146 

This bitwise operator was incorrect causing #6146. 

## The math
```php
const REQ_ENABLED = 0x01; // addon enabled, yay!
const REQ_DISABLED = 0x02; // addon disabled
const REQ_MISSING = 0x04; // addon missing from the manager
const REQ_VERSION = 0x08; // addon isn't the correct version
```

We pass `self::REQ_MISSING | self::REQ_VERSION` to `lookupRequirements`  or 
4 | 8 = 12

## Current behaviour with the given filter
For `$row['status'] = REQ_MISSING`
(4 & 12) === 12 -> _false 4 & 12 === 4_ 

For `$row['status'] = REQ_VERSION`
(8 & 12) === 12 -> _false 8 & 12 === 8_

## After with the given filter
For `$row['status'] = REQ_MISSING`
(4 | 12) === 12 -> _true_

For `$row['status'] = REQ_VERSION`
(8 | 12) === 12 -> _true_

With this change addons will correctly check their requirements versions and existance.

### DIFF
--- a/library/Vanilla/Addon.php
+++ b/library/Vanilla/Addon.php
@@ -784,7 +784,7 @@ class Addon {
         if ($count = count($issues)) {
             $subdir = $this->getSubdir();
 
-            trigger_error("The addon in $subdir has $count issues.", E_USER_NOTICE);
+            trigger_error("The addon in $subdir has $count issue(s).", E_USER_NOTICE);
             foreach ($issues as $issue) {
                 trigger_error($issue, E_USER_NOTICE);
             }
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -601,7 +601,7 @@ class AddonManager {
         // Filter the list.
         if ($filter) {
             $array = array_filter($array, function ($row) use ($filter) {
-                return ($row['status'] & $filter) === $filter;
+                return ($row['status'] | $filter) === $filter;
             });
         }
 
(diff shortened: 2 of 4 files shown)


tburry:
I made a commit to fix the issue @ac.

The issue was that the filter in **lookupRequirements()** should have been considered an OR filter where it was really an AND filter. Your solution was wrong because it would have succeeded when an addon had a status of 0 which is not really robust because the knowledge that that doesn't happen is coupled to **lookupRequirementsRecursive()**.

This is the exact type of issue that we should be trying to unit test in order to protect against in the future. This does two things:

1. Any refactoring in the future is protected against opening up the same bug.
2. Maybe some other code depends on the old behavior. If so, then this unit test will let us know that we have another set of requirements that has to be met and prevents us from ping-ponging between two incompatible bug fixes.

PR #6156, opened 2017-11-17
Add additional date filtering to API endpoints

initvector:
- [ ] Add new method for querying comments to `CommentModel`
- [ ] Add dateInserted parameter to /comments index
- [ ] Add dateUpdated parameter to /comments index
- [x] Add dateInserted parameter to /discussions index
- [x] Add dateUpdated parameter to /discussions index
- [x] Add dateInserted parameter to /users index
- [x] Add dateUpdated parameter to /users index

### DIFF
--- a/applications/dashboard/controllers/api/AbstractApiController.php
+++ b/applications/dashboard/controllers/api/AbstractApiController.php
@@ -19,6 +19,57 @@ abstract class AbstractApiController extends \Vanilla\Web\Controller {
     private $postFragmentSchema;
 
     /**
+     * If the parameter value is a valid date filter value, return an array of query conditions.
+     *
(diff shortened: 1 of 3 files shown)


DaazKu:
DateUpdate should be part of the discussion model if we can filter by it.

DaazKu:
I also saw some artifacts in the error message. Created an issue about it: #6168

DaazKu:
This is also missing some unit tests on the new functionalities.

initvector:
Tests have been added for `AbstractApiController::dateFilterField` and `CommentModel::lookup`. Additionally, "dateInserted" and "dateUpdated" fields have been added to endpoints' responses that filter by those fields, as necessary.

PR #6159, opened 2017-11-17
Add the ApiUtils helper class

tburry:
This class is meant to contain static helper functions for API endpoints.


tburry:
I'm adding this class now as-is so that you can add other methods to it @DaazKu.

Issue #5109, opened 2017-02-02
Create the /api/v2/addons endpoint


tburry:
<s>Depends on #5108.</s>

GET /api/v2/addons

- Get all enabled addons, filtering out those the hidden ones.
- Querystring all=1 will get all addons.
- Querystring search=xxx will search on name and description.

GET /api/v2/addons/\<key\>

- Get the details of a single addon. This will be just the row from the above endpoint.

PATCH /api/v2/addons/\<key\>

- Enable/disable an addon.
- Accepts parameter **enabled**
- Returns the result from the AddonModel->setEnabled

initvector:
Closed by #6160

PR #6160, opened 2017-11-17
Add some /api/v2/addons endpoints

tburry:
These endpoints are for managing addons.

### DIFF
--- a/applications/dashboard/controllers/api/AddonsApiController.php
+++ b/applications/dashboard/controllers/api/AddonsApiController.php
@@ -155,7 +155,8 @@ class AddonsApiController extends Controller {
             throw new \Garden\Web\Exception\NotFoundException("Addon");
         }
 
-        $row = $out->validate($this->filterOutput(reset($addons)));
+        $themeType = empty($query['themeType']) ? 'desktop' : $query['themeType'];
+        $row = $out->validate($this->filterOutput(reset($addons), $themeType));
         return $row;
     }
 
--- a/applications/dashboard/controllers/api/AddonsApiController.php
+++ b/applications/dashboard/controllers/api/AddonsApiController.php
@@ -1,0 +1,205 @@
+<?php
+
+/**
+ * @author Todd Burry <todd@vanillaforums.com>
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+use Garden\Schema\Schema;
+use Vanilla\Addon;
+use Vanilla\Models\AddonModel;
+use Vanilla\Web\Controller;
+
+/**
+ * Endpoints for managing addons.
+ */
+class AddonsApiController extends Controller {
+    /**
+     * @var AddonModel $addonModel
+     */
+    private $addonModel;
+
+    /**
+     * @var Schema $schema;
+     */
+    private $schema;
+
+    /**
+     * AddonsApiController constructor.
+     *
+     * @param AddonModel $addonModel The addon model dependency.
+     */
+    public function __construct(AddonModel $addonModel) {
+        $this->addonModel = $addonModel;
+    }
+
+    /**
+     * Get schema of an addon as returned by the API.
+     *
+     * @return Schema Returns an initialized schema.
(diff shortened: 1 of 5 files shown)


tburry:
Closes #5109.

applications/dashboard/controllers/api/AddonsApiController.php
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * Enable or disable an addon.
+     *
+     * @param string $addonID The addon ID.
+     * @param array $body The request body.
+     * @return array Returns an array of affected addons.
+     * @throws \Garden\Web\Exception\NotFoundException Throws an exception when the addon is not found.
+     */
+    public function patch($addonID, array $body) {
```
initvector: If this is the extent of the planned functionality, I think it would make more sense as a subresource action. No fields beyond the "enabled" attribute can be modified. This is essentially a toggle. We can treat the "enabled" state as a subresource (e.g. PATCH /api/v2/addons/:id/enabled or PUT /api/v2/addons/:id/enabled if you aren't a stickler for idempotency).

applications/dashboard/controllers/api/AddonsApiController.php
```diff
@@ -0,0 +1,210 @@
...
+            $options['themeType'] = $body['themeType'];
+        }
+        if ($body['enabled']) {
+            $r = $this->addonModel->enable($addon, $options);
+        } else {
+            $this->addonModel->disable($addon, $options);
+            $r = [$addon];
+        }
+
+        $r = array_map([$this, 'filterOutput'], $r);
```
initvector: This fails on addons with dependencies (e.g. Ideation). The result of `AddonModel::enable` for these types of addons has the first element being an instance of `Vanilla\Addon` and subsequent elements (newly-enabled requirements) being associative arrays. The type declaration for the first parameter of `AddonsApiController::filterOutput` is `Vanilla\Addon`.

applications/dashboard/controllers/api/AddonsApiController.php
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * Get a single addon.
+     *
+     * @param string $addonID The addon ID to get.
+     * @param array $query The query string.
+     * @return array Returns an addon info row.
+     * @throws \Garden\Web\Exception\NotFoundException Throws an exception when the addon is not found.
+     */
+    public function get($addonID, array $query) {
```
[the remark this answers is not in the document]
tburry: This is for the themeType parameter. I've added it.

applications/dashboard/controllers/api/AddonsApiController.php
```diff
@@ -0,0 +1,210 @@
...
+        $this->permission('Garden.Settings.Manage');
+
+        $id = $this->schema([
+            'addonID:s' => 'The ID of the addon.'
+        ], 'in')->setDescription('Get an addon.');
+
+        $in = $this->schema([
+            'themeType:s?' => [
```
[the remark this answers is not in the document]
tburry: It's one of those unfortunate, but necessary parameters in order to show which theme is enabled in the result. People that aren't using the API for feature parity won't need this, but in order to build an application that matches our functionality it is necessary unless we add a specific separate endpoint for themes.

applications/dashboard/controllers/api/AddonsApiController.php
tburry: I see your point, but this is symmetric with the get endpoints.

applications/dashboard/controllers/api/AddonsApiController.php
tburry: I fixed this in the PR, but need to hold off on unit tests. There is another PR coming that has a place for this kind of unit test.

tburry:
I think I've answered all of your points @r. Your general case-sensitivity comment has a unit test [here](https://github.com/vanilla/vanilla/pull/6160/files#diff-3f968fc26ded591b436ce9404ae4f7fcR127).

Issue #5124, opened 2017-02-02
Add conflict support to the Addon class


tburry:
- [x] The Addon class should support a key called "conflict" in its info array. It will be in the same format as the "require" array.
- [x] Add Addon->getConflicts()
- [x] Add checks to Addon->check() to make sure the "require" and "conflict" keys in the info array are arrays if present.
- [x] Unit test.

PR #6163, opened 2017-11-20
Add conflict support to addons

tburry:
Addons can add a **conflict** key in their addon.json to specify other addons that conflict with them. The key is in the same format at the **require** key. A few points:

- This is a change to the AddonModel which isn't currently hooked up to the various old managers so it won't have an immediate impact on the UI. This is part of a larger change to move to addon management via API.
- I modified the **require** and **conflict** array outputs from the /api/v2/addons endpoints. This isn't specifically necessary for this PR, but I didn't want to make yet another PR for this.
- Addons can now specify "*" as a version requirement like composer. This is useful for specifying conflicts.

*This PR was rebased. Make sure you do a force pull if you have already checked the branch out!*

### DIFF
--- a/applications/dashboard/controllers/api/AddonsApiController.php
+++ b/applications/dashboard/controllers/api/AddonsApiController.php
@@ -41,8 +41,19 @@ class AddonsApiController extends Controller {
      */
     protected function fullSchema() {
         if ($this->schema === null) {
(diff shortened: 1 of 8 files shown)


tburry:
Closes #5124.

PR #6184, opened 2017-11-23
[Backport] Escape backslashes to prevent the browser from converting them to slashes

DaazKu:
Backport vanilla/vanilla-patches#138


DaazKu:
Travis fails on php 7 because we are missing a fix on the unit test suite. Disregard plz.

PR #6193, opened 2017-11-27
Improve expand API parameter description

Charrondev:
The expand defintion wasn't very clear so I'm updating the description for a little more clarity.


initvector:
I've tweaked the phrasing, just a little bit, because the parameter is more commonly going to be provided as a comma-separated string of field names. The API docs will be responsible for stating the specifics of the [parameter's expected style/format](https://swagger.io/specification/#style-values-52).

PR #6195, opened 2017-11-28
Add the nav-pill class to the dashboard

tburry:
This is a small change for marking the API version in the dashboard. It
also includes a yarn upgrade.


Charrondev:
Your going to need to do a rebase and rebuild because the css maps are conflicting.

tburry:
@Charrondev done.

PR #6200, opened 2017-12-01
Fix setting user roles with the API

initvector:
The ability to set user roles over the API was broken. This update restores that ability.


initvector:
Backporting to [release/2.5](https://github.com/vanilla/vanilla/tree/release/2.5).

PR #6203, opened 2017-12-01
Fix failing discussion API tests

### LINKED ISSUE #6202: Update DiscussionsTest so that it uses its own 2 categories for testing
Current unit tests are broken because we introduced a change to the DiscussionModel where you [cannot assign a discussion to a non existing category](https://github.com/vanilla/vanilla/commit/7e9e8d33ac69248917732bc7d1573899c7a2df2c#diff-461fbde3929a6742e0a92301a01cdd04R1948).

This makes the test fails on patch where categoryID is upgrade from 1 to 2.

We should update the unit test to:
1. Use its own 2 categories to create discussions on.
2. Add a unit tests that properly fails when you try to assign a category that does not exist to a discussion.

initvector:
Vanilla's abstract API resource tests do some general modification of data during testing. One such test increments the value of any field ending in "ID". A recent patch requires a valid category ID, so blindly incrementing the `CategoryID` column no longer works.

This update fixes discussions API tests. Three test categories are created before the tests are run and tests are limited to those categories. In addition, a small tweak to `DiscussioModel::save` has been added to default its use-categories config check to `true`.

Closes #6202

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -1923,7 +1923,7 @@ class DiscussionModel extends Gdn_Model {
         $this->defineSchema();
 
         // If the site isn't configured to use categories, don't allow one to be set.
-        if (!c('Vanilla.Categories.Use')) {
+        if (!c('Vanilla.Categories.Use', true)) {
             unset($formPostValues['CategoryID']);
         }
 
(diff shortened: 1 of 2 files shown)


initvector:
Backporting to [release/2.5](https://github.com/vanilla/vanilla/tree/release/2.5)

PR #6204, opened 2017-12-03
Fix test compatibility with PHPUnit 6

initvector:
Vanilla's tests aren't compatible with PHPUnit v6. This update tweaks several items to make the tests compatible with PHPUnit v6, while maintaining compatibility with PHPUnit v5.

1. Replaced all references to `PHPUnit_Framework_TestCase` with the namespaced equivalent (`PHPUnit\Framework\TestCase`). This forward-compatible class name is compatible with PHP v5.
1. Added error class aliases, mapping relevant PHPUnit v6 namespaced classes (e.g. `PHPUnit\Framework\Error\Notice`) to their PHPUnit v5 counterparts (e.g. `PHPUnit_Framework_Error_Notice`).

The full suite of Vanilla tests should run under both PHPUnit v5 and v6.

### DIFF
--- a/tests/APIv0/AltTest.php
+++ b/tests/APIv0/AltTest.php
@@ -7,14 +7,12 @@
 
 namespace VanillaTests\APIv0;
 
-use Garden\Http\HttpClient;
-use Garden\Http\HttpResponse;
-use PDO;
+use PHPUnit\Framework\TestCase;
 
 /**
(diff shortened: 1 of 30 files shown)


DaazKu:
Follow up of #5606 which contains some improvements that we discarded from #5495

PR #6213, opened 2017-12-05
Backport: Fix setting user roles with the API

initvector:
This update backports #6200 to [release/2.5](https://github.com/vanilla/vanilla/tree/release/2.5).


DaazKu:
The unit tests fails because we didn't not backport [this](https://github.com/vanilla/vanilla/pull/6203) which only affects unit tests.

Issue #6225, opened 2017-12-07
Add utility function to convert query parameter to filter parameters using the schema


DaazKu:
> I'm noticing a lot of querystring variables that get passed right through to the where if they are set. These if statements should be consolidated to convenience function calls.
>
>When an event modifies a schema let it add an "x-filter" property to its variables and those will be passed through to the where clause by a standard function.
> 
```php
[
    'groupID:i?' => [
        'description' => '...',
        'x-filter' => true, // pass through to where as is.
    ],
    'groupID:i?' => [
        'description' => '...',
        'x-filter' => 'd.GroupID', // pass through to the where when the model needs to be coerced.
    ]
]
```

DaazKu:
I ended up doing a lot of thing to make this work (update garden-schema, refactor the DateFilterSchema etc...) Bumping the estimate to reflect that instead of creating 5 issues.

PR #6233, opened 2017-12-13
Add queryToFilters functionality and refactor DateFilterSchema

DaazKu:
This is a big PR that does many things.

- Add queryToFilters to ApiUtils.
    - You can set x-filter parameters on schemas to convert validated queries to where clauses. Pretty neat feature. 
    - Create unit tests for queryToFilters()
- Refactor DateFilterSchema
    - Move dateFilterField function from AbstractApiController to DateFilterSchema object since that's where it should belong.
    - Add properties to the schema and auto-convert the string notations to the expected schema structure.
    - Add default description on the schema (Add that default description to the provided description on initialization. To set a truly custom description one can do `$dateFilterSchema->setDescription()`).
    - Allow fields to be passed to the schema.
    - Rework and improve unit tests
- Update composer.json to get the latests version of garden-schema (needed by this PR)

Closes https://github.com/vanilla/vanilla/issues/6225


library/Vanilla/DateFilterSchema.php
```diff
@@ -160,16 +189,27 @@ public function parseSimple($date, $operator, ValidationField $field) {
...
 
-        if (Invalid::isInvalid($clean) && $field->isValid()) {
-            // This really shouldn't happen, but we want to protect against seeing the invalid object.
-            $field->addError('invalid', ['messageCode' => '{field} is invalid.', 'status' => 422]);
+                if (Invalid::isInvalid($clean) && $field->isValid()) {
+                    // This really shouldn't happen, but we want to protect against seeing the invalid object.
+                    $field->addError('invalid', ['messageCode' => '{field} is invalid.', 'status' => 422]);
+                }
+            } else {
+            $field->addError('invalid', ['messageCode' => '{field} is not a valid date filter.']);
```
initvector: Need to increase this indent, just a little bit.

library/Vanilla/DateFilterSchema.php
```diff
@@ -160,16 +189,27 @@ public function parseSimple($date, $operator, ValidationField $field) {
...
+        $field = new ValidationField($validation, $this->getSchemaArray(), '', $sparse);
+
+        if (is_string($data)) {
+            $data = $this->stringToSchema($data, $field);
+        }
 
-        $clean = $this->validateDateFilter($data, $field);
+        if (Invalid::isValid($data)) {
+            if (is_array($data)) {
+                $clean = $this->validateField($data, $field, $sparse);
```
initvector: Nesting the assignment of `$clean` beneath two conditions means there's the possibility it won't be set, even though it's the function's return value.

library/Vanilla/DateFilterSchema.php
```diff
@@ -181,26 +221,88 @@ public function validate($data, $sparse = false) {
...
+        $sortedSimpleOperators = $this->simpleOperators;
+        usort($sortedSimpleOperators, function($a, $b) {
+            if (strlen($a) > strlen($b)) {
+                return -1;
+            } else if (strlen($a) < strlen($b)) {
+                return 1;
+            }
+            return 0;
+        });
+        $simpleOperators = '(?<op>'.implode('|', array_map('preg_quote', $sortedSimpleOperators)).'|)';
```
initvector: This is going to give you something like: `(?<op>\<\=|\>\=|\<|\>|\=|)`. Is the trailing `|` there actually necessary?

library/Vanilla/DateFilterSchema.php
```diff
@@ -181,26 +221,88 @@ public function validate($data, $sparse = false) {
...
+            $field->addError('invalid', ['messageCode' => '{field} is not formatted as a valid date filter.']);
+        }
+
+        return $result;
+    }
+
+    /**
+     * If the parameter value is a valid date filter value, return an array of query conditions.
+     *
+     * @throws ServerException
```
initvector: Does it?

library/Vanilla/DateFilterSchema.php
```diff
@@ -181,26 +221,88 @@ public function validate($data, $sparse = false) {
...
+                        break;
                 }
-            } elseif (preg_match('/^(?<op><=|>=|>|<|)?\s*(?<value>'.$this->dateRegEx.')/i', $value, $match)) {
-                $result = $this->parseSimple($match['value'], $match['op'], $field);
-            } else {
-                $field->addError('invalid', ['messageCode' => '{field} is not formatted as a valid date filter.']);
             }
         } else {
-            $field->addTypeError('string');
+            throw new Exception('Invalid data supplied to DateFilterSchema');
```
initvector: This exception message doesn't make a lot of sense to me. Should it read `dateFilterField` instead of `DateFilterSchema`?

library/Vanilla/DateFilterSchema.php
```diff
@@ -58,54 +89,53 @@ private function parseRange($dates, $open, $close, ValidationField $field) {
...
+            $field->addError('invalid', ['messageCode' => '{field} date filter range must contain 2 datetime values.']);
             return Invalid::value();
         }
-        array_walk($dateArray, 'trim');
 
-        if (!preg_match("/{$this->dateRegEx}/i", $dateArray[0], $beginParts) ||
-            !preg_match("/{$this->dateRegEx}/i", $dateArray[1], $endParts)
-        ) {
+        $fakeField = new ValidationField($this->createValidation(), null, null);
+        if (Invalid::isInvalid($this->validateDatetime($dateArray[0], $fakeField))
```
initvector: `Schema::validateDatetime` will return an "invalid" value or an instance of `DateTimeImmutable`. I believe you could use these return values, instead of attempting to instantiate new instances of `DateTimeImmutable` based on the same input, further down.

library/Vanilla/DateFilterSchema.php
DaazKu: In that case an exception will be thrown so it doesn't matter.

library/Vanilla/DateFilterSchema.php
DaazKu: yes

DaazKu: 2017-01-01 is valid since no operator defaults to '='

PR #6236, opened 2017-12-14
Use inline display for bittersweet top menu

olivier-grech:
By default, the top-menu links in the bittersweet theme will be displayed in block instead of inline.
This change is intended to fix that.


charrondev:
Since this PR hasn't received any action in some time, I'm closing it in favour of https://github.com/vanilla/vanilla/issues/6819.

PR #6238, opened 2017-12-15
Add uploading user photos with the API

initvector:
This update adds the ability to set and remove a user's photo with the API.

### Additional
1. Added the ability to perform an in-place save of images with `ImageResizer`.
1. Added a method to retrieve the protected `ImageResizer::$typeExt` property.
1. Added the ability to get an upload's temporary filename to `UploadedFile`.
1. Added `$copy` parameter to `Gdn_Upload::saveAs` for allowing copying an upload a destination, instead of moving it.
1. Added `Controller::generateUploadPath` method for generating a valid, randomized upload path/filename, relative to the uploads directory.
1. Added `Controller::saveUpload` for saving uploads represented as instances of `UploadedFile`.
    * Allows for easy manipulation of the file's base name, which is handy for the full-size (p) and thumbnail (n) prefixes used by Vanilla for image uploads.
1. Added `UsersApiController::processPhoto` to prepare a photo (and its thumbnail) for saving.
1. Added `UsersApiController::savePhoto` to resize and save uploaded images.
1. Added the ability to set options (e.g. allowed extensions, maximum file size) to `UploadedFileSchema` constructor.
1. Added `Upload` upload class for tests to simulate user uploads.
1. Added `Vanilla\FileUtils` to abstract `is_uploaded_file` and `move_uploaded_file`.
1. Added `VanillaTests\Fixtures\FileUtils` to handle simulated uploads during testing.
1. Added tests for uploading and deleting user photos using the API.
1. Updated the directory creation in `Gdn_Upload::saveAs` to allow recursively creating directories, similar to what is currently used in [`Gdn_UploadImage::saveImageAs`](https://github.com/vanilla/vanilla/blob/bdf4b2b18364263ace7accbfaa620d3ae040f4be/library/core/class.uploadimage.php#L179).

### DIFF
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -366,7 +366,7 @@ class UsersApiController extends AbstractApiController {
     public function post_photo($id = null, array $body) {
         $this->permission('Garden.SignIn.Allow');
 
-        $photoUploadSchema = UploadedFileSchema::parse([]);
+        $photoUploadSchema = new UploadedFileSchema();
         $photoUploadSchema->setAllowedExtensions(array_values(ImageResizer::getTypeExt()));
 
         $in = $this->schema([
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -366,8 +366,9 @@ class UsersApiController extends AbstractApiController {
     public function post_photo($id = null, array $body) {
         $this->permission('Garden.SignIn.Allow');
 
-        $photoUploadSchema = new UploadedFileSchema();
-        $photoUploadSchema->setAllowedExtensions(array_values(ImageResizer::getTypeExt()));
+        $photoUploadSchema = new UploadedFileSchema([
+            'allowedExtensions' => array_values(ImageResizer::getTypeExt())
+        ]);
 
         $in = $this->schema([
             'photo' => $photoUploadSchema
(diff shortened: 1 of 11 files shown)


applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -318,6 +355,48 @@ public function post(array $body) {
...
+     *
+     * @param $id A valid user ID.
+     * @param array $body The request body.
+     * @throws ClientException if the image provided is not supported.
+     * @return array
+     */
+    public function post_photo($id = null, array $body) {
+        $this->permission('Garden.SignIn.Allow');
+
+        $photoUploadSchema = UploadedFileSchema::parse([]);
```
DaazKu: It would be more efficient to use `new UploadedFileSchema();` since Schema::parse() does unnecessary operations.
Also, UploadedFileSchema constructor does not take any parameters so supplying an empty array will just confuse anyone reading that code.

library/core/class.upload.php
```diff
@@ -318,13 +323,14 @@ public function saveAs($source, $target, $options = [], $copy = false) {
         if (!$handled) {
             $target = PATH_UPLOADS.'/'.$parsed['Name'];
             if (!file_exists(dirname($target))) {
-                mkdir(dirname($target));
+                mkdir(dirname($target), 0777, true);
```
DaazKu: Files should be 775 and directories 664.
Same below.

initvector: 0777 is actually the default file mode for [`mkdir`](http://php.net/mkdir). Any directories it was making in the past would've also been 0777. The only thing I actually changed here allowing for the creation of nested directories.

DaazKu: Let's create an issue about it

initvector: https://github.com/vanilla/vanilla/issues/6254

DaazKu: https://github.com/vanilla/vanilla/issues/6254

Issue #5177, opened 2017-02-13
[MySQL Strict Mode] compliance


nikoskip:
Hello,

I have a fresh installation of Vanilla 2.3 running along with MySQL 5.7.17 and I had a lot problems while I was playing with the administrator and the forum. Some actions triggered the error screen of Vanilla and it took me a while until I found that the problem was related with **strict mode** in MySQL.

My version of MySQL by default comes with this sql-mode:

`ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION`

So, finally I removed **STRICT_TRANS_TABLES** and for now, all is working as expected. 

I know it could take a lot of work to review the code and find this specific problems, but it would be nice if you guys add this requirement in the `README.md ` file. I found some related posts with the same problem:

- https://github.com/vanilla/vanilla/issues/4117
- https://open.vanillaforums.com/discussion/17730/vanilla-doesnt-work-well-in-mysql-strict-mode

Thanks! 

**UPDATE**

I also had to deactivate **ONLY_FULL_GROUP_BY** to get some core plugins work.

Charrondev:
To get everything working I had to disable **STRICT_TRANS_TABLES**, **ONLY_FULL_GROUP_BY**, and **NO_ZERO_DATE**. We should document this in the README.

DaazKu:
Please notify Ops when this Epic is closed.

pinchies:
Is it not possible for vanilla forum to manually set the desired SQL_mode on a per-session basis?

https://www.digitalocean.com/community/tutorials/how-to-prepare-for-your-mysql-5-7-upgrade

I was recently burned by this, because the "In this discussion" plugin is incompatible with Strict mode.

`PHP Fatal error:  Expression #1 of ORDER BY clause is not in GROUP BY clause and contains nonaggregated column 'jgaforum_vf1.c.DateInserted' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by in /home/jgaforum/public_html/library/database/class.database.php on line 421
`
By disabling that plugin, my issue was resolved. The cause is line 35 in vanilla/plugins/VanillaInThisDiscussion/class.inthisdiscussionmodule.php:

 ```
   public function getData($discussionID, $limit = 50) {
[...]

PR #6366, opened 2018-01-09
Add a note about MySQL strict mode to README

linc:
See  #5177

For now, mentioning it is better than nothing.


DaazKu:
I'm good with that but I also think we should try to put some effort on fixing those issues.
We should require devs to have strict mode enabled.

MySQL is shipped with strict mode enabled.

linc:
I agree, I'm just trying to limit open source support overhead in the meantime.

Issue #5832, opened 2017-08-09
Invalid default value for 'DateExpires'


RubbaBoy:
When installing Vanilla for the first time with a clean apache server and MySQL server, the error "Invalid default value for 'DateExpires'" occurs.

DaazKu:
Probably related: https://github.com/vanilla/vanilla/issues/5763

DaazKu:
See this article for a possible workaround: https://www.percona.com/blog/2016/10/18/upgrading-to-mysql-5-7-beware-of-the-new-strict-mode/

linc:
The workaround is to disable MySQL's strict mode.

linc:
Note from @DaazKu:

> This can be reproduced on MySQL and MariaDB but **not** on Percona

Vanilla uses Percona exclusively internally and for our automated testing, which is how this slipped by us. That was a major question for me.

PR #6376, opened 2018-01-09
Fix timestamp default values on AccessToken table

DaazKu:
Fixes: https://github.com/vanilla/vanilla/issues/5832

This update allow to set the default value of a timestamp to 'current_timestamp' or 'current_timestamp()' and refactor the AccessToken table to use that new feature.


linc:
Is this backwards compatible, or will it break existing sites until the update is run?

DaazKu:
Only the structure is changed so it won't do anything at all until the update is ran.

PR #6436, opened 2018-01-12
Add permissions to the client

### LINKED ISSUE #6416: Add permissions to client (Gdn_Controller)

tburry:
Will allow the client to modify the UI based on a user's permissions. Note that this isn't intended for security as all permissions must be verified on the server.

Closes #6416.


tburry:
@initvector I'm hoping you can review the server side stuff. This is mainly a refactor with a couple of changes:

1. I think the RolesApiController was not translating permission names properly. I switched the name scheme to account for the dots in the names.
2. The Vanilla\Permissions object now has a JsonSerialize method to allow the permissions to go into the client.

@Charrondev I'm guessing any client side bugs can be caught when we are developing with the new code. If you can look for glaring mistakes that would be great.

Charrondev:
I tweaked a couple of things, but it looks good. You can use `const` a lot more liberally. If you enable ESLint in PHP Storm it will warn you when you use a let/var on something that is never re-assigned.

I also wrapped the console.log in @core/gdn in a check for debug mode. The utility.log function will do that for you too.

PR #6442, opened 2018-01-15
Fix error message on profile post lists when all page content is filtered

### LINKED ISSUE #6107: User comments on profile paginate by total, not total viewable
**How the problem originally presented:** User with 600+ comments cannot view comments past page 3, getting message "The user has not commented yet".

@initvector investigation:

> This certainly seems like a permissions issue.
> 
> This user has not commented yet" is a bad message. What it actually means is it couldn't find any comments given the criteria we're using. If the user only has access to 60 comments, but the pagination thinks they have access to 600, it'll fail to find any comments after the 60th (e.g. "gimme 20 more comments, starting from 61").
> 
> The user has posted roughly 120 comments in categories they appear to no longer have access to.
> 
> The problem is the filtering appears to be taking place after the comments have been polled. We grab all the comments the user has created, then manually filter out the ones they don't have access to any more on a given page. This effectively means some pages in comment pagination can be empty, which coupled with that bad "error" message, stop you from advancing to the next batch of comments. For proof of this, you can manually jump ahead in the cited user's comment activity to see they do have more pages of comments they can access.

## Technical Details

CommentModel::getByUser2 fetches all comments by a user, then removes rows from the result set, based on permissions. As explained above, this can mean a user gets an empty result set, because they no longer have access to any of the comments on a particular page.

DaazKu:
Fixes #6107

First of, let's start by saying that the basic pager uses "more" which does not let you access "pages".
Since `/profile/discussions/{DISCUSISON_ID}/{USER_NAME}/p{NUMBER}` works, I added a detection that checks if the page had data before the permission filtering. If that's the case we display a more relevant message and let the "more" page active.

This solves the issue.

### DIFF
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -853,6 +853,7 @@ class VanillaHooks implements Gdn_IPlugin {
             $sender->View = 'profilecomments';
         }
         $sender->setData('Comments', $comments);
+        $sender->setData('UnfilteredCommentsCount', $commentModel->LastCommentCount);
 
         // Set the HandlerType back to normal on the profilecontroller so that it fetches it's own views
         $sender->HandlerType = HANDLER_TYPE_NORMAL;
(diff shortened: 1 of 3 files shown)


initvector:
Please make sure these translation strings are added to the [vanilla/locales](https://github.com/vanilla/locales) repo.

PR #6445, opened 2018-01-17
Fix session state during sign-in attempts when a user is banned

### LINKED ISSUE #5988: [SSO] There is no feedback that a session could not be started when connecting over SSO
We [start the session](https://github.com/vanilla/vanilla/blob/3ed9e2ba2e0f2f6bdfc6ba7c2988c1a0d60597d5/applications/dashboard/controllers/class.entrycontroller.php#L550), it fails and we redirect to the supplied target. 
0 feedback that the session could not start due to permission problems (like if the user was banned)

DaazKu:
Fixes #5988

- Session is now "invalid" if the user is banned or deleted.
- Throw the AfterSignIn event everywhere.
- Throw an exception on AfterSignIn if the user is banned.
    - This take care of #5988 
    - It is done that way (instead of in Session::start() to preserve backward compatibility). I'm also unsure as to why we are filling a ban array instead of throwing an exception but that tells me that we are probably checking the bad array later in the code.

How to test:
- Use entry/signin via PopUp
- Use entry/signin with no popup.
- Use VanillaConnect (which uses the /authenticate endpoint)
- Use any other SSO plugin (which uses the /entry/connect endpoint)

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -665,7 +665,7 @@ class DashboardHooks extends Gdn_Plugin {
      */
     public function base_afterSignIn_handler() {
         if (!Gdn::session()->isValid()) {
-            if ($ban = Gdn::session()->getPermissions()->getBan([])) {
+            if ($ban = Gdn::session()->getPermissions()->getBan()) {
                 throw new ClientException($ban['msg'], 401, $ban);
             } else {
                 if (!Gdn::session()->getPermissions()->has('Garden.SignIn.Allow')) {
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -10,6 +10,8 @@
 
 use Garden\Container\Container;
 use Garden\Container\Reference;
+use Garden\Web\Exception\ClientException;
+use Vanilla\Exception\PermissionException;
 
 /**
  * Event handlers for the Dashboard application.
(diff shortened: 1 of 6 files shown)


applications/dashboard/settings/class.hooks.php
```diff
@@ -656,6 +658,23 @@ public function gdn_dispatcher_appStartup_handler($sender) {
         $this->checkAccessToken();
     }
 
+    /**
+     * Check to see if a user is banned.
+     *
+     * @throws Exception if the user is banned.
+     */
+    public function base_afterSignIn_handler() {
+        if ($ban = Gdn::session()->getPermissions()->getBan([])) {
```
initvector: I think what we're trying to target here are user rows with their `Banned` or `Deleted` flags toggled on. This is going to trip for every ban type (e.g. 2FA, CSRF). Wouldn't it be better to test for just the ban types you're expecting? Maybe someone signs into the site and is "banned" (no permissions) until they enter their 2FA auth.

The empty array in the call to `getBan` is superfluous, because the default is already an empty array.

DaazKu: I'm pretty sure that all ban reasons should fail here.

initvector: It's my understanding that we just want to block deleted and banned users from accessing the site here.

My take on permission bans is that they're meant to strip user permissions, not necessarily block them from signing in. There's several ban types that aren't necessarily related to signing in (2FA, CSRF, updating, uninstall, private community, etc.). Addons can pretty much do whatever they like there. I think it's maybe casting the net too wide to reject anyone from signing into the site who has any kind of permission "ban". I don't think "you don't have any permissions" should always equate to "you cannot sign in".

DaazKu: Makes sense. I updated the code appropriately.

PR #6446, opened 2018-01-17
Fix category permission check when using the comments API endpoint

initvector:
`CommentModel::lookup` has a routine to filter out comments the user doesn't have access to. It does this by looking at the category the comment is in and comparing it against the result of `DiscussionModel::categoryPermissions`. The problem is, in its current state, the category ID isn't available. `CommentModel::lookup` is referencing an undefined property. Because of this, custom category permissions on a site can filter out *all* comments on the comments API endpoint for some users/roles (e.g. ["if `null` isn't in my list of allowed categories, remove this comment](https://github.com/vanilla/vanilla/blob/9491a2dca2f527ebc22c615ef02526afdd403e7b/applications/vanilla/models/class.commentmodel.php#L476)).

This update joins in the Discussion table when querying comments. Since category ID is now brought over from the discussions table as part of the row, comments can be properly filtered.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -465,7 +465,30 @@ class CommentModel extends Gdn_Model {
             return new Gdn_DataSet([]);
         }
 
-        $result = $this->getWhere($where, 'CommentID', $order, $limit, $offset);
+        // All fields should be associated with a table. If there isn't one, assign it to comments.
+        foreach ($where as $field => $value) {
+            if (strpos($field, '.') === false) {
+                $where["c.{$field}"] = $value;
+                unset($where[$field]);
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.commentmodel.php
```diff
@@ -465,7 +465,32 @@ public function lookup(array $where = [], $permissionFilter = true, $limit = nul
             return new Gdn_DataSet([]);
         }
 
-        $result = $this->getWhere($where, 'CommentID', $order, $limit, $offset);
+        // Avoid ambiguous fields.
+        $ambiguousFields = ['Body', 'DiscussionID', 'Format'];
```
DaazKu: This has the potential to break if we do an update and will probably fail with `DateInserted` anyway.

I suggest that you pad all $where with 'c.' instead.

Issue #6026, opened 2017-10-11
Category following


tburry:
We want the ability to follow categories to improve the UX on communities with many categories. This is not the same as Vanilla's current "unfollow" functionality, but will replace that.

## Requirements

- Users can follow or unfollow a category. Only signed in users can follow categories.
- The recent discussions page will have the ability to filter by just followed categories.
- The categories page will have the ability to filter by just followed categories. This will be a flat style category display.
- When filtering discussions or categories by all or followed, the user's selection should be saved so that going back to the URL shows their last selection. In this way users are customizing their UI.
- Users will be able to receive notifications for their followed categories which will be a more scalable way to implement advanced notifications. (message queue required)
- Category selection dropdowns should eventually display followed categories first.
- You can only follow a category with displayAs: discussions
- The dropdown in category lists and tables should have Follow/Unfollow options, replacing the "Mute" options.
- The category following functionality should be conditional based on the Vanilla.EnableCategoryFollowing config setting. There should be a way to toggle this setting in the dashboard.

## Design

- [Follow button](https://drive.google.com/drive/folders/12HWpLjOyAVvns5U2e6IJ2fIHkZ0uNmi_)
- View filter
- Message when the user hasn't followed any categories

## Implementation

- Implement `PUT /api/v2/categories/:id/follow` and `CategoryModel->follow()` (#6397).
- Add a filter for followed categories to `GET /api/v2/discussions` and `DiscussionModel->getWhere()` (#6398).
- Add a filter for followed categories to `GET /api/v2/categories` and `CategoryModel->getWhere()` (#6399).
- Add a filter for followed categories to `GET /api/v2/discussions/search`. The followed categories will be queried from the CategoryModel and passed into Sphinx (vanilla/internal#1346).
- Add following querystring options to /discussions and /categories (#6400).
- Add the follow/unfollow button to category page (#6401).
- Add filter dropdowns to the application to switch followed categories and all. The new base theme has this functionality planned already, but the current base theme is going to need some work later and ideation will need to be updated at that time. (#6402)
- Add config options to save the users previous selection so going to /categories or /discussions will select the current config option. (#6403)
- Add followed categories option to categories dropdowns. This will require the work done replacing the category select with a scalable category search component similar to the one in the Hootsuite app. (#6405)
- Add a followed categories option to advanced search. (vanilla/internal#1347)
- The view filter is now a core feature and must be integrated with ideation. (vanilla/internal#1348)

## Optimization

[...]

PR #6452, opened 2018-01-18
Add explicit category following

### LINKED ISSUE #6397: Add basic ability to follow to the categories
An overview of the "category following" feature can be found on the parent issue: #6026

- Add non-static `follow` method to `CategoryModel`.
- Users can follow or unfollow a category.
- Only signed-in users can follow a category.
- You can only follow categories where `DisplayAs` = discussions.
- The category following UI should display conditionally based on the `Vanilla.EnableCategoryFollowing` config setting. There should be a way to toggle this setting in the dashboard. The config and control will be added as part of this issue. The subsequent UI tasks for this feature will rely on the config being available.
- Add PUT /api/v2/categories/:id/follow
    - Given a category ID and a boolean follow field, the current user can follow or unfollow the specified category.
- Add API unit tests to verify functionality

initvector:
Vanilla currently has a version of category following that has all categories being "followed" by default. A user has to "mute" a category to unfollow it.

Vanilla will be moving away from its blacklist-style following and towards a whitelist-style. This update introduces selective category following, so a users choose what categories to follow, instead of which categories to unfollow. The specifics of the overall feature are outlined in #6026. The subset of functionality introduced here is only the core components for toggling a category followed or not by a user, getting a list of followed categories and determining if a category is followed by a user. Details can be found in the associated issue, #6397.

The `Vanilla.EnableCategoryFollowing` config is introduced here, defaulting to `false`, but not used. Subsequent UI portions of this feature will rely on this config.

*Note: The "mute" feature of Vanilla has already claimed fields with "follow" in the name (e.g. `Unfollow`, `Following`). Since we can't just rip out these fields, I've had to use fields that are relevant and but different. They might seem redundant (e.g. `Follow`, `Followed`).*

Closes #6397

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -338,17 +338,18 @@ class CategoryModel extends Gdn_Model {
             throw new InvalidArgumentException('Invalid $categoryID');
         }
 
+        $isFollowed = $this->isFollowed($userID, $categoryID);
+        if ($follow === null) {
+            $follow = !$isFollowed;
+        }
+        $follow = $follow ? 1 : 0;
+
         $category = static::categories($categoryID);
         if (!is_array($category)) {
             throw new InvalidArgumentException('Category not found.');
-        } elseif ($category['DisplayAs'] !== 'Discussions') {
+        } elseif ($category['DisplayAs'] !== 'Discussions' && !$isFollowed) {
             throw new InvalidArgumentException('Category not configured to display as discussions.');
         }
-
-        if ($follow === null) {
-            $follow = !$this->isFollowed($userID, $categoryID);
-        }
-        $follow = $follow ? 1 : 0;
 
         $this->SQL->replace(
             'UserCategory',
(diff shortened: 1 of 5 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -285,6 +295,96 @@ public static function getRootDisplayAs() {
...
+            throw new InvalidArgumentException('Invalid $userID');
+        }
+        if (!($categoryID = filter_var($categoryID, FILTER_VALIDATE_INT, $validationOptions))) {
+            throw new InvalidArgumentException('Invalid $categoryID');
+        }
+
+        $category = static::categories($categoryID);
+        if (!is_array($category)) {
+            throw new InvalidArgumentException('Category not found.');
+        } elseif ($category['DisplayAs'] !== 'Discussions') {
```
DaazKu: Unfollowing should work here. (unit test that)

initvector: Makes good sense. Done.

Issue #6374, opened 2018-01-09
Browse button to select an image is missing in recent activity and signatures


dannimajdalani:
## Issue

When Plugin signatures is enabled with Advanced Editor, browsing to profile/signature to upload an image to the signature is missing the browse button

## Signatures Screen
![signatures](https://user-images.githubusercontent.com/31856281/34739973-ff042444-f54b-11e7-94f5-860064f1cd43.png)

Similar case in Recent Activity Page /activity

## Recent Activity Screen
![activity](https://user-images.githubusercontent.com/31856281/34740049-26b94cf8-f54c-11e7-90d0-0ca9be9a793d.png)

## Normal Behaviour

![normal](https://user-images.githubusercontent.com/31856281/34740156-708a179a-f54c-11e7-8978-9653c4540db9.png)

LauraMakesArt:
This appears to also not have been resolved, (see: https://vanillaforums.teamwork.com/#tasks/8646770) I see the image upload on activity on profiles, now, but not on recent activity, or when creating polls.

charrondev:
Instead of going with the approach of manually adding image-upload why not just have it on by default I see 10+ places where we want it and can't think of many where we don't want it. 

@linc Do you know where we would *not* to spawn the editor without the image upload? As I understand it will still get filtered out by permission if not allowed.

charrondev:
After speaking with Todd, this will be added everywhere as a default param.

PR #6453, opened 2018-01-18
Fix image upload button not always appearing in Advanced Editor

DaazKu:
Update the advanced editor so that image uploads can be activated without file upload.
Fixes https://github.com/vanilla/vanilla/issues/6374

Specifying FileUpload allows both file and image upload.
Specifying ImageUpload only allows image upload.


LauraMakesArt:
Anyway we can get this issue backported?

vrjrobitaille:
Though this is merged, it does not appear to be deployed to staging, or production yet.

PR #6467, opened 2018-01-23
Add advanced category settings

### LINKED ISSUE #6394: Add a config option to turn on/off category following in the UI
Because category following can have theme problems with old themes we need to launch with it disabled.

- The config setting is Vanilla.EnableCategoryFollowing
- There needs to be an advanced settings in the manage categories page for this.

tburry:
This page allows category following to be turned on/off.

Closes #6394.

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -105,32 +105,47 @@ if (!function_exists('heading')) {
      * Handles url-ifying. Adds an optional button or return link.
      *
      * @param string $title The page title.
-     * @param string $buttonText The text appearing on the button.
+     * @param string|array $buttonText The text appearing on the button or an array of button definitions.
      * @param string $buttonUrl The url for the button.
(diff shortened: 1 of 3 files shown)


library/core/functions.render.php
```diff
@@ -105,32 +105,47 @@ function helpAsset($title, $description) {
...
         }
 
-        if ($buttonText !== '') {
-            if (val('class', $buttonAttributes, false) === false) {
-                $buttonAttributes['class'] = 'btn btn-primary';
+        $buttonsString = '';
+        foreach ($buttons as $button) {
+            $buttonText = $button['text'];
+            $buttonUrl = $button['url'];
+            $buttonAttributes = $button['attributes'];
```
initvector: Currently triggering a notice on the Manage Category page

> Undefined index: attributes

tburry: Okay I fixed

Issue #6402, opened 2018-01-11
Add filter dropdown to categories page to show followed categories or all


initvector:
Add filter dropdowns to the application to switch followed categories and all. The new base theme has this functionality planned already, but the current base theme is going to need some work later and ideation will need to be updated at that time.

The "followed" category list will be displayed as flat.

slafleche:
For discussions equivalent, see https://github.com/vanilla/vanilla/issues/6409

slafleche:
Using markup from https://github.com/vanilla/vanilla/issues/6414

slafleche:
Markup and CSS ready to go. Put on hold until we're ready to "wire it up".

PR #6469, opened 2018-01-23
Add UI for category following

### LINKED ISSUE #6400: Add followed-category query string parameters to discussions and categories index pages
A query parameter (e.g. "followed") needs to be added to /discussions and /categories for filtering the results on the basis of whether or not they are in the users' followed categories.

### LINKED ISSUE #6403: Add config for saving followed-category filtering for users
Add config options to disable saving a user's categories index/recent discussions filter preferences for followed categories. A user selecting to filter recent discussions to only their followed categories would see the page always filtered that way, even between visits, until the filter was altered.

initvector:
This update adds UI elements for category following.

A new config is also introduced in this update: `Vanilla.SaveFollowingPreferences`. When enabled (as by default), setting the "Following" filter on the recent discussions or root category index pages will save that preference for the current user. When they re-visit the page, the filters will be automatically re-applied.

A function for create the filter dropdown for categories as well as the follow button were added. Styling for these elements were mostly added in the styles-compat.scss file, which will always be added from now old to legacy themes.

Closes #6400 
Closes #6403 
Closes https://github.com/vanilla/vanilla/issues/6402


applications/vanilla/controllers/class.categorycontroller.php
```diff
@@ -33,6 +33,59 @@ public function follow($categoryID, $value, $tKey) {
...
 
+    /**
+     * Allows user to follow or unfollow a category.
+     *
+     * @param int $DiscussionID Unique discussion ID.
+     */
+    public function followed($categoryID = null) {
+        // Make sure we are posting back.
+        if (!$this->Request->isAuthenticatedPostBack()) {
+            //throw permissionException('Javascript');
```
DaazKu: Why is that commented?

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -120,6 +120,20 @@ public function index($Page = false) {
 
         $this->setData('Breadcrumbs', [['Name' => t('Recent Discussions'), 'Url' => '/discussions']]);
 
+        $followed = Gdn::request()->get('followed', null);
```
DaazKu: Code duplication: https://github.com/vanilla/vanilla/pull/6469/files#diff-297e705faae450995d693b7e5c568779R252

applications/vanilla/views/categories/all.php
```diff
@@ -1,10 +1,20 @@
 <?php if (!defined('APPLICATION')) exit();
 include_once $this->fetchViewLocation('helper_functions', 'categories');
-echo '<h1 class="H HomepageTitle">'.$this->data('Title').'</h1>';
+$title = $this->data('Title');
+$userID = Gdn::session()->UserID;
```
DaazKu: Where is  $userID used? There are multiple places where that's being done.

DaazKu: Probably an artifact left by refactoring the code.

applications/vanilla/views/discussion/index.php
```diff
@@ -21,6 +21,8 @@
 
 echo '<h1>'.$this->data('Discussion.Name').'</h1>';
 
+
```
DaazKu: Useless extra lines.

library/core/functions.render.php
```diff
@@ -651,6 +684,39 @@ function commentUrl($comment, $withDomain = true) {
...
+        }
+
+        if (!empty($defaultParams)) {
+            $defaultUrl = $baseUrl.'?'.http_build_query($defaultParams);
+        } else {
+            $defaultUrl = $baseUrl;
+        }
+
+        return filtersDropDown(
+            url('discussions'),
```
DaazKu: You can use $baseUrl here

library/core/functions.render.php
```diff
@@ -393,6 +393,39 @@ function category($depth = null, $category = null) {
...
+        }
+
+        if (!empty($defaultParams)) {
+            $defaultUrl = $baseUrl.'?'.http_build_query($defaultParams);
+        } else {
+            $defaultUrl = $baseUrl;
+        }
+
+        return filtersDropDown(
+            url('categories'),
```
DaazKu: You can use $baseUrl here

library/core/functions.render.php
```diff
@@ -938,6 +1069,79 @@ function ipAnchor($iP, $cssClass = '') {
...
+                $output .= '</li>';
+            }
+        }
+        $output .= <<<EOT
+              </ul>
+            </span>
+          </span>
+        </span>
+EOT;
+
```
DaazKu: Useless extra lines below

PR #6473, opened 2018-01-24
Add followed category filtering to mixed category layout

initvector:
This update adds the "Following" filter menu to the mixed category layout for filtering to followed categories.


slafleche:
My utility/update is broken on this branch... not sure if this is just a local problem yet.
I also have no categories show up on the /categories

slafleche:
The /utility/update is a problem on my local host.

False alarm for the empty /categories. We'll need to write a message when there are no categories, but that's a separate issue.

PR #6520, opened 2018-01-25
Fix image crop overflow

Charrondev:
Closes https://github.com/vanilla/support/issues/7

The image cropper was overflowing on mobile and desktop, with our current image cropper the image needs to be at a 1:1 pixel ratio to work properly and the contents are absolutely positioned inside, so I can't just resize the image.

This PR affects the image cropper on 2 different places on every theme (Using style-compat). Most noticeably improved are any responsive-themes and our basic mobile theme. Lithe does not offer a profile image editing at this time so it is unaffected. 

## Image Upload Popup

- Max width set to 100%, and overflow auto on the container to prevent overflow.
- Profile cropper has been hidden hidden at 600px breakpoint to make the popup usable for normal photo changing.

## Edit image profile page
- Max width set to 100%, and overflow auto on the container to prevent overflow.
- Profile cropper is always visible on this page.


Charrondev:
This issue does not just affect mobile. It affects desktop views as well if the image has a larger pixel with than the main content area. The image does not get shrunk (we can't do that right now without replacing the library). This fix just prevents it from overflowing on top of other content like this:

## Before
image

## After
image

The image is in a scrollable container now.

This is not "Oh this is perfect now". Ideally we could scale the image down. Instead this PR prevent a large image from totally breaking desktop and mobile themes.

slafleche:
The support issue specifically talks about mobile. I agree that there's an issue with some themes, but it's a different issue.

On desktop, I'm getting 2 scroll bars with your updates:
screen shot 2018-01-26 at 11 33 38

The cropper plugin doesn't take into consideration your max-width on the container, it still overflows and I'm unable to scroll into that area. It shouldn't have an `overflow: auto;`;

I'm not sure how the cropper works, but it might not work on a resized image, seeing as it probably measures in pixels for cropping and not in a relative measurement like a percentage.

Charrondev:
I've limited the scope of this PR to now just hide the cropper from mobile themes. Anything else can happen as support requests come in.

I've also opened an issue to replace the cropped in the future https://github.com/vanilla/vanilla/issues/6547

applications/dashboard/views/modules/cropimage.php
```diff
@@ -1,4 +1,4 @@
-<div class="padded">
+<div class="padded thumbnail-message">
```
slafleche: This is a bit too generic of a name. `change-picture-instructions` is clearer, since we may have thumbnails in multiple contexts.

applications/dashboard/scss/legacy/_style.scss
```diff
@@ -2488,6 +2488,17 @@ dl.About dd {
   display: block
 }
 
+/* ======================================================== Image Cropper */
+body.isMobile {
```
slafleche: Don't forget you can also change your profile picture on the `/profile` page. You should also be hiding: `. ChangePicture` (link on the thumbnail) and `. PictureLink` in the panel.

Charrondev: I don't see why we need to hide the page or the popup itself. Changing pictures works particularly well without the cropper and with things like the avatar pool.

slafleche: You're right, Keep the ability to change the picture, just hide the cropper.

applications/dashboard/scss/style-compat.scss
```diff
@@ -1,5 +1,5 @@
-// A style sheet for compatibility with new features.
 // This should always be added after style.css and not be overridden.
+
```
slafleche: If you leave your styles in `_style.css` they can be overwritten. Let's make a new folder "compatibility" and move your styles in there. Include them through `style-compat.scss`

Charrondev:
I've moved my changes as your requested. We'll likely need to fix some merge conflicts when the rich editor branch comes in, but that's expected.

PR #6530, opened 2018-01-26
Removed back links from comment form

slafleche:
The back links for comments are sometimes broken and are often hidden by themes. We've just removed them from Vanilla.

Closes https://github.com/vanilla/vanilla/pull/6516

### DIFF
--- a/applications/vanilla/views/post/comment.php
+++ b/applications/vanilla/views/post/comment.php
@@ -43,16 +43,16 @@ $this->fireEvent('BeforeCommentForm');
                         $CancelClass = 'Cancel';
                     }
 
-                    echo '<span class="'.$CancelClass.'">';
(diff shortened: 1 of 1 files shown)


initvector:
I'd rather see those lines removed, instead of commented. They aren't going to be lost, because they're in the repo's revision history. [If they need to be re-added, it's very easy to recover them.](https://github.com/vanilla/vanilla/blob/3381c1f4a8dbe449ac4bf8d1b350884b446c6224/applications/vanilla/views/post/comment.php#L46) In the event they aren't re-added (which I think is more likely), this dead code will be sitting in Vanilla, indefinitely.

I'd be okay with commented out lines, if we knew for sure it was going to be temporary.

DaazKu:
#6516 is a PR I made. Not an issue....

Issue #6480, opened 2018-01-24
Refactor header links according to rfc5988


DaazKu:
See https://tools.ietf.org/html/rfc5988#page-7
Also see: https://developer.github.com/v3/guides/traversing-with-pagination/

This means that we will have to redo: https://github.com/vanilla/vanilla/pull/6462

The header should be populated by an API views that is rendered when the content type is `application/json`

Breadcrumbs, paging and more will be handled by that view.

PR #6537, opened 2018-01-26
Update API to use Web Linking

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/6480

- Add a basic class to handle Web Linking
   - This class can be used application wise to add links. The new JsonView uses it to add the relevant header to the Data object before rendering it. Our old Head module can do the same thing.
- Create a custom view that allow to transform Data->meta to headers
- Remove setPageMeta function from APIUtils since it does basically nothing now. (Do not forget to merge https://github.com/vanilla/internal/pull/1385 too)

### DIFF
--- a/library/Vanilla/Web/JsonView.php
+++ b/library/Vanilla/Web/JsonView.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * @author Alexandre (DaazKu) Chouinard <alexandre.c@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license https://opensource.org/licenses/GPL-2.0 GPL-2.0
+ */
+
+namespace Vanilla\Web;
+
+use Garden\Web\ViewInterface;
+use Garden\Web\Data;
+
+/**
+ * Class JsonView
+ */
+class JsonView implements ViewInterface {
+
+    /** @var WebLinking */
+    private $webLinking;
+
+    /**
+     * JsonView constructor.
+     *
+     * @param WebLinking $webLinking
+     */
+    public function __construct(WebLinking $webLinking) {
+        $this->webLinking = $webLinking;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function render(Data $data) {
+
+        $paging = $data->getMeta('paging', []);
+
+        foreach(($paging['links'] ?? []) as $relType => $uri) {
+            $this->webLinking->addLink($relType, $uri);
+        }
+
+        $data->setHeader('Link', $this->webLinking->getLinkHeaderValue());
+
+        echo $data->render();
+    }
+}
--- a/library/Vanilla/Web/JsonView.php
+++ b/library/Vanilla/Web/JsonView.php
@@ -31,11 +31,24 @@ class JsonView implements ViewInterface {
      * {@inheritdoc}
      */
     public function render(Data $data) {
+        $paging = $data->getMeta('paging');
 
-        $paging = $data->getMeta('paging', []);
+        // Handle pagination.
+        if ($paging) {
+            $hasPageCount = isset($paging['pageCount']);
 
-        foreach(($paging['links'] ?? []) as $relType => $uri) {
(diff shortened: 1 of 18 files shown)


library/Vanilla/Web/WebLinking.php
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @link http://tools.ietf.org/html/rfc5988
+     * @link http://www.iana.org/assignments/link-relations/link-relations.xml
+     *
+     * @param string $uri Target URI for the link.
+     * @param string $rel Link relation. Either an IANA registered type, or an absolute URL.
+     * @param array $attributes Link parameters.
+     * @return WebLinking
+     */
+    public function addLink($rel, $uri, $attributes = []) {
```
tburry: The rel should be second and optional. On the spec only the URI is required. Same for others.

library/Vanilla/Web/JsonView.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function render(Data $data) {
+
+        $paging = $data->getMeta('paging', []);
+
+        foreach(($paging['links'] ?? []) as $relType => $uri) {
+            $this->webLinking->addLink($relType, $uri);
+        }
+
+        $data->setHeader('Link', $this->webLinking->getLinkHeaderValue());
```
tburry: What if the data has other link headers? Think about how a developer might want to extend this.

library/Vanilla/Web/WebLinking.php
DaazKu: That's more or less incorrect. 

From: https://tools.ietf.org/html/rfc5988#section-3

> In this specification, a link is a typed connection between two
   resources that are identified by Internationalised Resource
   Identifiers (IRIs) [RFC3987], and is comprised of:

>   o  A context IRI,
>
>   o  a link relation type (Section 4),
>
>   o  a target IRI, and
>
>   o  optionally, target attributes.

From: https://tools.ietf.org/html/rfc5988#section-5 specifies that 

> The Link Header Field
> ... It is semantically equivalent to the <LINK> element in HTML

From: https://www.w3.org/TR/2014/REC-html5-20141028/document-metadata.html#the-link-element
> A link element must have a rel attribute.
> ...
> The types of link indicated (the relationships) are given by the value of the rel attribute, which, if present, must have a value that is a set of space-separated tokens. The allowed keywords and their meanings are defined in a later section. If the rel attribute is absent, has no keywords, or if none of the keywords used are allowed according to the definitions in this specification, then the element does not create any links.

My understanding of all this is that a link that has no rel is not a valid link.
All the examples in the spec support that conclusion and multiple implementations too.

I also put rel first because it is more convenient for the removal function.
You can remove all links of a specific relation with one call.

library/Vanilla/Web/JsonView.php
DaazKu: They just have to call "add" on the webLinking object.

DaazKu: We will have to add logics in here to handle that.

tburry: Okay do it.

DaazKu: Done already: https://github.com/vanilla/vanilla/pull/6537/commits/f90b1d31679574039619cc47b61262341b429015

library/Vanilla/Web/WebLinking.php
initvector: While I can't find evidence in the [original Web Linking RFC](https://tools.ietf.org/html/rfc5988) explicitly stating the "relation type" is required, I also can't find a single practical example of a link header that does not include it. Also, the latest  [Web Linking RFC](https://tools.ietf.org/html/rfc8288) *does* explicitly state the following:

> The relation type of a link conveyed in the Link header field is conveyed in the "rel" parameter's value.  The rel parameter MUST be present but MUST NOT appear more than once in a given link-value; occurrences after the first MUST be ignored by parsers.

Requiring a relation type seems like a good idea to me, although I do acknowledge it might not fully adhere to the original RFC.

PR #6548, opened 2018-01-29
Add title attribute to seperators

Charrondev:
The editor toolbar was throwing a warning trying to access a nonexistent title attribute (for displaying to screenreaders). This PR adds title attributes to all of the dividers. 

Translation source here https://github.com/vanilla/locales/pull/50

### DIFF
--- a/plugins/editor/views/editor.php
+++ b/plugins/editor/views/editor.php
@@ -12,10 +12,8 @@ $html_arrow_down = '<span class="icon icon-caret-down" aria-hidden="true"></span
 $editor_file_input_name = $this->data('_editorFileInputName');
 
 foreach ($this->data('_EditorToolbar') as $button) {
-    $screenReaderMarkup = '';
-    if ($title = valr('attr.title', $button)) {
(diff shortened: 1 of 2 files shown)


slafleche:
The divider has no pertinent information for screen readers. It should not have a title attribute. It should, however, have `role="presentation"' to be ignored.

Charrondev:
I've adjusted this PR to just remove the warning (by checking if the title is set before accessing it).

plugins/editor/views/editor.php
```diff
@@ -12,7 +12,10 @@
 $editor_file_input_name = $this->data('_editorFileInputName');
 
 foreach ($this->data('_EditorToolbar') as $button) {
-   $screenReaderMarkup = '<span class="sr-only">' . $button['attr']['title'] . '</span>';
+    $screenReaderMarkup = '';
+    if ($title = valr('attr.title', $button)) {
```
slafleche: It's weird to be setting the value of $title in the if statement. For clarity, I would put it outside.

PR #6561, opened 2018-02-01
Create inline editor flyouts

### LINKED ISSUE #6499: Rich Editor - Inline Formatting Toolbar

Charrondev:
Closes #6499 

In order to accomplish this the following changes were made:

## Changes
- Create custom quill theme that does not use the normal toolbar module.
- Create custom react components (A generic toolbar, a wrapper to keep it floating inline, some icon components, and a toolbar button component)
- Mount the main react component in the editor theme.
- Slight adjustments to the markup / styles to make calculations easier.
- Fixed a doubled contenteditable occuring (quill makes its own but the outer one happened to includes many of quills modules like it's clipboard, which could then be "deleted" with backspace).

## What works
- The toolbar staying within the editor bounds.
- Tabbing into the toolbar.
- Setting bold, italic, and strike styles.

## What does work
- Escape key to clear the toolbar.
- There is some flashing when the the focus and it taken from the editor and given to the toolbar.
- Link and inline code styles.

All of these non functioning things will come in future PRs.


slafleche:
* Add resize listener on Window
* Turn off tabs, we want to tab into the menu

PR #6562, opened 2018-02-01
Add Emoji and Image Blots

slafleche:
I've added custom blots for the editor.


plugins/rich-editor/src/scripts/blots/ImageBlot.js
```diff
@@ -0,0 +1,22 @@
...
+
+    static value(node) {
+        return {
+            alt: node.getAttribute('alt'),
+            url: node.getAttribute('src')
+        };
+    }
+}
+
+ImageBlot.blotName = 'image';
```
Charrondev: We should name this differently from the built in image blot.

plugins/rich-editor/src/scripts/RichEditor.js
```diff
@@ -5,11 +5,26 @@
...
 
+// Blots
+Quill.register(EmojiBlot);
+Quill.register(ImageBlot);
+
+// Theme
 Quill.register("themes/vanilla", VanillaTheme);
 
+// Temporary function
+const makeElement = (tag, attrs) => {
```
Charrondev: Is this actually being used anywhere?

PR #6568, opened 2018-02-02
Fix broken pasting with emoji blot

### LINKED ISSUE #6567: Rich Editor - Pasting when text is selected (and flyout is open) deletes
## Steps to reproduce

- Copy text to clipboard.
- Select text.
- Paste

## Expected
Paste is inserted

## What happens
An empty EmojiBlot is inserted.

Charrondev:
Closes #6567 

---

Fixes the pasting issue by:

- Removing display none from quill's clipboard and using the sr-only mixin. This item has tabindex -1 on it already too so no one should ever be reaching it.
- Adding a className to the emoji blot. Apparently quill can only have a 1 item with same tagName, unless you specify the className

From this article: https://quilljs.com/guides/cloning-medium-with-parchment/#tweets
> Since our root Scroll Blot also uses a `<div>`, we also specify a className to disambiguate. Note Inline blots use <span> and Block Blots use `<p>` by default, so if you would like to use these tags for your custom Blots, you will have to specify a className in addition to a tagName.


plugins/rich-editor/src/scripts/blots/EmojiBlot.js
```diff
@@ -17,5 +17,6 @@ export default class EmojiBlot extends Embed {
     }
 }
 
-EmojiBlot.blotName = 'emoji';
+EmojiBlot.className = "emoji";
+EmojiBlot.blotName = 'emojis';
```
slafleche: Why isn't it the same as the class name? I'd rather keep plural for groups or lists, this is a singular blot.

PR #6571, opened 2018-02-05
Rich editor Emoji menu

slafleche:
Added emoji menu to rich editor.


plugins/rich-editor/src/scripts/components/EditorEmojiPicker.jsx
```diff
@@ -0,0 +1,108 @@
...
+    state;
+
+    /**
+     * @inheritDoc
+     */
+    constructor(props) {
+        super(props);
+
+        if (!this.count) {
+            this.count = 1;
```
Charrondev: This incrementing doesn't seem to actually work. AFAIK you can't access static members through the this property. I think you need a factory pattern to implement this in javascript. Do we even need these ID's? What are they used for?

plugins/rich-editor/src/scripts/components/EditorEmojiMenu.jsx
```diff
@@ -0,0 +1,67 @@
...
+        this.menuTitleID = props.menuTitleID;
+        this.emojiList = emojis;
+        this.closeMenu = props.closeMenu;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    render() {
+        if(this.props.isVisible) {
```
Charrondev: There seems to be a delay when you click the menu because it is essentially re-rendering it from scratch every time (although some elements may be cached). It would be more performant to render immediately once and just toggle display/visibility properties with CSS. In react a partial change is a lot more performant than fully re-rendering 1000+ DOM nodes.

plugins/rich-editor/src/scripts/components/EditorEmojiPicker.jsx
```diff
@@ -0,0 +1,108 @@
...
+    static propTypes = {
+        quill: PropTypes.instanceOf(Quill).isRequired,
+    };
+
+    /** @type {Quill} */
+    quill;
+
+    /**
+     * @type {Object}
+     * @property {RangeStatic} - The current quill selected text range..
```
Charrondev: This state annotation is incorrect for this class.

plugins/rich-editor/src/scripts/components/EditorEmojiPicker.jsx
```diff
@@ -0,0 +1,108 @@
...
+            isVisible: false,
+        };
+
+        this.escFunction = this.escFunction.bind(this);
+
+    }
+
+    escFunction(event){
+        if(event.keyCode === 27) {
+            this.closeMenu();
```
Charrondev: You're not passing the event through which is why e.preventDefault() isn't working in the closeMenu function. Could these 2 functions be merged together?

plugins/rich-editor/src/scripts/components/EditorEmojiMenu.jsx
slafleche: I'll make a separate issue for optimizing the emojis

plugins/rich-editor/src/scripts/components/EditorEmojiPicker.jsx
slafleche: It's for accessibility. Some aria attributes require an ID.

plugins/rich-editor/src/scripts/components/EditorEmojiMenu.jsx
```diff
@@ -0,0 +1,55 @@
...
+        super(props);
+        this.emojiList = emojis;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    render() {
+        if(this.props.isVisible) {
+            return <div id={this.props.menuID} className="richEditor-menu insertEmoji FlyoutMenu insertPopover" role="dialog" aria-labelledby={this.props.menuTitleID}>
```
Charrondev: menuID and menuTitleID aren't in the props validation.

PR #6573, opened 2018-02-05
Add inline link toolbar

Charrondev:
Adds an additional floating toolbar for inline links.

## Changes
- Adds link handling to the InlineEditorToolbar component.
- Pulls FloatingToolbar out into it's own component that worries solely about the positioning inside of the editor.

## To clarify
- What should we do about the focus state of the link input? It looks pretty bad.


slafleche:
* control/comment + k should be a shortcut to changing your selection to a link
* because we can't dismiss the inline toolbar, we're forced to not highjack the focus into the url field. This prevents people from editing the link text and it's also a problem for accessibility
* There are some edge cases to take care of like:
    * What happens if you have multiple links and you do a select all? 
    * What happens if you create a link, then select that link and another word and create another link
The above edge cases can be solved by just never giving access to that menu unless they explicitly ONLY have a link selected.

slafleche:
I've been told we're to follow the same pattern as Medium. So

* Once you've created a link, you can't edit it, you can only delete the "word" that was linked
* Add a title attribute on the link so you can see what it is after it was created
* If you select a link and or other text, the link icon will be active in the menu and clicking it will remove ALL links. 
* Removing a link from part of a word will remove the whole link. Example: your link is: "a link", selecting just "a" and clicking on the link in the menu will remove the link from "a link"

Charrondev:
I've fixed the travis build, and fixed the keyboard shortcut for the flyout.

PR #6581, opened 2018-02-06
Add file uploads to media endpoint

### LINKED ISSUE #6510: Add /media/file endpoint

initvector:
This update adds file uploading and querying to the media endpoint. The spec for this endpoint can be found here: [File Uploads](https://github.com/vanilla/roadmap/blob/master/rich-editor/implementation.md#file-uploads)

Closes #6510


applications/dashboard/controllers/api/MediaController.php
```diff
@@ -0,0 +1,306 @@
...
+            throw new NotFoundException('Media');
+        }
+        return $row;
+    }
+
+    /**
+     * Get a media row by its full URL.
+     *
+     * @param string $url The full media URL.
+     * @throws NotFoundException if the media item could not be found.
```
DaazKu: Missing @throws  Exception

applications/dashboard/controllers/api/MediaController.php
```diff
@@ -0,0 +1,306 @@
...
+                if (is_array($imageSize)) {
+                    $media['ImageWidth'] = $imageSize[0];
+                    $media['ImageHeight'] = $imageSize[1];
+                }
+        }
+
+        $ext = pathinfo(strtolower($upload->getClientFilename()), PATHINFO_EXTENSION);
+        $destination = $this->generateUploadPath($ext, true);
+        $uploadResult = $this->saveUpload($upload, $destination);
+        $media['Path'] = $uploadResult['Name'];
```
DaazKu: I think that the correct assignation would be `$media['Path'] = $uploadResult['SaveName'];` if I base myself on [this](https://github.com/vanilla/vanilla/blob/36faa6b09a6f8c95dee8ed888d9c240cbba0cc2a/plugins/editor/class.editor.plugin.php#L848)

applications/dashboard/controllers/api/MediaController.php
```diff
@@ -0,0 +1,310 @@
...
+        $this->permission('Garden.SignIn.Allow');
+
+        $uploadSchema = new UploadedFileSchema([
+            'allowedExtensions' => array_values(ImageResizer::getTypeExt())
+        ]);
+
+        $in = $this->schema([
+            'file' => $uploadSchema,
+            'type:s' => [
+                'description' => 'The upload type.',
```
DaazKu: It's actually the MIME type

DaazKu:
We should not allow non embed uploads to be deleted by the endpoint until we allow them to be uploaded.
Doing so is a bit confusing and potentially allow weird situations, that have not been tested or found yet, to occur.

initvector:
I am open to revisiting in the future, but my perspective is:

1. The spec states uploads with the endpoint should be limited to a certain type/ID. It does not make mention of restrictions on other actions.
1. The endpoint is for managing resources in the media table. Limiting GET and DELETE actions to only rows that have been uploaded with the endpoint seems unnecessary (and potentially counterintuitive/confusing for users). Limiting GETs to only what has been uploaded with the endpoint makes it blind to existing uploads. If we allow GETs on existing media, but not DELETEs, it seems arbitrary and confusing.

linc:
My thoughts:

1. One decision is mildly confusing to developers, while the other is mildly confusing to users. In that scenario, I'd rather choose the option that inconveniences developers.
1. It seems like restricting the DELETE to foreignType `embed` would technically be a backwards compatibility break.
1.  We intend to iterate on this functionality. Temporarily kneecapping it in the meantime seems an empty gesture.
1. Kludging a restriction seems like a thing that itself would then require more tests, which would then need to be removed later.
1. The level of uncertainty created here seems pretty minimal in relation to the wide-open nature of our addons playing in this area already.

Let's keep this as-is.

initvector:
I addressed some merge conflicts with master that arose after the review, so the original review (an approval) was dropped. I did not manually "dismiss" the previous review.

PR #6591, opened 2018-02-07
Add noopener to all of the things

### LINKED ISSUE #6205: Tabnabbing vulnerability on links with target="_blank"
Hey there,

It would be nice to have automated prevention against tabnabbing vulnerability for forum users' provided links.

To prevent that, it would have to detect links with `target="_blank"` and add `rel="noopener"` and maybe `rel="noreferrer"`.

Here's more info about the vulnerability and how to fix it: https://dev.to/ben/the-targetblank-vulnerability-by-example

Charrondev:
Closes #6205 

Adds a function to Gdn_Format that adds `rel="noopener"` to and links with `target="_blank"`. This processing will only be run on formats containing html.

I initially looked into doing this in htmlawed but that seems unmaintainable and doing this with regex seems like it would be a mess.

@tburry Had proposed doing it client side with a javascript solution, but since I couldn't put together a good whitelist or blacklist of places to run/not run the sanitization I feel that a server side solution is necessary to properly implement this.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -916,6 +916,7 @@ class Gdn_Format {
 
                 // Do HTML filtering before our special changes.
                 $result = $formatter->format($mixed, $options);
+                $result = self::sanitizeLinks($result);
             } else {
                 // The text does not contain HTML and does not have to be purified.
                 // This is an optimization because purifying is very slow and memory intense.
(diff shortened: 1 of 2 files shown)


DaazKu:
We actually want all links (not just user inputted) to be sanitized.
That was my understanding of the issue.

Charrondev:
I've updated this PR to include a javascript solution in addition to the server side solution. If a link needs to allow opener than it can set `data-allow-opener="true"` and it will be passed over.

Charrondev:
@nervoustwit Seems to have a local setup with embeds which may require this `data-allow-opener='true'` so he's agreed to do a little bit of testing on sign-in/register links.

nervoustwit:
Tested this locally, it works just fine.

library/core/class.format.php
```diff
@@ -983,6 +984,32 @@ protected static function replaceSpoilers($html, $replaceWith = '(Spoiler)') {
...
+     *
+     * @return string Returns the html with noopener added to any links with target="_blank".
+     */
+    protected static function sanitizeLinks($html) {
+        if (preg_match('/target="_blank"/i', $html)) {
+            $htmlDom = pQuery::parseStr($html);
+
+            foreach($htmlDom->query('a[target="_blank"]') as $targetBlankLink) {
+                $rel = $targetBlankLink->attr("rel");
+                if ($rel) {
```
DaazKu: You should check that noopener is not already present.

Charrondev:
I've added the check.

PR #6637, opened 2018-02-16
Refactor sort and filter UI for ideation

slafleche:
We want to update the UI for the ideation plugin for a more consistent look and feel across Vanilla.

Closes https://github.com/vanilla/internal/issues/1348

### DIFF
--- a/applications/vanilla/modules/class.discussionssortfiltermodule.php
+++ b/applications/vanilla/modules/class.discussionssortfiltermodule.php
@@ -177,9 +177,6 @@ class DiscussionsSortFilterModule extends Gdn_Module {
                 $dropdown->setTrigger($selected);
             }
 
-            //$dropdown->setView($this->dropdownView);
-            //$dropdown->setForceDivider(true); // Adds dividers between groups in the dropdown.
-
             $lastGroup = '';
             $index = 0;
             $filters = val('filters', $filterSet);
@@ -220,17 +217,12 @@ class DiscussionsSortFilterModule extends Gdn_Module {
                 $sortDropdown[] = [
                     'name' => val('text', $link),
                     'url' => url('/'.val('url', $link)),
-                    'active' => val('key', $filter) === $selectedValue,
+                    'active' => (val('key', $filter) === $selectedValue || ($selectedValue == false && $index == 0)),
                 ];
 
                 $lastGroup = $group;
                 $index++;
-
-
             }
-//            $dropdowns[] = $dropdown;
-
-            $break = "here";
 
             $dropdowns[] = $sortDropdown;
         }
--- a/applications/vanilla/modules/class.discussionssortfiltermodule.php
+++ b/applications/vanilla/modules/class.discussionssortfiltermodule.php
@@ -154,8 +154,7 @@ class DiscussionsSortFilterModule extends Gdn_Module {
             return [''];
         }
 
-        $sortDropdown = [];
-
+        $filterDropdown = [];
 
         $dropdowns = [];
         foreach($this->filters as $filterSet) {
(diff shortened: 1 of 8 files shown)


applications/vanilla/views/modules/discussionssortfilter.php
```diff
@@ -2,19 +2,42 @@
...
         ?>
         </span>
     <?php }
     if ($this->showFilters()) { ?>
         <span class="discussion-filters">
-        <?php foreach ($this->getFilterDropdowns() as $dropdown) {
-            echo $dropdown;
+        <?php
+        foreach ($this->getFilterDropdowns() as $dropdown) {
+            $label = val('name', val(0, $dropdown), "Filter");
```
initvector: Stick with single-quoting strings, unless you need to expand variables, etc (although I believe this was an oversight, since you're single-quoting everywhere else).

applications/vanilla/views/modules/discussionssortfilter.php
```diff
@@ -2,19 +2,42 @@
...
+                if ($isActive) {
+                    $foundActiveLink = true;
+                }
+            }
+
+            if(!$foundActiveLink) {
+                $sortLinks[0]['active'] = true;
+            }
+
+            echo linkDropDown($sortLinks, 'selectBox-ideationFilter', 'Sort');
```
initvector: We typically don't reference external addons in core code. Can the `selectBox-ideationFilter` class be removed?

applications/vanilla/modules/class.discussionssortfiltermodule.php
```diff
@@ -197,8 +203,26 @@ protected function getFilterDropdowns() {
                     [],
                     ['rel' => 'nofollow']
                 );
+
+                $link = $dropdown->getItems()[val('group', $filter)]['items'][val('key', $filter)];
+                $group = val('group', $filter);
+                if($lastGroup != $group && $index != 0 && ($index != $filterLength - 1)) {
```
initvector: Need a space between `if` and the opening parenthesis.

initvector: `($index != $filterLength - 1)` would probably make more sense if it were written as: `$index != ($filterLength - 1)`.  The whole comparison doesn't need to be grouped in parenthesis, but it would be logical to group the subtraction expression.

applications/vanilla/modules/class.discussionssortfiltermodule.php
```diff
@@ -197,8 +203,26 @@ protected function getFilterDropdowns() {
                     [],
                     ['rel' => 'nofollow']
                 );
+
+                $link = $dropdown->getItems()[val('group', $filter)]['items'][val('key', $filter)];
```
initvector: For the sake of readability, I would highly recommend breaking this up a little bit.

1. The result of `$dropdown->getItems` can be moved outside the `foreach` and assigned to a variable. That variable can be referenced in the loop, without needing to invoke the function with every iteration.
1. You're using the value of "group" on the next line. Maybe move the assignment to the top of the loop and use its value here. Similarly, you could assign all the values you'll be using from `$filter` (or other collections) into their own variable near the top of the loop, if you'll be using them as keys or for comparisons.

```php
$link = $items[$group]['items'][$filterKey] // Looks like you're assigning $key to another value. Otherwise, I'd recommend using that.
```

applications/vanilla/modules/class.discussionssortfiltermodule.php
```diff
@@ -128,18 +128,18 @@ protected function getSortData() {
                 }
             }
             $key = val('key', $sort);
-            $queryString = DiscussionModel::getSortFilterQueryString($this->selectedSort, $this->selectedFilters, $key);
+            $queryString = val('key', $sort) !== 'none' ? DiscussionModel::getSortFilterQueryString($this->selectedSort, $this->selectedFilters, $key) : '';
```
initvector: "none" should be replaced with `DiscussionModel::EMPTY_FILTER_KEY` (the constant for "none").

PR #6647, opened 2018-02-20
Create paragraph level formatting toolbar

### LINKED ISSUE #6501: Rich Editor - Paragraph Level Formatting toolbar
Implement the UI of the paragraph level toolbar. This should always be on the beginning of the currently selected line.

It's buttons should have the following options:

- H1
- H2
- code-block
- blockquote

### LINKED ISSUE #6572: Rich Editor - Fire event on "escape"
Each react component in the editor should use this event to close as needed when escape is hit.

Charrondev:
Closes #6501, Closes #6572 

Creates a paragraph level formatting toolbar for the editor. Also ensures that all existing flyouts close each other by firing an event.

I needed to make some updates to the BlockQuote, CodeBlock, and Spoiler blot so that they could interchange with each other. I've made a new generic blot for the content, and a helper function to generate their wrapper Blot.

There are still a few things to wrap up in these blots, but they aren't really relevant to this PR.


Charrondev:
I accidently made the PR to the wrong branch.

PR #6700, opened 2018-02-26
Include .htaccess.dist in phing build

### LINKED ISSUE #6277: Phing not including `.htaccess.dist` in the build
I see it in the include list, but it doesn't actually end up in the built zip file.

initvector:
The Phing build has an exception rule for files ending in "dist". This is primarily meant to keep unwanted files from dependencies or sub-builds from finding their way into the core Vanilla build. However, it has the additional effect of excluding .htaccess.dist from the build.

Phing's pattern matching rules are pretty rudimentary, so this update alters the "dist" exclusion rule to avoid applying it to root-level files. Since .htaccess.dist resides in the root directory of Vanilla, it is once again included in the build. Files ending in "dist" in subdirectories are still excluded.

Closes #6277

### DIFF
--- a/build/package-exclude.txt
+++ b/build/package-exclude.txt
@@ -9,10 +9,10 @@
 **/node_modules/**
 **/npm-shrinkwrap.json
 **/package.json
-**/*.dist
 **/*.scss
 **/*.map
 **/.yo-rc.json
+*/**/*.dist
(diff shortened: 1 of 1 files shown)


linc:
This makes the list of top 5 ironic things to miss a backport label on.

PR #6703, opened 2018-02-26
Update version requirements message in dashboard

### LINKED ISSUE #6365: Revisit required version verbiage in dashboard
To give users more advanced notice, #6270 changed the phrasing of the required version notifications in the for Vanilla 2.5, indicating Vanilla 2.6 will end Vanilla's support for PHP 5.6. The messaging should be altered for 2.6 to convey *current* version requirements.

initvector:
Vanilla 2.5 altered its version requirements warning in the dashboard to let users know PHP 5.6 support would be dropped with Vanilla 2.6. Now that Vanilla 2.6 is almost here, it's time to revert the warning message back to alerting users of the *current* version requirements.

This update alters the version requirements phrasing to state that anything less than PHP 7.0 is not current supported.

Closes #6365

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,14 +48,14 @@ If you professionally run a large community or enterprise forum, our cloud solut
 
 We strongly recommend:
 
-*  **PHP 7.0** or higher.
+*  **PHP 7.1** or higher.
 *  MySQL 5.6 or higher (or Percona/MariaDB equivalent).
 
-If your server is not running PHP 7.0 or higher, **you should address this soon**. While PHP 5.6 will receive security patches thru 2018, Vanilla may end support for it prior to that.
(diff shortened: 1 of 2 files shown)


linc:
This isn't the correct way of doing this.

The message should be updated to encourage use of 7.1 if you use less than 7.1. Then, we should update the bootstrap (environment?) file to block running if it's < 7.0, not allow a broken experience.

linc:
Should also check language in the README.

initvector:
1. [A hard limit on the 7.0 version requirement currently exists in environment.php.](https://github.com/vanilla/vanilla/blob/be7095d10edc4866c4715e72506adabcea2dc1c3/environment.php#L10)
1. I updated the version requirements message to more closely resemble the database version recommendation message.
1. The README has been purged of PHP 5.6 references. It has been updated to encourage users to 7.1, warning that PHP 7.0 reaches EOL in December 2018.

PR #6707, opened 2018-02-27
Rich Editor - Emoji picker optimization

slafleche:
I've done some optimizations for the emoji picker. It's a lot faster now!

We're now using react virtualized and we're using a class to show/hide the component, instead of recreating it each time.

Note there's an exception for the united nation flag in the emoji picker. For some reason that one emoji is the only one with is much bigger than the rest!


plugins/rich-editor/src/scripts/components/EditorEmojiMenu.jsx
```diff
@@ -25,33 +31,66 @@ export default class EditorEmojiMenu extends React.Component {
...
+                <a href="#" aria-label={t('Close')} onClick={this.props.closeMenu} className="Close richEditor-close">
+                    <span>×</span>
+                </a>
+            </div>
+            <div className="insertPopover-body">
+                <AutoSizer>
+                    {({ height, width }) => (
+                        <Grid
+                            containerRole = ''
+                            // containerStyle={{display: 'flex', flexWrap: 'wrap', alignItems: 'flex-start', justifyContent: 'flex-start'}}
```
Charrondev: You should remove these commented lines.

PR #6724, opened 2018-02-28
Make /categories/archive and /discussions/tagged noindex, follow

### LINKED ISSUE #6490: No indexing for /tagged pages
We've had a request to get pages under the `/tagged` uri to add the following meta tag:
```
<meta name="robots" content="noindex, follow">
```

tburry:
The category archives were already noindex, but I’m adding follow just
to be explicit.

Closes #6490.


tburry:
Upon looking at some webmaster tools reports I've noticed that the tagged pages actually are getting search traffic on some communities so I think making them noindex in core is a bad decision. Since the other change is non-functional I'm just going to close this PR.

Issue #6632, opened 2018-02-15
Rich Editor - Spoiler Blot


slafleche:
The spoiler blot is more complex than the other types of embeds we have. We want the container to be uneditable, but the contents to be editable.

- [x] Fix blot to allow more complex html structure
- [x] Add button on top of spoiler to goggle it
- [ ] Write JS to toggle spoiler
- [ ] Create unique ID needed for `aria-controls` and `aria-controlledby` attributes (ties to https://github.com/vanilla/vanilla/issues/6586)

Charrondev:
The requirements here are actually being discussed. After speaking with @tburry the spoiler will not be collapsable in the editor.

Proposal here:

https://github.com/vanilla/roadmap/pull/34/files

Charrondev:
Closed by #6725

PR #6725, opened 2018-02-28
Improve handling of multi-line block elements

### LINKED ISSUE #6582: Rich Editor - Can't get out of code block
There's some improvements to be made for the codeblock styles in the editor

- [ ] Don't allow links or any other styles inside of code blocks
- [ ] Selecting any part of the code block and then clicking on the code block in the menu gets rid of the whole code block.
- [ ] If you select regular text AND a codeblock, styles will only affect the non code block elements
- [ ] If only the code block is selected, the only styling option you have is to remove the code block

- [ ] Allow to "arrow out" of the block (i.e. be able to arrow your way out of the block to continue typing next to it
- [ ] Allow to clicking next to the block to exit out of it
- [ ] Don't capture tabs (we might add an option to support tabs in the future)
- [ ] Proper space between inline code block and other letters. (no left space if it's the first element in a paragraph, no right space if it's the last element, but we want proper space between other words

Charrondev:
Closes #6582 and #6632

## Creates proper nested blot structure

Provides an implementation of some nested Container blots. The key here was to work outwards from a line, instead of working downwards from the container. The `ContainerBlot`, `ContentBlot`, and `LineBlot` are used for the Quotes and Spoilers.

## Add custom keyboard handlers for escaping blocks

The handlers have been implemented according to the spec changes here:

https://github.com/vanilla/roadmap/pull/34/files#diff-55684cd001d40f12a0224f5924df3429R142

We can wait until that PR is merged first.

## Adds classnames to the embed blots

The embed blots with a div tagname and no classname were clashing with the container blots, so I had to add classes. I made the names a little more consistent and updated the tests to match.

## What this doesn't do

- Alter the paragraph item behaviour.
- Add server rendering for the CodeBlock or QuoteBlock or Spoiler.


Charrondev:
I've filed this issue to address that bug https://github.com/vanilla/vanilla/issues/6727

Charrondev:
Another issue I've noticed:

- The down/right arrow doesn't work when you're at the end of a line. I'm currently checking that you have to beginning for all of the arrow functions, whereas that should only apply for the top ones. The down ones should require you be at the end of the line.

Charrondev:
As far as skipping Code Blots with arrow I've filed this bug:

https://github.com/quilljs/quill/issues/1988

If there is no movement on it in the next couple of weeks, we can implement the fix in our custom code blot, and I can make a PR to the main quill repo.

I've filed this issue for the last you bug you pointed out.

PR #6748, opened 2018-03-02
Theme compatibility for "selectBox"

slafleche:
Here are some tweaks for a new component we'll be using.

I feel like this the best we can do without switching over to more modern positioning with flexbox. I tried hard to reuse existing classes as much as possible and get as much compatibility as we can. I only had to tweak one theme: https://github.com/vanilla/internal/pull/1459

Tied to: 
https://github.com/vanilla/internal/issues/1422
https://github.com/vanilla/internal/issues/1432


slafleche:
Wrong brach, I'll create a new PR.

slafleche:
See https://github.com/vanilla/vanilla/pull/6750

PR #6750, opened 2018-03-05
Theme compatibility for "selectBox"

slafleche:
Here are some tweaks for a new component we'll be using.

I tried hard to reuse existing classes as much as possible and get as much compatibility as we can for "free". There's another PR for compatibility with other core themes: https://github.com/vanilla/internal/pull/1459

Note that custom client themes will be done separately.

Tied to: 
https://github.com/vanilla/internal/issues/1422
https://github.com/vanilla/internal/issues/1432


applications/dashboard/scss/style-compat.scss
```diff
@@ -1,4 +1,6 @@
 // Compatibility
+@import "compatibility/_mixins.scss";
```
Charrondev: This should be named similarly to the other imports.

slafleche: What do you mean? "compatibility_mixins"?

Issue #6673, opened 2018-02-22
Add user content class onto all user content


Charrondev:
The class should be `userContent`. It should be applied to:

- [x] Discussions
- [x] Comments
- [x] Signatures
- [x] Private Messages
- [x] Warnings/Notes
- [x] Activity

PR #6758, opened 2018-03-06
Add userContent class

Charrondev:
Reference vanilla/vanilla#6673
Related vanilla/internal#1464
Related vanilla/addons#599

Add's user content class for non-plaintext content created by users. This class is used for some styles related the rich editor.

### DIFF
--- a/applications/dashboard/views/activity/helper_functions.php
+++ b/applications/dashboard/views/activity/helper_functions.php
@@ -151,7 +151,7 @@ if (!function_exists('WriteActivityComment')):
             <?php } ?>
             <div class="ItemContent ActivityComment">
                 <?php echo userAnchor($author, 'Title Name'); ?>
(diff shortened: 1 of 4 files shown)


applications/dashboard/views/activity/helper_functions.php
```diff
@@ -151,7 +151,7 @@ function writeActivityComment($comment, $activity) {
             <?php } ?>
             <div class="ItemContent ActivityComment">
                 <?php echo userAnchor($author, 'Title Name'); ?>
-                <div class="Excerpt"><?php echo Gdn_Format::to($comment['Body'], $comment['Format']); ?></div>
+                <div class="Excerpt userContent"><?php echo Gdn_Format::to($comment['Body'], $comment['Format']); ?></div>
```
slafleche: Can the excerpt be something other than plain text? It's not necessary to add `userContent` if it's plain text.

PR #6765, opened 2018-03-06
Add new addon type

Caylus:
In new addon.json files no distinction between plugins/themes/applications is made anymore, everything only has "type":"addon" defined.

This causes trouble in updatemodel::buildAddon when the Addon application tries to parse an addon.json file of a new plugin.
To be able to fix that, a new addon type constant is necessary.

### DIFF
--- a/conf/constants.php
+++ b/conf/constants.php
@@ -80,6 +80,7 @@ define('ADDON_TYPE_PLUGIN', 1);
 define('ADDON_TYPE_THEME', 2);
 define('ADDON_TYPE_LOCALE', 4);
 define('ADDON_TYPE_APPLICATION', 5);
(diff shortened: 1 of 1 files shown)


charrondev:
From @linc https://open.vanillaforums.com/discussion/36024/addon-directory-fixes

>That's basically accurate, yes. What's really happening is we've stopped delineating between 'application' vs. 'plugin' and are just calling them 'addons'. There used to be real differences between them, but the priority order is basically the only difference left at this point and we hope to remove the concept of 'applications' in the (distant) future. We recommend all new addons just use the 'plugins' folder and use those standards from now on.
>
> In this new worldview, 'theme' and 'locale' are basically special sub-types of addons.

PR #6779, opened 2018-03-08
Include autoloader using PATH_ROOT

jfarthing84:
Use PATH_ROOT instead of __DIR__ to include the composer autoloader.

### DIFF
--- a/environment.php
+++ b/environment.php
@@ -63,6 +63,6 @@ if (function_exists('mb_internal_encoding')) {
 }
 
 // Include the core autoloader.
-if (!include_once __DIR__.'/vendor/autoload.php') {
-    die("Could not find the autoloader. Did you forget to run 'composer install' in '".__DIR__."' ?\n");
+if (!include_once PATH_ROOT.'/vendor/autoload.php') {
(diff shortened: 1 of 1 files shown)


DaazKu:
Is there any reason in particular that made you do that change?

jfarthing84:
It is required to run Vanilla from a subdirectory, otherwise, one must copy `environment.php` along with `index.php` into the top-level directory.

charrondev:
I'm still not understanding the workflow here, is this for the case where you're using Vanilla as a composer dependency?

charrondev:
@jfarthing84 @chrisvanpatten Do either of you have the justifying use case for this change?

jfarthing84:
Other than the fact that PATH_ROOT is used for every other include in that file?

> It is required to run Vanilla from a subdirectory, otherwise, one must copy environment.php (along with index.php) into the top-level directory just to change that path.

charrondev:
The closest relavent thing I could find around this is this comment: https://github.com/vanilla/vanilla/pull/5674#issuecomment-309058410

@linc Where does this facilitate/break any of these use cases?

Issue #5867, opened 2017-08-17
Vanilla.Comments.AllowSelfDelete not allowing users to delete own comments


initvector:
The Vanilla.Comments.AllowSelfDelete config should allow users to delete their own comments, so long as the comment is still able to be edited. With this config enabled, a normal site user should be able to make a request to [`DiscussionController::deleteComment`](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/vanilla/controllers/class.discussioncontroller.php#L720) and have their comment deleted.

37f99dd tweaked the logic of determining if the comment can be edited. However, the ability to delete a comment is now based on [if the discussion can be edited](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/vanilla/controllers/class.discussioncontroller.php#L741). The check should be changed to verify the comment is still editable.

*Side note: Adding and editing comments is done in `PostController`. Maybe deleting comments should be moved there too.*

charrondev:
Digging in a little bit, the key piece here is that when the permission check for deleting a comment is done, it [checks the discussion instead](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/vanilla/controllers/class.discussioncontroller.php#L720).

Properly resolving the issue would look like the following:

- Move the `DiscussionModel`'s [editContentTimeout](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.discussionmodel.php#L252-L263) method up to `Gdn_Model`. The parameter names should be made more generic and the description updated.
- Update `DiscussionModel->canEdit` to use `Gdn_Model->editContentTimeout`  and remove `DiscussionModel->editContentTimeout`.
- Create a `canEdit` function on the `CommentModel` similar to [the one on the DiscussionModel](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.discussionmodel.php#L229-L243)
- Use this check in place of the [discussion check when deleting a comment](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/vanilla/controllers/class.discussioncontroller.php#L720).
- Use this same check in place the freeform checks in the `helper_functions` [getCommentOptions](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/views/discussion/helper_functions.php#L427-L479).

[...]

PR #6782, opened 2018-03-11
Respect Vanilla.Comment.AllowSelfDelete setting.

R-J:
This fixes https://github.com/vanilla/vanilla/issues/5867

It does not take any time outs into consideration and I haven't moved it to the postcontroller. Just a simple fix.

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -741,8 +741,7 @@ class DiscussionController extends VanillaController {
                 }
 
                 // Make sure that content can (still) be edited.
-                $canEdit = DiscussionModel::canEdit($discussion);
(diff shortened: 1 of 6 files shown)


charrondev:
Thanks for your contribution!

While this seems to work, it does so by short-circuiting a timeout check entirely. As a result it would allow a user (with a crafted request) to delete any of there content from any time.

This config setting is already checked a few lines before your change and actually doesn't need to be checked again. I've dug into the issue a bit and outlined what an expected fix would like in the original issue. https://github.com/vanilla/vanilla/issues/5867#issuecomment-374262583

We would gladly accept a pull request resolving the issue in its full scope, but are not currently looking to accept a small partial fix for the issue. The parent of that issue mentions moving the location of the delete method of a comment, but I think that does not need to be part of the scope of the fix.

PR #6797, opened 2018-03-14
Change hardcoded cache folder to constant value

R-J:
ComposerHelper::clearAddonManagerCache used `realpath(__DIR__.'/../../cache')` instead of the constants.php value `PATH_CACHE`

### DIFF
--- a/library/setup/ComposerHelper.php
+++ b/library/setup/ComposerHelper.php
@@ -18,13 +18,11 @@ class ComposerHelper {
      * Clear the addon manager cache.
      */
     private static function clearAddonManagerCache() {
-        $cacheDir = realpath(__DIR__.'/../../cache');
-
         $paths = array_merge(
-            [$cacheDir.'/addon.php'],
-            glob($cacheDir.'/locale/*.php'),
-            glob($cacheDir.'/theme/*.php'),
-            glob($cacheDir.'/*-index.php')
+            [PATH_CACHE.'/addon.php'],
+            glob(PATH_CACHE.'/locale/*.php'),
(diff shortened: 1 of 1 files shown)


Charrondev:
Related https://github.com/vanilla/vanilla/issues/6798

It might take a bit more to get this PR through. It seems to be failing our CI, becomes PATH_CACHE isn't defined in that environment.

initvector:
`ComposerHelper` operates outside Vanilla, under Composer. Because of this, it does not have access to Vanilla's constants.

https://github.com/vanilla/vanilla/blob/e7f487b1952198076e9d34f1801471f756a5906b/composer.json#L74-L75

As stated, `PATH_CACHE` lives in constants.php, within the config folder.

https://github.com/vanilla/vanilla/blob/e7f487b1952198076e9d34f1801471f756a5906b/conf/constants.php#L11-L14

However, that constant also relies on `PATH_ROOT`, which is defined in environment.php.

https://github.com/vanilla/vanilla/blob/e7f487b1952198076e9d34f1801471f756a5906b/environment.php#L24-L26

Replacing the fixed relative cache path with `PATH_CACHE` will break Composer installs and updates, not just on Travis but in most environments where it is run, because that constant isn't available. Additional refactoring to make constants more accessible will be required.

```ShellSession
composer install --optimize-autoloader
> Vanilla\Setup\ComposerHelper::preUpdate
Script Vanilla\Setup\ComposerHelper::preUpdate handling the pre-install-cmd event terminated with an exception

                                                               
  [ErrorException]                                             
  Use of undefined constant PATH_CACHE - assumed 'PATH_CACHE'  
```

PR #6817, opened 2018-03-16
Fix mention triggering in the editor

### LINKED ISSUE #6165: Mentions autocomplete box doesn't dismiss when you click away from it
If you are composing a post with an @mention, if you revisit that mention, then you tap away to type somewhere else, the autocomplete box follows you.

editorissue

mention-moving

### LINKED ISSUE #6783: @mention autocomplete cant search for names with accented characters
image

Adding an accented character to an @mention stops the lookup process, making it difficult to use on sites if they have a large userbase, and the accented character appears early in the name. In this example, after the Ä is included in the name, the autocomplete box goes away. 

https://vanillaforums.teamwork.com/#tasks/8317326

charrondev:
Closes #6783
Closes #6165

Fixes the broken @mention behaviour.

## Matching unicode characters

Unicode characters can now be matched, just like punctuation. Like punctuation they must be quoted. Suggestions will be offered containing unicode characters if the first few characters are valid even without the quotes (this is just the current behaviour).

## Moving mention dropdown

This was fixed both through a updated regex that now only looks at the current line of text. In order to fix the current line I had to fix the content-editable line detection that was broken. The issue was that a native `Range` when converted to a string, does not contain line breaks. At.who was trying to compensate by counting line breaks elements, but was counting line breaks in the entire editor, not up to the current cursor position. I got rid of the offset and now account for the line breaks in the `getPos` function by walking the editor's DOM tree backwards from the ending Node.

## Tests

JS tests have been added for the regex as well in legacy.test.js. Until we move the stuff in global.js into our build system, I can't fully test the function or use the same exact regex instance in the test and in the function itself.

I've manually tested this extensively as well, using our different editor types and browser stack.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -1900,17 +1900,32 @@ jQuery(document).ready(function($) {
                     // string as well as spaces. Spaces make things more
                     // complicated.
                     matcher: function(flag, subtext, should_start_with_space) {
-                        var regexStr = '(@(?=(")?)(\\w+|\\1(?:[^\\u0000-\\u001f\\u007f-\\u009f\\u2028]+?)\\1))(?:\\s|$)';
+
+                        // Split the string at the lines to allow for a simpler regex.
+                        var lines = subtext.split("\n");
+                        var lastLine = lines[lines.length - 1];
+
+                        var regexStr =
+                            '@' + // @ Symbol triggers the match
+                            '(?:(\\w+)' + // Any ASCII based letter characters
+                            '|' + // Or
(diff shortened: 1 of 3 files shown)


charrondev:
I've updated this to allow matching of unicode characters without needing quotes as you type. Spaces still require quotes though.

slafleche:
Another bug:
1. type 3 lines of gibberish
2. go back to first line, add space then @mention. No menu shows up.

Issue #6046, opened 2017-10-17
Fork Swagger to OpenAPI


tburry:
We want to move the swagger plugin to core, but not as a plugin.

- Move the code to the dashboard application.
- The endpoint should be /api/v2/open-api/v2 and /api/v2/open-api/v2/summary rather than /api/v2/swagger and /api/v2/swagger/summary

The model can still be the SwaggerModel for the v2 endpoint.

DaazKu:
From a support perspective I think that the menu item should probably stay named API v2.
Pretty much everything that refers to it has v2 in the name (public doc, the url, etc..)
We could confirm with sales and supports.

DaazKu:
Exclude authenticate and authenticator docs

PR #6823, opened 2018-03-19
Move the swagger plugin to core as swagger-ui

tburry:
Swagger will now be a core part of Vanilla. This move has a few changes:

- The API portion of the plugin is now core.
- The javascript is now included in the view rather than a separate
event.
- Some CSS tweaks that were removed earlier were added back in.
- The addon is enabled by default and hidden now.


tburry:
Closes #6046.

charrondev:
It seems this doesn't have the fixes from https://github.com/vanilla/internal/pull/1488. I discovered a potential security issue there and resolved it.

tburry:
@charrondev I copied your changes in.

charrondev:
My only question here is why are we including the JS in a view instead of just adding a deferred JS file through the controller? I see why we don't want to use and event here, but just adding it to head deferred would accomplish the same thing.

Issue #6770, opened 2018-03-06
Rich Editor - Check file upload permission client side


charrondev:
Check to see if file upload is allowed or not and hide UI accordingly.

slafleche:
PR: https://github.com/vanilla/vanilla/pull/6827

PR #6827, opened 2018-03-20
Rich Editor - check upload permission

slafleche:
Check if user is allowed to upload images in the rich editor.

Closes: https://github.com/vanilla/vanilla/issues/6770


plugins/rich-editor/class.rich-editor.plugin.php
```diff
@@ -54,7 +54,7 @@ public function gdn_form_beforeBodyBox_handler($sender, $args) {
         $controller->setData('editorData', [
             'editorID' => $editorID,
             'editorDescriptionID' => 'richEditor-'.$editorID.'-description',
-
+            'hasUploadPermission' => CheckPermission('uploads.add'),
```
charrondev: `checkPermission` should be lowercased.

DaazKu: *camelCased FTFY :P

PR #6834, opened 2018-03-21
Add the core app js if it exists

### LINKED ISSUE #6807: Core javascript app bundle is not loaded
It seems the the PHP that inserts the client scripts does not properly insert the core app bundle. It stops at the lib-core bundle and the core-bootstrap bundle. As a result this javascript only seems to load when running the hot reload/local dev server.

tburry:
Since the core js isn’t a true addon the check needs to be done
separately. This fix doesn’t currently have an impact since there is no
core app, but it should when code needs to add core app functionality.

Closes #6807.

### DIFF
--- a/applications/dashboard/models/class.assetmodel.php
+++ b/applications/dashboard/models/class.assetmodel.php
@@ -168,8 +168,14 @@ class AssetModel extends Gdn_Model {
             "/js/$basename/lib-core-$basename.js"
         ];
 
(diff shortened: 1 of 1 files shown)


tburry:
@charrondev can you cherry pick or merge this into your branch?

charrondev:
Yes. I merge master pretty frequently into feature/rich-editor-mvp.

PR #6835, opened 2018-03-21
Add filtering by archived status to categories API v2 endpoint

### LINKED ISSUE #6255: Add an archived parameter to /api/v2/categories
The parameter should be bool|null and default to false. Here is a summary of behavior:

| Archived  | Behavior |
| --------- | -------- |
| **false** | Exclude archived categories. |
| **true**  | Include only archived categories. |
| **null**  | Don't filter on archived. |

This filter only affects categories returned from the endpoint. If a category is specified in the **parentCategoryID** parameter it doesn't matter whether or not it's archived.

initvector:
This update adds the ability to filter results from the index of the categories endpoint by their archived status. Per the original issue:

1. If `archived` is false (default), only return non-archived categories from the current tree.
1. If `archived` is true, only return archived categories from the current tree.
1. If `archived` is null (an empty value), don't filter by the archived status. Return both archived and non-archived categories.

Closes #6255

### DIFF
--- a/applications/vanilla/controllers/api/CategoriesApiController.php
+++ b/applications/vanilla/controllers/api/CategoriesApiController.php
@@ -324,7 +324,7 @@ class CategoriesApiController extends AbstractApiController {
                 $limit
             );
 
-            $totalCountCallBack = function () use ($parent) {
+            $totalCountCallBack = function() use ($parent) {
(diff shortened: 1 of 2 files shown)


DaazKu:
Following our talk about boolean fields we should probably rename this to: isArchived

PR #6837, opened 2018-03-22
Cleanup where category condition.

prasad83:
When there is not category to filter - drop the where condition.

This was happening when (type=guest) was not set a default to any role and site was visited without any signin.


linc:
A week has passed without the CLA being signed, so I am closing this request for now.

PR #6840, opened 2018-03-22
Phase out category muting in favor of category following

initvector:
The UI for muting a category was phased out for category following with #6469. However, there still exists some functionality that needs to be walled off.

1. `CategoryController::follow` has been deprecated. This was formerly used to mute/unmute a category. Since we're shifting to following, this should no longer be used.
1. The recent discussions list no longer filters blacklisted (muted) categories.
1. Similar to the recent discussions page, `DiscussionsModule` no longer filters blacklisted (muted) categories.
1. `CategoryModel::categoryWatch` has been deprecated.
1. `CategoryModel::getVisibleCategories` has been added. The method gets categories a user has "view" access too, while allowing additional filtering (e.g. filtering by the `HideAllDiscussions` field).
1. `CategoryModel::getVisibleCategoryIDs` has been introduced as a drop-in replacement for `CategoryModel::categoryWatch`. It is essentially a wrapper for `CategoryModel::getVisibleCategories` that extracts category IDs. The event fired in `CategoryModel::categoryWatch`is now fired in this new method as a deprecated event. A new filter is fired to alter the return values of the method.
1. `CategoryModel::$Watching` has been deprecated.
1. `DiscussionModel::$Watching` has been deprecated.

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -397,7 +397,7 @@ class CategoryModel extends Gdn_Model {
     }
 
     /**
-     *
+     * Get followed categories, taking into account view permissions and the category's HideAllDiscussions flag, optionally.
      *
      * @param bool $honorHideAllDiscussion Whether or not the HideAllDiscussions flag will be checked on categories.
      * @return array|bool Category IDs or true if all categories are watched.
@@ -413,7 +413,7 @@ class CategoryModel extends Gdn_Model {
                 continue;
             }
 
-            if ($category['PermsDiscussionsView'] && $category['Following']) {
+            if ($category['PermsDiscussionsView'] && $category['Followed']) {
                 $watch[] = $categoryID;
             }
         }
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -443,6 +443,7 @@ class CategoryModel extends Gdn_Model {
      * Get a list of IDs of categories visible to the current user.
      *
      * @param array $options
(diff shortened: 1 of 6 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -428,6 +439,66 @@ public static function categoryWatch($honorHideAllDiscussion = true) {
...
+        $result = $this->eventManager->fireFilter('categoryModel_visibleCategories', $result);
+
+        return $result;
+    }
+
+    /**
+     * Get a list of IDs of categories visible to the current user.
+     *
+     * @see CategoryModel::categoryWatch
+     * @param array $options
```
DaazKu: A listing of the options or where to find them (probably better in this case) would be neat.

applications/vanilla/models/class.categorymodel.php
```diff
@@ -428,6 +439,66 @@ public static function categoryWatch($honorHideAllDiscussion = true) {
...
+     * @return array|bool An array of filtered category IDs or true if no categories were filtered.
+     */
+    public static function getVisibleCategoryIDs(array $options = []) {
```
DaazKu: We should probably make that a non static method and force the caller to use `CategoryModel::instance()-> getVisibleCategoryIDs()`

DaazKu:
With this update I'm now seeing the Social Groups category in the recent discussions list. I was not on master.

I also muted a category and tested on both master and this PR and muting already didn't do anything to the recent discussion list. At least with my current setup.

DaazKu:
Sounds like it was a memcached problem.

PR #6855, opened 2018-03-26
Refactor CommentModel->setWatch

R-J:
That method has a lot of nested if-clauses. I replaced them by `if (!condition) { return; }`

The other "bigger" change in this PR is that the following while block:
~~~
$markAsRead = true;
while ($discussion = $discussions->nextRow(DATASET_TYPE_ARRAY)) {
    if ($discussion['Read']) {
        continue;
    }
    $markAsRead = false;
    break;
}
~~~
with array functions:
~~~
if (in_array(false, array_column($discussions->resultArray(), 'Read'), true) {
    return;
}
~~~

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -752,14 +752,11 @@ class CommentModel extends Gdn_Model {
             return;
         }
 
-        // Loop over these discussions and see if any are still unread.
-        $markAsRead = true;
+        // Loop over these discussions and exit if there are any unread discussions
         while ($discussion = $discussions->nextRow(DATASET_TYPE_ARRAY)) {
-            if ($discussion['Read']) {
-                continue;
+            if (!$discussion['Read']) {
+                return;
             }
-            $markAsRead = false;
-            break;
         }
 
         // Mark this category read if all the new content is read.
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -752,9 +752,14 @@ class CommentModel extends Gdn_Model {
             return;
         }
 
-        // Search through all of these and exit if any are still unread.
-        if (in_array(false, array_column($discussions->resultArray(), 'Read'), true)) {
-            return;
+        // Loop over these discussions and see if any are still unread.
+        $markAsRead = true;
+        while ($discussion = $discussions->nextRow(DATASET_TYPE_ARRAY)) {
+            if ($discussion['Read']) {
+                continue;
+            }
+            $markAsRead = false;
+            break;
         }
 
         // Mark this category read if all the new content is read.
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -733,7 +733,7 @@ class CommentModel extends Gdn_Model {
 
                         // Find all discussions with content from after DateMarkedRead
                         $discussionModel = new DiscussionModel();
-                        $discussions = $discussionModel->get(0, 101, [
+                        $discussions = $discussionModel->get(0, $lookBackCount, [
                             'CategoryID' => $categoryID,
                             'DateLastComment>' => $dateMarkedRead
                         ]);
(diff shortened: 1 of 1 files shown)


charrondev:
I love seeing pull requests from the open source community, and we'll try to get to this as quickly as possible, but even if its a relatively simple thing, I would recommend you open issues to close with a PR in the future. This way we can outline what the success criteria are for a pull request before we come to review it.

@initvector Do you thing you could take a look at this next sprint?

R-J:
This PR does not fix any issue. It basically changes the current structure from:
~~~
if (condition) {
    ...
    if (condition) {
        ...
        if (condition) {
            ...
            if (condition) {
                ...
                if (condition) {
                     ...
                }
            }
        }
    }
}
~~~

to

~~~
if (!condition) {
    return;
}
....
if (!condition) {
    return;
}
....
if (!condition) {
    return;
}
....
if (!condition) {
    return;
}
~~~

I was looking at the referenced issue and started writing a solution when I found myself getting irritated by the number of nested `if`'s. Therefore I did this restructuring.

I would happily do a PR for the issue https://github.com/vanilla/vanilla/issues/6856 which would be easier to check than this PR and would solve a real issue. That might have been the better approach... Please don't waste any resources on this by now. I'll write a PR for the issue and afterwards work on this PR again!

R-J:
PR https://github.com/vanilla/vanilla/pull/6859 is there B)

charrondev:
Ah I see what you mean. Having cleanup work placed into a separate PR is definitely a good idea and can make reviews easier if the change afterwards isn't going to rewrite everything again.

Refactoring PRs should still be tied to an issue though. If the refactoring is large enough we may want to write unit tests before hand, to verify that nothing breaks in the process.

If the refactoring is to make an additional PR that fixes an issue easier, linking that issue and mentioning the purpose does the job as well.

applications/vanilla/models/class.commentmodel.php
```diff
@@ -654,118 +654,113 @@ public function getUserScore($commentID, $userID) {
      * @param int $totalComments Total in entire discussion (hard limit).
      */
     public function setWatch($discussion, $limit, $offset, $totalComments) {
+        $userID = Gdn::session()->UserID;
+        if (!$userID) {
```
charrondev: I had a nagging suspicion that this wouldn't work out (because we tend to use the `$id > 0` checks all over the code base, but since out recent authenticator rewrite it seems pretty clear that we correct handle the weird case where a userID might be < 0 in our `Session->start()`).

applications/vanilla/models/class.commentmodel.php
```diff
@@ -654,118 +654,113 @@ public function getUserScore($commentID, $userID) {
...
+        unset($discussionModel);
+
+        // Abort if we get back as many as we asked for, meaning a
+        // lot has happened.
+        if ($discussions->numRows() > $lookBackCount) {
+            return;
+        }
+
+        // Search through all of these and exit if any are still unread.
+        if (in_array(false, array_column($discussions->resultArray(), 'Read'), true)) {
```
charrondev: I'd like to see the `markAsRead` check still spread out over multiple lines the way it was previously. While this is certainly shorter, it much less legible to me.

The previous check
```php
                            // Loop over these and see if any are still unread
                            $markAsRead = true;
                            while ($discussion = $discussions->nextRow(DATASET_TYPE_ARRAY)) {
                                if ($discussion['Read']) {
                                    continue;
                                }
                                $markAsRead = false;
                                break;
                            }
```

applications/vanilla/models/class.commentmodel.php
```diff
@@ -654,118 +654,118 @@ public function getUserScore($commentID, $userID) {
...
+            return;
+        }
+
+        // Loop over these discussions and see if any are still unread.
+        $markAsRead = true;
+        while ($discussion = $discussions->nextRow(DATASET_TYPE_ARRAY)) {
+            if ($discussion['Read']) {
+                continue;
+            }
+            $markAsRead = false;
```
charrondev: I think we can still shorten this anyways. As is you aren't actually checking for the value of `$markAsRead` anymore. I'm going try a different type of loop here.


```php
// Bail out if there are any unread discussions
while ($discussion = $discussions->nextRow(DATASET_TYPE_ARRAY)) {
    if (!$discussion['Read']) {
        return
    }
}
```

Issue #6856, opened 2018-03-26
Categories cannot been marked Read automatically


R-J:
The CommentModel->setWatch method is used for determining if a category is shown as "Read" or not. That method contains the following lines:
~~~
$dateMarkedRead = val('DateMarkedRead', $category);
if ($dateMarkedRead) {
~~~
As a result only categories which have been marked as read manually can be shown as read.

I think it should be handled like that:
~~~
$wheres = ['CategoryID' => $categoryID];
if (val('DateMarkedRead', $category)) {
    $wheres['DateLastComment>'] = $dateMarkedRead;
}
$discussions = $discussionModel->get(0, $lookBackCount + 1, $wheres);
~~~

I will do a PR if https://github.com/vanilla/vanilla/pull/6855 has been approved or declined.

charrondev:
- Related https://github.com/vanilla/vanilla/issues/4013
- Related https://github.com/vanilla/vanilla/issues/2357

theknigge:
@charrondev do you have an idea when you can fix that?

charrondev:
@R-J had taken a pretty good swing at it with https://github.com/vanilla/vanilla/pull/6859 but seems to have made some mistake in how he branched the 2 PRs. I tried to reconcile his PR but ultimately could not fix the conflicts in a reasonable amount of time.

Personally it's not on our immediate roadmap, although I understand the annoyance. We are accepting contributions though. @R-J's would have been merged (and mostly fixed the issue) if the conflicts could have been resolved.

charrondev:
Fixed by https://github.com/vanilla/vanilla/pull/8574

PR #6859, opened 2018-03-26
Let All Categories Show "Read" Information Correctly

R-J:
Fix https://github.com/vanilla/vanilla/issues/6856

Check was only done when DateMarkedRead was set manually before. This would lead to false negatives for every category which was never been marked read.
Also changed the LookBackCount being used as the limit for the get operation. The hard coded value leads to false results for forums with more than 50 discussion per page.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -716,7 +716,7 @@ class CommentModel extends Gdn_Model {
             }
 
             /**
-             * Fuzzy way of trying to automatically mark a cateogyr read again
+             * Fuzzy way of trying to automatically mark a category read again
              * if the user reads all the comments on the first few pages.
              */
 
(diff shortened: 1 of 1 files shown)


charrondev:
@R-J Are you going to fix the conflicts on this one soon?

charrondev:
Have you tried rebasing instead of merging? That might go more smoothly.

charrondev:
I'm sorry @R-J I've really tried to merge/rebase this for you, but it's just not happening. https://github.com/vanilla/vanilla/pull/6855 pretty much 100% changed this function and this was not based on that. This PR will need to be tried again from master.

Recreating an ~10 line change (excluding whitespace) will be a lot easier than trying to merge these  together.

R-J:
@charrondev At the end it has even been just 5 lines :-D

PR #6862, opened 2018-03-26
Add pagination link tags to page when using pager

### LINKED ISSUE #6522: Pagination with rel=“next” and rel=“prev”
According to this Google blog post, we should be adding links with "next" and "prev" links in the header when we have pagination.

https://webmasters.googleblog.com/2011/09/pagination-with-relnext-and-relprev.html

initvector:
`PagerModule` has been updated to add rel=prev/next link tags to the page head, as necessary, when the pager module is used.

1. `PagerModule::addRelLinks`, a private method, has been added. This method is responsible for updating the page head and setting the link response header. It is currently only called in `PagerModule::configure`.
1. Adding the link tags should only occur once. It doesn't matter if you write the pager once or ten times.
1. The first page of the pager has been normalized to have no page number. This is done for search engines, to avoid seemingly duplicating content (e.g. /discussions and /discussions/p1 being identical).

Closes #6522

### DIFF
--- a/applications/dashboard/modules/class.pagermodule.php
+++ b/applications/dashboard/modules/class.pagermodule.php
@@ -112,11 +112,11 @@ class PagerModule extends Gdn_Module {
     }
 
     /**
-     * Add prev/next rel meta pagination tags to the document head. Also, add Link header to the response.
+     * Add prev/next relationship links to document and the response headers.
      *
      * @param Gdn_Controller $controller
      */
-    private function addPageMeta(Gdn_Controller $controller) {
+    private function addRelLinks(Gdn_Controller $controller) {
         static $pending = true;
 
         // Make sure this only happens once.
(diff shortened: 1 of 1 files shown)


applications/dashboard/modules/class.pagermodule.php
```diff
@@ -100,9 +105,53 @@ public function __construct($sender = '') {
...
+                $head->addTag('meta', [
+                    'rel' => 'prev',
+                    'href' => $prevHref
+                ]);
+                $this->webLinking->addLink('prev', $prevHref);
+            }
+
+            if ($this->hasMorePages()) {
+                $nextHref = $this->pageUrl($currentPage + 1);
+                $head->addTag('meta', [
```
slafleche: According to the doc, it looks like it's supposed to be a `link` https://webmasters.googleblog.com/2011/09/pagination-with-relnext-and-relprev.html

applications/dashboard/modules/class.pagermodule.php
```diff
@@ -100,9 +105,53 @@ public function __construct($sender = '') {
...
+                    'rel' => 'next',
+                    'href' => $nextHref
```
slafleche: Shouldn't these get wrapped in a "url" to be absolute in case they're used where there's be an extra slug?

slafleche: Absolute URLs would be better to avoid conflicts with other plugins.

PR #6865, opened 2018-03-27
Allow react components to mount on pages

tburry:
- Allows react components to be specified in the data-react attribute of elements.
- The default App component is used for adding pages.
- The core application module has functions for client-side interop between components.


src/scripts/Main/NotFoundPage.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+    render() {
+        return <div className="Center SplashInfo">
+            <DocumentTitle title={this.title}/>
+            <div>{this.message}</div>
+        </div>;
+    }
+}
+
+NotFoundPage.defaultProps = {
```
charrondev: We can set these as static values on the class now. eg:

```js
class NotFoundPage {
  static propTypes = {
    type: PropTypes.string,
    title: PropTypes.string,
    message: PropTypes.string,
  };
}
```

addon.json
```diff
@@ -8,24 +8,26 @@
...
         "exports": {
             "app": [
+                "./src/scripts/application",
                 "./src/scripts/events",
                 "./src/scripts/gdn",
                 "./src/scripts/permissions",
                 "./src/scripts/utility",
                 "./src/scripts/dom-utility",
                 "axios",
                 "react",
```
charrondev: We should probably add a couple more pieces here as well. `react-dom`, `react-redux`, `react-router-dom` and `redux`.

Issue #6806, opened 2018-03-14
Rich Editor - Accessible toolbars


slafleche:
Follow aria guidelines for keyboard navigation in editor menus
- [x] The inline and paragraph menus have `role="menu"`
- [x] Follow [Aria spec](https://www.w3.org/TR/wai-aria-practices-1.1/#toolbar) for making menus accessible
- [x] Tab order needs to keep natural flow
- [x] If escape closes a popup, we need to reset focus on element that opened the popup in the first place

This issue replaces https://github.com/vanilla/vanilla/issues/6404

PR #6866, opened 2018-03-27
Rich editor - Improve menu accessibility

slafleche:
Improve paragraph and inline menu accessibility in rich editor:

Features:
- [ ] Added description to rich editor. I chose a "modular" approach so plugins could "plug" in their snippet of functionality in there.

- [ ] Paragraph menu is a ["true" menu](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) 
  - [x] On the menu toggle, you can use arrow down to open and arrow up to close
  - [x] On the menu toggle, you can click with a mouse, hit enter or hit space
  - [x] Escape closes menu and sets focus on menu toggle if focus was on menu toggle, or in menu
  - [x] Arrowing Left or Up goes to last option in menu and loops
  - [x] Arrowing Right or Down goes to next option in menu and loops
  - [x] Tabbing does not loop, but allows you to access menu
  - [x] Proper `role="menu"` and `role="menuitem"` are set
  - [x] aria-pressed is set on each option
  - [x] Toggle button has `aria-controls`, `aria-haspoups`, `aria-expanded`, `aria-label`
  - [ ] When openning the menu, we set focus on the first element

- [ ] Inline menu is almost "true" menu. Since it's always open when you've got a selection, I added a description in the editor description. I also opted for an "alert" message when it gets inserted in the page to let the user know it's there. There's no proper toggle here or focus to set since the user could still be typing
	- [x] The escape behaviour changed a bit. If you escape, it sets your cursor at the end of your selection, so you get "out" of the menu
	  - [x] Arrowing Left or Up goes to last option in menu and loops
  	  - [x] Arrowing Right or Down goes to next option in menu and loops
  	  - [x] Tabbing does not loop, but allows you to access menu
  	  - [x] Proper `role="menu"` and `role="menuitem"` are set
  	  - [x] aria-pressed is set on each option

  Locale PR: https://github.com/vanilla/locales/pull/58

  Closes: https://github.com/vanilla/vanilla/issues/6806


plugins/rich-editor/src/scripts/components/EditorToolbar.jsx
```diff
@@ -114,30 +116,64 @@ export class EditorToolbar extends React.PureComponent {
         this.quill.off(Quill.events.SCROLL_OPTIMIZE, this.quillOptimizeHandler);
     }
 
+    /**
+     * Handle key presses
+     * @param {React.SyntheticEvent} e
```
charrondev: This docblock isn't correct.

plugins/rich-editor/src/scripts/components/ParagraphEditorToolbar.jsx
```diff
@@ -73,7 +75,11 @@ export class ParagraphEditorToolbar extends React.PureComponent {
 
         // Quill can directly on the class as it won't ever change in a single instance.
         this.quill = props.quill;
-
+        UniqueID.enableUniqueIds(this);
```
charrondev: The withEditor function injects the editors common unique id in as a prop. You should be able to get with `this.props.editorID`. Take a look at the EmojiPopover for an example.

charrondev:
>  Added description to rich editor. I chose a "modular" approach so plugins could "plug" in their snippet of functionality in there.

Am I missing something with this? They just look like they're hardcoded translation strings.

> When opening the menu, we set focus on the first element

Is this supposed to work with the arrow key to open as well?

---

One other thing I noticed is that it doesn't seem possible to tab over to the `x` in the link menu.

slafleche:
> > Added description to rich editor. I chose a "modular" approach so plugins could "plug" in their snippet of functionality in there.
> 
> Am I missing something with this? They just look like they're hardcoded translation strings.

I meant that it could dynamically be created. Breaking the instructions into multiple pieces is easier to deal with and allows parts to be enabled or disabled. 

> > When opening the menu, we set focus on the first element
> 
> Is this supposed to work with the arrow key to open as well?

The only 100% "true" menu as far as the aria spec is concerned is the paragraph menu. The inline menu doesn't have a toggle button and I've added the description and the "alert" to explain the functionality.

charrondev:
A couple things left:

- The `x` in the link menu is still missing a tabindex 0.
- Using the down arrow to open the paragraph menu does not set your focus on the first item. It is supposed as illustrated here https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-2/menubar-2.html.

PR #6875, opened 2018-03-28
Change redirect behaviour for users logging in through Facebook.

nervoustwit:
Facebook has recently changed their Redirect_Uri policy for Facebook Login. In the past and in grandfathered Facebook Login Apps you can have the "Use Strict Mode for Redirect URIs" set to "No" in an app allowing your app to accept Redirec_Uri's with variable query strings attached to them (e.g. https://myforum.com/entry/connect/facebook?Target=%2Fdiscussion%2Fsome-discussion). Now when you create a new Facebook Login app the "Use Strict Mode for Redirect URIs" is toggled to "Yes" and you cannot set it to "No". And if you switch a grandfathered account to "Yes" you cannot set it back to "No". This breaks the log in process entirely.

This PR removes adding "?Target" strings to Redirect_Uri sent to Facebook Login when a user is logging in. 

The affect for clients means that where ever a user is in the forum when he clicks on Facebook Login he will be redirected back to the home page of the forum when his log in is complete.

### DIFF
--- a/plugins/Facebook/class.facebook.plugin.php
+++ b/plugins/Facebook/class.facebook.plugin.php
@@ -431,7 +431,7 @@ class FacebookPlugin extends Gdn_Plugin {
         }
 
         // This isn't a trusted connection. Don't allow it to automatically connect a user account.
-        saveToConfig('Garden.Registration.AutoConnect', false, false);
+//        saveToConfig('Garden.Registration.AutoConnect', false, false);
 
         $form = $sender->Form; //new gdn_Form();
         $iD = val('id', $profile);
@@ -441,6 +441,7 @@ class FacebookPlugin extends Gdn_Plugin {
         $form->setFormValue('FullName', val('name', $profile));
         $form->setFormValue('Email', val('email', $profile));
         $form->setFormValue('Photo', "//graph.facebook.com/{$iD}/picture?width=200&height=200");
+        $form->setFormValue('Target', val('Target', $state, '/'));
         $form->addHidden('AccessToken', $accessToken);
 
         if (c('Plugins.Facebook.UseFacebookNames')) {
@@ -542,7 +543,7 @@ class FacebookPlugin extends Gdn_Plugin {
         }
 
         if ($query) {
-            $redirectUri .= '&'.$query;
+            $redirectUri .= '?'.$query;
         }
 
         // Get a state token.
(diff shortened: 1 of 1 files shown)


DaazKu:
The redirect should be put in the state parameter: https://github.com/vanilla/vanilla/blob/6ac2898bd66f6b1c695825662ba97205e6acda59/plugins/Facebook/class.facebook.plugin.php#L555

plugins/Facebook/class.facebook.plugin.php
```diff
@@ -431,7 +431,7 @@ public function base_connectData_handler($sender, $args) {
         }
 
         // This isn't a trusted connection. Don't allow it to automatically connect a user account.
-        saveToConfig('Garden.Registration.AutoConnect', false, false);
+//        saveToConfig('Garden.Registration.AutoConnect', false, false);
```
DaazKu: What was the reason for this for being disabled?
If the config is turned on for another plugin this may cause problems no?

plugins/Facebook/class.facebook.plugin.php
```diff
@@ -542,7 +543,7 @@ public function authorizeUri($query = false, $redirectUri = false) {
         }
 
         if ($query) {
-            $redirectUri .= '&'.$query;
+            $redirectUri .= '?'.$query;
```
DaazKu: In theory, unless I'm mistaken, someone could still whitelist an URL containing a query string. 
We should detect if $redirectUri contains '?' and in that case add '&' instead of '?'

plugins/Facebook/class.facebook.plugin.php
```diff
@@ -552,7 +553,7 @@ public function authorizeUri($query = false, $redirectUri = false) {
             'client_id' => $appID,
             'redirect_uri' => $redirectUri,
             'scope' => $scopes,
-            'state' => json_encode(['token' => $stateToken]),
+            'state' => json_encode(['token' => $stateToken, 'Target' => $this->getTargetUri()]),
```
DaazKu: Let's use camel case field names like token does.
You can do `$form->setFormValue('Target', val('target', $state, '/'));` above.

plugins/Facebook/class.facebook.plugin.php
nervoustwit: Was testing and it was annoying me, I meant to put it back.

PR #6888, opened 2018-03-28
Fix for addons id param to conform with other controllers

### LINKED ISSUE #6719: [API doc] Fix addonApiController idParam not being conform with the other controllers
Addon idParameter should probably be named "id" like everything else.
screen shot 2018-02-27 at 2 41 31 pm

cchabilall83:
fixes #6719

Addon idParameter was addonID instead of id

We changed the rename $addonID to $id to resolve the display issue

### DIFF
--- a/applications/dashboard/controllers/api/AddonsApiController.php
+++ b/applications/dashboard/controllers/api/AddonsApiController.php
@@ -178,17 +178,17 @@ class AddonsApiController extends Controller {
     /**
      * Get a single addon.
      *
-     * @param string $addonID The addon ID to get.
+     * @param string $id The addon ID to get.
      * @param array $query The query string.
      * @return array Returns an addon info row.
      * @throws \Garden\Web\Exception\NotFoundException Throws an exception when the addon is not found.
      */
-    public function get($addonID, array $query) {
+    public function get($id, array $query) {
         $this->permission('Garden.Settings.Manage');
 
-        $id = $this->schema([
-            'addonID:s' => 'The ID of the addon.'
-        ], 'in')->setDescription('Get an addon.');
+//        $id = $this->schema([
+//            'addonID:s' => 'The ID of the addon.'
+//        ], 'in')->setDescription('Get an addon.');
 
         $in = $this->schema([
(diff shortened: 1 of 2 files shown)


applications/dashboard/controllers/api/AddonsApiController.php
```diff
@@ -178,17 +178,17 @@ public function index(array $query) {
...
      * @throws \Garden\Web\Exception\NotFoundException Throws an exception when the addon is not found.
      */
-    public function get($addonID, array $query) {
+    public function get($id, array $query) {
         $this->permission('Garden.Settings.Manage');
 
-        $id = $this->schema([
-            'addonID:s' => 'The ID of the addon.'
-        ], 'in')->setDescription('Get an addon.');
+//        $id = $this->schema([
```
DaazKu: Wut? Why did you comment the schema?

cchabilall83: I forgot to uncomment it. I thought it wasn't used anywhere, and it was  overwriting the $id param ($addonID) right a way.

DaazKu: Yeah. You need to rename addonID to id.

tburry:
To try and explain a bit here. I put the constraint on parameters named `$id` for a reason and that is to [remove ambiguities when paths get nested](https://docs.vanillaforums.com/developer/apiv2/resource-routing/#parameter-ambiguities). I put in specific unit tests for those cases and these changes broke them.

Also keep in mind that the constraints worked the way they did because the code was done before we were at PHP7. I had relaxed the constraint in #6922 now that we have type hints. However, there are still ambiguities when `$id` parameters are strings now, but that is still backwards compatible because previously `$id` parameters could only be natural numbers. Endpoints that require an `$id` string parameter will have to be careful with what they support on sub paths.

linc:
Also, this was not put in this sprint and I don't recall it being discussed at the sprint planning meeting.

DaazKu:
@linc It was from last sprint and we were blocked because $id could not be a string.

Issue #6915, opened 2018-03-29
SwaggerModel needs to handle type-hinted endpoints


tburry:
When swagger calls an endpoint it needs to call with a dummy value that is appropriate rather than null.

PR #6922, opened 2018-03-30
Respect scalar type hints in endpoint parameters

tburry:
PHP 7’s scalar type hints can tell us what type a path parameter
expects. This provides much of the functionality that route constraints
required.

- Check for scalar type hints when testing parameter constraints.
- Change the default id constraint against its regex only when there is
no type hint.
- Set an appropriate type-hinted argument value when generating the
swagger documentation (this fixes documentation for the /authenticators
endpoint if you care to un-exclude that one in the SwaggerModel for
testing).
- Use the scalar type hint for default documentation parameters.

### DIFF
--- a/applications/dashboard/models/ReflectionAction.php
+++ b/applications/dashboard/models/ReflectionAction.php
@@ -162,7 +162,35 @@ class ReflectionAction {
             }
 
             // Default the call args.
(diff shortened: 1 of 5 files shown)


tburry:
Closes #6915

PR #6924, opened 2018-03-30
Add the POST /authenticate/password endpoint

tburry:


applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -459,4 +459,44 @@ public function ssoDataSchema() {
...
+     *
+     * @param array $body The username/password to sign in.
+     * @return array Returns the signed in user.
+     * @throws \Exception Throws all exceptions to be dispatched as error responses.
+     */
+    public function post_password(array $body): array {
+        $this->permission();
+
+        $in = $this->schema([
+            'username:s' => 'The user\'s Username or email address.',
```
DaazKu: What that UC "U" intentional?

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -459,4 +459,44 @@ public function ssoDataSchema() {
...
+            'password:s' => 'The user\'s password.',
+            'persist:b' => [
```
DaazKu: Shouldn't we also add the setCookie option like we did for `post()`?

tburry: I had thought we removed that. It's not in the markdown anymore at least.

PR #6927, opened 2018-03-30
Add the POST /users/request-password endpoint

tburry:
This adds the endpoint with the following changes:

- Some of the code in /entry/password request has been refactored into
the UserModel so that both endpoints can use it.
- You can now only request a password for a username if email address
is not unique and username is unique. This is to crack down on some
attack vectors.
- Labelling on the forms have been changed to reflect the new
restrictions.
- A new TestLogger is now registered for APIv2 tests so that logging
can be checked during testing.

### DIFF
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -590,13 +590,12 @@ class UsersApiController extends AbstractApiController {
         $in = $this->schema([
             'email:s' => 'The email/username of the user.',
         ]);
+        $out = $this->schema([], 'out');
 
         $body = $in->validate($body);
 
         $this->userModel->passwordRequest($body['email']);
         $this->validateModel($this->userModel, true);
-
-        return '';
     }
 
     /**
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -579,6 +579,27 @@ class UsersApiController extends AbstractApiController {
     }
 
     /**
+     * Send a password reset email.
+     *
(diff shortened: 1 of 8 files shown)


applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -578,6 +578,27 @@ public function put_ban($id, array $body) {
...
+     *
+     * @param array $body The POST body.
+     * @throws Exception Throws all exceptions to the dispatcher.
+     */
+    public function post_requestPassword(array $body) {
+        $this->permission();
+
+        $in = $this->schema([
+            'email:s' => 'The email/username of the user.',
+        ]);
```
DaazKu: You are missing an out schema.
```php
$this->schema([], 'out');
```

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -578,6 +578,27 @@ public function put_ban($id, array $body) {
...
+
+        $body = $in->validate($body);
+
+        $this->userModel->passwordRequest($body['email']);
+        $this->validateModel($this->userModel, true);
+
+        return '';
```
DaazKu: No need to return anything if there's nothing to return. `null` would probably be more appropriate than an empty string. If you keep the return statement the DocBlock needs to be updated.

tests/TestLogger.php
```diff
@@ -0,0 +1,76 @@
...
+                }
+            }
+            if ($found) {
+                return $item;
+            }
+        }
+        return null;
+    }
+
+    public function clear() {
```
DaazKu: DocBlock

PR #6928, opened 2018-03-30
Set participant's email to empty if permission check fails

cchabilall83:
fixes#208

The participants email was being exposed to members without permission

To ensure the the participants email is exposed, we loop through the participants and verify if they have permission to view the personal information, if not the email is sent to an empty string

### DIFF
--- a/applications/conversations/controllers/class.messagescontroller.php
+++ b/applications/conversations/controllers/class.messagescontroller.php
@@ -368,6 +368,14 @@ class MessagesController extends ConversationsController {
 
         $this->Conversation = $this->ConversationModel->getID($conversationID);
         $this->Conversation->Participants = $this->ConversationModel->getRecipients($conversationID);
+
+        //Verify if participant email should be visible
+        foreach ($this->Conversation->Participants as $participant) {
+            if (!Gdn::session()->checkPermission('Garden.PersonalInfo.View')) {
+                $participant->Email = '';
+            }
+        }
+
         $this->setData('Conversation', $this->Conversation);
 
         // Bad conversation? Redirect


applications/conversations/controllers/class.messagescontroller.php
```diff
@@ -368,6 +368,14 @@ public function index($conversationID = false, $offset = -1, $limit = '') {
 
         $this->Conversation = $this->ConversationModel->getID($conversationID);
         $this->Conversation->Participants = $this->ConversationModel->getRecipients($conversationID);
+
+        //Verify if participant email should be visible
```
DaazKu: Missing before and period at the end.
`// Verify if participant email should be visible.`

applications/conversations/controllers/class.messagescontroller.php
```diff
@@ -368,6 +368,14 @@ public function index($conversationID = false, $offset = -1, $limit = '') {
...
+        foreach ($this->Conversation->Participants as $participant) {
+            if (!Gdn::session()->checkPermission('Garden.PersonalInfo.View')) {
```
DaazKu: You are going a permission check multiple times for no reason. The code should be:
```php
if (!Gdn::session()->checkPermission('Garden.PersonalInfo.View')) {
    foreach ($this->Conversation->Participants as $participant) {
        $participant->Email = '';
    }
}
```

The permission check is also wrong. It should be `Garden.Email.View`.

DaazKu: `Garden.Email.View` determine if one can receive emails... `Garden.PersonalInfo.View` was correct

PR #6939, opened 2018-04-03
Fix polyfill loading

charrondev:
This PR fixes polyfill loading. Polyfills are now conditionally loaded through an inline script that will append the polyfill script (in the head and deferred) after itself if certain conditions aren't met. It will then get parsed before all other bundled scripts.

As a result we no longer really need the onReady handler, but we can deal with that separately.

I also brought in the CustomEvent polyfill from the rich editor branch.

This can be pretty easily tested by adding something like this

```html
          <script>
                delete window.NodeList.prototype.forEach;
                delete window.Promise;
                delete window.Iterator
          </script>
```

Before all of the other scripts (easiest through the default.master.tpl).


applications/dashboard/modules/class.headmodule.php
```diff
@@ -137,9 +137,10 @@ public function addTag($tag, $properties, $content = null, $index = null) {
          *  - numeric: This will be the script's sort.
          *  - string: This will hint the script (inline will inline the file in the page.
          *  - array: An array of options (ex. sort, hint, version).
+         * @param string $content The content of the script tag.
          *
          */
-        public function addScript($src, $type = 'text/javascript', $addVersion = true, $options = []) {
+        public function addScript($src, $type = 'text/javascript', $addVersion = true, $options = [], $content = null) {
```
tburry: I think you can just call addScript with `['content' => '...']` in the $options parameter.

library/core/class.controller.php
```diff
@@ -1863,6 +1863,26 @@ public function renderMaster() {
                 // This is done in the controller rather than the asset model because the translations are not linked to compiled code.
                 $this->Head->addScript(url('/api/v2/locales/'.rawurlencode(Gdn::locale()->current())."/translations?js=1&x-cache=1&h=$busta", true), 'text/javascript', false, ['defer' => 'true']);
 
+                $polyfillFileUrl = asset("/js/polyfills/core-polyfills.js", false, $addVersion);
```
tburry: Can you make this a method in the AssetModel so this can be re-used on non Gdn_Controller classes?

Issue #6945, opened 2018-04-04
Required PHP Extensions Are Never Ensured To Be Available


R-J:
You can install Vanilla just fine without mbstring, curl and gd extension but all of them will cause not easy to debug errors later on. Let's be honest: no one really reads requirements so even if they are all named there, people would install Vanilla on systems that doesn't satisfy the requirements.

Therefore requirements should either be tested during installation or there should be some kind of auto-enabled plugin (like stubcontent is) which tests requirements and shows warning messages in the dashboard.

charrondev:
Before we moved to a docker setup internally (https://github.com/vanilla/vanilla-docker) we would occasionally run into this type of thing when onboarding new people, so I'm definitely in agreement with this.

Just to summarize here the required extensions are:

- `mbstring`
- `gd`
- `curl`

Additionally we use the `intl` extension for the Multilingual plugin and in `Gdn_Smarty` although we have a fallback if the extension is not installed. If we're going to enforce these extensions maybe we should consider requiring that extension as well and get rid of those checks? What is the scenario where we don't care to have it installed?

charrondev:
Additionally I'm going to condense your other related issues into this one. The 

#6944 
> GD is obviously a requirement (see here) but it is not listed as a requirement in the README

#6943 isn't quite 100% resolved by this fixing this issue, but its close enough I think. @linc If you think the portion of people that are missing the GD extension but already have a setup forum and haven't noticed profile pictures not working correctly then we can re-open this one.
> Please see this discussion: https://open.vanillaforums.com/discussion/comment/252352/#Comment_252352
>
>A user got a 404 status return when he tried to access /profile/picture simply because he hadn't > installed the GD extension. Not sure if 500 or 501 would be correct, but both are better than 404.

charrondev:
So we had our sprint meeting this morning and it seems that some of these extensions are used in particular rare cases and are not actually required for every installation. As a result we don't want to fully block installation of Vanilla without them.

We would rather have a warning on the installation, but not fully prevent installation, then handle the individual cases where these extensions are used with proper error messages/warnings.

You want to weigh in @linc 

**Errors**
- `mbstring` - This seems to be peppered throughout `library` and sometimes the function calls are wrapped in `if()` statements and other times they aren't. I'd say this is definitely required and should probably error in installation.
- `gd` - Seems to be required for image uploads. That's a pretty core part of Vanilla and there doesn't seem to be any reasonable fallback.

[...]

PR #6946, opened 2018-04-04
Verify required PHP extensions are available during setup

R-J:
This PR should fix https://github.com/vanilla/vanilla/issues/6945

By now only two of five required PHP extensions have been taken care of during setup. This PR defines an array of required extensions and loops through this array to check for the modules status. If a module is not enabled, an error is added to the form.


R-J:
P.S.: `Form->addError()` translates the error message by itself so there is no need to add a `t()` anywhere.

DaazKu:
We are gonna review this at a later time.
We want to have a discussion on how this should be best handled. (In setup or later or whatever)

linc:
I'm assigning developers to look at this by the end of the month. Apologies for the delay.

PR #6953, opened 2018-04-04
Move at-who javascript into a bundled file

### LINKED ISSUE #5676: Create legacy js bundle for global js bits that need to be unit tested
This way we can move pieces out from global.js such as the recent regex change from the recent at who issues.

Related:
#6783
#6817

charrondev:
Closes #5676 

Basically takes all of the at.who related things from global.js and moves them into a new bundled file, where the important to test functions are exported.

Additional parts of global.js that get pulled out in the future should have there own files and be handled in `legacy/index.ts`. The code style has been updated and some functions pulled up and named.

Apparently the function nesting was getting to be a bit much in global.js so a custom `String.prototype` extension was actually in the middle of the setup function for at who. That was left in `global.js`.


tburry:
Sorry about the merge conflicts. I snuck in some PRs which is bad etiquette.

PR #6968, opened 2018-04-05
Have ParagraphMenu update its icon to match current formatting

### LINKED ISSUE #6808: Rich Editor - Dynamic paragraph formatting icon
The paragraph formatting icon should be dynamic based on type of block you are in:

> Block formatting bar. A button follows the cursor along the side of the editor. Clicking the butting brings up a toolbar of paragraph formatting options. The icon for this should match the currently applied paragraph format. If there is no formatting then it is the pilcrow (¶).

charrondev:
Closes #6808

What was previously the "pilcrow" icon for the paragraph menu has been replaced with a dynamic icon that reflects the state of the current range/selection.

A couple of the Toolbar related components have been converted to typescript. I also removed an unused `handleKeypress` method from the `Toolbar` component. It was getting its onKeyDown handler from a prop instead.


plugins/rich-editor/src/scripts/Editor/Generic/MenuItem.tsx
```diff
@@ -0,0 +1,110 @@
+/**
+ * @author Adam (charrondev) Charron <adam.c@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license https://opensource.org/licenses/GPL-2.0 GPL-2.0
+ */
+
+import React from "react";
+import * as PropTypes from "prop-types";
```
slafleche: Do we still need these?

plugins/rich-editor/src/scripts/Editor/Generic/MenuItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+        super(props);
+        this.onBlur = (props.isLast && props.onBlur) ? props.onBlur : () => undefined;
+        this.buttonRole = props.role || "button";
+    }
+
+    public render() {
+        const { propertyName, isActive, onClick } = this.props;
+        const Icon = Icons[propertyName];
+        const buttonClasses = classnames("richEditor-button", "richEditor-menuItem", {
+            isActive: isActive || false,
```
slafleche: Is this the official style or are we going to use the `!!` style?

plugins/rich-editor/src/scripts/Editor/Generic/Toolbar.tsx
```diff
@@ -237,11 +194,11 @@ export class Toolbar extends React.PureComponent {
     /**
      * Handle the simple on/off inline formats (eg. bold, italic).
      *
-     * @param {string} itemKey - The key of the item.
-     * @param {Object} itemData - The item to modify.
-     * @param {Object} range - The range to update.
+     * @param itemKey - The key of the item.
```
slafleche: Shouldn't these params be refactored to typescript format?

plugins/rich-editor/src/scripts/Editor/Generic/MenuItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+        });
+
+        return <button
+            ref={(ref: HTMLButtonElement) => { this.domButton = ref; }}
+            className={buttonClasses}
+            type="button"
+            aria-label={t('richEditor.menu.' + this.props.propertyName)}
```
slafleche: The translation isn't coming through. 
screen shot 2018-04-06 at 13 46 34

plugins/rich-editor/src/scripts/Editor/Generic/Toolbar.tsx
charrondev: Notice the types have been removed. I left text comments for the parameters, which is still valid here. We should definitely not keep the type annotations in the comments though. Parameter comments are optional, but in some cases nice to have.

plugins/rich-editor/src/scripts/Editor/Generic/MenuItem.tsx
charrondev: I've updated these translations

PR #6969, opened 2018-04-06
Fix edge case crawl errors

### LINKED ISSUE #6796: Overzealous Page Crawl Issue Related to Pagination?
I have noticed an issue,  where Google is indexing pages via pagination beyond content, when pages are not even relevant.

For example: It's indexing page 2263 of a 3 page discussions. Here are some examples from from open source community found in Google Search and in Google webmaster tools:

https://open.vanillaforums.com/discussion/3908/guestpost-v1-3/p2263
https://open.vanillaforums.com/discussion/2789/blacklist-forums-hiding-the-vanilla-text/p1527

Screen shot from Google Webmaster tools of some examples:

image

I also have examples in my own hosted forum as well.

tburry:
There are several edge cases where crawlers may start crawling non-existent pages of discussions, categories, etc. See the individual commits for more information.

There are two main components to this PR:

1. Render a single previous link for numbered pagers when the specified page is greater than the number of pages.
2. Throw not found errors on select pages when the specified page is greater than the number of pages.

The not found errors provide some edge cases that should be taken into consideration:

1. When there are no records the first page should not throw an error.
2. Auto offset into discussions can throw things for a loop.
3. Off-by-one errors.

Closes #6796

### DIFF
--- a/applications/dashboard/modules/class.pagermodule.php
+++ b/applications/dashboard/modules/class.pagermodule.php
@@ -14,6 +14,8 @@ use Vanilla\Web\WebLinking;
  * Builds a pager control related to a dataset.
  */
 class PagerModule extends Gdn_Module {
+    const PREV_NEXT_CLASS = 'PrevNextPager';
+    const NUMBERED_CLASS = 'NumberedPager';
 
     /** @var WebLinking */
     private $webLinking;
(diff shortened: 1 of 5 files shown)


tburry:
I've added a bunch of line comments to my code to hopefully explain what I've done. I'll say though that this needs some basic click around QA. I ran into some major gotchas while doing the fix that were not obvious from a code point of view, but were blindingly obvious when I clicked around.

PR #6970, opened 2018-04-06
Fix undefined $timeLeft variable notices

tburry:
I just started seeing a lot of notices crop up with a missing variable. I’m not sure when the change was done, but the underlying function in Gdn_Model was added in ccae291fffc03636884b26d51475f57dbc0e4d4e.

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -742,7 +742,7 @@ class DiscussionController extends VanillaController {
                 }
 
                 // Make sure that content can (still) be edited.
-                if (!$this->CommentModel->canEdit($comment)) {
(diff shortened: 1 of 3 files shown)


initvector:
I was only seeing the notice for comments. `getDiscussionOptions` should've already been handling the variable.

https://github.com/vanilla/vanilla/blob/248e98bb7e31aa56c37c1423229d0d3c6c5accaf/applications/vanilla/views/discussion/helper_functions.php#L227-L229

`getCommentOptions` was not.

https://github.com/vanilla/vanilla/blob/248e98bb7e31aa56c37c1423229d0d3c6c5accaf/applications/vanilla/views/discussion/helper_functions.php#L441-L443

`CommentModel::canEdit` was added in #6782, but there were some strange calls that lead to these notices. I've tracked down a couple more and addressed them here.

Issue #6802, opened 2018-03-14
Rework AuthenticateApiController


DaazKu:
We need to rework that controller to match this: https://github.com/vanilla/roadmap/blob/master/api/authenticate.md

PR #6974, opened 2018-04-06
Update authenticate and authenticators API

DaazKu:
Partially address #6802

Check the specs: https://github.com/vanilla/roadmap/blob/master/api/

- Remove session related stuff from `/authenticate`
  - Disabled connect feature in AuthenticateController because of that.
- Add `/authenticate/authenticators` stuff
- Refactor POST `/authenticate` so that authenticators `type` and `id` are under the 'authenticate' field.
- Add a constraint on authenticator where type MUST equals id if the authenticator is unique. This helps getting authenticators by ID that that constraint will help "blacklisting" certain authenticator IDs in the UI.
- Add back getAuthenticatorByID in AuthenticatorModel. Always defaults to unique authenticator if the ID would clash.
- Add `isUserLinked()` function to SSOAuthenticators. This allows us to know if a user is linked or not to an authenticator using GET `/authenticate/authenticators` and `/authenticate/authenticators/:id`
- Add unit tests and update old ones.


library/Vanilla/Authenticator/PasswordAuthenticator.php
```diff
@@ -35,7 +35,7 @@ public function __construct(UserModel $userModel) {
     protected static function getAuthenticatorTypeInfoImpl(): array {
         return [
             'ui' => [
-                'photoUrl' => '/applications/dashboard/design/images/password-authenticator.svg',
+                'photoUrl' => '/applications/dashboard/design/images/password-Authenticator.svg',
```
initvector: Probably wanna keep this lowercase

library/Vanilla/Models/AuthenticatorModel.php
```diff
@@ -159,7 +184,24 @@ public function getAuthenticators() {
     }
 
     /**
-     * Get available authenticator classes.
+     * Get the list of ID of unique authenticators.
+     *
+     * @return array
+     */
+    public function getUniqueAuthenticatorIDs() {
```
initvector: Is it safe to slap a return type on this method?

tests/APIv2/Authenticate/LinkUserTest.php
```diff
@@ -145,15 +155,47 @@ public function testLinkUserWrongPassword() {
         $this->assertEquals($this->currentUser['userID'], $body['userID']);
     }
 
+    public function testUnlinkUser() {
```
initvector: Could use some doc-blockin'

applications/dashboard/controllers/api/AuthenticateApiController.php
```diff
@@ -127,133 +130,128 @@ public function delete($authenticator, $authenticatorID = '', array $query) {
...
-     * @param array $query
-     * @return array
      * @throws Exception
+     * @param array $query
+     * @return Authenticator
      */
-    public function get_session($authSessionID, array $query) {
-        $this->permission();
+    public function index_authenticators(array $query) {
+         $this->permission();
```
initvector: One too many spaces, I think

applications/dashboard/controllers/class.authenticatecontroller.php
```diff
@@ -104,26 +104,32 @@ public function initialize() {
...
         if ($response['authenticationStep'] === 'authenticated') {
             $redirectURL = (val('target', $this->request->getQuery(), '/'));
         } else {
-            $target = val('target', $this->request->getQuery());
-            if ($target) {
-                $target = '&target='.$target;
-            }
-            $redirectURL = "/authenticate/connectuser?authSessionID={$response['authSessionID']}{$target}";
+            throw new Gdn_ErrorException('Connect user feature is not implemented.');
+//            $target = val('target', $this->request->getQuery());
```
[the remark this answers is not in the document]
DaazKu: Nah. I'm working in there next sprint and is really doesn't do much to have these completely removed vs commented.

Issue #6559, opened 2018-02-01
Addon Key for Toggle


austins:
For Vanilla 2.5, it looks like the _addon name_ is being used for the toggle slug on the `/settings/applications` page. For example, I have an app called "Basic Pages" and the toggle link is `/settings/applications/all/Basic%20Pages`. This causes an "Application not found." error.

Should the _addon key_ be used for the toggle slug instead? It would then be `/settings/applications/all/basicpages`, which will allow the addon to be looked up successfully. Even so, there seems to be more underlying code in the enable/disable application methods that prevent it from successfully toggling.

Charrondev:
Are you using the [addon.json](https://docs.vanillaforums.com/developer/addons/addon-info/) format for defining your application's metadata?

austins:
Yes, I'm using the addon.json format like so:

~~~
{
  "key": "basicpages",
  "name": "Basic Pages",
  "description": "Basic Pages is an application that provides a way for you to create basic public pages for static content in Garden.",
  "version": "2.1.8",
  "url": "http://vanillaforums.org/addon/basicpages-application",
  "license": "GNU GPL v2",
  "authors": [
    {
      "name": "Austin S.",
      "homepage": "https://github.com/austins"
    }
  ],
  "icon": "basicpages.png",
  "type": "addon",
  "require": {
    "vanilla": ">=2.5"
  },
  "setupController": "setup"
}
~~~

charrondev:
@austins This issue doesn't have anyone internal that seems to care for it. I'm not going to close it but I'd just recommend not using applications anymore.

Addons should be able to be converted to a plugin with a few minor changes and setting the priority in addon.json to be 10.

PR #6987, opened 2018-04-09
Addon Key for Toggle

austins:
Fixes issue #6559.

1. For addons that have spaces in their names such as "Basic Pages", the toggle link would contain "Basic%20Pages". With this, Gdn::addonManager()->lookupAddon() wouldn't be able to find the addon because the multiCache array contains each addon's real key or raw key, not their name. The fix is to use Key or KeyRaw for the toggle link.
2. Vanilla's "kludge that sets keyRaw until we use key everywhere" sets $info['keyRaw'] to the addon's name that may contain spaces. It should set it to the name without a space as versions of Vanilla before 2.5 (without addon.json) had their key being their names without spaces like "BasicPages". The fix is to remove spaces in the name. This allows the application to properly be enabled or disabled as the config property had been "EnabledApplications.BasicPages" without any spaces.

Successfully tested my changes for compatibility with old style addons without addon.json and also new style addons with only addon.json. Lookup and toggling working good.

### DIFF
--- a/applications/dashboard/views/settings/helper_functions.php
+++ b/applications/dashboard/views/settings/helper_functions.php
@@ -180,7 +180,7 @@ function writeAddonMedia($addonName, $addonInfo, $isEnabled, $addonType, $filter
         $label = sprintf(t('Enable %s'), $screenName);
     }
 
-    $url = '/settings/'.$addonType.'/'.$action.'/'.$addonName;
+    $url = '/settings/'.$addonType.'/'.$action.'/'.val('Key', $addonInfo, $addonInfo['KeyRaw']);
 
     $media->setToggle(slugify($addonName), $isEnabled, $url, $label);
 
--- a/library/Vanilla/Addon.php
+++ b/library/Vanilla/Addon.php
@@ -157,7 +157,7 @@ class Addon {
                 // Kludge that sets keyRaw until we use key everywhere.
                 if ($info['oldType'] === 'application') {
                     if (!isset($info['keyRaw'])) {
-                        $info['keyRaw'] = $info['name'];
+                        $info['keyRaw'] = str_replace(' ', '', $info['name']);
                     }
                 } else {
                     if ($addonFolder !== $info['key']) {


DaazKu:
I don't understand what problem this PR is supposed to fix.
Toggle URLs for applications/plugins are not generated using the name.

screen shot 2018-04-09 at 5 49 51 pm

The variable is poorly named and in reality what we pass there is the key or the keyRaw.
Even for [old addons](https://github.com/vanilla/vanilla/blob/f30141c503422b84040c26463fffa18c0ea3a4c5/library/core/class.pluginmanager.php#L272) -> https://github.com/vanilla/vanilla/blob/f11483eca91486cb1e3e75d0e70fe0a5ff5821ca/library/core/class.applicationmanager.php#L61 -> https://github.com/vanilla/vanilla/blob/32fe13d05dee9dc2bf731b942a68eba5bd466c5f/applications/dashboard/views/settings/plugins.php#L14

austins:
@DaazKu, let me double check. I was using [Basic Pages 2.1.8](https://github.com/austins/Vanilla-App-BasicPages) (from the master branch) as it has addon.json only and I haven't uploaded it to the directory yet. I was testing the changes on Vanilla 2.5, but let me test out with Vanilla's master branch too.

austins:
@DaazKu, Vanilla 2.5.1 and the latest master (2.7a1?) have the issue with [Basic Pages v2.1.8](https://github.com/austins/Vanilla-App-BasicPages/releases/tag/v2.1.8) that has addon.json only. I haven't uploaded this version to the directory yet due to another issue with the directory.



You can see the toggle link above is `/settings/applications/all/Basic Pages`.

$addon returns null here:
https://github.com/austins/vanilla/blob/75bf122c9a46912e8aa3a15b6df103a2a7fa90a4/applications/dashboard/controllers/class.settingscontroller.php#L94

DaazKu:
You are missing `keyRaw` in addon.json (since it is an application).

To convert old apps/plugins I suggest that you use https://github.com/vanilla/vanilla-cli.
From /vanilla-cli you can do `./bin/vanilla addon-json`.

Here is the result:
```
{
    "name": "Basic Pages",
    "description": "Basic Pages is an application that provides a way for you to create basic public pages for static content in Garden.",
    "version": "2.1.7",
    "url": "http://vanillaforums.org/addon/basicpages-application",
    "license": "GNU GPL2",
    "setupController": "setup",
    "settingsUrl": "pagessettings/allpages",
    "key": "basicpages",
    "type": "addon",
    "priority": 10,
    "authors": [
        {
            "name": "Austin S.",
            "homepage": "https://github.com/austins"
        }
    ],
    "require": {
        "vanilla": ">=2.1"
    },
    "keyRaw": "BasicPages"
}
```

FYI applications are getting deprecated soon™ to be replaced by addons.

I will try to have a post or some sort of documentation created on how to update applications/plugins properly.

austins:
Adding "keyRaw" did the trick to fix the toggle. That wasn't in the documentation, so I look forward to seeing your post on updating applications/plugins properly. Thank you! Anyway, it seems like converting the application to a plugin/addon may be the better way to go when updating pre-Vanilla 2.5 applications.

Strange how the toggle link for applications gets generated differently than those for plugins. For example, if you look at Button Bar, its _only_ key is "buttonbar" in addon.json, but its link gets generated without a space ("ButtonBar") because it uses the plugin's folder name. If you rename ButtonBar folder to ButtonBar2, the link will show this. With applications, it just uses the _name_ in addon.json without removing spaces.

In [the documentation](https://docs.vanillaforums.com/developer/addons/addon-quickstart/#2-define-your-addon), it says the key value "must exactly match the folder name, including capitalization. Some systems are case-insensitive and will ignore differences in capitalization, but others are not, so make sure they match!"

It still looks like toggle links don't use the key or keyRaw; otherwise, the link for ButtonBar would be "buttonbar". Folder name is better than addon name for sure, but the key is unique and a proper identifier for an addon that should be used instead. Even so, the key that's set for ButtonBar does not match its folder name.

austins:
@DaazKu, adding "keyRaw" did the trick to fix the toggle. That wasn't in the documentation, so I look forward to seeing your post on updating applications/plugins properly. Thank you! Anyway, it seems like converting the application to a plugin/addon may be the better way to go when updating pre-Vanilla 2.5 applications.

Strange how the toggle link for applications gets generated differently than those for plugins. For example, if you look at Button Bar, its _only_ key is "buttonbar" in addon.json, but its link gets generated without a space ("ButtonBar") because it uses the plugin's folder name. If you rename ButtonBar folder to ButtonBar2, the link will show this. With applications, it just uses the _name_ in addon.json without removing spaces.

In [the documentation](https://docs.vanillaforums.com/developer/addons/addon-quickstart/#2-define-your-addon), it says the key value "must exactly match the folder name, including capitalization. Some systems are case-insensitive and will ignore differences in capitalization, but others are not, so make sure they match!"

It still looks like toggle links don't use the key or keyRaw; otherwise, the link for ButtonBar would be "buttonbar". Folder name is better than addon name for sure, but the key is unique and a proper identifier for an addon that should be used instead. Even so, the key that's set for ButtonBar does not match its folder name.

PR #7032, opened 2018-04-12
Create Quill PHP renderer

charrondev:
Please Squash merge this when ready. This is the part of the rich editor that is not the plugin. Everything else has been moved to https://github.com/vanilla/rich-editor.

This represents mostly the PHP renderer for Quill. It is mostly completed, but could use some additional tests, particularly security ones. The CSS and javascript required for the rendered output is also present here.


initvector:
There are an awful lot of static methods in here. I believe these were done for the sake of performance (e.g. avoid instantiating X blots just for their `matches` method), but static methods don't lend themselves well to DI and can make testing more difficult. If there's anyway to revisit this implementation to avoid (or at least sharply limit the usage of) static methods, I think it would be a worthwhile endeavor.

PR #7039, opened 2018-04-16
Refactor core js files into typescript

tburry:
Rename our big core files to ts/tsx and add appropriate type hints.


src/scripts/Components/DocumentTitle.tsx
```diff
@@ -15,23 +14,32 @@ import { getMeta } from "@core/application";
...
+            return this.props.children;
+        } else {
+            return <h1>{this.props.title}</h1>;
+        }
+    }
+
+    /**
+     * Calculate the status bar title from the props.
+     *
+     * @param {IProps} props The props used to calculate the title.
```
charrondev: We should probably skip the comment type annotations and just put them in the function declaration. `+    private getHeadTitle(props: IProps): string {`

Issue #7014, opened 2018-04-12
Restore the breadcrumb under the editor


LauraMakesArt:
https://github.com/vanilla/vanilla/pull/6516
https://github.com/vanilla/support/issues/1

We implemented this change and have had nothing but confusion since. When the changed was proposed I assumed we were only removing the breadcrumb from groups. Removing it completely eliminates all our bottom navigation, and the end users immediately noticed.

DaazKu:
Todd
> The breadcrumb should be built in the controller and send to the view as data

PR #7043, opened 2018-04-16
Move "category" back link definition into discusion controller

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/7014

We added a new property that can be redefined using events.
Views that have been redefined won't reflect that change but nothing will break!

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -95,6 +95,7 @@ class DiscussionController extends VanillaController {
         }
 
         $this->setData('Category', $Category);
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -95,6 +95,7 @@ public function index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
         }
 
         $this->setData('Category', $Category);
+        $this->setData('Editor.BackLink', anchor($Category['Name'], categoryUrl($Category)));
```
charrondev: Was the removal of the the the `htmlspecialchars` here intentional?

DaazKu: Nope. I forgot to add it back. Good catch.

PR #7044, opened 2018-04-16
Authenticator - Added Sign In Page

slafleche:
This PR is for the new UI for the "Sign In" page `/authenticate/signin`.

The `/authenticate/authenticators` endpoint is *incomplete*, but returns the data we need for the UI for testing (see setup in: https://github.com/vanilla/vanilla/pull/7038).

This PR also includes some fixes/refactoring for ID generation and translation strings.

Related: https://github.com/vanilla/locales/pull/60


src/scripts/Interfaces/componentIDs.ts
```diff
@@ -14,7 +14,7 @@ export function uniqueID(props:IComponentID, uniqueSuffix:string, allowNoID?:boo
 
     if (!allowNoID) {
         if ((!props.ID && !props.parentID) || (props.ID && props.parentID)) {
-            throw new Error(`You must have *either* ID or parentID`);
+            throw new Error(`You must have *either* 'id' or 'parentID'`);
```
charrondev: ID should be uppercased here to match the props field.

PR #7046, opened 2018-04-17
Add the /api/v2/dashboard/menus endpoint

tburry:
This endpoint lists all of the menus in the dashboard. It is going to be necessary in order to add a client-side nav module and addon management page.

This is a bit of yack shaving, but to give a bit of context on where this PR is coming from:

- We've the /api/v2/addons resource for a while which is necessary for a client-side addons page.
- Since enabling/disabling an addon can affect the menu we need a way of refreshing it.
- We need to refactor the addons page in general to support the new feature in the API such as conflicts. And once we change the page we can remove a bunch of old code.

### DIFF
--- a/applications/dashboard/controllers/api/DashboardApiController.php
+++ b/applications/dashboard/controllers/api/DashboardApiController.php
@@ -1,0 +1,76 @@
+<?php
+/**
+ * @author Todd Burry <todd@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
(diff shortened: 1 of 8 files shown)


tburry:
Just a note that the DashboardNavModule is a bit of a mess. I cleaned up as much as possible in order to make the endpoint and figure that a refactor of that code is beyond the scope of this PR. I think if we ever want to do such a thing it should be done when the module is moved 100% into React.

PR #7048, opened 2018-04-17
Add ability to override theme per contoller

Csabbencs:
Although class.controller.php said it was possible the override theme from any controller, you couldn't do so, because the assetmodel would never know about it.

To test it:
1. Don't pull the code first, but add 
`$this->Theme = isMobile() ? 'mobile' : desktop';`
to any controller before 
`$this->Render();`

2. Change 'mobile' or 'desktop' to any theme of your kind, i.e. 'bittersweet' and see that the current theme won't change.

3. Pull and merge my code.

4. Repeat step 2 and see that the theme now changes.

### DIFF
--- a/applications/dashboard/models/class.assetmodel.php
+++ b/applications/dashboard/models/class.assetmodel.php
@@ -79,7 +79,7 @@ class AssetModel extends Gdn_Model {
      * @return array
      * @throws Exception
      */
-    public function getCssFiles($themeType, $basename, $eTag, &$notFound = null, $dynamicCurrentTheme = null) {
+    public function getCssFiles($themeType, $basename, $eTag, &$notFound = null, $currentTheme = null) {
         $notFound = [];
         $basename = strtolower($basename);
 
@@ -138,7 +138,7 @@ class AssetModel extends Gdn_Model {
                 $paths[] = [false, $folder, $options];
 
             } else {
-                list($path, $urlPath) = self::cssPath($filename, $folder, $themeType, $dynamicCurrentTheme);
+                list($path, $urlPath) = self::cssPath($filename, $folder, $themeType, $currentTheme);
                 if ($path) {
                     $paths[] = [$path, $urlPath, $options];
                 } else {
@@ -258,7 +258,7 @@ class AssetModel extends Gdn_Model {
      * @param string $themeType mobile or desktop
      * @return array|bool
      */
-    public static function cssPath($filename, $folder = '', $themeType = '', $dynamicCurrentTheme = null) {
+    public static function cssPath($filename, $folder = '', $themeType = '', $currentTheme = null) {
         if (!$themeType) {
             $themeType = isMobile() ? 'mobile' : 'desktop';
         }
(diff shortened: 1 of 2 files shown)


library/core/class.controller.php
```diff
@@ -1809,7 +1809,8 @@ public function renderMaster() {
                     // style.css and admin.css deserve some custom processing.
                     if (in_array($CssFile, $CssAnchors)) {
                         // Grab all of the css files from the asset model.
-                        $CssFiles = $AssetModel->getCssFiles($ThemeType, ucfirst(substr($CssFile, 0, -4)), $ETag);
+                        $dummy = null;
+                        $CssFiles = $AssetModel->getCssFiles($ThemeType, ucfirst(substr($CssFile, 0, -4)), $ETag, $dummy, $this->Theme);
```
linc: Just use `null` inline rather than setting a dummy variable first.

applications/dashboard/models/class.assetmodel.php
```diff
@@ -79,7 +79,7 @@ public function addCssFile($filename, $folder = false, $options = false) {
      * @return array
      * @throws Exception
      */
-    public function getCssFiles($themeType, $basename, $eTag, &$notFound = null) {
+    public function getCssFiles($themeType, $basename, $eTag, &$notFound = null, $dynamicCurrentTheme = null) {
```
linc: Probably just `$currentTheme` is sufficient. Adding the word "dynamic" is just making it longer without adding information.

applications/dashboard/models/class.assetmodel.php
```diff
@@ -286,6 +286,10 @@ public static function cssPath($filename, $folder = '', $themeType = '') {
 
         // 3. Check the theme.
         $theme = Gdn::themeManager()->themeFromType($themeType);
+
+        // Let override theme dynamically
+        if (isset($dynamicCurrentTheme) && $dynamicCurrentTheme != $theme) $theme = $dynamicCurrentTheme;
```
linc: Our coding standards require the use of a new line + braces on all logic statements. So always:

```
if (condition) {
   // do this
}
```

library/core/class.controller.php
Csabbencs: I did that in the first place but then it dumped with:  'Cannot pass parameter null by reference', because `$AssetModel->getCssFiles` expects a reference for parameter 4 (I was not brave enough to change that. I can't see the implications), but I can change that if you say so. Can you advise?

applications/dashboard/models/class.assetmodel.php
Csabbencs: Fine, will do.

applications/dashboard/models/class.assetmodel.php
Csabbencs: Fine, will do.

library/core/class.controller.php
linc: You're correct. I went digging and as much as this method signature is extraordinarily poorly designed, I dare not alter it now given the costs & possible repercussions of that. Now I'm wondering if we're approaching this the wrong way from the start. I'll have a developer take a look at this or I'll circle back later when I can unwind this a bit more.

linc: As an aside, I traced that pass-by-ref to a change by Tim in 2015, which I assume means our cloud infrastructure is doing something with it.

tburry: The convention is `$_` as a dummy argument and you don't have to declare it first.

PR #7049, opened 2018-04-17
Add the Attributes class

tburry:
API output that includes a free-form attributes array has an issue in JSON encode where it will encode as `[]` instead of `{}` if it is empty. This is fine for PHP clients, but a bug in most any other API client.

The Attributes class is a thin wrapper for an array that handles empty data properly and also simplifies some of the dbdecode functionality.


library/Vanilla/InternalRequest.php
```diff
@@ -0,0 +1,147 @@
...
+ * @author Todd Burry <todd@vanillaforums.com>
+ * @copyright 2009-2017 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+namespace Vanilla;
+
+use Garden\Web\RequestInterface;
+
+class InternalRequest implements RequestInterface {
```
initvector: This shouldn't be here, should it? I'm not seeing any references to this class in the updates.

PR #7053, opened 2018-04-17
Change authenticator bootstrapping to container config

tburry:
We should always be configuring dependencies in the container rather than instantiating them in the bootstrap. This ensures that they are instantiated as late as possible which can also mean that they aren’t instantiated at all if they aren’t needed.


DaazKu:
I've put the addCall on the existing rule and updated unit tests

PR #7066, opened 2018-04-18
Enable tags on Ideas

cchabilall83:
Currently ideas do not allow for tagging.  This update will ideation to use the afterDiscussionFormsOptions event to enable tagging.

See vanilla/internal#1535

### DIFF
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -312,7 +312,7 @@ class VanillaHooks implements Gdn_IPlugin {
             return;
         }
 
-        if (in_array($Sender->RequestMethod, ['discussion', 'editdiscussion', 'question'])) {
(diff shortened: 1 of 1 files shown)


initvector:
It bums me out that adding tagging for a discussion type is like this. We shouldn't have to manually add discussion types and referencing discussion types from addons isn't great. It doesn't look like it's updated often ([the last time was apparently six years ago](https://github.com/vanilla/vanilla/commit/0ac75e34114dc07db486ebcd05f60dc977ec482c)), but I can't think of a good reason for walling off tag-able discussion types with hardcoding.

Before we merge, will you please make a new issue for improving addon access to updating tag-able discussion types?

cchabilall83:
https://github.com/vanilla/internal/issues/1559

It's not fully spec'd out, just gives an overview of the issue

PR #7071, opened 2018-04-19
Add options argument to conversations for group invite emails

cchabilall83:
Currently group invites send and email with a link directly to the message instead of the group homepage. 
 
These changes will allow the conversation to  detect if the message being sent is a group invite and modify link sent in the email to direct the user to the group home page so they can join the group.  An options array is added to save method in the conversationmodel and notifyusers method in the conversationsmodel.  

[Issue#1533](https://github.com/vanilla/internal/issues/1533)


cchabilall83:
Changes made  in #7563

PR #7073, opened 2018-04-19
New styles for @mention

slafleche:
New styles for @mentions in the rich editor.

Related: 
- https://github.com/vanilla/rich-editor/pull/1
- https://github.com/vanilla/roadmap/pull/50


applications/dashboard/scss/editor/_richEditor.scss
```diff
@@ -148,11 +152,11 @@ $embedded-quote_fg: #A3A3A3 !default;
 }
 
 // button reset
+.atMentionList-suggestion,
```
charrondev: Wouldn't it be better to give the suggestion a richEditor-button class, then than to add to the list here?

slafleche: `.richEditor-button` gets other styles that we don't want, like a set width and height since it expects to only contain an SVG.

PR #7074, opened 2018-04-19
Add allowedUrl utility function

charrondev:
This Pr adds an allowed URL function to application. It's in application because in the future we may want to restrict the domains allowed based on configuration settings (Eg. whitelisted, blacklisted URLs). For now it just validates a URL using the `validator.js` library. Essentially its replacing  a kind of long regex that would have to handle a bunch of edge cases and provides some nice configuration options.

Needed for https://github.com/vanilla/rich-editor/pull/4

Essentially right now its a wrapper for a piece of that library with some preset options so that in the future all checks will be centrally located.


src/scripts/application.ts
```diff
@@ -84,6 +85,27 @@ export function translate(str: string, defaultTranslation?: string): string {
...
+ * Determine if a string is an allowed URL.
+ *
+ * In the future this may be extended to check if we want to whitelist/blacklist various URLs.
+ *
+ * @param input - The string to check.
+ */
+export function isAllowedUrl(input: string): boolean {
+    // Options https://github.com/chriso/validator.js#validators
+    const options = {
+        protocols: ["http", "https", "ftp"],
```
slafleche: Do we really want to support `ftp`?

PR #7076, opened 2018-04-19
Group search update

slafleche:
This PR is needed for the upcoming group search feature.


charrondev:
The search is no longer to the right for me with the default theme

image

It seems to be caused by the `width: 100%` on the .pageHeading-main

slafleche:
@charrondev did you pull internal?

PR #7077, opened 2018-04-20
Fix broken sign-out URL in embed comment form

bleistivt:
Current master breaks embedded comment forms when the user is logged in.

     The "Gdn_Auth" object does not have a "xgetAuthenticator" method. 

This PR makes the signout URL creation similar to the signin URL:

https://github.com/bleistivt/vanilla/blob/cee99aa4c0e4440da139f9a498c7481fe64667cf/applications/vanilla/views/discussion/helper_functions.php#L628


DaazKu:
That function is part of the "old" code so it's fine it that works.

Issue #7026, opened 2018-04-12
Rich Editor - Tweak position of paragraph menu


slafleche:
For new discussions, we want the paragraph menu to be in the margin. The core themes will need to be tweaked to accommodate this.

PR #7078, opened 2018-04-20
Tweaked Rich Editor styles for pilcrow positioning

slafleche:
Theses styles affect the gutters of pages with an editor. They are in the editor styles, but affect elements in Vanilla. Not ideal, but this way we don't need to worry about it breaking anyone else's theme without the plugin.

I took the liberty of hiding the user icon next to forms when the Rich Editor is turned on. I would like to always get rid of it, but it's particularly bad with the pilcrow menu. 

Closes: https://github.com/vanilla/vanilla/issues/7026  with related: https://github.com/vanilla/rich-editor/pull/5


applications/dashboard/scss/editor/_richEditor.scss
```diff
@@ -676,3 +677,65 @@ $embedded-quote_fg: #A3A3A3 !default;
...
+                max-width: 100%;
+            }
+
+            .FormWrapper {
+                padding: {
+                    left: $richEditor-pilcrow_gutter;
+                }
+
+                .richEditor {
+                    padding-left: $richEditor-pilcrow_gutter;
```
charrondev: Shouldn't this be removed? It's still padded inwards even with the extra gutter on the side.

image

PR #7093, opened 2018-04-23
GooglePlus: Double url() call

dannimajdalani:
Double URL call causing a problem with Sub-communities

### DIFF
--- a/plugins/GooglePlus/class.googleplus.plugin.php
+++ b/plugins/GooglePlus/class.googleplus.plugin.php
@@ -246,7 +246,7 @@ class GooglePlusPlugin extends Gdn_Plugin {
      */
     public function signInButton($type = 'button') {
         $target = Gdn::request()->post('Target', Gdn::request()->get('Target', url('', '/')));
-        $url = url('entry/googleplusauthredirect?Target='.$target);
+        $url = url('entry/googleplusauthredirect?Target='.$target, true);
 
         $result = socialSignInButton('Google', $url, $type, ['rel' => 'nofollow']);
         return $result;


plugins/GooglePlus/class.googleplus.plugin.php
```diff
@@ -246,7 +246,7 @@ public function setup() {
      */
     public function signInButton($type = 'button') {
         $target = Gdn::request()->post('Target', Gdn::request()->get('Target', url('', '/')));
-        $url = url('entry/googleplusauthredirect?Target='.$target);
+        $url = url('entry/googleplusauthredirect?Target='.$target, true);
```
tburry: If the issue is a double call to url() why not just remove this call?

nervoustwit: I concur.

nervoustwit:
Looks good.

plugins/GooglePlus/class.googleplus.plugin.php
```diff
@@ -246,7 +246,7 @@ public function setup() {
      */
     public function signInButton($type = 'button') {
         $target = Gdn::request()->post('Target', Gdn::request()->get('Target', url('', '/')));
-        $url = url('entry/googleplusauthredirect?Target='.$target);
+        $url = 'entry/googleplusauthredirect?Target='.$target;
```
charrondev: If we're getting rid of the wrapping url function we should rename this variable. It's no longer gaurenteed to be a valid  URL at this point. We should call it something like `$path` or `$destination`.

DaazKu: - Todd [reasoning](https://github.com/vanilla/vanilla/pull/7093#discussion_r183521024) was right but by supplying an absolute URL there's no further call to the `url()` function.

The double encoding happens only because of the `anchor()` function:
```
        $prefix = substr($destination, 0, 7);
        if (!in_array($prefix, ['https:/', 'http://', 'mailto:']) && ($destination != '' || $forceAnchor === false)) {
            $destination = Gdn::request()->url($destination, $withDomain, $sSL);
        }
```
So the first solution should have been merging as is.

- `socialSignInButton()` expects an URL not a path. We do use $url to describe relative paths in multiple places though.

My point here is that resolving this "perfectly" is a question of perspective and I wouldn't block the resolution of this PR because the variable names does not perfectly fit.

charrondev: Ah that's true. `socialSignInButton` definitely should be requesting `$path` instead of `$url` in this case though.

PR #7104, opened 2018-04-24
Add PageInfo class

### LINKED ISSUE #7004: Scrape API should not return data URIs of images
I tried scraping google and the response is this:

```json
{
    "url": "https://www.google.ca/search?source=hp&ei=HCLOWsjRCMXq_AaakJ2wCQ&q=asdfasdf&oq=asdfasdf&gs_l=psy-ab.3..0l7.588.919.0.953.9.4.0.0.0.0.190.190.0j1.1.0....0...1.1.64.psy-ab..8.1.189.0...0.nockmJxxbPU",
    "type": "site",
    "name": "asdfasdf - Google Search",
    "body": "\u00a0\u00a0",
    "photoUrl": "data:image/gif;base64,R0lGODlhAQABAIAAAP///////yH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==",
    "height": null,
    "width": null,
    "attributes": []
}
```

This would also be resolved by only looking at opengraph tags.

### LINKED ISSUE #7030: Create a class with functionality of fetchPageInfo
The capabilities of `fetchPageInfo` need to be refactored into a class.

initvector:
This update adds the `PageInfo` class, which more or less is intended to replace the `fetchPageInfo` function.

1. The `HttpRequest` class from Garden HTTP is used for the requests.
1. `pQuery` has been dropped in favor of DOMDocument.
1. The class only looks for images in OpenGraph tags. No more scouring a page for every random image.
1. The default embed class now uses `PageInfo`, which means the media endpoint is also using it.
1. Tests have been added to verify document processing yields desired results.

Closes #7004
Closes #7030

### DIFF
--- a/library/Vanilla/Embeds/LinkEmbed.php
+++ b/library/Vanilla/Embeds/LinkEmbed.php
@@ -7,16 +7,23 @@
 namespace Vanilla\Embeds;
 
 use Exception;
+use Vanilla\PageInfo;
 
 /**
  * Generic link embed.
  */
 class LinkEmbed extends Embed {
 
+    /** @var PageInfo */
+    private $pageInfo;
+
     /**
      * LinkEmbed constructor.
+     *
+     * @param PageInfo $pageInfo
      */
-    public function __construct() {
+    public function __construct(PageInfo $pageInfo) {
+        $this->pageInfo = $pageInfo;
         parent::__construct('link', 'link');
     }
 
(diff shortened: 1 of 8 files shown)


tburry:
Can we rename this to something more descriptive like WebScraper?

library/Vanilla/WebScraper.php
```diff
@@ -0,0 +1,206 @@
...
+    public function pageInfo(string $url): array {
+        $info = [
+            'Title' => '',
+            'Description' => '',
+            'Images' => []
+        ];
+
+        $response = $this->getUrl($url);
+
+        if ($response->isResponseClass('2xx') === false) {
```
charrondev: Couldn't this just be `(!$response->isResponseClass('2xx'))`?

PR #7112, opened 2018-04-26
Add translation support for "All" and "Following" in category following filter menus

dannimajdalani:
Category-Following drop down filter and Label of "All" & "Following" was not translated, I added the key for 'Following' in [#62](https://github.com/vanilla/locales/pull/62/files), 'All' key is already added.
cat_follow

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -811,7 +811,7 @@ if (!function_exists('filtersDropDown')) {
      *     ** 'param': URL parameter associated with the filter.
      *     ** 'value': A value for the URL parameter.
      * @param string $extraClasses any extra classes you add to the drop down
-     * @param string $default The default label for when no filter is active.
+     * @param string|null $default The default label for when no filter is active. If `null`, the default label is "All".
      * @param string|null $defaultURL URL override to return to the default, unfiltered state.
      * @param string $label Text for the label to attach to the cont
      * @return string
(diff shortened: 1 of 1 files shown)


initvector:
Why'd you choose to translate everything in the `linkDropDown` function, instead of specifically targeting "All" and "Following" higher up, like in `categoryFilters`, `discussionFilters` and `filtersDropDown`?

I believe functions/methods in Vanilla *usually* assume incoming strings are already translated, having been translated before the call to that function/method is made (whatever creates the original text is responsible for translating it before handing it off).  `linkDropDown` is also general-use function ("Write a link drop down control.") and not tied to category following, specifically. Modifying that to translate everything, instead of targeting the specific strings cited in the original issue, may be casting the net too wide and have unintended side effects or cause already-translated strings to be "translated" again.

dannimajdalani:
Related [62](https://github.com/vanilla/locales/pull/62)

library/core/functions.render.php
```diff
@@ -858,7 +858,7 @@ function filtersDropDown($baseUrl, array $filters = [], $extraClasses = '', $def
             // Add the default link to the top of the list.
             array_unshift($links, [
                 'active' => $active === null,
-                'name' => $default,
+                'name' => t($default),
```
initvector: I would update the default to `null` (instead of "All") and translate it at the beginning of the function. That way, we're only translating text generated in this function.

```php
function filtersDropDown($baseUrl, array $filters = [], $extraClasses = '', $default = null, $defaultUrl = null, $label = 'View') {
    if ($default === null) {
        $default = t('All');
    }
```

PR #7120, opened 2018-04-26
Add getData event to TagModule

cchabilall83:
In order to allow addons to customize which tags are being displayed, we have to allow them to hook in and modify the data.

### DIFF
--- a/applications/vanilla/modules/class.tagmodule.php
+++ b/applications/vanilla/modules/class.tagmodule.php
@@ -103,14 +103,18 @@ class TagModule extends Gdn_Module {
 
         $this->autoContext();
 
-        $discussion = DiscussionModel::instance()->getID($this->ParentID);
+        $this->EventArguments['tags'] = TagModel::instance()->getDiscussionTags($this->ParentID, false);
+        $this->EventArguments['_TagData'] = &$this->_TagData;
+        $this->fireEvent('getData');
+
+        if($this->_TagData) {
+            return;
+        }
+
         $tagCacheKey = "TagModule-{$this->ParentType}-{$this->ParentID}";
         switch ($this->ParentType) {
             case 'Discussion':
                 $tags = TagModel::instance()->getDiscussionTags($this->ParentID, false);
-                if ($discussion->Type == 'Idea') {
-                    $tags = $this->filterIdeaStatusTags($tags);
-                }
                 break;
             case 'Category':
                 $tagQuery->join('TagDiscussion td', 't.TagID = td.TagID')
@@ -156,21 +160,6 @@ class TagModule extends Gdn_Module {
      */
     public function assetTarget() {
         return 'Panel';
-    }
-
-    public function filterIdeaStatusTags($tags) {
-
-        $discussionTagIDs = array_column($tags,'TagID');
-
-        $filteredTags = Gdn::sql()->select('*')
-            ->from('Tag')
-            ->whereIn('TagID', $discussionTagIDs)
-            ->where('Type <>', 'Status')
-            ->orderBy('CountDiscussions', 'desc')
-            ->get()
-            ->resultArray();
-
-        return $filteredTags;
     }
 
     /**
--- a/applications/vanilla/modules/class.tagmodule.php
+++ b/applications/vanilla/modules/class.tagmodule.php
@@ -103,10 +103,9 @@ class TagModule extends Gdn_Module {
 
         $this->autoContext();
 
-        $this->EventArguments['tags'] = TagModel::instance()->getDiscussionTags($this->ParentID, false);
+        $this->EventArguments['id'] = $this->ParentID;
         $this->EventArguments['_TagData'] = &$this->_TagData;
         $this->fireEvent('getData');
-
         if($this->_TagData) {
             return;
         }
--- a/applications/vanilla/modules/class.tagmodule.php
+++ b/applications/vanilla/modules/class.tagmodule.php
@@ -102,7 +102,7 @@ class TagModule extends Gdn_Module {
         $tagQuery = Gdn::sql();
 
         $this->autoContext();
-
+        //Allow addon to manipulate the data being rendered
         $tagData = null;
         $this->EventArguments['ParentID'] = $this->ParentID;
         $this->EventArguments['tagData'] = &$tagData;
--- a/applications/vanilla/modules/class.tagmodule.php
+++ b/applications/vanilla/modules/class.tagmodule.php
@@ -102,7 +102,7 @@ class TagModule extends Gdn_Module {
         $tagQuery = Gdn::sql();
 
         $this->autoContext();
-        //Allow addon to manipulate the data being rendered
+        // Allow addon to manipulate the data being rendered.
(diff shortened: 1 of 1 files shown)


applications/vanilla/modules/class.tagmodule.php
```diff
@@ -103,6 +103,13 @@ public function getData() {
 
         $this->autoContext();
 
+        $this->EventArguments['id'] = $this->ParentID;
```
initvector: For the sake of simplicity, I'd keep these names the same. You'll know the ID maps to the ParentID of the tags and not an ID of a tag.

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -103,6 +103,13 @@ public function getData() {
...
+        $this->EventArguments['_TagData'] = &$this->_TagData;
```
initvector: I would avoid attempting to give addons a backdoor to directly alter a protected property on the class. You can set a local tag data variable to a null value and pass that to the event. Addons hooking in can check to see if the value is still null before acting, so they don't overwrite each other's data. After all the handlers have been run through, `TagModule::getData` can check to see if addons populated the local variable and set its `_TagData` property as necessary.

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -103,6 +103,13 @@ public function getData() {
...
+        $this->fireEvent('getData');
+        if($this->_TagData) {
+            return;
```
initvector: This method has some pretty specific criteria for its data. It should be an instance of `Gdn_DataSet` with a type of `DATASET_TYPE_ARRAY`. Data populated as the result of the getData event should be validated and possibly converted, as necessary, so we're sure what the module is expecting for its data is actually what it's getting.

initvector:
I updated the title and description to be a little more generic and not so specific to Ideation.

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -103,6 +103,15 @@ public function getData() {
 
         $this->autoContext();
 
+        $tagData = null;
+        $this->EventArguments['ParentID'] = $this->ParentID;
```
initvector: ParentType is available and used by this method, so it's probably worth passing along too.

cchabilall83:
@initvector this is also related to [PR](https://github.com/vanilla/internal/pull/1552).

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -103,6 +103,15 @@ public function getData() {
...
+        $this->EventArguments['tagData'] = &$tagData;
+        $this->fireEvent('getData');
+        if(is_array($tagData) && $tagData) {
```
linc: Coding standard nitpick: add a space after `if`.

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -103,6 +103,15 @@ public function getData() {
 
         $this->autoContext();
```
linc: Somewhere in here I feel like an inline comment would be appropriate about why this change is being made. "// Allow an addon to hijack the setting of our data, which aborts the normal database queries." or something to that effect.

applications/vanilla/modules/class.tagmodule.php
initvector: What if no tags are supposed to be displayed (e.g. the only tag is a "status" tag), so `$tagData` is an empty array (falsy)? In its current state, wouldn't the code continue on to the standard routine?

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -102,6 +102,16 @@ public function getData() {
         $tagQuery = Gdn::sql();
 
         $this->autoContext();
+        //Allow addon to manipulate the data being rendered
+        $tagData = null;
+        $this->EventArguments['ParentID'] = $this->ParentID;
+        $this->EventArguments['tagData'] = &$tagData;
+        $this->fireEvent('getData');
+        if ($tagData !== null) {
```
initvector: The expectation is that, after this method is run, *something* will be in the `_TagData` property. The logic here doesn't look like it will allow that. First, we're checking to see if it's `null`. That's good. We're attempting to confirm an addon modified the value.  Next, we're making sure the value is an array. If it is, we're updating the `_TagData` property with the value and exiting the method. If it isn't an array, we're just exiting the method. `_TagData` is never updated with anything.

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -102,6 +102,16 @@ public function getData() {
...
+            if (is_array($tagData))
+            $this->_TagData = new Gdn_DataSet($tagData, DATASET_TYPE_ARRAY);
```
initvector: Something about this formatting seems off...

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -102,6 +102,16 @@ public function getData() {
         $tagQuery = Gdn::sql();
 
         $this->autoContext();
+        //Allow addon to manipulate the data being rendered
+        $tagData = null;
+        $this->EventArguments['ParentID'] = $this->ParentID;
+        $this->EventArguments['tagData'] = &$tagData;
```
initvector: Please see https://github.com/vanilla/vanilla/pull/7120#discussion_r185049470

applications/vanilla/modules/class.tagmodule.php
initvector: That definitely wasn't the case [when I commented on it](https://github.com/vanilla/vanilla/blob/22718c1bba70a1df954bea0959a42890ed3a039f/applications/vanilla/modules/class.tagmodule.php#L110).

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -102,6 +102,18 @@ public function getData() {
         $tagQuery = Gdn::sql();
 
         $this->autoContext();
+        //Allow addon to manipulate the data being rendered
+        $tagData = null;
+        $this->EventArguments['ParentID'] = $this->ParentID;
+        $this->EventArguments['ParentID'] = $this->ParentType;
```
initvector: Why are you setting ParentID twice?

applications/vanilla/modules/class.tagmodule.php
```diff
@@ -102,6 +102,18 @@ public function getData() {
...
+        $this->EventArguments['tagData'] = &$tagData;
+        $this->fireEvent('getData');
+        if ($tagData !== null) {
+            if (is_array($tagData)) {
```
initvector: At this point, since we're checking if it's not null, then checking if it's an array, couldn't we just check to see if it's an array (i.e. drop the is-null condition and bring out the `is_array` check)? If there's a reason to keep the null check, couldn't you combine the condition from the inner block?

cchabilall83: I think we drop the is_null check, but I don't think it's necessary

PR #7142, opened 2018-04-30
Add TwitterEmbed class

### LINKED ISSUE #7131: Add TwitterEmbed class
Twitter needs an server-side embed class. It will use Twitter cards, so Twitter script(s) will need to be included. See [`YouTubeEmbed`](https://github.com/vanilla/vanilla/blob/master/library/Vanilla/Embeds/YouTubeEmbed.php).

initvector:
This update adds support for scraping Twitter URLs for card data.

1. Currently, only status URLs are supported, similar to what's seen in `Gdn_Format`.
1. Twitter's oEmbed responses are pretty limited. For now, I don't think they're necessary for what Vanilla will be doing, so I'm not including them.
1. No scripts are presently used. When the client-side embed strategy is more fleshed out, they can be added.
1. Tests have been included for rendering markup. These are based on what's currently in `Gdn_Format` and will be updated, as necessary.
1. The media API v2 endpoint's scrape/embed type is now dynamic.

Closes #7131


charrondev:
This looks good. If you just fix the merge conflict I'll approve it.

PR #7144, opened 2018-04-30
add translations to size and format error messages

igraziatto:
Issue https://github.com/vanillaforums/immediate-media/issues/3

Not backported yet.


igraziatto:
I have fixed the conditional and made a PR on locales https://github.com/vanilla/locales/pull/63

PR #7147, opened 2018-05-01
[Backport] Add translations to size and format error messages

igraziatto:
Issue: https://github.com/vanillaforums/immediate-media/issues/3
Original PRs:  https://github.com/vanilla/vanilla/pull/7144, #7152

### DIFF
--- a/cache/.htaccess
+++ b/cache/.htaccess
@@ -1,0 +1,1 @@
+Deny from all


cache/.htaccess
```diff
@@ -1 +0,0 @@
-Deny from all
```
charrondev: We need to make sure this is deleted and put it back on master branch as well.

DaazKu: Why did we do that?

PR #7148, opened 2018-05-01
Fix username validation when signing in with SSO

### LINKED ISSUE #5866: Manually entering a username while connecting can dodge requirements
If connecting via SSO and you need to manually enter a username, it's possible the username validation and requirement restrictions may not be applied. This might be possible with other SSO addons, but I've experienced it specifically with jsConnect.

Steps to reproduce:

1. Enable autoconnect on a site.
1. Attempt to connect a new account, using a name that already exists on the site.
1. You'll be prompted to enter the password for the existing user or create a new name.
1. If you enter a new name, you can enter pretty much whatever you want. It appears to skip all validation.

This seems to be due to the following:
1. [`EntryController::connect` asking SSO addons to interpret and modify the incoming data.](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/dashboard/controllers/class.entrycontroller.php#L438)
1. [Addons, like jsConnect, use this event to inject the Name and Email fields.](https://github.com/vanilla/addons/blob/8725d2c1f080f6aa8a8bf599034281864f2857ca/plugins/jsconnect/class.jsconnect.plugin.php#L420)
1. [`EntryController::connect` performs some connection logic, based on the presence of the Name or Email fields.](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/dashboard/controllers/class.entrycontroller.php#L558)
1. [This ultimately trickles down to a condition where creating a user or not is determined.](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/dashboard/controllers/class.entrycontroller.php#L714)
1. [A user is potentially created based on ConnectName (the field the user entered), although it's never properly validated.](https://github.com/vanilla/vanilla/blob/41dd256854a93da28bc5f5b0274bea1c3fb6b646/applications/dashboard/controllers/class.entrycontroller.php#L729)

initvector:
SSO has been updated to utilize `UserModel` updates introduced in #7163.

1. If the username comes from a trusted authentication provider, it must not be a blacklisted name.
1. If the username comes from an untrusted authentication provider, or if the username comes from a user during connection, it must not be blacklisted __*and*__ it must match the regex validation pattern.

Tests and some addons have been updated, as necessary, to generate default valid usernames.

*Warning: At 500 lines, `EntryController::connect` is a complex, delicate beast. It's not something to be taken lightly. Please take your time in carefully reviewing the code.*

### Testing

This can have a significant impact on SSO. Please test, throughly. I've primarily tested with social connect addons and jsConnect. I recommend testing with as many configurations as you can think of.

Here's a few scenarios to test with trusted __*and*__ untrusted providers:

1. Connecting with no username from the authentication provider.
1. Connecting with an existing username, providing a new name.
1. Connecting with an existing username, connecting to that user with its password.
1. Connecting with a blacklisted username.
1. Connecting with an invalid username (e.g. one that does not match the username validation regex)

Closes #5866

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -477,6 +477,16 @@ class EntryController extends Gdn_Controller {
             }
         }
 
+        // Validate username fields, in order of relevancy.
+        foreach (['ConnectName', 'Name'] as $usernameField) {
+            if ($this->Form->getFormValue($usernameField, null) !== null) {
+                $this->Form->validateRule($usernameField, 'function:validateAgainstUsernameBlacklist');
+                $this->Form->validateRule($usernameField, 'ValidateUsername');
+                // Only validate the most relevant field.
+                break;
+            }
+        }
+
         // Make sure the minimum required data has been provided by the connection.
         if (!$this->Form->getFormValue('Provider')) {
             $this->Form->addError('ValidateRequired', t('Provider'));
@@ -806,11 +816,6 @@ class EntryController extends Gdn_Controller {
                     if (c('Garden.Registration.NameUnique')) {
                         // Check to see if there is already a user with the given name.
                         $user = $userModel->getWhere(['Name' => $connectName])->firstRow(DATASET_TYPE_ARRAY);
-                    }
-
-                    if (!$user) {
-                        // Using a new username, so validate it.
-                        $this->Form->validateRule('ConnectName', 'ValidateUsername');
                     }
                 }
             } else {
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -477,16 +477,6 @@ class EntryController extends Gdn_Controller {
             }
         }
 
-        // Validate username fields, in order of relevancy.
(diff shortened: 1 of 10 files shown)


DaazKu:
> This includes information communicated by the provider, as well as user-provided values

Can we have that vetoed and documented somewhere? I'm working on the unit tests for SSOModel and I thought that information from **trusted providers** should bypass validation and user submitted information should be validated.

DaazKu:
Also, I'm gonna hijack the review of this issue.

initvector:
> Also, I'm gonna hijack the review of this issue.

The more the merrier! But I want @nervoustwit to take a look too.

DaazKu:
Here is how I thought it worked:

- Data submitted by the Provider
    - Provider is Trusted -> bypass validation
    - Provider is Not Trusted -> validate data
- Data submitted by the User -> validate data

initvector:
I'm onboard with that logic if that's the general consensus, at least in regard to basic validation of the username (e.g. regex pattern matching). That said, I think blacklisted usernames shouldn't be allowed, regardless of the point of origin.

PR #7155, opened 2018-05-02
Allow Gdn_Dispatcher to dispatch to dash-cased controllers

### LINKED ISSUE #7109: Allow dash-cased controllers and methods in Gdn_Dispatcher
The new dispatcher converts dash-cased URL parts to CapitalCased controllers and camelCased methods. Trivially, this is done by removing dashes from the URL parts during dispatch, however there are a few considerations here:

1. SEO has to be considered. We can allow every combination of dashes to work in URLs. We Should try and check against the actual casing of the controllers/methods to make sure the dashes always precede a capital letter or number.
2. For backward compatibility, paths without dashes should always match. To avoid SEO concerns we should make sure the canonical URL is now set with dashes. It is mu suspicion that the controller that properties used to calculate the default canonical URL are not properly cased which will need to be fixed and tested. Look through our main controllers (CategoryController, DiscussionsController, DiscussionController) for examples of camel case URLs and make sure they work. Try and think of other URLs that may have an issue.

It's possible that #2 will be enough of a solution since proper canonical URLs will prevent SEO issues. If you wish you can at lease make sure the following dashed URL parts never match:

- Paths with dashes at the beginning or the end of the string (ex. /-discussions).
- Paths with multiple dashes in a row (ex. /discussions/bookmarked--popin).

tburry:
This PR allows us to dispatch to controllers/methods that contain dashes in their URL. That change itself is fairly trivial, but there is a bit of work handling canonical URLs when the dashes are not properly added.

This breaks backwards compatibility with some canonical URLs, but I think the new ones are more robust. This is just for default canonical URLs, not ones that are explicitly set.

I've grabbed the old default canonical URL code for comparison. You can browse around with the debug on and look for notices where the old and new canonical URLs differ. We want to ensure that our "main" canonical URLs are the same while other pages are fine.

Closes #7109.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -205,7 +205,7 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
      * @return string Returns the filtered name.
      */
     private function filterName($name) {
-        if (empty($name) || $name[0] === '-' || substr($name, -1) === '-') {
+        if (empty($name) || $name[0] === '-' || substr($name, -1) === '-' || preg_match('`--`', $name)) {
             return $name;
         }
 
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -204,7 +204,7 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
      * @param string $name The name to convert.
      * @return string Returns the filtered name.
      */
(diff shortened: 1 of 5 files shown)


tburry:
Forgot to remove the WIP so I punted until next sprint.

PR #7175, opened 2018-05-04
Add APIv2 smart ID support

### LINKED ISSUE #7108: Implement smart ID middleware for API v2
There is documentation on how smart IDs are formatted in the roadmap.

The smart IDs should be implemented as middleware that alters the request before it makes it to the API.

tburry:
Add smart ID support to the API v2. Go to the blocking roadmap issue for a description of how it works.

- I tried to make the middleware extensible without exploding the number of classes we have to have. To this end I allow basic smart ID resolvers that do simple DB lookup and callback resolvers that can do whatever they want.

- I opted not to inject models into the middleware because I don't want every model hanging around for every request. In the future we may want to explore some sort of model factory that allows us to map record types to models since we have many kludges throughout our app for this.

- User ID smart IDs do some basic security checking on email addresses. I opted not to do an extra check on whether authenticators exist or are enabled. If an authenticator does not exist then there is a generic not found error so the single query is functionally equivalent. In terms of an authenticator being enabled/disabled I'm not actually sure we should be looking at this for smart IDs. In any case I want to wait for the new authentication to be complete before doing something more sophisticated.

Closes #7108.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -151,6 +151,14 @@ $dic->setInstance('Garden\Container\Container', $dic)
         return new \Garden\Web\PreflightRoute('/api/v2', true);
     })])
     ->addCall('setAllowedOrigins', ['isTrustedDomain'])
+    ->addCall('addMiddleware', [new Reference('@smart-id-middleware')])
+
+    ->rule('@smart-id-middleware')
+    ->setClass(\Vanilla\Web\SmartIDMiddleware::class)
+    ->setConstructorArgs(['/api/v2/'])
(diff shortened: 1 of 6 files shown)


tburry:
A note here that the smart ID middleware and user ID resolvers are reasonably well unit tested. I did not unit test the integration with Vanilla though which means I may have got a database wrong. I did some basic QA with the three core smart ID resources though.

tburry:
I had to change the core dispatcher middleware implementation a bit. Middleware applied to the dispatcher could not throw exceptions which is not optimal so I put an exception handler around the call.

PR #7181, opened 2018-05-07
Format drafts as plaintext instead of text

### LINKED ISSUE #6690: Rich Editor - Handle plaintext formatting
Implement the Gdn_Format::plaintext for the quill renderer.

tburry:
Gdn_Format::text() is meant to work on HTML, not any old format.

Closes #6690.


tburry:
Not that this is not the exact implementation the issue expects, but I figured we could get to that later if we want the optimization and this is still the correct fix for drafts.

Issue #7122, opened 2018-04-26
Integrate new embed manager with the Quill renderer


charrondev:
- Use the HTML from the Embed inside of the Quill embed renderer.
- Remove duplicated/now unnecessary embed blots from Quill renderer.
- Ensure that the new embeds return the proper HTML structure.
- Ensure that the following existing embeds all work
  - Link
  - Video (Youtube, Vimeo minimum)
  - Image

PR #7187, opened 2018-05-08
Rework client and server side embeds

charrondev:
Goes with https://github.com/vanilla/rich-editor/pull/15
Closes https://github.com/vanilla/vanilla/issues/7122

This PR reworks the existing client side embeds.

## Frontend
- Adds utility functions for
  - ensuring a script is present and loaded in the head of the document.
- Moves the following utility functions
  - Emoji fallback utilities into `@core/dom`
  - Simplify fraction utility from the rich editor plugin into `@core/utility`.
- Makes CSS changes for in-editor embeds.
- Refactors and simplifies the FileUploader class so that it works with the new Promise based/async rendering approach in the rich editor.
- Moves existing new user content embed rendering functions into `@core/app/user-content/embeds` and moves the existing user content twitter rendering function from `global.js` as well. This is currently shared, although we may alter in the future if/when we update all of the old embeds to use the EmbedManager.

## Backend
- Add server-side External Embed Quill blot for rendering content generated using the media-scrape endpoint.
  - Integrates with Embed Manager to render content
- Add abstract `VideoEmbed` class to make it easier to create video embeds in the future
  - Includes utility methods and code generation for generic video embed usage
- Add getter for head module in Gdn_Controller (directly accessing the property made me feel dirty)
- ~Add EmbedManager::getScripts~ Removed in favor of a purely client-side solution.
- ~Add ExternalBlot::addScripts~ Removed in favor of a purely client-side solution.


library/Vanilla/Quill/Blots/Embeds/ExternalBlot.php
```diff
@@ -0,0 +1,74 @@
...
+        /** @var EmbedManager embedManager */
+        $this->embedManager = Gdn::getContainer()->get(EmbedManager::class);
+    }
+
+    /**
+     * Add client-side scripts to the current controller.
+     *
+     * @param array $scripts
+     */
+    private function addScripts(array $scripts) {
```
charrondev: We should be able to get rid of this entirely. These scripts are all being lazy loaded by the client.

library/Vanilla/Quill/Parser.php
```diff
@@ -30,6 +30,7 @@ class Parser {
         Blots\BulletedListBlot::class,
         Blots\OrderedListBlot::class,
         Blots\Embeds\EmojiBlot::class,
+        Blots\Embeds\ExternalBlot::class,
         Blots\Embeds\ImageBlot::class,
```
charrondev: We should be able to remove all of the block embed blots after this one.

charrondev:
So a few things that still need resolving here:

- [ ] The video embed blots are displaying the old HTML. We will need to update this.
  - Move the URL generation into the match request so the generated embed video URL comes down in the API response (I'll then update the client side blots so that the videos work in the editor.
  - Use the new HTML structure from the VideoBlot.
- [ ] Remove unnecessary blots (supplanted by the embed manager).
  - VideoBlot
  - LinkEmbedBlot
  - ImageBlot
- [ ] Resolve an issue with block level formats being applied the ExternalBlot (and creating fatal errors).

PR #7198, opened 2018-05-09
Add nofollow to Quill links

initvector:
This update:

1. Adds a `rel` attribute to the anchor tags generated by the link rendering blot and sets its value to "nofollow".
1. Ditches the `target` attribute.

Tests have been updated, accordingly.


charrondev:
Actually I don't think we're supposed to have target blank at all on links from the editor. Currently target blank is completely banned unless a certain config settings is on, and in that case it's not inserted automatically, it just "allowed" if a user put it themselves in an HTML type of insert, in which case noopener would be also added.

PR #7218, opened 2018-05-11
Add default parameters to Gdn_ErrorException and fix cleanErrorArguments()

tburry:
With PHP 7.1 functions can no longer be called with too few arguments. This was causing issues with the Gdn_ErrorException constructor. I copied the defaults of its base class to help with this.

While fixing this issue I uncovered an issue with the cleanErrorArguments() function. It has no protection for infinite recursion and also doesn’t clean the security blacklist on object properties. I fixed this, but decided to make another version because I wanted a no side-effects version of the function.

### DIFF
--- a/library/core/functions.error.php
+++ b/library/core/functions.error.php
@@ -9,7 +9,9 @@
  */
 
 /**
- * Class Gdn_ErrorException
+ * This class is for internal use within the global error handler only. Don't use this class elsewhere.
+ *
+ * An extension of the **ErrorException** that includes context variables.
  */
 class Gdn_ErrorException extends ErrorException {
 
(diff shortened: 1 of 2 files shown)


library/core/functions.error.php
```diff
@@ -692,6 +695,49 @@ function cleanErrorArguments(&$var, $blackList = ['configuration', 'config', 'da
...
+    /**
+     * This is an internal function not to be used outside of error printing.
+     *
+     * @param $var
+     * @param array $blackList
+     */
+    function __cleanErrorArguments($var, $blackList = ['configuration', 'config', 'database', 'password']) {
+        $seen = [];
+
+        $fn = function ($var, int $nest = 0) use (&$fn, $blackList, &$seen) {
```
charrondev: Do we normally allow variable shadowing like this? It makes things a little bit harder to follow?

PR #7220, opened 2018-05-11
Fix Category following filter displaying on QnA plugin

cchabilall83:
fixes vanilla/addons#629

The category-following filter is currently by default true and with and  the option to opt-out.  This allows for the filter to be displayed on views that it shouldn't.  The filter should only be displayed on the category index and recent discussions pages.

This fix checks if category following is enabled and then allows the view to opt in or out accordingly.  In order to opt-in, the SelfUrl will have to be added to the $categoryFilterList.  The DiscussionsController::setCategoryFollowingList can be used to set any request to opt-in.

This approach was taken because some plugins call the DiscussionsController::index method (ie. QnA ) and there was no clear way to distinguish whether the filter should be displayed (opted into) besides the requested url to the controller.

I've tested this using the QnA, Resolved, Participated, Ideation and Reaction plugins.  Subcommunities were tested as well.

### DIFF
--- a/applications/vanilla/controllers/class.discussionscontroller.php
+++ b/applications/vanilla/controllers/class.discussionscontroller.php
@@ -31,7 +31,7 @@ class DiscussionsController extends VanillaController {
     protected $categoryIDs;
 
     /** @var boolean Value indicating if the category-following filter should be displayed when rendering a view */
-    public $enableFollowingFilter;
+    public $enableFollowingFilter = false;
 
     /** @var array list of pages that opt in to the category following filter */
     private $categoryFilterList = ["discussions"];
@@ -351,7 +351,6 @@ class DiscussionsController extends VanillaController {
     public function initialize() {
         parent::initialize();
         $this->ShowOptions = true;
-        $this->enableFollowingFilter = false;
         $this->Menu->highlightRoute('/discussions');
         $this->addJsFile('discussions.js');
 
@@ -856,7 +855,7 @@ class DiscussionsController extends VanillaController {
      *
      * @param string $url
      */
-    public function setCategoryFollowingList(string $url = null) {
+    public function addCategoryFollowingUrl(string $url) {
         if ($url) {
             array_push($this->categoryFilterList, $url);
         }
--- a/applications/vanilla/controllers/class.discussionscontroller.php
+++ b/applications/vanilla/controllers/class.discussionscontroller.php
@@ -33,6 +33,9 @@ class DiscussionsController extends VanillaController {
     /** @var boolean Value indicating if the category-following filter should be displayed when rendering a view */
     public $enableFollowingFilter;
 
+    /** @var array list of pages that opt in to the category following filter */
+    private $categoryFilterList = ["discussions"];
+
 
     /**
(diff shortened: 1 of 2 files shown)


cchabilall83:
After review we will go with a different approach to address this issue.
We will use the current path of the page to determine if it should opt-into displaying the filter.  There will be a whitelist property on the discussions controller and setter method so that any page that wants to opt into the filter and be added to the whitelist.  The index method with the validate against the the whitelist.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -841,4 +850,15 @@ public function tagged() {
         $this->View = c('Vanilla.Discussions.Layout') == 'table' && $this->SyndicationMethod == SYNDICATION_NONE ? 'table' : 'index';
         $this->render($this->View, 'discussions', 'vanilla');
     }
+
+    /**
+     * Allows pages to opt-in to the category following filter.
+     *
+     * @param string $url
+     */
+    public function setCategoryFollowingList(string $url = null) {
```
initvector: This name is kinda misleading. The word "set" makes me think I'm overwriting whatever value might currently be there.

initvector: Why'd you make `$url` optional? It seems like you'd want to make that required, since all this function does is take the value you give it and load it into the array.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -343,6 +351,7 @@ public function unread($page = '0') {
     public function initialize() {
         parent::initialize();
         $this->ShowOptions = true;
+        $this->enableFollowingFilter = false;
```
initvector: Could you assign this default when declaring the property?

linc:
This appears to have been introduced as a problem about two years ago: https://github.com/vanilla/vanilla/pull/4276

That PR references https://github.com/vanilla/vanilla/issues/4275, which is an unrelated problem of the dropdown not appearing at all under certain conditions (a new discussion in a category of type "Discussions"). Therefore, I think this subtree behavior was indeed accidental and not necessarily part of the fix. Let's confirm this PR works correctlt in a category of type "Discussions" and type "Nested", then we can call this case-closed and merge it.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -131,10 +136,13 @@ public function index($Page = false) {
                 null,
                 Gdn::request()->get('save')
             );
+            if (in_array($this->SelfUrl, $this->categoryFilterList)) {
```
linc: `// We need to check against a whitelist due to addons calling this method ad hoc.`

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -30,6 +30,12 @@ class DiscussionsController extends VanillaController {
     /** @var array Limit the discussions to just this list of categories, checked for view permission. */
     protected $categoryIDs;
 
+    /** @var boolean Value indicating if the category-following filter should be displayed when rendering a view */
```
linc: `Whether to show the category following filter.`

PR #7228, opened 2018-05-16
Editor: updated mimetypes list

dannimajdalani:
Added mimetypes for  "psd","swf", "doc", "xls", "ppt", "rar",


dannimajdalani:
#7229

DaazKu:
swf is potentially dangerous.
I don't know for the others.

DaazKu:
Also, where were those extra mime type taken from??!?

Not saying these are wrong but I lookup to httpd mimes types when I want to find something usually. https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

PR #7230, opened 2018-05-16
Add validation filters and the BodyFormatValidator class

tburry:
Several improvements to data validation are included. These include:

1. Refactor the `Gdn_Validation` class. One of the key improvements is allowing validation rule callbacks to update the value being validated. This can be done to improve formatting, alter the value's type and more.
1. Add the `BodyFormatValidator` class, which is used to validate a specific format. Currently, only rudimentary support for validating the "rich" format is provided. Instances of this class may be called as a function.
1. Add support for automatically using `BodyFormatValidator` from `Gdn_Validation`, if configured as an available rule.
1. Add `Gdn_Validation` to the container, configuring it to use `BodyFormatValidator`.
1. Add `Gdn_Validation` parameters to the constructors of several models for the sake of dependency injection with an instance preconfigured to use the `BodyFormatValidator` class.
1. Add the `Vanilla\Invalid` class to allow for more explicit "invalid" validation results.
1. Add code to make `validateTime` a useful function. Prior to this, the function had only been a placeholder in Vanilla. Now, it is a function that can validate a time string.
1. Add tests for verifying validation.

### DIFF
--- a/applications/dashboard/models/class.activitymodel.php
+++ b/applications/dashboard/models/class.activitymodel.php
@@ -61,9 +61,11 @@ class ActivityModel extends Gdn_Model {
 
     /**
      * Defines the related database table name.
+     *
+     * @param Gdn_Validation $validation The validation dependency.
      */
-    public function __construct() {
-        parent::__construct('Activity');
+    public function __construct(Gdn_Validation $validation = null) {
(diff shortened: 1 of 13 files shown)


linc:
This PR requires an issue or description to explain its purpose.

PR #7232, opened 2018-05-17
Improve README instructions for upgrades and installs for 2.6

linc:
Update our installation and migration instructions to break out per-version instructions better and provide more obvious linking to other sections.

Addresses feedback from @cdepage and this open community discussion: https://open.vanillaforums.com/discussion/36260

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,9 @@ Please consider using [maintenance mode](#using-maintenance-mode) before running
 
 1. Backup your database, `.htaccess` and `conf/config.php` file somewhere safe.
 1. Upload the new release's files so they overwrite the old ones.
-1. Delete all files in `/cache`.
+1. Delete all files in `/cache` (except `.htaccess` if you use Apache).
 1. Follow all version-specific instructions below. It is **critcal** you delete the listed files.
-1. Go to `example.com/utility/update` to run any database updates needed. (404? See next paragraph.) If it fails, try it a second times by refreshing the page.
+1. Go to `example.com/utility/update` to run any database updates needed. (404? See next paragraph.) If it fails, try it a second time by refreshing the page.
 
 If you run into a problem, see [Getting Help](#getting-help) below.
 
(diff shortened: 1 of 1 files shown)


linc:
Gonna backport this to the 2.6 branch for the next release toooooooo

README.md
```diff
@@ -70,42 +70,75 @@ by default for all smartphones & tablets. Heck, it even works on the PlayStation
...
-Follow these steps to upgrade Vanilla when a new stable release is announced.
+Follow these steps to upgrade Vanilla when a new stable release is announced. These instructions assume you're using SFTP to manually copy files to a server.
+
+Please consider using [maintenance mode](#using-maintenance-mode) before running database updates if your database is very large (millions of users or comments).
+
+1. Backup your database, `.htaccess` and `conf/config.php` file somewhere safe.
+1. Upload the new release's files so they overwrite the old ones.
+1. Delete all files in `/cache`.
+1. Follow all version-specific instructions below. It is **critcal** you delete the listed files.
+1. Go to `example.com/utility/update` to run any database updates needed. (404? See next paragraph.) If it fails, try it a second times by refreshing the page.
```
charrondev: second times -> second time

README.md
```diff
@@ -70,42 +70,75 @@ by default for all smartphones & tablets. Heck, it even works on the PlayStation
...
 ## Upgrading
 
-Follow these steps to upgrade Vanilla when a new stable release is announced.
+Follow these steps to upgrade Vanilla when a new stable release is announced. These instructions assume you're using SFTP to manually copy files to a server.
+
+Please consider using [maintenance mode](#using-maintenance-mode) before running database updates if your database is very large (millions of users or comments).
+
+1. Backup your database, `.htaccess` and `conf/config.php` file somewhere safe.
+1. Upload the new release's files so they overwrite the old ones.
+1. Delete all files in `/cache`.
```
charrondev: I don't if we have a better solution to this problem, but if deleting all of the files in `/cache` includes deleting an `.htaccess` file currently. If you're on git, it comes up as a diff, and could prevent people from doing a `git pull` in the future without additional interventions.

Worse is that people running on apache with `.htaccess` are then no-longer denying access to the cache folder.

README.md
```diff
@@ -70,42 +70,75 @@ by default for all smartphones & tablets. Heck, it even works on the PlayStation
...
 
-* Backup your database, `.htaccess` and `conf/config.php` file somewhere safe.
-* Upload the new release's files so they overwrite the old ones.
-* Delete all files in `/cache`.
-* Go to `yourforum.com/utility/update` to run any database updates needed. (404? See next paragraph.) If it fails, try it a second times by refreshing the page.
-* If any "Deleted Files" are listed on the upgrade page, manually delete the listed files.
+If your forum still uses URLs including `?p=`, support for this URL structure has ended. Follow these steps to switch to the simpler format:
 
-If your forum still uses URLs including `?p=`:
+1. Confirm your server is setup to handle rewrites. On Apache, using the `.htaccess` file provided will accomplish this. Additional setup is required on nginx and other platforms.
```
charrondev: This should link to our nginx docs as well.

README.md
linc: I added a note at the top that these instructions are for copy/pasting files, not a git-based setup, fwiw. There's like 8 other things wrong with these instructions if you're using git, nevermind that you shouldn't be using git directly in a web directory.

charrondev: That's true, but this still could open servers up to allow people access the cache directory.

Either that `.htaccess` is useless, and we should get rid of it, or it's important and they shouldn't delete it.

PR #7233, opened 2018-05-18
Changing file_get_contents in getProfile() to cURL

zlorkovic:
As cURL is required by Vanilla and it's used in this plugin, good idea is to change file_get_contents to cURL as some servers have allow_url_fopen set to false due security reasons.


linc:
Thanks for your contribution! Apologies for the delayed reply. If you can make @DaazKu's suggested fix I'd greatly appreciate it and then we can get this merged. Otherwise I'll see if I can get a developer to fix it up in an upcoming sprint.

zlorkovic:
Done. I've changed it/added to all FB graph cURL requests.

Issue #5641, opened 2017-06-01
Unable to update draft category if Vanilla.Categories.DoHeadings enabled


initvector:
This behavior has been specifically observed on the drafts page, but probably exists elsewhere.

The discussion form uses [`Gdn_Form::categoryDropDown` to display categories](https://github.com/vanilla/vanilla/blob/7c96e5e24d6e6a566d38271f650281686bcbfac7/applications/vanilla/views/post/discussion.php#L30) for the user to select. One of the `$Options` it passes along is "Context", [which is the current category's subtree](https://github.com/vanilla/vanilla/blob/7c96e5e24d6e6a566d38271f650281686bcbfac7/applications/vanilla/controllers/class.postcontroller.php#L141). The depth of children in this tree is adjusted to be relative to the category provided (i.e. Cat 2 might be at depth 2 if it's under Cat A, but if we're getting the subtree of Cat 2, it'll be at depth 1). If a site has `Vanilla.Categories.DoHeadings` enabled, `Gdn_Form::categoryDropDown` fails to recognize this depth adjustment, assumes its a top-level category and [disables the top level of this subtree](https://github.com/vanilla/vanilla/blob/7c96e5e24d6e6a566d38271f650281686bcbfac7/library/core/class.form.php#L553).

### Steps to recreate
1. Enabled `Vanilla.Categories.DoHeadings`.
1. Create a draft, saving it to a second-level category.
1. View a list of your drafts. Edit the one you just created.
1. You should see the category your originally selected is no longer an option. It is disabled.

DaazKu:
Seems like https://github.com/vanilla/vanilla/issues/4547 was never properly addressed.

linc:
I'm experiencing this issue and it's worse than described. If you've selected the child-most category in a tree, it de-selects the category and traps you there, rendering the draft discussion un-postable.

<img width="403" alt="screen shot 2018-04-02 at 4 33 05 pm" src="https://user-images.githubusercontent.com/117672/38214864-f3a199ae-3693-11e8-9aa7-f47633b6de9c.png">

linc:
FYI this is my category tree

<img width="226" alt="screen shot 2018-04-02 at 4 42 58 pm" src="https://user-images.githubusercontent.com/117672/38215192-f7acb8f2-3694-11e8-81c3-9211bf94d55e.png">

linc:
This appears to have been introduced as a problem about two years ago: https://github.com/vanilla/vanilla/pull/4276

[...]

PR #7238, opened 2018-05-22
Fix editing drafts in sub catergories with do headings enabled

cchabilall83:
fix vanilla/vanilla#5641

When Vanilla.Categories.DoHeadings is enabled, editing a draft created in sub categories is disabled and doesn't allow for the draft to be edited.  The category the draft was created is now considered a top-level category and is therefore disabled.

This update verifies if a post is a draft and doesn't disable the category from the dropdown list in the form.

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -553,6 +553,18 @@ class Gdn_Form extends Gdn_Pluggable {
         // Show root categories as headings (ie. you can't post in them)?
         $doHeadings = val('Headings', $options, c('Vanilla.Categories.DoHeadings'));
 
+        // If DoHeadings is enabled check if post is a draft so that children,
+        // so that children categories won't be considered a top-level
+        // category and disabled.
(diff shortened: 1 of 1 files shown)


cchabilall83:
After review this current fix doesn't resolve the bigger issue.  The issue is with how we get the dept of the sub-category.  This needs be to be address and the when we edit the draft the dropdown should be the same as when we created the draft

linc:
This appears to have been introduced as a problem about two years ago: https://github.com/vanilla/vanilla/pull/4276

That PR references https://github.com/vanilla/vanilla/issues/4275, which is an unrelated problem of the dropdown not appearing at all under certain conditions (a new discussion in a category of type "Discussions"). Therefore, I think this subtree behavior was indeed accidental and not necessarily part of the fix. Let's confirm this PR works correctlt in a category of type "Discussions" and type "Nested", then we can call this case-closed and merge it.

PR #7240, opened 2018-05-22
Fix cached announcements on Recent Discussions sometimes being filtered out

### LINKED ISSUE #7121: Announcements no longer pinned on top of recent-discussions
# Details
 
Recent change introduced in #6948 is causing announcements not to get pinned on top of recent-discussions page

## Before 

before_change
## After

after_change

initvector:
`DiscussionModel::getAnnouncements` is a very delicate method. It performs some unusual caching practices (namely using a fixed cache key, even for dynamic criteria) that didn't play well with the move from category watching to category following. The result is announced discussions might not be properly pinned when caching is enabled and something is filtering categories, like the Subcommunities addon. The announced discussion IDs are cached in one area, then retrieved and filtered out in another area.

This update makes a couple of minor tweaks to make Recent Discussions function a bit like it did in 2.5, while still utilizing the methods introduced with category following.

1. The call to `DiscussionModel::getAnnouncements` gets its own conditions array. This array only gets the conditions the method would've received in 2.5. This prevents caching of only a limited subset of announcements.
1. `DiscussionModel::discussionSummaryQuery` (used by `DiscussionModel::getAnnouncements`) now uses `CategoryModel::getVisibleCategoryIDs`, instead of `DiscussionModel::categoryPermissions`. This is the newer method for performing this action and is hook-able by addons.

Closes #7121

### DIFF
--- a/applications/vanilla/controllers/class.discussionscontroller.php
+++ b/applications/vanilla/controllers/class.discussionscontroller.php
@@ -148,7 +148,7 @@ class DiscussionsController extends VanillaController {
 
         // Check for individual categories.
         $categoryIDs = $this->getCategoryIDs();
-        $where = [];
+        $where = $announcementsWhere = [];
         if ($this->data('Followed')) {
             $followedCategories = array_keys($categoryModel->getFollowed(Gdn::session()->UserID));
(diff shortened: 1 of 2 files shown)


initvector:
Related #7241

PR #7259, opened 2018-05-25
Add /users/:id/confirm-email to APIv2

### LINKED ISSUE #7170: Add the POST /users/:id/confirm-email endpoint
This one came up and the APIv1 version isn't very good. Our current /entry/emailconfirm signs the user in after confirming their email. The APIv2 version of the endpoint won't do that.

cchabilall83:
This update adds /users/:id/confirmEmail to API v2 to confirm new users email address after registration.  Currently functionality signs the user in after confirmation, this will return true or throw an exception on failure

Closes #7170

### DIFF
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -597,15 +597,15 @@ class UsersApiController extends AbstractApiController {
         $this->userModel->passwordRequest($body['email']);
         $this->validateModel($this->userModel, true);
     }
-
     /**
-     * Confirm user email address after registration
+     * Confirm a user email address after registration.
      *
      * @param int $id The ID of the user.
      * @param array $body The POST body.
-     * @throws Exception
+     * @throws ClientException if email has been confirmed.
+     * @throws Exception if confirmationCode doesn't match.
      * @throws NotFoundException if unable to find the user.
-     * @return array
+     * @return array the response body.
      */
     public function post_confirmEmail($id, array $body) {
         $this->permission();
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -613,7 +613,7 @@ class UsersApiController extends AbstractApiController {
         $this->idParamSchema('in');
         $in = $this->schema( [
             'confirmationCode:s' => 'Email confirmation code'
-        ])->setDescription('Confirm a users current email address by using a confirmation code');
+        ], 'in')->setDescription('Confirm a users current email address by using a confirmation code');
         $out = $this->schema(['userID:i', 'email:s', 'emailConfirmed:b'], 'out');
 
         $row = $this->userByID($id);
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -608,7 +608,7 @@ class UsersApiController extends AbstractApiController {
     public function post_confirmEmail ($id, array $body) {
         $this->permission('Garden.Users.Edit');
 
-        $in = $this->schema( ['emailKey:s' => ''])->setDescription('Code sent via email');
+        $in = $this->schema( ['emailKey:s' => 'key generated on registration'])->setDescription('Confirm new user email upon registration');
         $out = $this->schema(['emailConfirmed:b' => 'The current email confirmed value.'], 'out');
 
         $row = $this->userByID($id);
@@ -616,14 +616,14 @@ class UsersApiController extends AbstractApiController {
 
         $emailKey = '';
         $userData = $this->normalizeInput($body);
-        if (array_key_exists('emailKey', $userData)) {
-            $emailKey = $userData['emailKey'];
+        if (array_key_exists('EmailKey', $userData)) {
+            $emailKey = $userData['EmailKey'];
         }
 
         $this->userModel->confirmEmail($row, $emailKey);
-        $this->validateModel($this->userModel, true);
+        $this->validateModel($this->userModel);
 
-        $result = $out->validate($row);
+        $result = $out->validate($this->userByID($id));
         return $result;
     }
 
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -608,8 +608,8 @@ class UsersApiController extends AbstractApiController {
      * @return array the response body.
      */
     public function post_confirmEmail($id, array $body) {
-        $this->permission();
-        
+        $this->permission(\Vanilla\Permissions::BAN_CSRF);
+
         $this->idParamSchema('in');
         $in = $this->schema( [
             'confirmationCode:s' => 'Email confirmation code'
@@ -617,7 +617,7 @@ class UsersApiController extends AbstractApiController {
         $out = $this->schema(['userID:i', 'email:s', 'emailConfirmed:b'], 'out');
 
         $row = $this->userByID($id);
-        if ($row['Confirmed'] == 1){
+        if ($row['Confirmed']) {
             throw new ClientException('This email has already been confirmed');
         }
 
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -611,9 +611,9 @@ class UsersApiController extends AbstractApiController {
         $this->permission(\Vanilla\Permissions::BAN_CSRF);
 
         $this->idParamSchema('in');
-        $in = $this->schema( [
+        $in = $this->schema([
             'confirmationCode:s' => 'Email confirmation code'
-        ],'in')->setDescription('Confirm a users current email address by using a confirmation code');
+        ], 'in')->setDescription('Confirm a users current email address by using a confirmation code');
         $out = $this->schema(['userID:i', 'email:s', 'emailConfirmed:b'], 'out');
 
         $row = $this->userByID($id);
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -599,6 +599,35 @@ class UsersApiController extends AbstractApiController {
     }
 
     /**
+     * @param $id
+     * @param array $body
+     * @return array
+     * @throws Exception
+     * @throws NotFoundException if unable to find the user.
+     */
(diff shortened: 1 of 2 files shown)


applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
         $this->validateModel($this->userModel, true);
     }
 
+    /**
+     * @param $id
```
initvector: Need some more detail in this doc block. Explain what we're doing here. Add brief descriptions to the parameters (you can copy from other methods in the class).

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+     * @param array $body
+     * @return array $result The response body
+     * @throws Exception
+     * @throws NotFoundException if unable to find the user.
+     */
+    public function post_confirmEmail ($id, array $body) {
+        $this->permission('Garden.Users.Edit');
```
initvector: Probably don't want users with the ability to edit other users being the only ones who can confirm their email address. What permissions does `EntryController::emailConfirm` require?

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
         $this->validateModel($this->userModel, true);
     }
 
+    /**
+     * @param $id
+     * @param array $body
+     * @return array $result The response body
```
initvector: I don't think this tag is in the right format. [Checkout its docs](http://docs.phpdoc.org/references/phpdoc/tags/return.html).

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+     * @throws Exception
+     * @throws NotFoundException if unable to find the user.
+     */
+    public function post_confirmEmail ($id, array $body) {
```
initvector: No space between the method name and its parameters.

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+        $this->permission('Garden.Users.Edit');
+
+        $in = $this->schema( ['emailKey:s' => 'key generated on registration'])->setDescription('Confirm new user email upon registration');
```
initvector: The line is a little long, so the formatting could use some breathing room. Maybe something like...

```php
$in = $this->schema([
    'emailKey:s' => 'key generated on registration'
])->setDescription('Confirm new user email upon registration');
```

initvector: I think these descriptions should be more generic. The vast majority of uses are going to be for new users, but that doesn't mean that's all it can be used for. Some suggestions:

1. emailKey - Email confirmation code
1. Description - Confirm a users current email address by using a confirmation code

initvector: I'm thinking out-loud here, but maybe we can rename the "email key" parameter to "confirmation code", "code" or something similar. We use "email key" internally, but we have a chance to address renaming some things to a more user-friendly version as we expose them via the API. The word "key" strikes me as an index, which this isn't. To me, this is a randomly-generated "code".

Nothing that necessarily has to change, just opening it up for discussion.

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+        $out = $this->schema(['emailConfirmed:b' => 'The current email confirmed value.'], 'out');
+
+        $row = $this->userByID($id);
+        $body = $in->validate($body);
+
+        $emailKey = '';
```
initvector: If we made it past validating the body against the input schema, you shouldn't need to initialize this to an empty value. `$body` should definitely have an "emailKey" entry.

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+        $userData = $this->normalizeInput($body);
```
initvector: This isn't necessary. If you were `$body` (or `$userData`) off to a method that was built to handle field names as they exist in the database, you'd normalize them. However, we aren't doing any of that here. You can grab the incoming confirmation code without normalizing the array.

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+        if (array_key_exists('EmailKey', $userData)) {
+            $emailKey = $userData['EmailKey'];
```
initvector: As mentioned earlier, if you've made it this far, you definitely already have the value in `$body["emailKey"]`. You can pass it, as is, to `UserModel::confirmEmail`.

tests/APIv2/UsersTest.php
```diff
@@ -129,6 +129,37 @@ public function testDeletePhoto() {
         $this->assertStringEndsWith('/applications/dashboard/design/images/defaulticon.png', $user['photoUrl']);
     }
 
+    /**
+     * Test confirm email is successful
+     */
+    public function testConfirmEmailSucceed() {
+        $emailKey = ['emailKey' => 'test123'];
```
initvector: You can generate an email confirmation code, like the ones Vanilla would create on its own, by using `UserModel::confirmationCode`. I recommend using that to make this test closer to a real scenario.

tests/APIv2/UsersTest.php
```diff
@@ -129,6 +129,37 @@ public function testDeletePhoto() {
...
+        $user = $this->testPost();
+
+        /** @var \UserModel $userModel */
+        $userModel = self::container()->get('UserModel');
+        $userModel->saveAttribute($user['userID'], 'EmailKey', 'test123');
+
+        $response = $this->api()->post("{$this->baseUrl}/{$user['userID']}/confirmEmail", $emailKey);
+        $this->assertEquals($response->getBody(), ['emailConfirmed' => true]);
```
initvector: I believe users created with the users API endpoint are going to have their email address confirmed by default. In that case, this test would confirm the request didn't trigger an exception, but not that it necessarily affected the value of `emailConfirmed`.

tests/APIv2/UsersTest.php
```diff
@@ -129,6 +129,37 @@ public function testDeletePhoto() {
...
+     */
+    public function testConfirmEmailSucceed() {
+        $emailKey = ['emailKey' => 'test123'];
+        $user = $this->testPost();
+
+        /** @var \UserModel $userModel */
+        $userModel = self::container()->get('UserModel');
+        $userModel->saveAttribute($user['userID'], 'EmailKey', 'test123');
+
+        $response = $this->api()->post("{$this->baseUrl}/{$user['userID']}/confirmEmail", $emailKey);
```
initvector: The original issue stated the URL should look more like: /users/:id/confirm-email

You shouldn't have to change anything, but make sure you're testing with that URL.

tests/APIv2/UsersTest.php
```diff
@@ -129,6 +129,37 @@ public function testDeletePhoto() {
...
+        $this->assertEquals($response->getBody(), ['emailConfirmed' => true]);
+    }
+
+    /**
+     * Test confirm email fails
+     *
+     * @expectedException \Exception
```
initvector: This should be more specific. How would you know if this exception was thrown because of the bad email confirmation key or if it was thrown because of a random error in the dispatcher? Right now, all this test confirms is that something went wrong. You could use the `@expectedExceptionMessage` tag for greater specificity.

tests/APIv2/UsersTest.php
```diff
@@ -129,6 +129,37 @@ public function testDeletePhoto() {
         $this->assertStringEndsWith('/applications/dashboard/design/images/defaulticon.png', $user['photoUrl']);
     }
 
+    /**
+     * Test confirm email is successful
```
initvector: Make sure your descriptions end in a full stop.

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+     * @param array $body
+     * @return array $result The response body
+     * @throws Exception
+     * @throws NotFoundException if unable to find the user.
+     */
+    public function post_confirmEmail ($id, array $body) {
+        $this->permission('Garden.Users.Edit');
+
+        $in = $this->schema( ['emailKey:s' => 'key generated on registration'])->setDescription('Confirm new user email upon registration');
+        $out = $this->schema(['emailConfirmed:b' => 'The current email confirmed value.'], 'out');
```
initvector: `EntryController::confirmEmail` includes user ID and the actual email address that was confirmed, in addition to emailConfirmed. Let's include those in the output too.

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -598,6 +598,35 @@ public function post_requestPassword(array $body) {
...
+
+        $row = $this->userByID($id);
```
initvector: If the user is currently confirmed, should we let them attempt to confirm again, anyway? `EntryController::confirmEmail` doesn't seem to care, but I can't think of a good reason to allow it. The user is still going to run into an error, but it'll be a misleading one ("We couldn't confirm your email. Check the link in the email we sent you...").

applications/dashboard/controllers/api/UsersApiController.php
cchabilall83: resolved

tests/APIv2/UsersTest.php
cchabilall83: resolved

tests/APIv2/UsersTest.php
cchabilall83: resolved

tests/APIv2/UsersTest.php
cchabilall83: confirmationCode is private, not sure a work around for that

initvector: My mistake! This is fine, as is, then.

Issue #6504, opened 2018-01-25
Rich Editor - Mobile View


charrondev:
According to the mockups (except without the emoji picker).

All features are technically still supported (in the case of editing), but may not have UI for them. This means there is no UI for anything except for:

- Plaintext
- Embed
- Image Upload.

PR #7260, opened 2018-05-25
Rich Editor Mobile Styles

slafleche:
Mobile styles for editor. To be merged along with https://github.com/vanilla/rich-editor/pull/26

Closes https://github.com/vanilla/vanilla/issues/6504

Note that there are still issues to resolve, this is a first iteration. Issues will be filed. For now we have:
* https://github.com/vanilla/rich-editor/issues/29
* https://github.com/vanilla/rich-editor/issues/28


charrondev:
So a few things that I'm noticing as I use it on my actual device:

- We need to prevent scrolling of the body while in full-screen. It feels pretty weird otherwise. It looks like CSS-only solutions for this no longer work on mobile safari. touchstart/move handlers on the body are necessary.
- I think we're going to have to 100% ensure a minimum font-size of 16px on the rich editor contenteditable. The zooming is really a pain.

A couple things I'd like to see issues opened for:

- You can't delete embeds from a touchscreen device. We need to either show the keyboard while they are selected or show a tooltip to delete them.

- https://github.com/vanilla/vanilla/issues/7271

applications/dashboard/src/scripts/dom.ts
```diff
@@ -319,3 +321,51 @@ export function ensureScript(scriptUrl: string) {
         }
     });
 }
+
+export function getNextTabbableElement(
```
charrondev: A utility function like this should be added with unit tests and shouldn't log anything to the console. We can probably just kick this down the road until we go to resolve the focus issues on mobile.

slafleche:
The body scrolling doesn't matter because we're going to change our approach next sprint. It won't be fixed anymore.

As for the font size, I think that'll be up to the theme. We'll likely get a lot of negative feedback from clients if we update our core themes to that size. We can consider it for future themes.

charrondev:
I'd just like to see the merge conflicts fixed, and an issue filed for this and I'll be happy to merge it.

> You can't delete embeds from a touchscreen device. We need to either show the keyboard while they are selected or show a tooltip to delete them.

PR #7275, opened 2018-05-29
Rich Editor -  Hover and Focus styles

slafleche:
Hover and focus styles for new UI elements in rich editor.

To be clear this is not meant to fix any pre-existing UI elements and it's not meant as final styling for the new base theme. This is strictly adding missing styles where there were none.

Related: https://github.com/vanilla/rich-editor/pull/31

Closes: https://github.com/vanilla/rich-editor/issues/23


charrondev:
A few notes:

- For the toolbars the active and hover states are the same.
- There seems to be an issue with the focus state flashing when an item is pressed, especially when clicking things like the paragraph menu or the embed menu where the focus is then moved. You had mentioned putting a transition/animation delay on these?

slafleche:
1. Fixed some of the inconsistencies
2. I tried to add a delay, but because we set focus afterwards, it looks clunky. For this PR let's stick with default styling and we can do more complex styles in the theme
3. Can you reproduce? It looks good on my end

PR #7276, opened 2018-05-29
Add JSON LD to discussion pages

### LINKED ISSUE #6872: /media/scrape API endpoint is missing details about own site
In order to complete the design for our own links embedded,
image

we need some additional data about these pages. Missing is:

- username
- timestamp
- user url.

These can come back as attributes. Additionally the type should be different. Maybe instead of type `site` we call it type `vanilla`.

initvector:
The primary goal of this update is to add Vanilla discussion data to the result of /media/scrape requests. To that end, several things have been done:

1. `HeadModule` has been updated to include `setJsonLD` and `getJsonLD` methods. These methods set and get JSON LD data, respectively. If the module has JSON LD at the time of render, it is output in a script tag with a type of "application/ld+json". The data must go either in the head or body of the page. I opted for the head module, since that seemed to be the most reliable to render the data.
1. Added `DiscussionModel::structuredData`. Given a discussion row, this method will generate structured data per the [DiscussionForumPosting](http://schema.org/DiscussionForumPosting) schema.
1. `DiscussionController::index` has been updated to call `DiscussionModel::structuredData` and use its result to set JSON LD in `HeadModule`.
1. `PageScraper` has been altered to allow registering of metadata handlers. There are currently two handlers: `OpenGraphParser` and `JsonLDParser`. `OpenGraphParser` is essentially the OpenGraph capabilities that previously existed in `PageScraper`, moved out into a dedicated class. `JsonLDParser` is a JSON LD parser, currently limited to parsing items with a type matching [DiscussionForumPosting](http://schema.org/DiscussionForumPosting). There was already decent code coverage with existing tests, so I do not anticipate any breaking changes here.
1. Tests have been added for the new JSON LD parsing.

Closes #6872


charrondev:
With this PR the `/media` endpoint no longer shows up in swagger for me.

PR #7289, opened 2018-05-31
Fix identifier escaping in Gdn_SqlDriver

tburry:
This is a port of https://github.com/vanilla/vanilla-patches/pull/249


DaazKu:
No need for a port. We just need to bring vanilla-patches into vanilla. @linc we'll do that early next week.

PR #7291, opened 2018-05-31
Fix Advanced Editor uploads failing if unable to determine MIME type

### LINKED ISSUE #7270: Editor: File upload fails when AllowedFileExtension is not added to mimetypes
## Details

Whenever a file is uploaded with an extension that exists in AllowedFileExtensions but does not have a record in [mimetypes.php](https://github.com/vanilla/vanilla/blob/master/plugins/editor/mimetypes.php) upload fails.

We recently did a security patch [#198](https://github.com/vanilla/vanilla-patches/pull/198) that adds MIME type validation to editor uploads.

## Requested Fix

When adding a custom file extension to AllowedFileExtensions list, a record should be automatically added to mimetypes file without having to go through updating mimetypes file everytime a new custom extension is required.

dannimajdalani:
Added a default MIME type of 'application/octet-stream' in case:

- Uploaded file is an allowed file extension
   &
- Extension has an unknown MIME

Closes #7270

### DIFF
--- a/plugins/editor/class.editor.plugin.php
+++ b/plugins/editor/class.editor.plugin.php
@@ -741,8 +741,10 @@ class EditorPlugin extends Gdn_Plugin {
 
         $mimeType = $fileData['type'];
         $allowedMimeTypes = $this->getAllowedMimeTypes();
+        // If mimetype is not in the allowedMimeTypes list, set it to application/octet-stream
+        // before saving it to the database.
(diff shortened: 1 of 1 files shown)


DaazKu:
screen shot 2018-06-04 at 2 11 44 pm

The file is not saved with the proper mime type.

initvector:
Already has two approvals, so I'm gonna go ahead and merge.

PR #7297, opened 2018-06-04
Add Instagram Embed

cchabilall83:
This feature allows for the embedding of instagram of posts using instagram's omembed endpoint

- I use Instagram's oEmbed is used to retrieve the html, parse it for the necessary data
- I use Instagram's embed.js library to fully render the image client side.  The   instgrm.Embeds.process() is used to intialize the embed code
- I added "html" to $validAttributes in oembed() in the Embed class
- I've added test data to the the MediaTest and EmbedManagerTest

Closes Vanilla/Vanilla#7210


library/Vanilla/Embeds/InstagramEmbed.php
```diff
@@ -0,0 +1,77 @@
...
+     * @inheritdoc
+     */
+    public function renderData(array $data): string {
+        $instagramPostID = $data['attributes']['postID'] ?? '';
+        $instagramPostID = htmlspecialchars($instagramPostID);
+        $width = $data['width'] ?? 412;
+        $height = $data['height'] ?? 510;
+
+        $result = <<<HTML
+<div class="embed-image embed embedImage">
```
charrondev: Why are we using image classes here?

cchabilall83: wasn't sure which classes should be applied, so i mimicked how we did it in the image embed.  I was going to discuss with Stephane tomorrow how to handle it.  Suggestions?

slafleche:
@charrondev I'll take a look today for the CSS and then hand it back to you

slafleche:
Alright Chris, your styles should be good now! Any future embeds should follow this same convention. Let me know if you have questions.

applications/dashboard/src/scripts/app/user-content/embeds/instagram.ts
```diff
@@ -0,0 +1,47 @@
...
+// Setup instagram embeds.
+onContent(convertInstagramEmbeds);
+registerEmbed("instagram", renderInstagram);
+
+/**
+ * Renders posted instagram embeds.
+ */
+async function convertInstagramEmbeds() {
+    await ensureScript("//platform.instagram.com/en_US/embeds.js");
+    window.instgrm.Embeds.process();
```
charrondev: You need to check that `window.instgrm` is actually available here.

PR #7298, opened 2018-06-04
Editor Mobile styles v2

slafleche:
Our first attempt at mobile styles with fixed positioned editor didn't work. Here's a second attempt with a sticky menu instead.

**Important information about the sticky menu:**
We currently don't have a proper responsive theme with this type of sticky menu to test. If you want to try it out, you need to have the following in your HTML:

```HTML
<div class="stickyHeaderWrap">
   <div class="stickyHeader">
    <!-- your header -->
    </div>
</div>
```

Both `.stickyHeaderWrap` and `.stickyHeader` need to have a height set in CSS. Also make sure your theme is **responsive**.

_______

Related: https://github.com/vanilla/internal/pull/1604, https://github.com/vanilla/rich-editor/pull/38

Closes: https://github.com/vanilla/rich-editor/issues/35


package.json
```diff
@@ -4,6 +4,7 @@
     "license": "GPL-2.0",
     "private": true,
     "dependencies": {
+        "@types/react-virtualized": "9.18.3",
```
charrondev: This should be in dev-dependencies

applications/dashboard/src/scripts/dom.ts
```diff
@@ -490,3 +491,73 @@ export function watchFocusInDomTree(rootNode: Element, callback: (hasFocus: bool
...
+                    window.requestAnimationFrame(data => {
+                        lastScrollPos = currentScrollPos;
+                        currentScrollPos = Math.max(window.scrollY, 0);
+                        handleStickyHeaderState(header, {
+                            currentScrollPos,
+                            lastScrollPos,
+                        });
+                    });
+                },
+                200,
```
charrondev: I was experiencing some jitteryness on my iOS with a 200ms debounce. Setting it 100ms resolves the issue.

PR #7300, opened 2018-06-04
Delete package-lock.json

charrondev:
We don't use package locks, and one of the pinned versions in this package lock has a security vulnerability (likely because it hasn't been regenerated in ages).

We already have a `yarn.lock` in this repo that does not have the vulnerable package in it.


DaazKu:
Add the lock(s) to .gitignore if we don't use them.

charrondev:
I don't think we should. If someone uses npm to install dependencies and makes a PR, having a `package-lock.json` is smoke tests that what they built with will not align with what everyone else builds with.

The ideal scenario here is:

- Do not ignore the `package-lock.json`.
- If someone installs with `npm` and they or we see this new file generated, someone notices something is wrong.
- Once we start adding additional checks in CI, one of them can be to ensure no `package-lock.json` files are admitted.

PR #7305, opened 2018-06-04
Improve JS test harness

charrondev:
The JS test harness has been improved in a few different ways:

- Reporter outputs less now. Mocha `minimal` reporting is a summary, and output from failing tests only. Successful tests do not all show.
- Adds a `test:debug` option that enables the debug port on a normal chrome instance.
- Adds instruction for debugging with Chrome Dev Tools and attaching VSCode.
- Fixes sourcemaps. This was accomplish by doing 2 things:
  - Passing transpiled typescript to babel after initial transpilation. I was hoping to not need `awesome-typescript-loader` and just be able to use `@babel/preset-typescript` for this, but it is still in beta (for months!! https://github.com/babel/babel/milestone/15).
  - Having a single test entrypoint that the rest of the tests are loaded from.
- A particularly troublesome FormData that didn't want to be treated as an `Iterable` in certain headless browsers. I've switched it to use a forEach loop instead which is more compatible. This is mainly because I'm trying to avoid loading polyfills in the test environment right now.

There was one single slightly unlrelated thing that snuck it's way in here, but the `renderEmbed()` method now throws an error instead of silently logging it. This is to accomodate some upcoming changes in the rich editor.


tests/javascript/test-setup.ts
```diff
@@ -0,0 +1,11 @@
+// require all modules ending in "_test" from the
```
cdepage: Should this be ".test" instead of "_test" ?

tests/javascript/webpack.test.config.js
```diff
@@ -39,13 +39,19 @@ module.exports = {
             {
                 test: /\.tsx?$/,
```
cdepage: While I agree that all React components will end in .tsx, there could be supporting files that end in .ts that would also be good to unit test. Should this be updated to /\.(ts|tsx)?$/, ?

charrondev: That's what the ? mark at the end is for. The x is an optional character.

charrondev:
I'm slapping a hold on this until I can fix the issues on CI.

tests/javascript/webpack.test.config.js
cdepage: d'oh.... regex fail on my end.

PR #7313, opened 2018-06-05
Allow search to opt out of SQL table escaping (fix database error from added escaping)

charrondev:
This PR fixes the the normal search's plaintext SQL search.

Because we don't have a `fromRaw` or `from(SqlObject)` or `fromSql` type function, we are doing a search and replace that broke because of the additional escaping. 

https://github.com/vanilla/vanilla/blob/master/applications/dashboard/models/class.searchmodel.php#L152-L159

@tburry's preferred fix was allowing to opt out of escaping on the `from()` method.

Other solutions that were initially considered:
- Changed the find/replace query
- Allowing from to take a SQLObject that presumably already be escaped to the correct amount.

### DIFF
--- a/applications/dashboard/models/class.searchmodel.php
+++ b/applications/dashboard/models/class.searchmodel.php
@@ -151,7 +151,7 @@ class SearchModel extends Gdn_Model {
         // Perform the search by unioning all of the sql together.
         $sql = $this->SQL
             ->select()
-            ->from('_TBL_ s')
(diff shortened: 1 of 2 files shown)


DaazKu:
Where is the relevant unit test?

DaazKu:
I understand the need for the patch on 2.6. Let's make sure we have one on master though.

PR #7342, opened 2018-06-11
Fix/reverting 7325

topcheese13:
This backported PR needs to be reverted.

### DIFF
--- a/applications/vanilla/controllers/class.discussionscontroller.php
+++ b/applications/vanilla/controllers/class.discussionscontroller.php
@@ -151,7 +151,7 @@ class DiscussionsController extends VanillaController {
 
         // Check for individual categories.
         $categoryIDs = $this->getCategoryIDs();
-        $where = $announcementsWhere = [];
+        $where = [];
         if ($this->data('Followed')) {
             $followedCategories = array_keys($categoryModel->getFollowed(Gdn::session()->UserID));
             $visibleCategoriesResult = CategoryModel::instance()->getVisibleCategoryIDs(['filterHideDiscussions' => true]);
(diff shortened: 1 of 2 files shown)


initvector:
The update to `DiscussionModel::discussionSummaryQuery` exposed an object pollution issue. In this case:

1. [`BookmarkedModule::getData` modifies the SQL property of `DiscussionModel`, which is a shared instance of `Gdn_MySQLDriver`, before calling `DiscussionModel::get`.](https://github.com/vanilla/vanilla/blob/2a4ac8a6fa65660bd020fa9904f5f41b47b863d3/applications/vanilla/modules/class.bookmarkedmodule.php#L45)
1. [`DiscussionModel::get` calls out to `DiscussionModel::discussionSummaryQuery`](https://github.com/vanilla/vanilla/blob/2a4ac8a6fa65660bd020fa9904f5f41b47b863d3/applications/vanilla/models/class.discussionmodel.php#L430).
1. [`DiscussionModel::discussionSummaryQuery` then makes a call to `CategoryModel::getVisibleCategoryIDs`.](https://github.com/vanilla/vanilla/blob/2a4ac8a6fa65660bd020fa9904f5f41b47b863d3/applications/vanilla/models/class.discussionmodel.php#L337) At this point, any database queries made by `CategoryModel` still have the discussion ID conditions introduced in `BookmarkedModule::getData`. This can lead to a fatal error.

If possible, it's probably better to update `BookmarkedModule::getData` to [make better use of the `$where` parameter of `DiscussionModel::get`](https://github.com/vanilla/vanilla/blob/2a4ac8a6fa65660bd020fa9904f5f41b47b863d3/applications/vanilla/modules/class.bookmarkedmodule.php#L50), instead of directly altering the shared SQL instance.

PR #7343, opened 2018-06-11
Add embed for imgur

### LINKED ISSUE #7209: Create imgur embed
- The embed renderer in `plugins/rich-editor/embeds/`
- The server rendered blot from the EmbedManager `library/Vanilla/Embeds/`
- The JS needed to render the embed normally in `src/scripts/app/user-content/embeds`

cchabilall83:
This feature allows for the embedding of imgurs embeds

The imgur link is parsed to extract the postID and a call in made to imgur to get the json for that particular post, from the response it is determined if this post is in album or not.  
The data extracted from the media scrape is used to reconstruct imgurs  embed code and imgurs embed js script is used to render the post.
Data for imgurs embeds has been added to the the MediaTest and EmbedManagerTest as well as Quill Santize tests.
Closes #7209

### DIFF
--- a/applications/dashboard/src/scripts/app/user-content/embeds/imgur.ts
+++ b/applications/dashboard/src/scripts/app/user-content/embeds/imgur.ts
@@ -14,7 +14,7 @@ registerEmbed("imgur", renderImgur);
 /**
  * Renders posted imgur embeds.
  */
-async function convertImgurEmbeds() {
+export async function convertImgurEmbeds() {
     const images = Array.from(document.querySelectorAll(".imgur-embed-pub"));
     if (images.length > 0) {
         await ensureScript("//s.imgur.com/min/embed.js");
@@ -40,7 +40,7 @@ export async function renderImgur(element: Element, data: IEmbedData) {
     await ensureScript("//s.imgur.com/min/embed.js");
 
     const url = "imgur.com/" + data.attributes.postID;
-    const isAlbum = false;
+    const isAlbum = data.attributes.isAlbum;
     const dataSet = isAlbum ? "a/" + data.attributes.postID : data.attributes.postID;
 
     if (!window.imgurEmbed) {
@@ -52,6 +52,9 @@ export async function renderImgur(element: Element, data: IEmbedData) {
     blockQuote.setAttribute("lang", "en");
     blockQuote.dataset.id = dataSet;
     blockQuote.setAttribute("href", url);
-    convertImgurEmbeds();
+
     element.appendChild(blockQuote);
+    setImmediate(() => {
+        convertImgurEmbeds().then();
+    });
 }
--- a/applications/dashboard/src/scripts/app/user-content/embeds/imgur.ts
+++ b/applications/dashboard/src/scripts/app/user-content/embeds/imgur.ts
@@ -1,0 +1,47 @@
+/**
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license https://opensource.org/licenses/GPL-2.0 GPL-2.0
+ */
+
+import { registerEmbed, IEmbedData } from "@dashboard/embeds";
+import { ensureScript } from "@dashboard/dom";
+import { onContent } from "@dashboard/application";
+
+// Setup imgur embeds.
+onContent(convertImgurEmbeds);
+registerEmbed("imgur", renderImgur);
+
+/**
+ * Renders posted imgur embeds.
+ */
+async function convertImgurEmbeds() {
+    await ensureScript("//s.imgur.com/min/embed.js");
+
+    if (!window.imgurEmbed) {
+        throw new Error("The Imgur post failed to load");
+    }
+}
+
+/**
+ * Render a single imgur embed.
+ */
+export async function renderImgur(element: Element, data: IEmbedData) {
+    await ensureScript("//s.imgur.com/min/embed.js");
+
+    const url = "imgur.com/" + data.attributes.postID;
+    console.log(data.attributes.postID);
+    const isAblum = data.attributes.isAblum;
+    const dataSet = isAblum ? "a/" + data.attributes.postID : data.attributes.postID;
+
+    if (!window.imgurEmbed) {
+        throw new Error("The Imgur post failed to load");
+    }
+
+    const blockQuote = document.createElement("blockquote");
+    blockQuote.classList.add("imgur-embed-pub");
+    blockQuote.setAttribute("lang", "en");
+    blockQuote.dataset.id = dataSet;
+    blockQuote.setAttribute("href", url);
+
+    element.appendChild(blockQuote);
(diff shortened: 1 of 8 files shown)


applications/dashboard/src/scripts/app/user-content/embeds/imgur.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Render a single imgur embed.
+ */
+export async function renderImgur(element: Element, data: IEmbedData) {
+    await ensureScript("//s.imgur.com/min/embed.js");
+
+    const url = "imgur.com/" + data.attributes.postID;
+    const isAlbum = data.attributes.isAlbum;
+    const dataSet = isAlbum ? "a/" + data.attributes.postID : data.attributes.postID;
+
+    if (!window.imgurEmbed) {
```
initvector: This and the `await` now seem redundant, since this load/check is being performed in `convertImgurEmbeds`.

bootstrap.php
```diff
@@ -244,16 +244,15 @@
     ->addCall('addEmbed', [new Reference(Vanilla\Embeds\TwitterEmbed::class)])
     ->addCall('addEmbed', [new Reference(Vanilla\Embeds\YouTubeEmbed::class)])
     ->addCall('addEmbed', [new Reference(Vanilla\Embeds\VimeoEmbed::class)])
-    ->addCall('addEmbed', [new Reference(Vanilla\Embeds\InstagramEmbed::class)])
```
initvector: Why are we getting rid of Instagram?

bootstrap.php
```diff
@@ -244,16 +244,15 @@
...
     ->addCall('addEmbed', [new Reference(Vanilla\Embeds\SoundCloudEmbed::class)])
     ->addCall('addEmbed', [new Reference(Vanilla\Embeds\ImageEmbed::class), Vanilla\Embeds\EmbedManager::PRIORITY_LOW])
     ->setShared(true)
 
     ->rule(Vanilla\PageScraper::class)
     ->addCall('registerMetadataParser', [new Reference(Vanilla\Metadata\Parser\OpenGraphParser::class)])
     ->addCall('registerMetadataParser', [new Reference(Vanilla\Metadata\Parser\JsonLDParser::class)])
-    ->setShared(true)
-;
+    ->setShared(true);
```
initvector: Why'd you move the semicolon? It's better to keep it separate in large chained calls, like this.

library/Vanilla/Embeds/ImgurEmbed.php
```diff
@@ -0,0 +1,70 @@
...
+                throw new Exception('Unable to get post ID.', 400);
+            }
+        }
+        // Get the json for th imgur post
+        $jsonResponse = $this->httpRequest($url . ".json");
+        if ($jsonResponse->getStatusCode() == 200) {
+            $decodedResponse = json_decode($jsonResponse);
+            $isAlbum = val('is_album', $decodedResponse->data->image);
+            if ($isAlbum === null) {
+                throw new Exception('Unable to get album.', 400);
```
initvector: This will fail, because there isn't an `Exception` class under the `Vanilla\Embeds` namespace. Same for the post ID error, above.

library/Vanilla/Embeds/ImgurEmbed.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function matchUrl(string $url) {
+        $data = [];
+
+        if ($this->isNetworkEnabled()) {
+            preg_match('/https?:\/\/imgur\.com\/gallery\/(?<postID>[a-z0-9]+)/i', $url, $matches);
```
initvector: This doesn't seem to be as flexible as it should be. It works on "gallery" URLs, but not other types of URLs. For example:

1. Works: https://imgur.com/gallery/10HROiq
1. Does not work: https://imgur.com/10HROiq
1. Does not work: https://m.imgur.com/gallery/10HROiq
1. Does not work: https://imgur.com/a/LQVun5v

library/Vanilla/Embeds/ImgurEmbed.php
```diff
@@ -0,0 +1,70 @@
...
+                if (!array_key_exists('attributes', $data)) {
+                    $data['attributes'] = [];
+                }
+                $data['attributes']['postID'] = $matches['postID'];
+            } else {
+                throw new Exception('Unable to get post ID.', 400);
+            }
+        }
+        // Get the json for th imgur post
+        $jsonResponse = $this->httpRequest($url . ".json");
```
initvector: No spaces around the concatenation operator.

library/Vanilla/Embeds/ImgurEmbed.php
```diff
@@ -0,0 +1,70 @@
...
+        if ($jsonResponse->getStatusCode() == 200) {
+            $decodedResponse = json_decode($jsonResponse);
```
initvector: `json_decode` has an [`$assoc` parameter](http://php.net/manual/en/function.json-decode.php#refsect1-function.json-decode-parameters) that interprets the JSON as an associative array. This is preferred, since it's typically better memory usage, although you likely won't notice it at the scale of these responses.

library/Vanilla/Embeds/ImgurEmbed.php
```diff
@@ -0,0 +1,70 @@
...
+            $isAlbum = val('is_album', $decodedResponse->data->image);
```
initvector: Where you can, try to avoid using `val` in favor of PHP's [coalescing operator](http://php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op).

tests/Library/Vanilla/Embeds/EmbedManagerTest.php
```diff
@@ -103,7 +105,44 @@ public function provideRenderedData() {
...
+                    'url' =>'https://imgur.com/gallery/OsirufX',
+                    'type' =>'imgur',
+                    'name' => null,
+                    'body' => null,
+                    'photoUrl' => null,
+                    'height' => null,
+                    'width' => null,
+                    'attributes' => [
+                        'postID' => 'a/OsirufX',
+                        'isAlbum' => false,
```
initvector: Shouldn't `postID` be "OsirufX" and `isAlbum` be `true`?

library/Vanilla/Embeds/ImgurEmbed.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/**
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL-2.0
+ */
+
+namespace Vanilla\Embeds;
+
+/**
+ * Instagram Embed.
```
initvector: Is it, though?

PR #7353, opened 2018-06-13
Fix categoryDropDown when editing drafts with DoHeadings enabled

cchabilall83:
Close Vanilla/Vanilla#5641
When you save a draft in child category with Vanilla.Categories.DoHeadings enabled (Parent Category is considered heading so you can't post in it), you're unable to edit the draft.  The root of the issue can be traced back to the postcontroller discussion method where the draft's category id is used to find the category subtree for that specific draft and assigned the context.  The context is passed on to the categoryDropDrown method, where the it now assumes the top level of the subtree is the parent category and disables it.  Removing the assignment of context from the postcontroller fixes this issue.

I've tested with categories displayed as heading, nested, flat and discussion and could not find any adverse effects.

There is same issue in the subcommunities plugin but the fix is not as apparent.

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -149,8 +149,7 @@ class PostController extends VanillaController {
             if (val('DisplayAs', $this->Category) == 'Discussions' && !$draftID) {
                 $this->ShowCategorySelector = false;
             } else {
-                // Get all our subcategories to add to the category if we are in a Header or Categories category.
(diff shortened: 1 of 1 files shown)


linc:
Is this an accurate summary?

This solves the issue in core, but not in Subcommunities. That said, the status of the (same) issue in Subcommunities isn't changed by this whatsoever.

linc:
If that's accurate, file the issue on Subcommunities and cross-reference it, then let's push ahead with review of this.

Issue #7341, opened 2018-06-11
Unknown column 'd.discussionID in where clause' when bookmarking a discussion with a multisite setup


dannimajdalani:
## Details
Enable Subcommuntiy plugin
Bookmark a discussion

<img width="286" alt="screen shot 2018-06-11 at 1 40 39 pm" src="https://user-images.githubusercontent.com/31856281/41247793-27ecd7e4-6d7d-11e8-90c4-cae21dddc16f.png">

LauraMakesArt:
@dannimajdalani and @francisbo Please make sure that you add additional details to this issue before you forget about them, we'll work on this next sprint.

LauraMakesArt:
https://github.com/vanilla/vanilla/pull/7342#issuecomment-396402858

I believe Ryan's comments on this PR are relevant to this issue.

topcheese13:
Should we also revert this PR https://github.com/vanilla/vanilla/pull/7342#issuecomment-396402858 (2 commits) from master ? @initvector @linc

initvector:
I'd rather see the object pollution issues fixed on master, instead of rolling back the fix that exposes them.

topcheese13:
As long as this is on somebody's radar, that we have this broken thing on master !

initvector:
> As long as this is on somebody's radar, that we have this broken thing on master !

That's what issues are for! 😄

PR #7359, opened 2018-06-14
Fix error message "Unknown column" appearing when bookmarking a discussion

topcheese13:
Closes https://github.com/vanilla/vanilla/issues/7341

Remove separate addition of a where clause prior to the get function that created pollution of SQL objects. Simply added the where condition as a parameter of the `get` function.

I removed the original condition `'w.Bookmarked' => '1'` in the where parameter of the get function as we already fetch the bookmarkIDs with the same condition a little higher in the code.

```
$bookmarkIDs = Gdn::sql()
                ->select('DiscussionID')
                ->from('UserDiscussion')
                ->where('UserID', Gdn::session()->UserID)
                ->where('Bookmarked', 1)
                ->get()->resultArray();
```

That condition seemed useless.

### DIFF
--- a/applications/vanilla/modules/class.bookmarkedmodule.php
+++ b/applications/vanilla/modules/class.bookmarkedmodule.php
@@ -42,12 +42,10 @@ class BookmarkedModule extends Gdn_Module {
                 $discussionModel = new DiscussionModel();
                 DiscussionModel::categoryPermissions();
 
-                $discussionModel->SQL->whereIn('d.DiscussionID', $bookmarkIDs);
-
                 $bookmarks = $discussionModel->get(
                     0,
                     $this->Limit,
-                    ['w.Bookmarked' => '1']
+                    ['d.DiscussionID' => $bookmarkIDs]
                 );
                 $this->setData('Bookmarks', $bookmarks);
             } else {


initvector:
> That condition seemed useless.

I know it seems functionally useless, and it's *probably* safe to remove, but do you think it might've been a performance optimization?

topcheese13:
I don't think it's optimizing anything as their is no indexing of the `Bookmarked` column. 

```
CREATE TABLE `GDN_UserDiscussion` (
  `UserID` int(11) NOT NULL,
  `DiscussionID` int(11) NOT NULL,
  `Score` float DEFAULT NULL,
  `CountComments` int(11) NOT NULL DEFAULT '0',
  `DateLastViewed` datetime DEFAULT NULL,
  `Dismissed` tinyint(4) NOT NULL DEFAULT '0',
  `Bookmarked` tinyint(4) NOT NULL DEFAULT '0',
  `Participated` tinyint(4) NOT NULL DEFAULT '0',
  PRIMARY KEY (`UserID`,`DiscussionID`),
  KEY `FK_UserDiscussion_DiscussionID` (`DiscussionID`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

PR #7362, opened 2018-06-15
Throw relevant message when SQLDriver cannot be instantiated

DaazKu:
In some weird case it's possible that vanilla could be hit with no proper configuration. When that's the case we should throw a relevant error message instead of "masking the error".

This is especially relevant since we added type hinting on method where we expect Gdn_SQLDriver.
We get the following error message:
```
Uncaught TypeError: Argument 1 passed to CategoryCollection::__construct() 
must be an instance of Gdn_SQLDriver, instance of stdClass given...
```
and it is quite confusing.

I did a fresh install with no problems and the `testAltInstall()` test works no problems.

I'd just like some context about https://github.com/vanilla/vanilla/commit/8d02cbd1d432f9e72c468686bf225388b4847f6d before determining if this fix is "ok" or if it will break some things I am unaware of.

### DIFF
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -514,7 +514,7 @@ class Gdn_Database {
             $name = $this->Engine.'Driver';
             $this->_SQL = Gdn::factory($name);
             if (!$this->_SQL) {
-                $this->_SQL = new stdClass();
+                throw new Exception("Could not instantiate database driver '$name'.");
             }
             $this->_SQL->Database = $this;
         }


DaazKu:
I've never got any context from anyone on the matter. It's probably safe to just throw an exception like this update does.

alex-mtl:
Just small note about this issue. Last august i've discussed this issue with @kaecyra 
And after some investigation he wrote a message to me:

> You were right its coming from outside. I patched our varnish config which was naively written
> It was assuming that the Host: header would be present but would be an IP

@DaazKu  you can ask @kaecyra for details and do some infrastructure config update I think to have a real fix to avoid these error messages.

PR #7368, opened 2018-06-18
Fix inability to restore some discussions from the moderation queue

dannimajdalani:
Fixes [#1592](https://github.com/vanilla/internal/issues/1592)

**Issue 1**

When Premoderated category plugin is enabled, after a discussion get edited for a second time, we get the error "Field 'InsertUserID' doesn't have a default value|Gdn_Database|Query|replace `GDN_Discussion` "
this got fixed by adding the missing fields to the array before saving the data to GDN_Log

**Issue 2**

After a discussion gets approved, it is marked as closed when the value of DateLastComment is set to "0000-00-00 00:00:00"
https://github.com/vanilla/vanilla/blob/be20cf00c2a40e35f69c5dc187bdde9b7f3731a1/applications/vanilla/models/class.discussionmodel.php#L1012

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -1009,7 +1009,8 @@ class DiscussionModel extends Gdn_Model {
         }
 
         // Allow for discussions to be archived
-        if ($discussion->DateLastComment && Gdn_Format::toTimestamp($discussion->DateLastComment) <= $archiveTimestamp) {
+        $dateLastCommentTimestamp = $discussion->DateLastComment ? Gdn_Format::toTimestamp($discussion->DateLastComment) : null;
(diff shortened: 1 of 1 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1009,7 +1009,8 @@ public function calculate(&$discussion) {
         }
 
         // Allow for discussions to be archived
-        if ($discussion->DateLastComment && Gdn_Format::toTimestamp($discussion->DateLastComment) <= $archiveTimestamp) {
+        $dateLastCommentTimestamp = $discussion->DateLastComment ? Gdn_Format::toTimestamp($discussion->DateLastComment) : null;
```
topcheese13: I'm not sure you need the ternary operator here ?

Wouldn't

```
$dateLastCommentTimestamp = Gdn_Format::toTimestamp($discussion->DateLastComment);
```

be enough ?

PR #7372, opened 2018-06-19
Add Giphy Embed

### LINKED ISSUE #7126: Create Giffy embed
- The Blot in Quill
- The server rendered blot from the EmbedManager
- The JS needed to make the embed appear both in an active editor and in the rendered content.

cchabilall83:
This feature allows for the embedding of Giphy posts

The Giphy link is parsed to extract the postID and stored. 
A call is made to Giphy's oembed endpoint to retrieve the pertinent information about the post.
Giphy's embed html is reconstructed using the information gathered and rendered client side using the  giphyRenderer function.
Data for Giphy embeds has been added to the the MediaTest and EmbedManagerTest as well as Quill Santize tests.
Closes #7126


library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,77 @@
...
+
+            $oembedData = $this->oembed("https://giphy.com/services/oembed?url=" . urlencode($url));
+            if ($oembedData) {
+                $data = $this->normalizeOembed($oembedData);
+            }
+
+        if ($post) {
+                $data['attributes']['postID'] = $post['postID'];
+            }
+            $data['type'] = "giphy";
```
initvector: This should already get populated by the embed manager, right?

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,77 @@
...
+            $data['attributes']['url'] = $url;
```
initvector: Do we need the URL here? It should already be a top-level field.

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,77 @@
...
+                '/https?:\/\/([a-zA-Z]+\.)?gi?phy?\.(com|is)\/(([\w]+\/)?(([\w]+-)*+)?(?<postID>[a-zA-Z0-9]+))(\/giphy.gif)?/i',
+                $url,
+                $post
+            );
+
+            if (!$post['postID']) {
+                throw new Exception('Unable to get post ID.', 400);
+            }
+
+            $oembedData = $this->oembed("https://giphy.com/services/oembed?url=" . urlencode($url));
```
initvector: No spaces around the concatenation operator.

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,77 @@
...
+
+        return $data;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function renderData(array $data): string {
+        $height = htmlspecialchars($data['height']) ?? 1;
+        $width = htmlspecialchars($data['width']) ?? 1;
```
initvector: You want to test for the existence of a value before using it. Here, you're passing a potentially undefined value to `htmlspecialchars`.

```php
htmlspecialchars($data['height'] ?? 1)
```

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,77 @@
...
+        $padding = ($height/$width) * 100;
+        $url = "https://giphy.com/embed/".$data['attributes']['postID'];
```
initvector: For the sake of defensive coding, I wouldn't assume `postID` is set in `attributes`.

tests/APIv2/MediaTest.php
```diff
@@ -375,6 +375,7 @@ public function provideTypeUrls() {
             ['https://www.instagram.com/p/BizC-PPFK1m', 'instagram'],
             ['https://soundcloud.com/syrebralvibes/the-eden-project-circles', 'soundcloud'],
             ['https://imgur.com/gallery/10HROiq', 'imgur'],
+            ['https://giphy.com/gifs/super-smash-bros-ultimate-jwSlQZnsymUW49NC3R', 'giphy'],
```
initvector: It looks like your regex in `matchUrl` supports some URL variations. We should test for those additional formats here, so we know if they fail to match in the future.

applications/dashboard/src/scripts/app/user-content/embeds/giphy.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Renders giphy embeds.
+ */
+export async function giphyRenderer(element: HTMLElement, data: IEmbedData) {
+    if (data.attributes.postID == null) {
+        throw new Error("Giphy embed fail, the post could not be found");
+    }
+
+    element.classList.add("embedGiphy");
+    element.style.width = `${data.width}px` || "100%";
```
initvector: This won't quite work, will it? If `data.width` is empty, won't "px" still cause "100%" not to kick in?

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,101 @@
...
+
+use Exception;
+
+/**
+ * giphy Embed.
+ */
+class GiphyEmbed extends Embed {
+
+    /** @inheritdoc */
+    protected $domains = ['giphy.com', 'media.giphy.com', 'gph.is'];
```
initvector: I don't think you need "media.giphy.com" in there. "giphy.com" should cover it.

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * giphyEmbed constructor.
+     */
+    public function __construct() {
+        parent::__construct('giphy', 'gif');
```
initvector: Shouldn't the render type be "image", like the other image-based embeds?

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,101 @@
...
+            $post = $this->parseURL($url);
+
+            $oembedData = $this->oembed("https://giphy.com/services/oembed?url=".urlencode($url));
+            if ($oembedData) {
+                $data = $this->normalizeOembed($oembedData);
+            }
+
+            // Obtain the real post ID from media url provided by oembed call
+            if ($domain == 'gph.is') {
+                $post = $this->parseURL($data['url']);
```
initvector: With this logic, you're assuming `$data` is populated. However, if `$oembedData` were falsy, `$data` would be empty.

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,101 @@
...
+            '/https?:\/\/([a-zA-Z]+\.)?gi?phy?\.(com|is)\/(([\w]+\/)?(([\w]+-)*+)?(?<postID>[a-zA-Z0-9]+))(\/giphy.gif)?/i',
+            $url,
+            $post
+        );
+
+        if (!$post['postID']) {
+            throw new Exception('Unable to get post ID.', 400);
+        }
+
+        return $post;
```
initvector: It looks like the entire purpose of this method is to extract the post ID from a URL. Why would it need to return an array? Could it return just the ID?

library/Vanilla/Embeds/GiphyEmbed.php
```diff
@@ -0,0 +1,101 @@
...
+     * @inheritdoc
+     */
+    public function renderData(array $data): string {
+        $height = $data['height'] ?? 1;
+        $encodedHeight = htmlspecialchars($height);
+
+        $width = $data['width'] ?? 1;
+        $encodedWidth = htmlspecialchars($width);
+
+        $padding = ($encodedHeight/$encodedWidth) * 100;
```
initvector: Before attempting to divide the height by the width, you should verify they're numerical values (`is_numeric`). Otherwise, I think you'll possibly be attempting to divide using actual strings.

tests/Library/Vanilla/Embeds/EmbedManagerTest.php
```diff
@@ -162,6 +164,26 @@ public function provideRenderedData() {
...
+                    "url" => "https://giphy.com/gifs/super-smash-bros-ultimate-jwSlQZnsymUW49NC3R",
+                    "type" => "giphy",
+                    "name" => null,
+                    "body" => null,
+                    "photoUrl" => null,
+                    "height" => 270,
+                    "width" => 480,
+                    "attributes" => [
+                        "postID" => "jwSlQZnsymUW49NC3R",
+                        "url" => "https://giphy.com/gifs/super-smash-bros-ultimate-jwSlQZnsymUW49NC3R"
```
initvector: You don't need URL in here, do you?

tests/Library/Vanilla/Quill/Sanitize/ExternalEmbed/GiphySanitizeTest.php
```diff
@@ -0,0 +1,37 @@
...
+                "insert" => [
+                    "embed-external" => [
+                        "url" => null,
+                        "type" => "giphy",
+                        "name" => null,
+                        "body" => null,
+                        "photoUrl" => null,
+                        "height" => $content,
+                        "width" => $content,
+                        "attributes" => ["postID" => $content, "url" => $content]
```
initvector: Same kinda thing as above. There isn't any need to keep another copy of URL here, is there?

library/Vanilla/Embeds/GiphyEmbed.php
cchabilall83: If width or height is not numeric or not provided, i'm not sure what a proper fallback would be.  Stephane suggested rendering it as an image, but I believe that's out of this scope for this issue

applications/dashboard/src/scripts/app/user-content/embeds/giphy.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Renders giphy embeds.
+ */
+export async function giphyRenderer(element: HTMLElement, data: IEmbedData) {
+    if (data.attributes.postID == null) {
+        throw new Error("Giphy embed fail, the post could not be found");
+    }
+    const width = data.width + "px";
+    element.classList.add("embedGiphy");
+    element.style.width = width || "100%";
```
charrondev: This still will not kick it. "px" is a truthy value.

It needs to be `data.width ? width : "100%";`

PR #7374, opened 2018-06-20
Add Getty Embed

### LINKED ISSUE #7127: Create getty images embed
- The embed renderer in `plugins/rich-editor/embeds/`
- The server rendered blot from the EmbedManager `library/Vanilla/Embeds/`
- The JS needed to render the embed normally in `src/scripts/app/user-content/embeds`

cchabilall83:
Added Getty embed functionality

The url is parsed to extract the post id and the a api called to Getty's oembed endpoint.
The information returned from the call is then parsed and then used to render the post client side.  
To render a post the data passed from the embed has to be loaded into a global object before we load the widgets.js script, not the script will fail.

Embeddable Links
https://www.gettyimages.ca/license/470623275
https://www.gettyimages.ca/event/dress-rehearsal-of-jessica-langs-alvin-ailey-american-dance-theater-debut-775177143#alvin-ailey-dancers-rehearse-en-before-its-world-premiere-at-the-h-picture-id974035560
https://www.gettyimages.ca/detail/news-photo/usain-bolt-of-jamaica-competes-in-the-mens-100-meter-news-photo/589521390#usain-bolt-of-jamaica-competes-in-the-mens-100-meter-semifinal-on-day-picture-id589521390
https://www.gettyimages.ca/detail/photo/duvet-cat-royalty-free-image/628070598

**Some getty images can't be embeded.  It's not apparent and the api returns different errors.
examples**
https://www.gettyimages.ca/license/909486626
https://www.gettyimages.ca/license/735894463

closes #7127


applications/dashboard/scss/editor/_embed.scss
```diff
@@ -264,6 +265,15 @@ $embed-focus-color: #0291DB;
     width: $embedSoundCloud_width;
 }
 
+// Getty
+
+.embed-getty {
+    width: 100%;
+    max-width: $embedGetty_width;
+}
+embedExternal-content.embed-getty{
```
charrondev: This is not a good CSS-selector at all. I'd just remove it.

For reference this will lookup the following:

```html
<embedExternal-content class="embed-getty"></embedExternal-content>
```

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,117 @@
...
+namespace Vanilla\Embeds;
+
+use Exception;
+/**
+ * Getty Embed.
+ */
+class GettyEmbed extends Embed {
+
+    /** @inheritdoc */
+    protected $domains = ['gettyimages.ca', 'gty.im'];
```
charrondev: What about "gettimages.com"? They actually seem to have `gettyimages.{}` for a ton of country specific domains.

image

I don't think we care that much about the rest of them, but we at least need them the `.com` domain.

charrondev: Don't forget the tests.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function renderData(array $data): string {
+        $url = "//www.gettyimages.com/detail/".$data['attributes']['postID'];
+        $encodedURL = htmlspecialchars($url);
+
+        array_walk_recursive($data, function(&$row) {
```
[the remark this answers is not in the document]
cchabilall83: There are so many fields required on the client side, that's why I went this method.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+            if ($oembedData) {
+                $data = $oembedData;
+            }
+
+            if (array_key_exists('html', $data)) {
+                $data['attributes'] = $this->parseResponseHtml($data['html']);
+            }
+
+            if ($post) {
+                $data['attributes']['postID'] = $post['postID'];
```
initvector: We start with an empty `$data` array. __*If*__ we get back some good oEmbed data, we set the result to `$data`. __*If*__ `$data` has an "html" attribute, we parse it and populate its "attributes" field with the result.

With the way things are structured, neither of those things are guaranteed to have happened by the time you reach this line. You're potentially referencing a non-existent key ("attributes") and then treating that potentially non-existent value as an array, referencing another key ("postID").

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function renderData(array $data): string {
+        $url = "//www.gettyimages.com/detail/".$data['attributes']['postID'];
+        $encodedURL = htmlspecialchars($url);
+        $encodedData = json_encode($data);
+        $sanitizedData = htmlspecialchars($encodedData);
+        $encodeID = htmlspecialchars($data['attributes']['id']);
+
+        $result = <<<HTML
+<a id="{$encodeID}" data-json={$sanitizedData} class='gie-single js-gettyEmbed' href="{$encodedURL}">Embed from Getty Images</a>
```
initvector: I'm not a big fan of blindly stuffing the `data-json` attribute, if we can help it. I think we only need eight elements, so it's not making sense to me why we would include almost twice that many. Could you use several specific `data-*` attributes, instead of one giant monolithic one?

initvector: The `data-json` attribute should probably be quoted.

initvector: Let's stick with a consistent quote style. The `class` attribute is using "single quotes", while everything else is double.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return array $data
+     */
+    private function parseResponseHtml(string $html): array {
+        $data =[];
+
+        if (preg_match('/id:\'(?<id>[a-zA-Z0-9-_]+)\'/i', $html, $id)) {
+            $data['id'] = $id['id'];
+        }
+
+        if (preg_match( '/sig:\'(?<sig>[a-zA-Z0-9-_]+=)\'/i', $html, $sig)) {
```
initvector: > Need to get rid of the space after the opening parenthesis.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+            $data['sig'] = $sig['sig'];
+        }
+
+        if (preg_match( '/items:\'(?<item>[0-9-]+)\'/i', $html, $item)) {
```
initvector: > Need to get rid of the space after the opening parenthesis.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+            $data['items'] = $item['item'];
+        }
+
+        if (preg_match( '/caption: (?<isCaption>true|false)/i', $html, $isCaptioned)) {
```
initvector: > Need to get rid of the space after the opening parenthesis.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+            $data['isCaptioned'] = $isCaptioned['isCaption'];
+        }
+
+        if (preg_match( '/is360: (?<is360>true|false)/i', $html, $is360)) {
```
initvector: > Need to get rid of the space after the opening parenthesis.

library/Vanilla/Embeds/GettyEmbed.php
```diff
@@ -0,0 +1,108 @@
...
+            $data['is360'] = $is360['is360'];
+        }
+
+        if (preg_match( '/tld:\'(?<tld>[a-zA-Z]+)\'/i', $html, $tld)) {
```
initvector: > Need to get rid of the space after the opening parenthesis.

cchabilall83:
@initvector  can we merge this?

PR #7380, opened 2018-06-21
PromotedContent: fixed score type validation

### LINKED ISSUE #7312: /api/v1/discussions/promoted: endpoint returning all available discussions
## Details

Calling the promoted endpoint with api/v1/discussions/promoted.json?selector=score&selection=2 should return discussions with score >2. It is currently returning all discussions including discussion with a Null score.

https://github.com/vanilla/vanilla/blob/f8c3b26b71e41e177cea6c22cef87706e41683be/applications/vanilla/modules/class.promotedcontentmodule.php#L435

dannimajdalani:
Closes #7312 

When calling the promotedcontent api with a score selector, the api was returning all discussions. Validation of  $minScore was failing because of its string value. 
**To test**
api/v1/discussions/promoted.json?selector=score&selection=2
should return all discussion with score >2

### DIFF
--- a/applications/vanilla/modules/class.promotedcontentmodule.php
+++ b/applications/vanilla/modules/class.promotedcontentmodule.php
@@ -431,10 +431,7 @@ class PromotedContentModule extends Gdn_Module {
             $minScore = val('Score', $parameters, null);
         }
 
-        if (!is_integer($minScore)) {
-            $minScore = false;
-        }
-
+        $minScore = filter_var($minScore, FILTER_VALIDATE_INT);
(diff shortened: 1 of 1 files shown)


topcheese13:
Your description should explain why you needed to change to make it work not repeat what was said in the issue. I'm assuming it's to allow string integer, i.e. `"15"`. 

```
if (!is_integer($minScore)) {
    $minScore = false;
}
```
to
```
$minScore = filter_var($minScore, FILTER_VALIDATE_INT);
```

PR #7381, opened 2018-06-21
Fix for colour pickers in modals

slafleche:
Vanilla doesn't initialize colour inputs in modals. The simplest fix is just to always check for color inputs in modals in the dashboard.


slafleche:
Closing this issue for now. We're not doing any changes in the short term.

PR #7389, opened 2018-06-26
Add delete method to users API v2 endpoint

### LINKED ISSUE #7384: Setting the user delete method / strategy in API v2
Currently the vanilla api doesn't seem to support setting the delete method for users:

- Keep content
- Blank user
- Delete user content

### Functionality present in the v1 api:

image

### Missing from api v2
(unless I'm misunderstanding the "expand" option?)

image

garygreen:
Fixes #7384

### DIFF
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -57,16 +57,26 @@ class UsersApiController extends AbstractApiController {
      * Delete a user.
      *
      * @param int $id The ID of the user.
+     * @param array $body The request body.
      * @throws NotFoundException if the user could not be found.
      */
-    public function delete($id) {
+    public function delete($id, array $body) {
         $this->permission('Garden.Users.Delete');
+        
+        $this->idParamSchema()->setDescription('Delete a user.');
 
-        $in = $this->idParamSchema()->setDescription('Delete a user.');
+        $in = $this->schema([
+            'deleteMethod:s?' => [
+                'description' => 'The deletion method / strategy.',
+                'enum' => ['keep', 'wipe', 'delete'],
+                'default' => 'delete',
+            ]
+        ], 'in');
         $out = $this->schema([], 'out');
+        $body = $in->validate($body);
 
         $this->userByID($id);
-        $this->userModel->deleteID($id);
+        $this->userModel->deleteID($id, ['DeleteMethod' => $body['deleteMethod']]);
     }
 
     /**


garygreen:
Hmm looks like there is a problem with this PR, the `deleteMethod` is only received when you send with query string parameters, it doesn't appear to be set when you send a with json body. @tburry / @linc could you shed some light on how to get the delete method to receive the proper parameters?

garygreen:
@DaazKu Cheers buddy, works a treat. All ready for review now :-)

charrondev:
While this may work, the schema doesn't actually show correctly in swagger. Be sure to double check in `/settings/swagger` that your fields are showing correctly.

If you need an example of an ID schema with a body as well take a look at `UsersApiController::patch()` and `UsersApiController::userPatchSchema()`.

garygreen:
Not sure what swagger is tbh. Is that something I can access thru the dashboard? Is it a documentation thing? I can certainly add/fix it but would like to get official response from Vanilla if this is something they are interested in adding first.

charrondev:
Developer with @vanilla/developers here. I'd accept this contribution if it works and properly shows up in the dashboard documentation.

I realize now it isn't actually released in the 2.6 open source release, but it is available in the dashboard by default on master branch (what will be 2.7 release in the near future).

Navigate to `/settings/swagger` in the dashboard for an in-dashboard dynamic documentation. It's the same setup that we use for our API reference here: https://docs.vanillaforums.com/help/apiv2/swagger/, except built dynamically for your local site. If the schema's are set up correctly, you new endpoint, and its parameters should be visible on the `/settings/swagger` of your local machine.

charrondev:
You might also want to check out this doc if you haven't already https://docs.vanillaforums.com/developer/apiv2/api-endpoints/#the-anatomy-of-a-controller-action.

If you feel like something is missing there, [opening an issue on our docs repo](https://github.com/vanilla/docs/issues/new) can help us fill in the gaps.

garygreen:
Thanks for the information. So it looks like I just need to add a 3rd parameter to the schema (which matches the function name) for it to appear in swagger?

I tried accessing swagger locally but for some reason it appears to be loading and then very quickly redirecting onto a 404 page

swagger-404

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -57,16 +57,26 @@ public function __construct(
...
+            'deleteMethod:s?' => [
+                'description' => 'The deletion method / strategy.',
+                'enum' => ['keep', 'wipe', 'delete'],
+                'default' => 'delete',
+            ]
+        ], 'in');
+        $body = $in->validate($body);
+
         $in = $this->idParamSchema()->setDescription('Delete a user.');
         $out = $this->schema([], 'out');
```
initvector: I think you'll be able to get this to properly display in the API docs if you juggle the order a little, similar to other endpoints.

```php
$this->idParamSchema()->setDescription('Delete a user.');
$in = $this->schema([
    'deleteMethod:s?' => [
        'description' => 'The deletion method / strategy.',
        'enum' => ['keep', 'wipe', 'delete'],
        'default' => 'delete',
    ]
], 'in');
$out = $this->schema([], 'out');

$body = $in->validate($body);
```

garygreen:
Ok I made that change and it appears to be in swagger correctly now:

image

initvector:
Splendid. Thanks @garygreen !

PR #7397, opened 2018-06-27
Add Bookmark Discussion Integration Test

cchabilall83:
These ApiV0 integrations tests, tests the users ability to a bookmark on a discussion and remove a bookmark on a discussion. The remove bookmark test (testRemoveDiscussionBookMark()) depends on the add bookmark test (testDiscussionAddBookMark()) to setup a bookmarked discussion.

closes Vanilla#7087

### DIFF
--- a/tests/APIv0/SmokeTest.php
+++ b/tests/APIv0/SmokeTest.php
@@ -62,6 +62,23 @@ class SmokeTest extends BaseTest {
     }
 
     /**
+     * Get a single discussion from an array of Discussions.
+     *
+     * @param array $discussions
+     * @return array $discussion
+     * @throws \Exception
+     */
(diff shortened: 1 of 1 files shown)


tests/APIv0/SmokeTest.php
```diff
@@ -61,6 +61,23 @@ public function setTestUser($testUser) {
         return $this;
     }
 
+    /**
+     * Get a single discussion from an array of Discussions.
+     *
+     * @param array $discussions
+     * @return array $discussion
```
initvector: The format of the [@return tag](http://docs.phpdoc.org/references/phpdoc/tags/return.html) is:

```
@return [Type] [<description>]
```

Including the name of the variable you're returning isn't necessary, because the name doesn't come along as part of the result.

PR #7428, opened 2018-07-03
Fixed broken soundcloud embed urls

cchabilall83:
This fix allows for SoundCloud "sets" and "user" pages to be embeddable.  The embed url is different in each case, as is the the specific id that is used.  The parsehtml function was modified to filter out whether the url is a track, set or user and returns a post id and embed url for each case.  Additional testing was added to ensure the proper response.

Test links

Closes Vanilla#7366

### DIFF
--- a/library/Vanilla/Embeds/SoundCloudEmbed.php
+++ b/library/Vanilla/Embeds/SoundCloudEmbed.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @copyright 2009-2018 Vanilla Forums Inc.
- * @license GPL-2.0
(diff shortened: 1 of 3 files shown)


library/Vanilla/Embeds/SoundCloudEmbed.php
```diff
@@ -65,7 +66,7 @@ public function renderData(array $data): string {
     /**
      * Parses the oembed repsonse html for permalink and other data.
```
charrondev: We should add a larger detailed comment here about the different types of soundcloud embeds we are trying to parse here and give an example of each type.

PR #7435, opened 2018-07-04
Add CodePen Embed

cchabilall83:
This feature adds  CodePen embed functionality.

A request is made to CodePen's API endpoint, and the data returned is parsed and used to reconstruct CodePen's iframe embed.  The height, id, class, iframe src and some styles where extracted from the html code sent back in the response.  The iframe is responsive so not additional css styles are needed.  

Additional media, embed, quill tests were added.

Closes Vanilla#7408

### DIFF
--- a/applications/dashboard/src/scripts/app/user-content/embeds/codepen.ts
+++ b/applications/dashboard/src/scripts/app/user-content/embeds/codepen.ts
@@ -3,7 +3,7 @@
  * @license https://opensource.org/licenses/GPL-2.0 GPL-2.0
  */
 
-import { registerEmbed, IEmbedData, FOCUS_CLASS, IEmbedElements } from "@dashboard/embeds";
+import { registerEmbed, IEmbedData, IEmbedElements } from "@dashboard/embeds";
 
 registerEmbed("codepen", codePenRenderer);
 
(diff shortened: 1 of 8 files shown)


applications/dashboard/src/scripts/app/user-content/embeds/codepen.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Renders codepen embeds.
+ */
+export async function codePenRenderer(elements: IEmbedElements, data: IEmbedData) {
+    const contentElement = elements.content;
+    const height = data.height ? data.height : 300;
+    const iframe = document.createElement("iframe");
+    iframe.setAttribute("id", data.attributes.id);
+    iframe.setAttribute("src", data.attributes.embedUrl);
+    iframe.setAttribute("height", height.toString());
+    iframe.setAttribute("class", "cp_embed_iframe");
```
initvector: This class isn't required, is it?

library/Vanilla/Embeds/CodePenEmbed.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public function renderData(array $data): string {
+        $height = $data['height'] ?? '';
+        $embedUrl = $data['attributes']['embedUrl'] ?? '';
+        $style = $data['attributes']['style'] ?? '';
+        $id = $data['attributes']['id'];
+
+        $encodedHeight = htmlspecialchars($height);
+        $encodedEmbedUrl = htmlspecialchars($embedUrl);
+        $encodedStyle = htmlspecialchars($style);
```
initvector: I'm not a big fan of allowing CodePen to insert arbitrary styles into Vanilla pages. Can you update `parseHtml` to only extract certain whitelisted styles?

cchabilall83: I think we mainly care about width and overflow, so I will parse those out.

PR #7441, opened 2018-07-09
Add Token Signing Trait and UserAuthenticationNonceModel functionality

cchabilall83:
The Token Signing trait consists of utility methods that are shared across the **AccessTokenModel** and **UserAuthenticationNonceModel**.  The methods (see below) were extracted from the AccessTokenModel.  There was more functionality added to the UserAuthenticationNonceModel as well, this includes the **update**, **getNonce**, **issue**, **consume** an **verify** methods.

Additionally the TokenModel fixture was created to unit test the the newly created trait with the TokenSigningTests class.  UserAuthenticationNonceModelTest class  was created to test the UserAuthenticationNonceModel and AccessTokenModelTest class was modify to account for the changes made.

Methods Extracted from AccessTokenModel

- secret (variable)
- getSecret
- setSecret
- base64urlEncode
- base64urlDecode
- randomToken
- verifyTokenSignature
- signToken
- tokenError

**These were not listed in the issue but I notice there were used by the extracted methods.**

- toTimestamp
- encodeDate
- decodeDate

Closes Vanilla#7402

### DIFF
--- a/applications/dashboard/models/class.userauthenticationnoncemodel.php
+++ b/applications/dashboard/models/class.userauthenticationnoncemodel.php
@@ -1,19 +1,23 @@
 <?php
 /**
- * UserAuthenticationNonce model.
- *
+ * @author Chris Chabilall chris.c@vanillaforums.com
  * @copyright 2009-2018 Vanilla Forums Inc.
  * @license http://www.opensource.org/licenses/gpl-2.0.php GNU GPL v2
  */
 
 /**
- * Handles user data.
+ * UserAuthenticationNonce model.
+ * Handles user data and issuing and consuming nonce's.
  */
 class UserAuthenticationNonceModel extends Gdn_Model {
 
     use \Vanilla\PrunableTrait, \Vanilla\TokenSigningTrait;
 
+    /**
+     * @var string $tokenIdentifier Used to deteremine what type of token is generated.
+     */
     protected static $tokenIdentifier = "nonce";
+
     /**
      * Class constructor. Defines the related database table name.
      *
@@ -33,11 +37,6 @@ class UserAuthenticationNonceModel extends Gdn_Model {
     public function insert($fields) {
         $this->prune();
 
-        if (!isset($fields['Timestamp'])) {
-            $fields['Timestamp'] = date(MYSQL_DATE_FORMAT);
-        }
-
-        $this->encodeRow($fields);
         parent::insert($fields);
         if (!empty($this->Database->LastInfo['RowCount'])) {
             $result = $fields['Nonce'];
@@ -51,32 +50,32 @@ class UserAuthenticationNonceModel extends Gdn_Model {
      * {@inheritdoc}
      */
     public function update($fields, $where = false, $limit = false) {
-        $this->encodeRow($fields);
         $result = parent::update($fields, $where, $limit);
         return $result;
     }
 
     /**
-     * Get an access token by its numeric ID.
+     * Get a matching nonce from the UserAuthenticationNonce table.
      *
-     * @param int $nonce
-     * @param string $datasetType
-     * @param array $options
+     * @param int $nonce The nonce to be looked up.
+     * @param string $datasetType The datatype returned from the query.
      * @return array|bool
      */
-    public function getNonce($nonce, $datasetType = DATASET_TYPE_ARRAY, $options = []) {
+    public function getNonce($nonce, $datasetType = DATASET_TYPE_ARRAY) {
         $row = $this->getWhere(['Nonce' => $nonce])->firstRow($datasetType);
         return $row;
     }
 
 
     /**
-     * @param string $expires
-     * @param string $type
-     * @return string
-     * @throws Gdn_UserException
+     * Issue a signed Nonce.
+     *
+     * @param string $expires The expiration time of the nonce.
+     * @param string $type The type of the of token.
+     * @return string $nonce the signed nonce.
+     * @throws Gdn_UserException Unable to generate a signed nonce.
      */
-    public function issue($expires = '5 minutes', $type = 'system') {
+    public function issue($expires = '5 minutes', $type = 'system') : String {
         if ($expires instanceof  DateTimeInterface) {
             $expireDate = $expires->format(MYSQL_DATE_FORMAT);
         } else {
@@ -89,13 +88,13 @@ class UserAuthenticationNonceModel extends Gdn_Model {
             throw new \Exception("Unable to generate Nonce", 500);
         }
 
-        $token = $this->insert([
+        $result = $this->insert([
             'Nonce' => $nonce,
             'Token' => $type,
             'Timestamp' => $expireDate,
         ]);
 
-        if (!$token) {
+        if (!$result) {
             throw new Gdn_UserException($this->Validation->resultsText(), 400);
         }
 
@@ -103,7 +102,12 @@ class UserAuthenticationNonceModel extends Gdn_Model {
     }
 
     /**
-     * @param $nonce
+     * Consumes the nonce and invalidates the timestamp so it can't be used again.
+     * The timestampo column is not nullable and the zero dates aren't allowed, so the date
+     * is set to the Unix epoch time.
+     *
+     * @param string $nonce The nonce to be consumed.
+     * @throws Exception Unable to find nonce.
      */
     public function consume(string $nonce) {
         $row = $this->getNonce($nonce);
@@ -112,15 +116,18 @@ class UserAuthenticationNonceModel extends Gdn_Model {
                 ['Timestamp' => Gdn_Format::toDateTime($this->toTimestamp('1971-01-01 00:00:01'))],
                 ['Nonce' => $nonce]
             );
+        } else {
+            throw new \Exception("Unable to find Nonce", 500);
         }
     }
 
     /**
-     * @param string $nonce
-     * @param bool $consume
-     * @param bool $throw
-     * @return bool
-     * @throws Exception
+     * Verifies a nonce can be consumed and consumes it if the flag is true.
+     *
+     * @param string $nonce The nonce to verify.
+     * @param bool $consume Flag to determine if the nonce is to be consumed.
+     * @param bool $throw Whether or not to throw an exception on a verification error.
+     * @return bool If the nonce is verified it returns true, false if otherwise.
      */
     public function verify(string $nonce = '', bool $consume = true, bool $throw = false): bool {
         // First verify the token without going to the database.
@@ -147,5 +154,4 @@ class UserAuthenticationNonceModel extends Gdn_Model {
 
         return true;
     }
-
 }
--- a/applications/dashboard/models/class.userauthenticationnoncemodel.php
+++ b/applications/dashboard/models/class.userauthenticationnoncemodel.php
@@ -11,8 +11,9 @@
  */
 class UserAuthenticationNonceModel extends Gdn_Model {
 
-    use \Vanilla\PrunableTrait;
+    use \Vanilla\PrunableTrait, \Vanilla\TokenSigningTrait;
 
+    protected static $tokenIdentifier = "nonce";
     /**
      * Class constructor. Defines the related database table name.
      */
@@ -31,7 +32,117 @@ class UserAuthenticationNonceModel extends Gdn_Model {
         if (!isset($fields['Timestamp'])) {
             $fields['Timestamp'] = date(MYSQL_DATE_FORMAT);
         }
+        if (empty($fields['Token'])) {
+            $fields['Token'] = $this->randomToken();
+        }
 
-        return parent::insert($fields);
+        $this->encodeRow($fields);
+        parent::insert($fields);
+        if (!empty($this->Database->LastInfo['RowCount'])) {
+            $result = $fields['Token'];
+        } else {
+            $result = false;
+        }
+
+        return $result;
+    }
+    /**
+     * {@inheritdoc}
+     */
+    public function update($fields, $where = false, $limit = false) {
+        $this->encodeRow($fields);
+        return parent::update($fields, $where, $limit);
+    }
+
+    /**
+     * Get an access token by its numeric ID.
+     *
+     * @param int $nonce
+     * @param string $datasetType
+     * @param array $options
+     * @return array|bool
+     */
+    public function getNonce($nonce, $datasetType = DATASET_TYPE_ARRAY, $options = []) {
+        $row = $this->getWhere(['Nonce' => $nonce])->firstRow($datasetType);
+        return $row;
+    }
+
+
+    /**
+     * @param string $expires
+     * @param string $type
+     * @return string
+     * @throws Gdn_UserException
+     */
+    public function issue($expires = '5 minutes', $type = 'system') {
+        if ($expires instanceof  DateTimeInterface) {
+            $expireDate = $expires->format(MYSQL_DATE_FORMAT);
+        } else {
+            $expireDate = Gdn_Format::toDateTime($this->toTimestamp($expires));
+        }
+
+        $token = $this->insert([
+            'Type' => $type,
+            'Timestamp' => $expireDate,
+        ]);
+
+        if (!$token) {
+            throw new Gdn_UserException($this->Validation->resultsText(), 400);
+        }
+
+        $Nonce = $this->signToken($token, $expireDate);
+
+        return $Nonce;
+    }
+
+    /**
+     * @param $nonce
+     */
+    public function consume(string $nonce) {
+        $row = $this->getNonce($nonce);
+        if ($row) {
+            $this->update(
+                ['Timestamp' => '0000-00-00 00:00:00'],
+                ['Nonce' => $nonce],
+                1
+            );
+        }
+    }
+
+    /**
+     * @param string $nonce
+     * @param bool $consume
+     * @param bool $throw
+     * @return bool
+     * @throws Exception
+     */
+    public function verify(string $nonce = '', bool $consume = true, bool $throw = false) {
+        // First verify the token without going to the database.
+        if (!$this->verifyTokenSignature($nonce, self::$tokenIdentifier, $throw)) {
+            return false;
+        }
+
+        $token = $this->trim($nonce);
+
+        $row = $this->getToken($token);
+
+        if (!$row) {
+            return $this->tokenError('The nonce was not found in the database.', 401, $throw);
+        }
+
+        // Check the expiry date from the database.
+        $dbExpires = $this->toTimestamp($row['Timestamp']);
+        if ($dbExpires === '0000-00-00 00:00:00') {
+            return $this->tokenError('Nonce was already used.', 401, $throw);
+        } elseif ($dbExpires < time()) {
+            return $this->tokenError('Nonce has expired.', 401, $throw);
+        }
+
+        if ($consume) {
+            $this->consume($nonce);
+        }
+
+        return $row;
+
     }
 }
--- a/applications/dashboard/models/class.userauthenticationnoncemodel.php
+++ b/applications/dashboard/models/class.userauthenticationnoncemodel.php
@@ -12,6 +12,10 @@ class UserAuthenticationNonceModel extends Gdn_Model {
 
     use \Vanilla\PrunableTrait;
     use \Vanilla\TokenSigningTrait;
+
+    /**
+     * Timestamp a consumed nonce will set to.
+     */
     const CONSUMED_TIMESTAMP = "1971-01-01 00:00:01";
 
     /**
@@ -19,7 +23,7 @@ class UserAuthenticationNonceModel extends Gdn_Model {
      *
      * @param string $secret The secret used to sign access tokens for the client.
      */
-    public function __construct($secret = '') {
+    public function __construct($secret = null) {
         parent::__construct('UserAuthenticationNonce');
         $this->setPruneField('Timestamp');
         $this->setPruneAfter('45 minutes');
@@ -37,12 +41,8 @@ class UserAuthenticationNonceModel extends Gdn_Model {
         if (!isset($fields['Timestamp'])) {
             $fields['Timestamp'] = date(MYSQL_DATE_FORMAT);
         }
-        parent::insert($fields);
-        if (!empty($this->Database->LastInfo['RowCount'])) {
-            $result = true;
-        } else {
-            $result = false;
-        }
+        $result = parent::insert($fields);
+
         return $result;
     }
 
@@ -50,7 +50,7 @@ class UserAuthenticationNonceModel extends Gdn_Model {
      * Issue a signed Nonce.
      *
      * @param string $expires The expiration time of the nonce.
-     * @param string $type The type of the of token.
+     * @param string $type The type of nonce.
      * @return string $nonce The signed nonce.
      * @throws Gdn_UserException Unable to generate a signed nonce.
      */
@@ -65,7 +65,7 @@ class UserAuthenticationNonceModel extends Gdn_Model {
             'Timestamp' => $expireDate,
         ]);
 
-        if (!$result) {
+        if ($result === false) {
             throw new Gdn_UserException($this->Validation->resultsText(), 400);
         }
         return $nonce;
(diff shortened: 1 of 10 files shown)


library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Trait TokenSigningTrait.
+ *
+ * Methods to be used for token generation and signing.
+ */
+trait TokenSigningTrait {
+    /**
+     * @var string $secret The secret used to sign the token.
+     */
+    public $secret;
```
initvector: Should be private.

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+        return $this;
+    }
+
+    /**
+     * Base64 Encode a string, but make it suitable to be passed in a url.
+     *
+     * @param string $str The string to encode.
+     * @return string The encoded string.
+     */
+    protected static function base64urlEncode($str) {
```
initvector: Should be private.

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+        return trim(strtr(base64_encode($str), '+/', '-_'), '=');
+    }
+
+    /**
+     * Decode a string that was encoded using base64UrlEncode().
+     *
+     * @param string $str The encoded string.
+     * @return string The decoded string.
+     */
+    protected static function base64urlDecode($str) {
```
initvector: Should be private.

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+    public function randomToken() {
+        $token = self::base64urlEncode(openssl_random_pseudo_bytes(24));
+        return $token;
+    }
+
+    /**
+     * Verify a token's signature.
+     *
+     * @param string $accessToken The full access token.
+     * @oaran string $tokenIdentifier the model that is calling verify signature.
```
initvector: Might need to update that tag.

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+     * @param bool $throw Whether or not to throw an exception on a verification error.
+     * @return bool Returns **true** if the token's expiry date and signature is valid or **false** otherwise.
+     * @throws \Exception Throws an exception if the token is invalid and {@link $throw} is **true**.
+     */
+    public function verifyTokenSignature($accessToken, $tokenIdentifier, $throw = false) {
```
initvector: You should not need the `$tokenIdentifier` parameter. Per the issue:

> Most of the methods can be used as-is, but some of them mention "access token" by name. Turn this into a protected static property or something like that. The trait can use "token" while the models can use "access token" and "nonce".

You would need to have a protected static property on this trait: `$tokenIdentifier` (or whatever). You would then reference that property in these error messages, instead of concatenating in this parameter. Classes implementing this trait could override the property, automatically customizing their error messages.

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Generate and sign a token.
+     *
+     * @param string $expires When the token expires.
+     * @return string
+     */
+    public function randomSignedToken($expires = '2 months') {
+        return $this->signToken($this->randomToken(), $expires);
+    }
+
```
initvector: Don't need this empty line.

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -28,10 +37,121 @@ public function __construct() {
...
+     *
+     * @param int $nonce The nonce to be looked up.
+     * @param string $datasetType The datatype returned from the query.
+     * @return array|bool
+     */
+    public function getNonce($nonce, $datasetType = DATASET_TYPE_ARRAY) {
+        $row = $this->getWhere(['Nonce' => $nonce])->firstRow($datasetType);
+        return $row;
+    }
+
```
initvector: Don't need that empty line.

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -28,10 +37,121 @@ public function __construct() {
...
+        parent::insert($fields);
+        if (!empty($this->Database->LastInfo['RowCount'])) {
+            $result = $fields['Nonce'];
+        } else {
+            $result = false;
+        }
+
+        return $result;
+    }
+    /**
```
initvector: Need an empty line between these two.

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -1,25 +1,34 @@
...
      * Class constructor. Defines the related database table name.
+     *
+     * @param string $secret The secret used to sign access tokens for the client.
      */
-    public function __construct() {
+    public function __construct($secret = '') {
         parent::__construct('UserAuthenticationNonce');
         $this->setPruneField('Timestamp');
         $this->setPruneAfter('45 minutes');
+        $this->secret = $secret ?: c('Garden.Cookie.Salt');
```
initvector: Use the trait's setter for this property. Don't directly access it (it should be private).

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,199 @@
...
+        return $arr[1];
+    }
+
+    /**
+     * Generate and sign a token.
+     *
+     * @param string $expires When the token expires.
+     * @return string
+     */
+    public function randomSignedToken($expires = '2 months') {
```
initvector: Why would this be better in here, instead of staying in `AccessTokenModel`? It looks like it might only be used there.

plugins/vanillaconnect/addon.json
```diff
@@ -7,7 +7,7 @@
     "settingsUrl": "/settings/vanillaconnect",
     "usePopupSettings": false,
     "settingsPermission": "Garden.Settings.Manage",
-    "hidden": true,
+    "hidden": false,
```
initvector: Why did this change?

tests/Models/UserAuthenticationNonceModelTest.php
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @throws \Gdn_UserException
+     */
+    public function testConsume() {
+        $model = new UserAuthenticationNonceModel('hhh');
+        $issuedNonce = $model->issue();
+        $model->consume($issuedNonce);
+        $consumedNonce = $model->getNonce($issuedNonce);
+
+        $this->assertEquals("1971-01-01 00:00:01", $consumedNonce['Timestamp']);
```
initvector: Instead of comparing the dates, I think it'd be better to expect this test to throw an exception of a specific type and with a specific message, similar to several of the other tests. Issue the nonce, consume it, then attempt to verify it.

tests/fixtures/src/TokenModel.php
```diff
@@ -0,0 +1,27 @@
...
+class TokenModel {
+
+    use TokenSigningTrait;
+    public $tokenIdentifier;
+    public $secret;
+
+    public function __construct($secret = '') {
+        $this->secret = 'sss';
+    }
+
```
initvector: Don't need this empty line.

tests/fixtures/src/TokenModel.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ */
+class TokenModel {
+
+    use TokenSigningTrait;
+    public $tokenIdentifier;
+    public $secret;
+
+    public function __construct($secret = '') {
+        $this->secret = 'sss';
```
initvector: Same as in the other classes that implement this trait: don't directly access it. Use the setter.

initvector: Why does your constructor have a `$secret` parameter if it's not used?

tests/Models/UserAuthenticationNonceModelTest.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace VanillaTests\Models;
+
+use VanillaTests\SharedBootstrapTestCase;
+use UserAuthenticationNonceModel;
+use VanillaTests\SiteTestTrait;
+
+/**
+ * Test the {@link AccessTokenModel}.
```
initvector: Are we testing `AccessTokenModel` or `UserAuthenticationNonceModel`?

tests/Models/TokenSigningTests.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+namespace VanillaTests\Models;
+
+use VanillaTests\SharedBootstrapTestCase;
+use VanillaTests\SiteTestTrait;
+use VanillaTests\Fixtures\TokenModel;
+
+/**
+ * Test the {@link TokenModel}.
```
initvector: I think we're supposed to be testing the `TokenSigningTrait`.

tests/Models/TokenSigningTests.php
```diff
@@ -0,0 +1,69 @@
...
+ * Used to test the token generation and signing utility methods.
+ */
+class TokenSigningTests extends SharedBootstrapTestCase {
```
initvector:
```php
class TokenSigningTraitTest extends SharedBootstrapTestCase {
```

Make sure you update your filename too.

tests/Models/TokenSigningTests.php
```diff
@@ -0,0 +1,69 @@
...
+    use SiteTestTrait;
+
+    /**
+     * Tests random token generation and signing.
+     */
+    public function testVerifyRandomTokenSignature() {
+        $model = new TokenModel();
+        $model->tokenIdentifier ='nonce';
```
initvector: This should probably be set in the class.

library/Vanilla/TokenSigningTrait.php
cchabilall83: I use this in the TokenSigningTraitTest

library/Vanilla/TokenSigningTrait.php
cchabilall83: The reason I went with is parameter was because I kept getting .  Maybe i'm not implementing it properly.

Fatal error: VanillaTests\Fixtures\TokenModel and Vanilla\TokenSigningTrait define the same property ($tokenIdentifier) in the composition of VanillaTests\Fixtures\TokenModel. However, the definition differs and is considered incompatible. Class was composed in /srv/vanilla-repositories/vanilla/tests/fixtures/src/TokenModel.php on line 17
unit-test-php-fpm    |

cchabilall83: i was able to set it in the constructor, nevermind

initvector: There's a [documented limitation of trait properties](http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.properties) that I was unaware of:

> If a trait defines a property then a class can not define a property with the same name unless it is compatible (same visibility and initial value), otherwise a fatal error is issued.

Updating the value in the constructor is a valid way to get around this, but if we're going that route, let's not make it a static property. It doesn't make a lot of sense to me for Vanilla to update a static property every time the object is instantiated.

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -28,10 +37,121 @@ public function __construct() {
...
+    public function issue($expires = '5 minutes', $type = 'system') : String {
+        if ($expires instanceof  DateTimeInterface) {
+            $expireDate = $expires->format(MYSQL_DATE_FORMAT);
+        } else {
+            $expireDate = Gdn_Format::toDateTime($this->toTimestamp($expires));
+        }
+
+        $token = $this->randomToken();
+        $nonce = $this->signToken($token, $expireDate);
+        if (!$nonce) {
```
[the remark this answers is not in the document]
cchabilall83: Should I remove this exception, since we throw an exception in signToken?

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -28,10 +37,121 @@ public function __construct() {
...
+    }
+
+    /**
+     * Get a matching nonce from the UserAuthenticationNonce table.
+     *
+     * @param int $nonce The nonce to be looked up.
+     * @param string $datasetType The datatype returned from the query.
+     * @return array|bool
+     */
+    public function getNonce($nonce, $datasetType = DATASET_TYPE_ARRAY) {
```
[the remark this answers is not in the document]
cchabilall83: I can't access parent:getID in UserAuthenticationNonceModelTest, is there a work around?

applications/dashboard/models/class.userauthenticationnoncemodel.php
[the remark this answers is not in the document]
initvector: Yeah. Either `signToken` will return the signed token or it'll throw an exception, which would exit the method and start bubbling up the stack until it was caught. I don't think you'll ever get a falsy value. Unless I'm missing something, the contents of this condition don't seem reachable.

applications/dashboard/models/class.userauthenticationnoncemodel.php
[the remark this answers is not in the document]
initvector: You should be inheriting `getID` from the parent class. You wouldn't access it with `parent::getID()`. You access it with `$this->getID()`. It's a public method, so you should be able to access it from anywhere.

library/core/class.accesstokenmodel.php
```diff
@@ -26,6 +24,7 @@ public function __construct($secret = '') {
         parent::__construct('AccessToken');
         $this->PrimaryKey = 'AccessTokenID';
         $this->secret = $secret ?: c('Garden.Cookie.Salt');
+        $this->tokenIdentifier ='access token';
```
initvector: Spacing... ☹️

tests/Models/UserAuthenticationNonceModelTest.php
```diff
@@ -0,0 +1,42 @@
...
+        $model = new UserAuthenticationNonceModel('hhh');
+        $nonce = $model->issue();
+        $this->assertEquals(true,  $model->verify($nonce));
+    }
+
+    /**
+     * That a nonce can be consumed.
+     *
+     * @expectedException \Exception
+     * @expectedExceptionMessage Nonce has expired.
```
initvector: I think you'll find this fails when you use the proper exception message: Nonce was already used.

tests/fixtures/src/TokenModel.php
```diff
@@ -0,0 +1,27 @@
...
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license http://www.opensource.org/licenses/gpl-2.0.php GNU GPL v2
+ */
+
+namespace VanillaTests\Fixtures;
+
+use Vanilla\TokenSigningTrait;
+
+/**
+ * Class TokenModel.
```
initvector: Again, let's kill off class names in class doc blocks.

tests/fixtures/src/TokenModel.php
```diff
@@ -0,0 +1,27 @@
...
+ * Used for TokenSigningTraitTest.
```
initvector: If this is *only* for testing that trait, the model should probably have a name to reflect that. As it stands, its name implies it's a general-use class for testing tokens. I think the description should be updated to reflect it's a class for testing tokens *or* the class should be renamed to better indicate it's only for testing that one trait.

tests/Models/TokenSigningTraitTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $model = new TokenModel();
+        $token = $model->randomSignedToken('last month');
+        $this->assertEquals(false, $model->verifyTokenSignature($token, true));
+    }
+
+    /**
+     * An altered token signature shouldn't verify.
+     *
+     * @expectedException \Exception
+     * $expectedExceptionMessage Invalid signature.
```
initvector: Doc tags start with an `@`.

tests/Models/TokenSigningTraitTest.php
```diff
@@ -0,0 +1,64 @@
...
+    public function testBadSignature() {
+        $model = new TokenModel();
+        $token = $model->randomSignedToken().'!';
+        $this->assertEquals(false, $model->verifyTokenSignature($token,true));
+    }
+
+    /**
+     * A nonsense token shouldn't verify.
+     *
+     * @expectedException \Exception
```
initvector: Could use an `@expectedExceptionMessage` in here.

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -31,7 +37,90 @@ public function insert($fields) {
...
+     * Issue a signed Nonce.
+     *
+     * @param string $expires The expiration time of the nonce.
+     * @param string $type The type of the of token.
+     * @return string $nonce The signed nonce.
+     * @throws Gdn_UserException Unable to generate a signed nonce.
+     */
+    public function issue($expires = '5 minutes', $type = 'system'): string {
+        $token = $this->randomToken();
+        $nonce = $this->signToken($token, $expires);
```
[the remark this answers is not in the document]
cchabilall83: I've moved randomSignedToken back to the accesstokenmodel

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -1,25 +1,35 @@
...
      * Class constructor. Defines the related database table name.
+     *
+     * @param string $secret The secret used to sign access tokens for the client.
      */
-    public function __construct() {
+    public function __construct($secret = null) {
         parent::__construct('UserAuthenticationNonce');
         $this->setPruneField('Timestamp');
         $this->setPruneAfter('45 minutes');
+        $this->setSecret($secret);
```
initvector: Bringing back what I said before: Can we have this default to the value of `Garden.Cookie.Salt`?

I'm not seeing any current usage of `UserAuthenticationNonceModel` passing anything to the constructor. That would mean after this change, there wouldn't be a secret to sign any of the nonces.

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -31,7 +41,85 @@ public function insert($fields) {
...
+     * Consumes the nonce and invalidates the timestamp so it can't be used again.
+     *
+     * @param string $nonce The nonce to be consumed.
+     * @throws Exception if unable to find nonce.
+     */
+    public function consume(string $nonce) {
+        $row = $this->getID($nonce, DATASET_TYPE_ARRAY);
+        if ($row) {
+            // Timestamp cannot be null or zero. Use a constant date for consumed nonces.
+            parent::update(
```
initvector: What would be the benefit of calling the parent's `update` method, instead of the inherited method (e.g. `$this->update()`)?

applications/dashboard/models/class.userauthenticationnoncemodel.php
```diff
@@ -31,7 +41,85 @@ public function insert($fields) {
...
+                ['Timestamp' => self::CONSUMED_TIMESTAMP],
+                ['Nonce' => $nonce]
+            );
+        } else {
+            throw new \Exception("Unable to find Nonce", 500);
```
initvector: For the sake of consistency, let's make this a lower-case "nonce".

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,186 @@
...
+ * @license http://www.opensource.org/licenses/gpl-2.0.php GNU GPL v2
+ */
+
+namespace Vanilla;
+
+/**
+ * Methods to be used for token generation and signing.
+ */
+trait TokenSigningTrait {
+    /** @var string $secret The secret used to sign the token. */
```
initvector: Don't need to recap the property name here

tests/Models/TokenSigningTraitTest.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * An expired token shouldn't verify.
+     *
+     * @expectedException \Exception
+     * @expectedExceptionMessage Your nonce has expired.
+     */
+    public function testExpiryDate() {
+        $model = new TokenTestingModel();
+        $token = $model->signToken($model->randomToken(), 'last year');
+        $this->assertEquals(false, $model->verifyTokenSignature($token, true));
```
initvector: If we're expecting exceptions, it doesn't seem like these assertions are actually doing anything. Do they need to be there or can we remove them?

library/Vanilla/TokenSigningTrait.php
```diff
@@ -0,0 +1,186 @@
...
+     */
+    public function verifyTokenSignature($accessToken, $throw = false) {
+        $parts = explode('.', $accessToken);
+
+        if (empty($accessToken)) {
+            return $this->tokenError('Missing '.$this->tokenIdentifier, 401, $throw);
+        }
+
+        if (count($parts) !== 4) {
+            return $this->tokenError($this->tokenIdentifier.' missing parts.', 401, $throw);
```
initvector: This error message could potentially read "nonce missing parts.", which doesn't align with the other messages in this method all starting with a capital letter. You can prefix the message with "Your", like similar messages, below.

tests/fixtures/src/TokenTestingModel.php
```diff
@@ -0,0 +1,23 @@
...
+class TokenTestingModel {
+
+    use TokenSigningTrait;
+
+    /**
+     * TokenModel constructor.
+     */
+    public function __construct() {
+        $this->setSecret('sss');
+        $this->tokenIdentifier ='nonce';
```
initvector: ⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️
⬜️😭😭😭⬜️😭😭😭⬜️😭😭😭⬜️😭😭😭⬜️😭😭😭⬜️
⬜️😭⬜️⬜️⬜️😭⬜️😭⬜️😭👾😭⬜️😭⬜️⬜️⬜️😭⬜️⬜️⬜️
⬜️😭😭😭⬜️😭😭😭⬜️😭😭😭⬜️😭⬜️⬜️⬜️😭😭😭⬜️
⬜️⬜️⬜️😭⬜️😭⬜️⬜️⬜️😭⬜️😭⬜️😭⬜️⬜️⬜️😭⬜️⬜️⬜️
⬜️😭😭😭⬜️😭⬜️⬜️⬜️😭⬜️😭⬜️😭😭😭⬜️😭😭😭⬜️
⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️⬜️

tests/Models/TokenSigningTraitTest.php
cchabilall83: the previous tests on the accesstokenmodel had assertions, I left them.  I don't think they're doing much, get rid of them?

initvector: So long as PHPUnit doesn't complain about it (sometimes it doesn't like assertion-less tests), getting rid of the assertions in tests where we're expecting exceptions should be fine.

applications/dashboard/models/class.userauthenticationnoncemodel.php
cchabilall83: I was thinking of something like this

    public function setSecret($secret) {
        $this->secret = $secret ?: c('Garden.Cookie.Salt');
        return $this;
    }

initvector: I would stick to letting the class determine how a default should be set. `AccessTokenModel` and `UserAuthenticationNonce` could maybe default to that config, but maybe another class implementing the trait would have a separate source for its default secret.

tests/Models/TokenSigningTraitTest.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * An altered token signature shouldn't verify.
+     *
+     * @expectedException \Exception
+     * @expectedExceptionMessage Your nonce has an invalid signature.
+     */
+    public function testBadSignature() {
+        $model = new TokenTestingModel();
+        $token = $model->signToken($model->randomToken(), '2 months').'!';
+        $model->verifyTokenSignature($token,true);
```
initvector: 🤔

cchabilall83: lol i was just about to correct that

library/core/class.accesstokenmodel.php
```diff
@@ -26,6 +24,7 @@ public function __construct($secret = '') {
         parent::__construct('AccessToken');
         $this->PrimaryKey = 'AccessTokenID';
         $this->secret = $secret ?: c('Garden.Cookie.Salt');
```
initvector: This should really be doing the same secret setup as the nonce class.

tests/Models/TokenSigningTraitTest.php
```diff
@@ -0,0 +1,65 @@
...
+        $model = new TokenTestingModel();
+        $token = $model->signToken($model->randomToken(), '2 months').'!';
+        $model->verifyTokenSignature($token, true);
+    }
+
+    /**
+     * A nonsense token shouldn't verify.
+     *
+     * @expectedException \Exception
+     * @expectedExceptionMessage nonce missing parts.
```
initvector: I'm concerned this exception message isn't correct and the test could still be passing.

cchabilall83: for some reason the test fails when it should

PR #7461, opened 2018-07-13
Fix sharing discussions

initvector:
A recent refactor moved to retrieving discussions as arrays, instead of objects, in `getRecord`. Most of the references were updated, but it looks like `ShareUrl` slipped through the cracks. This update ensures it is properly set.


dannimajdalani:
https://github.com/vanilla/support/issues/56

PR #7476, opened 2018-07-17
Add discussionsApiController_beforeGet event in /api/v2/discussions/:id

cchabilall83:
This allows groups to hook into /api/v2/discussions/:id and override the permissions so that group members can view secret group discussions.

related to vanilla/internal#1629
blocks vanilla/internal#1662

### DIFF
--- a/applications/vanilla/controllers/api/DiscussionsApiController.php
+++ b/applications/vanilla/controllers/api/DiscussionsApiController.php
@@ -223,7 +223,7 @@ class DiscussionsApiController extends AbstractApiController {
         $out = $this->schema($this->discussionSchema(), 'out');
 
         $query = $in->validate($query);
-        
+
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -223,7 +223,7 @@ public function get($id, array $query) {
         $out = $this->schema($this->discussionSchema(), 'out');
 
         $query = $in->validate($query);
-        
+
         $this->getEventManager()->fireFilter('discussionsApiController_beforeGet', $this, $id, $query);
```
initvector: `DiscussionsApiController::index` has a similar event, but it's named [indexFilters](https://github.com/vanilla/vanilla/blob/9a59d5a23762499598165793b070cd1019dd9c37/applications/vanilla/controllers/api/DiscussionsApiController.php#L428). Let's use that same convention here and name this "getFilters".

PR #7484, opened 2018-07-19
Fix query escaping throwing error when deleting a role

topcheese13:
Fixes https://github.com/vanilla/support/issues/41

The changes to escaping broke this, even though we never needed to escape the "field" of the having clause in the first place.


linc:
Is it EVER useful to have the escape flag set to `true` when calling `having()` now, or is it always broken?

topcheese13:
>Is it EVER useful to have the escape flag set to true when calling having() now, or is it always broken?

It's never useful to escape the "field" part of the query, I don't think we ever set the field from user interaction anywhere.

Issue #7447, opened 2018-07-10
Implement redux standard for new authenticate pages


charrondev:
After discussion with @tburry We've come along the following format for our redux standard. This should be implemented for the `/authenticate` pages (sign-in, and forgot password) and documented in our documentation.

## Action Types

```ts
interface IApiReponse<T = any> {
    data: T;
    status: number;
    headers: any;
}

interface IFieldError {
    message: string; // translated message
    code: string; // translation code
    status: number; // HTTP status
    field: string;
}

interface IApiError {
    message: string;
    status: number;
    headers: any;
    errors: {
        [key: string]: IFieldError[];
    };
}
interface IApiAction<M = any> {
    type: string;
    meta: M;
}

interface IApiSuccessAction<M = any, P = any> extends IApiAction<M> {
    payload: IApiReponse<P>;
}

interface IApiErrorAction<M = any> extends IApiAction<M> {
    type: string;
    payload: IApiError;
}

interface IApiRequestAction<M = any> extends IApiAction<M> {}
```

## Action examples

```ts

const successAction: IApiSuccessAction = {
    type: "GET_DISCUSSIONS_LIST_SUCCESS",
    payload: {
        data: {
            discussionId: 1234,
        },
        status: 200,
        headers: {},
    },
    meta: {
        params: {},
    },
};

const errorAction: IApiErrorAction = {
    type: "GET_DISCUSSIONS_LIST_ERROR",
    payload: {
        message: "",
        errors: {
            "collection.field": [
                {
                    message: "error",
                    code: "error",
                    status: 400,
                    field: "field",
                },
            ],
        },
        status: 200,
        headers: {},
    },
    meta: {},
};

const requestAction: IApiRequestAction = {
    type: "GET_DISCUSSIONS_LIST_REQUEST",
    meta: {},
};
```

## Utilities

A new utility function will be be defined to automatically generate action types and action creators. It will take the following shape.

**createApiActions**
```ts
function createApiActions<ResponseType = any, MetaType = any>(
    requestType: string,
    successType: string,
    errorType: string,
): {
    actionTypes: string[];
    actionCreators: {
        request: (requestOptions: MetaType) => IApiRequestAction<MetaType>;
        success: (
            requestOptions: MetaType,
            response: IApiReponse<ResponseType>,
        ) => IApiSuccessAction<MetaType, ResponseType>;
        error: (requestOptions: MetaType, error: IApiError) => IApiErrorAction<MetaType>;
    };
} {
[...]

PR #7485, opened 2018-07-19
Refactor /authenticate pages onto redux

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7447
Goes with https://github.com/vanilla/rich-editor/pull/84 and https://github.com/vanilla/internal/pull/1665  _If these aren't merged at the same time, they will not be able to build until those PRs are merged._

The existing authentication pages (signin, password, recoverpassword) have been refactored onto redux. Some utilities have been added to facilitate this. One in particular is preloading a successful action into the page in the server response using `Gdn_Controller->addClientApiAction()`.

## Testing

Things shouldn't be too different here. I've tested the following on my localhost:

- Successful form submissions work.
    - User is authenticated.
    - User is redirected.
    - Password reset emails are sent.
- Authenticators are preloaded without a network
    - No network request should be made for authenticators when loading the /authenticate/signing page directly.
    - A api call should be made from the client if there is a navigation through react router (eg. load the recover password page directly, then navigate the signin page using the signin link).
- Error handling
	- Field specific errors get given to the correct fields
	- Global error messages do not appear if there are field specific errors.

## Error handling and invalidation

- Components no longer do error handling themselves.
- Errors only get invalidated when a new form submission response returns from the server. This simplifies things a bit, and as @tburry had mentioned helps prevent users from forgetting what the error was when the start typing (previously we removed field errors as soon as the user starting typing in the input)
- Error handling is mostly handled in the reducers.
- Some utilities for error handling have been added. `indexArrayByKey`, `getFieldErrors`, and `getGlobalErrorMessage`. These replace all of the error handling that was being done in the component.
- One of the recover password API endpoints wasn't correctly returning the field that was invalid (when a valid email, that didn't exist was inserted). This has been fixed by specifying the field.

## Moving to Redux

- Reducers and actions for `/authenticate/authenticators`, `/authenticate/password`, and `/users/request-password` have been added.
- Utilities have been added to cut out the boilerplate of creating common actions and thunks. `generateApiActionCreators()`, `apiThunk()`.

## Other

- Many type definitions and interfaces have been moved into the @types folder
- The authentication controller was loading a bunch of CSS and Js files that had no business being loaded on these new pages. The following scripts have been removed:
	- `jquery.gardenhandleajaxform.js` - We are doing our own form submission with redux and axios.
	- `jquery.atwho.js` - No at-mentions (from the old editor) are present on these pages.
	- `authenticate.js` - All authentication code for these new pages is present in automatically loaded bundles.
    - `authenticate.css` - This was an empty CSS file. I deleted the CSS file as well.
    - `authenticator.css` - This was a useless CSS file. I deleted this CSS file too.


slafleche:
I requested to recover my password. When I click on "sign in" again, There's an unstyled "loading"

screen shot 2018-07-19 at 16 41 31

charrondev:
> The errors sometimes return HTML and are currently being escaped:

From my discussion with Todd we will be looking to interpolate these correctly, but will need to develop a proper solution for it https://github.com/vanilla/vanilla/issues/7487. We may need to do this for knowledge management before we circle back on authenticator stuff.

> I requested to recover my password. When I click on "sign in" again, There's an unstyled "loading"

So this seems to have come in from @tburry as the beginning of an attempt to have a generic loader. Since we nixing this for now, and that likely won't get fully planned for a bit, I've nixed the unfinished loader. I'm not filing an issue at this time though, because the requirements are unclear.

> After you click on sign in, if you click again on recover password, you don't get to enter your e-mail address. It should be reset each time.

I've resolved this.

Additionally I have 1 more PR for the ui-tests plugin (update similar to the rich editor).https://github.com/vanilla/internal/pull/1665

Issue #7268, opened 2018-05-28
Fix XSS In External Embeds (in editor, client js)


DaazKu:
Post the following as Body:
```json
[{
    "insert": {
        "embed-external": {
            "url": "javascript:alert(/xss/.source)",
            "type": "link",
            "name": "Example Domain",
            "body": "This domain is established to be used for illustrative examples in documents. You may use this\n    domain in examples without prior coordination or asking for permission.",
            "photoUrl": "javascript:alert(/xss/.source)",
            "height": null,
            "width": null,
            "attributes": []
        }
    }
}, {
    "attributes": {
        "link": "javascript:alert(/xss/.source)"
    },
    "insert": "http://www.example.com"
}, {
    "insert": "\n\n"
}]
":"\n\n"}]
```

![screen shot 2018-05-28 at 3 29 44 pm](https://user-images.githubusercontent.com/2412909/40627428-ffd3f238-628c-11e8-9568-9520492df0a3.png)

DaazKu:
BONUS
----
Related: https://github.com/vanilla/vanilla-patches/issues/264

```
[{
    "insert": {
        "embed-external": {
            "url": "javascript:alert(/xss/.source)",
            "type": "link",
            "name": "Example Domain",
            "body": "This domain is established to be used for illustrative examples in documents. You may use this\n    domain in examples without prior coordination or asking for permission.",
            "photoUrl": "https://httpbin.org/basic-auth/user/passwd",
            "height": null,
            "width": null,
            "attributes": []
        }
    }
},, {
    "insert": "\n\n"
}]
":"\n\n"}]
```

charrondev:
Additionally the body wasn't being sanitized at all in the editor.

```json
[
    {
        "insert": {
            "embed-external": {
                "data": {
                    "url": "javascript:alert(/xss/.source)",
                    "type": "link",
                    "name": "Example Domain",
                    "body":
                        "<script>alert('Gotchya!!')</script>This domain is established to be used for illustrative examples in documents. You may use this\n    domain in examples without prior coordination or asking for permission.",
                    "photoUrl": "javascript:alert(/xss/.source)",
                    "height": null,
                    "width": null,
                    "attributes": []
                },
                "loaderData": { "type": "link", "url": "http://test.com" }
            }
        }
    },
    { "attributes": { "link": "about:blank" }, "insert": "http://www.example.com" },
    { "insert": "\n\n" }
]
```

[...]

PR #7496, opened 2018-07-23
Fix rich editor XSS issues

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7268 

This PR fixes a whole swath of potential XSS issues in the Rich Editor embeds and in the Quill Renderer.

### Example body that triggers XSS before

Clicking the link here would previously trigger and alert.

```json
[
    {
        "insert": {
            "embed-external": {
                "data": {
                    "url": "javascript:alert(/xss/.source)",
                    "type": "link",
                    "name": "Example Domain",
                    "body":
                        "<script>alert('Gotchya!!')</script>This domain is established to be used for illustrative examples in documents. You may use this\n    domain in examples without prior coordination or asking for permission.",
                    "photoUrl": "javascript:alert(/xss/.source)",
                    "height": null,
                    "width": null,
                    "attributes": []
                },
                "loaderData": { "type": "link", "url": "http://test.com" }
            }
        }
    },
    { "attributes": { "link": "about:blank" }, "insert": "http://www.example.com" },
    { "insert": "\n\n" }
]
```

A few examples of what has been fixed here:

### XSS in the server embeds

We were not properly sanitizing our links for valid protocols. I've added new methods in `@dashboard/utility.ts` and `Gdn_Format`: `sanitizeUrl()`. It takes a url and validates it's protocol against a whitelist.

If the protocol does not match the whitelist, `unsafe:` is prepended onto the url. This is what our current filtering solution `Htmlawed` does.

I added a new assertion to the `SanitizeTest`: `assertSanitizedLinks()`. It inserts a javascript scheme url and ensures that a javascript href never gets outputted. I then updated all of the embeds failing the test to properly sanitize their links.

Many embed test were not properly testing all properties for sanitization. I've updated these tests to try and insert bad content in all positions.

### CSS escaping

I've remove the `cssSpecialChars` functions. A doc will follow later this week, but it will include a policy of no insertion of user content into inline styles, obviating the need for this method which garbles the url and does not fully solve this issue. https://github.com/vanilla/vanilla-patches/issues/264. Instances of inlined background images have been replaced with `` tags.

### Updates to global JS.

https://github.com/vanilla/vanilla-patches/issues/313 is only fixed for the rich editor after this.

I've updated to the rich editor embeds to not need this client side transformation which opened images from the rich editor to XSS. I held out of refactoring that piece to use our

### XSS in the client embeds

There was so much unsanitized content that could be loaded into the the user content embeds. I've eleminated almost all of these by converting the embed rendering functions to use React, which escapes all content by default. Additionally these got the same link sanitization as the server side blots and links, even if the links aren't actually actionable in most browsers' implementations of contenteditable.


library/core/class.format.php
```diff
@@ -2462,24 +2462,39 @@ public static function rich(string $deltas): string {
...
-            "^" => "\\00005e", "`" => "\\000060", "{" => "\\00007b", "|" => "\\00007c", "}" => "\\00007d",
-            "~" => "\\00007e",
-        ];
+    public static function sanitizeUrl(string $url): string {
+        $isSafe = false;
 
-        $result = str_replace(array_keys($specialChars), array_values($specialChars), $string);
-        return $result;
+        foreach (self::SAFE_PROTOCOLS as $protocol) {
+            if (strpos($url, $protocol) === 0) {
```
initvector: Would using [`parse_url`](http://php.net/parse_url) be a better fit here?

```
echo parse_url('http://google.com', PHP_URL_SCHEME) // "http"
echo parse_url('https://google.com', PHP_URL_SCHEME) // "https"
echo parse_url('mailto:user@example.com', PHP_URL_SCHEME) // "mailto"
echo parse_url('tel:5555555555', PHP_URL_SCHEME) // "tel"
```

You could grab the "scheme" and perform a simple `in_array` check. Beware the result of `parse_url` could be `null` or `false`.

PR #7497, opened 2018-07-24
Add Integration tests for save, delete draft and post discussion from draft

cchabilall83:
This adds integrations for drafts, It covers the following draft options.

Save a draft from the New Discussion form (including Title, Category, and Body).
Post a saved draft discussion as-is.
Delete a draft.

Added setSpamCount method so that floodcontrol wouldn't be triggered.

Closes Vanilla#7088


tests/APIv0/SmokeTest.php
```diff
@@ -419,10 +518,12 @@ public function testDiscussionAddBookMark() {
      */
     public function testRemoveDiscussionBookMark() {
         $api = $this->api();
-        $api->setUser($this->getTestUser());
+        $admin =  $this->api()->queryUserKey('Admin', true);
+        //use admin user so flood control isn't triggered
```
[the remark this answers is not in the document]
initvector: Since APIv2 tests are "internal" requests and APIv0 tests are not, you can't use the same methods. However, you can configure the flood control to be more permissive by using the "Vanilla.Discussion.SpamCount" and "Vanilla.Discussion.SpamTime" config keys.

tests/APIv0/SmokeTest.php
```diff
@@ -365,6 +376,121 @@ public function testPostRestrictedDiscussion() {
...
+
+        $r = $api->post(
+            '/post/discussion.json',
+            $draft
+        );
+        $statusCode = $r->getStatusCode();
+        $this->assertEquals(200, $statusCode);
+
+        $draftModel = new \DraftModel();
+        $postedDraft = $draftModel->getWhere(['Name' => $draft['Name']])->firstRow(DATASET_TYPE_ARRAY);
```
initvector: Doesn't DraftID come over as part of the response (`$r`) when saving a draft? Could that be used here, instead of trying to match on the name?

tests/APIv0/SmokeTest.php
```diff
@@ -418,10 +543,13 @@ public function testDiscussionAddBookMark() {
      * @large
      */
     public function testRemoveDiscussionBookMark() {
+        $this->setSpamCount(10);
```
initvector: What about doing this in `beforeSetupClass` or something, and setting it to a much higher number, so flood control would effectively be disabled for all tests in this class?

tests/APIv0/SmokeTest.php
cchabilall83: We don't get a draftID in the response, it's very limited.  

$r->getBody returns
{
"isHomepage" : false,
"Type" : "Discussion"
}

tests/APIv0/SmokeTest.php
cchabilall83: i can move this to the BaseTest,  i'll set it to 100.

tests/APIv0/SmokeTest.php
initvector: That's unusual. Do you have any idea why? When saving a draft on the frontend, you get the following response:

```json
{
    "DiscussionID": "",
    "DraftID": 2,
    "FormSaved": true,
    "DeliveryType": "VIEW",
    "Data": "[...]",
    "InformMessages": [
        {
            "CssClass": "Dismissable AutoDismiss",
            "Message": "Draft saved at 11:00AM"
        }
    ],
    "ErrorMessages": "",
    "RedirectTo": "",
    "RedirectUrl": ""
}
```

cchabilall83: I was missing some params in my request.

PR #7502, opened 2018-07-24
Add new mention notifications and tests

charrondev:
Closes https://github.com/vanilla/rich-editor/issues/65

Acceptance criteria are in the issue.

Implements mention notifications for the Quill rendered mentions for discussions and comments.

## Manual testing

Create posts and comments with mentions using the Rich Editor and verify that the notifications are sent according to the acceptance criteria of the original issue.

## Implementation Details

I've opted to add another parse method to the Quill parser `parseMentions`. Additionally there is an other utility method added to `Gdn_Format` (soon to be `\Vanilla\Formatting\FormatUtility`) to instantiate the parser, do a json decode, and return the names. This will get an array of usernames mentioned in a post. This is then conditionally used in the CommentModel and DiscussionModel, instead of our global `getMentions`.

`getMentions()` is not passed the format, and I don't feel our global functions should be instantiating a quill parser, which is why I've opted to it this way. I could not find other usages of `getMentions()`. All other aspects of mention handling remains as part of our old system.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -1241,7 +1241,12 @@ class CommentModel extends Gdn_Model {
             }
 
             // Notify any users who were mentioned in the comment.
-            $Usernames = getMentions($Fields['Body']);
+            if ($Fields['Format'] === 'Rich') {
+                $Usernames = Gdn_Format::getRichMentionUsernames($Fields['Body']);
+            } else {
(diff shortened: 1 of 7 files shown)


charrondev:
To clarify, because we had an in-person conversation:

Whatever criteria the notifications get sent with is untouched in this PR. The existing notification system may or may not be ideal, but I'm re-using the existing one from the advanced editor, and just providing the set of usernames that may or may not get a notification differently.

PR #7530, opened 2018-07-26
Admin theme tweaks

tburry:
- Change to Open Sans.
- Remove super light text.
- Remove main menu height and descriptions.
- The title bar is now thinner.


applications/dashboard/scss/admin.scss
```diff
@@ -4,8 +4,10 @@
 //
 // @author    Becky Van Bussel <beckyvanbussel@gmail.com>
 //
+@import url('https://fonts.googleapis.com/css?family=Open+Sans:400,400i,600,600i,700,700i');
 
-@import 'src/variables';
```
charrondev: What happens with this src/variables sheet? Is it going away since it's not being used anymore?

applications/dashboard/js/dashboard.js
```diff
@@ -858,26 +858,26 @@ $(document).on('contentLoad', function(e) {
      * @param element - The scope of the function.
      */
     function navbarHeightInit(element) {
-        var $navbar = $('.js-navbar', element);
+        // var $navbar = $('.js-navbar', element);
```
charrondev: Do we still need to keep this function? We should just delete it if it doesn't do anything.

applications/dashboard/scss/src/_navbar.scss
```diff
@@ -62,42 +69,12 @@
       }
     }
 
-    &.hover,
+    //&.hover,
```
charrondev: Can we delete this now?

charrondev:
A few other little things in here too:

## Media titles

We should bolden these to be bolder than the normal content, or make a color change to denote the difference. The class is `.media-title`.

This makes a big difference on our tables.

|Before|After|
|---|---|
|image|image|

The primary issue here actual seems to be that the font-weight set on these is 500, which is not present in Open Sans. It should be 600.

charrondev:
We also seem to have lost a title here.

|My local on this PR|Staff forum|
|---|---|
|image|image|

slafleche:
The misalignment on the tabs are from a border with the same colour as the background

applications/dashboard/scss/admin.scss
tburry: I kept it there because style.css is using it for some reason. I want to be able to tweak the settings for either theme without affecting the other. Whether the _variables.scss can be removed is not the purview of this PR.

applications/dashboard/scss/src/_buttons.scss
```diff
@@ -85,6 +85,10 @@ button::-moz-focus-inner {
       border-color: $btn-link-disabled-color;
     }
   }
+
+  .header-block & {
```
slafleche: No need for the `&`

charrondev:
I've gone through and tweaked all of the other `font-weight: 500`s to be `font-weight:600`s. That caught the last of things that had lost their semi-bold/bold formatting.

PR #7534, opened 2018-07-27
 Deprecate val() function

alex-mtl:
Add @deprecated tag to val() function

Transform few val() call to use ??  instead

Fixes: https://github.com/vanilla/vanilla-patches/issues/299


library/core/class.dispatcher.php
```diff
@@ -391,7 +391,7 @@ private function analyzeRequest($request) {
         if ($controllerName) {
             // The controller was found based on the path.
             $result['controller'] = $controllerName;
-        } elseif (Gdn::pluginManager()->hasNewMethod('RootController', val(0, $parts))) {
+        } elseif (Gdn::pluginManager()->hasNewMethod('RootController', $parts[0]??false)) {
```
DaazKu: https://docs.vanillaforums.com/developer/contributing/coding-standard-php/#1-overview

Should be: `$parts[0] ?? false`

alex-mtl: Sorry, updated...

library/core/class.dispatcher.php
```diff
@@ -472,9 +472,9 @@ private function getEnabledApplicationFolders() {
      * @return array Returns an array in the form `[$controllerName, $parts]` where `$parts` is the remaining path parts.
      * If a controller cannot be found then an array in the form of `['', $parts]` is returned.
      */
-    private function findController($parts) {
+    private function findController(array $parts) {
```
linc: This is fine in this case (because this method is private), but I wanna point out that Vanilla is littered with (bad) array/object interchangeability that can make this dangerous in many situations. It's also not part of the goal of this PR and it's omitted from the description, which increases the risk.

linc:
Please excuse my irrelevant deprecated comment contribution. :o)

alex-mtl:
@linc By the way your comment is interesting to me. I think I will confirm next week that val() has no impact on memory usage ;-) sorry guys :-) but to me the function itself seem useless and not needed even if there is no memory "leaks" or "issues" and we should better avoid to use it in future

applications/dashboard/library/class.nestedcollection.php
```diff
@@ -440,14 +440,14 @@ private function addItem($type, $item) {
...
+        foreach ( $item['key'] as $i => $key_part) {
 
-            if ($i === count(val('key', $item)) - 1) {
+            if ($i === count($item['key'] ?? false) - 1) {
                 // Add the item here.
                 if (array_key_exists($key_part, $items)) {
                     // The item is already here so merge this one on top of it.
                     if ($items[$key_part]['type'] !== $type)
-                        throw new \Exception(val('key', $item)." of type $type does not match existing type {$items[$key_part]['type']}.", 500);
+                        throw new \Exception($item['key'] ?? ''." of type $type does not match existing type {$items[$key_part]['type']}.", 500);
```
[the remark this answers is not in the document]
alex-mtl: Sorry, can not get your point here

applications/dashboard/views/modules/dropdown.php
```diff
@@ -3,50 +3,54 @@
...
+                        if ($iBadge = $item['badge'] ?? false) {
+                            echo ' '.wrap($iBadge, 'span', ['class' => 'Alert']);
                         }
                         ?></a>
                 </li>
             <?php }
-            if (val('type', $item) == 'divider') { ?>
-                <li role="presentation" <?php if (val('cssClass', $item)) { ?> class="<?php echo val('cssClass', $item); ?>"<?php } ?>>
+            if ($item['type'] ?? '' == 'divider') { ?>
+                <li role="presentation" <?php if ($iCssClass = $item['cssClass'] ?? false ) { ?> class="<?= $iCssClass; ?>"<?php } ?>>
```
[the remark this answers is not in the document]
DaazKu: https://open.vanillaforums.com/discussion/31810/how-the-vanilla-dev-staff-decides-important-issues

applications/dashboard/library/class.nestedcollection.php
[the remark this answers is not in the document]
initvector: Using `val`, the value of `$item['key']` would've been prepended to the rest of the exception string. Now, if `$item['key']` exists, it's used as the full exception message. It's not prepended to the existing message string. If it doesn't exist, the value on the right side of the null coalescing operator is used.

```
>>> $item = ['key' => 'value']
=> [
     "key" => "value",
   ]
>>> $item['key'] ?? ''." of type [...] does not match existing type [...]."
=> "value"
>>> $item['not-a-key'] ?? ''." of type [...] does not match existing type [...]."
=> " of type [...] does not match existing type [...]."
```

You probably want to group your value coalescing in parenthesis.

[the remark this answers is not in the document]
DaazKu:
```
>>> echo $test ?? 'Prepend'."Sring"
PrependSring⏎
>>> $test = 1
=> 1
>>> echo $test ?? 'Prepend'."Sring"
1⏎
>>> 
```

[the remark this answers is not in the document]
DaazKu: Ryan beat me to it. Long story short: ?? has a super low priority: http://php.net/manual/en/language.operators.precedence.php

applications/dashboard/library/class.nestedcollection.php
```diff
@@ -533,7 +534,10 @@ private function isActive($item) {
         } else {
             $highlightRoute = url($this->highlightRoute);
         }
-        return (val('url', $item) && (trim(url(val('url', $item)), '/') == trim($highlightRoute, '/')));
+        if ($ret = ($url = $item['url'] ? true : false)) {
```
initvector: Legibility is important. While the original code was cluttered, I think it was probably easier to determine the intent when reading it. We were verifying we had a "URL" and that it matched the "highlighted route", when formatted as a full URL. Now, there are two assignments and a ternary in a single condition of an `if` statement. It seems like a lot to unpack/understand its intent at first glance. Keep in mind that more junior developers may be working behind you.

If *I* had to refactor the same code into an easier-to-understand format, I'd probably do something similar to this:
```php
$url = $item['url'] ?? false;
if ($url) {
    $result = trim(url($url), '/')) == trim($highlightRoute, '/');
} else {
    $result = false;
}
return $result;
```

applications/dashboard/library/class.nestedcollection.php
```diff
@@ -425,12 +425,13 @@ private function touchKey(&$item) {
      * @param array $item The item to add to the array.
      * @throws Exception
      */
-    private function addItem($type, $item) {
+    private function addItem($type, Array $item) {
```
initvector: Let's stick with using lowercase type keywords

applications/dashboard/views/modules/dropdown.php
```diff
@@ -3,50 +3,54 @@
...
-                    <a role="menuitem" class="dropdown-menu-link <?php echo val('cssClass', $item); ?>" tabindex="-1" href="<?php echo url(val('url', $item)); ?>" <?php echo attribute(val('attributes', $item, [])) ?>><?php
-                        if (val('icon', $item)) {
-                            echo icon(val('icon', $item));
+            <?php  if (($item['type'] ?? '') == 'link') { ?>
+                <li role="presentation" <?php
+                    if ($ilistItemCssClass = ($item['listItemCssClass'] ?? false) || empty($item['icon'])) {
+                       ?>class="<?php
+                        echo trim($ilistItemCssClass.(empty($item['icon']) ? ' no-icon' : '')); ?>"<?php
+                    } ?>>
+                    <a role="menuitem" class="dropdown-menu-link <?php  $item['cssClass'] ?? ''; ?>" tabindex="-1" href="<?php echo url($item['url'] ?? ''); ?>" <?php echo attribute($item['attributes'] ?? []) ?>><?php
```
initvector: I think we lost an `echo` in here.

```php
<?php  $item['cssClass'] ?? ''; ?>
```

applications/dashboard/library/class.nestedcollection.php
initvector: The value of `$item['url']` should be a string. You're essentially extracting it into `$url` as a boolean. On the next line, you're calling `trim` and `url` on `$url`, which means you're passing a boolean value for string parameters.

bleistivt:
While I also think that it is a good decision to retire `val()`, especially since there is the `??`-operator now, I would like to emphasize what @linc said about array/object interchangeability:

I just searched all of my plugin code for `val()`-usage and the first occurrences of `val()` that I found cannot be replaced.

The `User`-argument in `activityModel_beforeSendNotification_handler` can be an object or an array depending on the caller (`ActivityModel::sendNotification()` vs `ActivityModel::email()`).

So my question is: Will one of the SQL result types become deprecated?
What should be used for new code? Objects of `Gdn_DataSet` (default) or PHP arrays (easier to manipulate with language tools)?

alex-mtl:
I am very new to the project, but from discussions we had recently with the team, I would say the intention is to move "strict types" direction. Whenever we code any part of the application in any method we need to clearly declare what type our function require (array || object).

Most findings of val() calls I've seen - we can detect the usage of Array or Object, but not both.

About your question about Gdn_DataSet we have no plans at the moment to get rid of it. But we have some ideas about  to change few core models to use DATASET_TYPE_ARRAY by default (let say as 1st step)

linc:
Yes, array is the preference now. We have not altered the default only to avoid unpredictable side effects.

Issue #7527, opened 2018-07-26
Add rich format to advanced editor dropdown conditionally


charrondev:
When rich editor is enabled the advanced editor dropdown is disabled and the current format being use is displayed.

cchabilall83:
@charrondev 
If the advanced editor is enabled will the rich editor be disabled?

charrondev:
Current acceptance criteria from @tburry and @linc 


1) Enabling Rich Editor adds 'rich text' to the formatting dropdown in the advanced editor plugin and auto-selects it.
2) Selecting a formatting option other than 'rich text' functionally disables the Rich Editor (new posts default to whatever format is selected).
3) Rendering for all formats including rich still work even if they are not the currently selected main format. (This is already the case. You should not have to change anything to make this happen).

charrondev:
@tburry I'd like to add one more success criteria here. Since multiple plugins are now contributing to available formats, we should move that formats dropdown out of the advanced editor plugin and into the "Posting" page of the dashboard. Plugins such as the rich editor and advanced editor could then add their formats to that list.

cchabilall83:
@charrondev The last success criteria should be put into it's own separate issue as it would change the complexity of the current issue.

Issue #7521, opened 2018-07-26
Rich editor does not have a quote solution


charrondev:
We should be able to put together a naive quote solution similar to what we already had.

charrondev:
## Proposal 1 - Client side mostly

1. Disable the Quotes plugin when the Rich Editor is enabled.
2. Add a configuration dashboard page for the Rich Editor. This can contain the format that is selected (if rich editor is enabled), and a checkbox to enable/disable quotes.
3. Add hooks in the rich editor plugin where the quotes plugin does.

```php
    public function discussionController_beforeDiscussionDisplay_handler($sender) {
        $this->renderQuotes($sender);
    }
    public function postController_beforeDiscussionDisplay_handler($sender) {
        $this->renderQuotes($sender);
    }
    public function discussionController_beforeCommentDisplay_handler($sender) {
        $this->renderQuotes($sender);
    }
    public function postController_beforeCommentDisplay_handler($sender) {
        $this->renderQuotes($sender);
    }
```
4. Have this button fire an event with the ID of the post or comment, and the type.
5. Have the rich editor listen on this event, and if it is the primary editor in the page (there can be multiple) do the following:
  - Call the requisite API endpoint to fetch the full data for the post.
  - Insert a the body content from the post as if it was copy/pasted
  - Quote that body contents (quill will remove, fix any conflicting formatting)
  - Insert an @mention of the username in the top of the quote.

charrondev:
The quotes plugin:
- Can stay turned on.
- Should still inject its javascript file in the page.


Things that need to be stripped out when we create a quote with the quote button

- Button

PR #7543, opened 2018-07-30
Make editor formatting settings global in the post page

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7527

Moves the formatting options from the editor plugin to the dashboard posting page.

## Changes

- Adds config options `Garden.InputFormatter`, `Garden.InputFormatter`, `Plugins.editor.ForceWysiwyg` to the /settings/posting page.
- Advanced Editor settings are gone. In place of its previous settings page is a redirect to the /settings/posting page.
- Splits out the label descriptions into separate pieces without HTML in them and wraps them in translation functions.

## Related

- Allowing Rich Editor to hook into these settings: https://github.com/vanilla/rich-editor/pull/90
- Translation source strings in the locales repo: https://github.com/vanilla/locales/pull/69

## Notes

- I've verified that all 3 of the config settings persist when saved.
- The force WYSIWYG option still only appears when the wysywyg format is enabled.
- The IPB formatting plugins still successfully hooks into these settings. This is why I am firing the `GetFormats` events twice. Once as the `VanillaSettingsController` and once as the `EditorPlugin`.
- I've made `Text` the only option by default.

### DIFF
--- a/applications/vanilla/controllers/class.vanillasettingscontroller.php
+++ b/applications/vanilla/controllers/class.vanillasettingscontroller.php
@@ -55,9 +55,10 @@ class VanillaSettingsController extends Gdn_Controller {
         $validation = new Gdn_Validation();
         $configurationModel = new Gdn_ConfigurationModel($validation);
         $configurationModel->setField([
+            'Vanilla.Categories.MaxDisplayDepth',
             'Garden.InputFormatter',
             'Garden.MobileInputFormatter',
-            'Vanilla.Categories.MaxDisplayDepth',
+            'Plugins.editor.ForceWysiwyg',
             'Vanilla.Discussions.PerPage',
             'Vanilla.Comments.PerPage',
             'Garden.Html.AllowedElements',
--- a/applications/vanilla/controllers/class.vanillasettingscontroller.php
+++ b/applications/vanilla/controllers/class.vanillasettingscontroller.php
@@ -40,10 +40,23 @@ class VanillaSettingsController extends Gdn_Controller {
         // Check permission
         $this->permission('Garden.Settings.Manage');
 
+        $formats = ['Text']; // Check for additional formats
+        $this->EventArguments['formats'] = &$formats;
+
+        $this->fireEvent('GetFormats');
+        // This was moved from the editor plugin. In order to maintain compatibility with existing event hooks
+        // We will need to fire this as EditorPlugin.
+        $this->EventArguments['formats'] = &$formats;
+        $this->fireAs('EditorPlugin');
+        $this->fireEvent('GetFormats');
+        $this->setData('formats', $formats);
+
         // Load up config options we'll be setting
         $validation = new Gdn_Validation();
         $configurationModel = new Gdn_ConfigurationModel($validation);
         $configurationModel->setField([
+            'Garden.InputFormatter',
+            'Garden.MobileInputFormatter',
             'Vanilla.Categories.MaxDisplayDepth',
             'Vanilla.Discussions.PerPage',
             'Vanilla.Comments.PerPage',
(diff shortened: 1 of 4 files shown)


applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -40,11 +40,25 @@ public function posting() {
...
 
+        $formats = ['Text']; // Check for additional formats
+        $this->EventArguments['formats'] = &$formats;
+
+        $this->fireEvent('GetFormats');
+        // This was moved from the editor plugin. In order to maintain compatibility with existing event hooks
+        // We will need to fire this as EditorPlugin.
+        $this->EventArguments['formats'] = &$formats;
+        $this->fireAs('EditorPlugin');
+        $this->fireEvent('GetFormats');
```
initvector: The event manager has a `fireDeprecated` method that might be useful here. It'll trigger deprecation notices if any handlers are detected hooking into this event.

applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -40,11 +40,25 @@ public function posting() {
         // Check permission
         $this->permission('Garden.Settings.Manage');
 
+        $formats = ['Text']; // Check for additional formats
+        $this->EventArguments['formats'] = &$formats;
+
+        $this->fireEvent('GetFormats');
```
initvector: Since we're only modifying a single value, what do you think about using `fireFilter` instead?

applications/vanilla/views/vanillasettings/posting.php
```diff
@@ -23,6 +29,52 @@
...
+                <?php echo $form->dropDown('Garden.InputFormatter', $formatOptions); ?>
+            </div>
+        </li>
+        <li class="form-group">
+            <?php
+            $forceWysiwygLabel = 'Reinterpret All Posts As Wysiwyg';
+            $forceWysiwygNote1 =  t('ForceWysiwyg.Notes1', 'Check the below option to tell the editor to reinterpret all old posts as Wysiwyg.');
+            $forceWysiwygNote2 = t('ForceWysiwyg.Notes2', 'This setting will only take effect if Wysiwyg was chosen as the Post Format above. The purpose of this option is to normalize the editor format. If older posts edited with another format, such as markdown or BBCode, are loaded, this option will force Wysiwyg.');
+            $label = '<p class="info">'.$forceWysiwygNote1.'</p><p class="info"><strong>'.t('Note:').' </strong>'.$forceWysiwygNote2.'</p>';
+            echo $form->toggle('Plugins.editor.ForceWysiwyg', $forceWysiwygLabel, [], $label);
```
initvector: Does this addon-specific config value need to be in a core settings page? Do you think any other addons might use this value? If so, it might be worth creating a core counterpart. If not, and it's specific to Advanced Editor, maybe we could move this setting back into Advanced Editor's settings page.

charrondev: It is directly tied to and appears on this page when the wysiwyg format is enabled (handled in our existing dashboard JS). Making this the only Advanced Editor settings (likely ever again) doesn't seem right. I figure if/when we offer an equivalent conversion method for the Rich Editor we can unify it under 1 global setting. Likely `Garden.ConvertLegacyFormattingOptions` or something like that.

 I didn't want to change this value because it is likely being used somewhere that it shouldn't be and I didn't want to blow up the scope by tracking down all usages of this config value and ensuring that existing values carry over.

initvector: If you want to keep it on this page, that's cool, but I'm still not onboard with hardcoding addon configs into core settings pages. This setting is seemingly dependent on the Advanced Editor being enabled. I think there's the potential for real confusion if someone doesn't have the Advanced Editor enabled, but is trying to make use of this config. What about hooking in from Advanced Editor to add the field to this form?

charrondev:
I've modified the new events to be fired with `fireFilter` and `fireDeprecated`.

> Does this addon-specific config value need to be in a core settings page?

In light of us needing to add a quoting setting here for the Rich Editor with https://github.com/vanilla/vanilla/issues/7521 I've opted to add the extra event needed to hook this from the advanced editor side. The event is called `postingSettings_formatSpecificFormItems`. If you have a better name I'm game to change it.

This allows the html to inserted into the view from the rich editor plugin and the for the plugin to add its fields to the ConfigurationModel.

applications/vanilla/views/vanillasettings/posting.php
```diff
@@ -2,9 +2,17 @@
...
 <h1><?php echo t('Posting Settings'); ?></h1>
+
 <?php
 /** @var Gdn_Form $form */
 $form = $this->Form;
+/** @var \Garden\EventManager $eventManager */
+$eventManager = Gdn::getContainer()->get(\Garden\EventManager::class);
+$formats = $this->data('formats');
+$formatOptions = [];
+foreach ($formats as $formatName) {
```
initvector: I believe you can achieve a similar result with:

```php
$formatOptions = array_combine($formats, $formats);
```

Issue #7544, opened 2018-07-30
/api/v2/media/scrape should respect the meta charset tag


tburry:
For this URL: https://hub.github.com/ there is no content type header, but a `<meta charset="utf-8">` tag at the top.

- Currently, the scrape endpoint doesn't respect the meta tag.
- If you look at the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta) for the meta tag you will also notice `<meta http-equiv="Content-Type" content="text/html; charset=IANAcharset">` is an alternative way to represent the character set.
- The docs give a warning about character sets so I recommend using a whitelist of "UTF-8", "US-ASCII", "ISO-8859-1".

PR #7545, opened 2018-07-30
Fix UTF-8 page scraping for HTML 5 pages

tburry:
Unfortunately, DOMDocument is built with a version of libxml that does not understand HTML5 too well. This cause UTF-8 characters to not be read properly. In an attempt to make our scraper work with the majority of HTML5 documents this small change adds a charset and hopefully doesn’t kill efficiency.

I am well aware that this does not 100% solve the problem. If anyone wants to take a stab at this issue I am fine with that given:

1. Some profiling is done to make sure a solution doesn’t make large documents kill our servers.
2. There are no regression bugs. I recommend adding more character cases to the test files or even a non utf-8 test file.

### DIFF
--- a/library/Vanilla/PageScraper.php
+++ b/library/Vanilla/PageScraper.php
@@ -51,7 +51,17 @@ class PageScraper {
 
         $document = new DOMDocument();
         $rawBody = $response->getRawBody();
+
+        // Add charset information for HTML 5 pages.
+        // https://stackoverflow.com/questions/39148170/utf-8-with-php-domdocument-loadhtml#39148511
+        if (!preg_match('`\s*<?xml`', $rawBody)) {
+            $encoding = mb_detect_encoding($rawBody);
+            $rawBody = "<?xml version=\"1.0\" encoding=\"$encoding\"?>$rawBody";
+        }
+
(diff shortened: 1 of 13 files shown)


tburry:
Closes #7544.

library/Vanilla/PageScraper.php
```diff
@@ -190,4 +187,107 @@ private function userAgent() {
...
+     * Create the DOM from an HTML string.
+     *
+     * @param string $html The HTML to create the body from.
+     * @return DOMDocument Returns the document.
+     * @throws Exception Throws an exception if the HTML is so malformed that it couldn't be parsed.
+     */
+    private function createDom(string $html): DOMDocument {
+        // Add charset information for HTML 5 pages.
+        // https://stackoverflow.com/questions/39148170/utf-8-with-php-domdocument-loadhtml#39148511
+        if (!preg_match('`^\s*<?xml`', $html) && $encoding = mb_detect_encoding($html)) {
```
alex-mtl: do we really need t his here? we have another detection inside fixDocumentEncoding function

alex-mtl:
If possible we need to create few more test cases with big html files 50Kb, 100Kb, 500Kb 1Mb

Issue #7516, opened 2018-07-25
Rich editor format buttons are inconsistent


charrondev:
A few scenarios from @slafleche https://github.com/vanilla/vanilla/issues/7162

## Requirements:

- A spec for what unformats other things and formats on top.
- A spec for what unformats all of itself or just to its boundary.

## Implementation of fixes

- Create individual functions for formatting a given format for each range (that wrap quills built `format()` which is inconsistent with our custom multi-line blots)
- Use these for each format button
- Have unit tests for all of these functions, particularly against the edge cases such as link, code, codeblock, blockquote, spoiler.

### Description
Switching to spoiler, when you already had paragraph styles keeps those styles in the spoiler and there's no way to get rid of it.

### Issue severity
Moderate

### Steps to reproduce
1. Write some text
2. Give it a paragraph style, like code block or heading
3. Make it a spoiler

The spoiler wraps the block styles and there's no way to undo it. Clicking another style in the menu remove the spoiler.

---

### Description

Inline code block removes all paragraph styles

### Issue severity
Moderate

### Steps to reproduce
1. Make either: a heading, a quote or a spoiler. 
2. Select part of all of the text inside
3. Make it an inline block

The paragraph style disappears. If we don't support inline styles in any of these, it shouldn't be available in the menu instead of breaking the paragraph style.

charrondev:
Another one

### Description 
List formatting cleared when I use inline code formatting.

### Steps to reproduce.

1. Create a numbered list.
2. Select text within the list and choose the inline code formatting.
3. The list formatting disappears around the item you chose.

PR #7546, opened 2018-07-30
Fix quill rendering of heading with inline formats in it

charrondev:
I came across this while working on the in-editor formatting https://github.com/vanilla/vanilla/issues/7516.

Heading blots were not rendering properly if there were multiple separate inline formatted items inside of them. Luckily this structure of operations is already a solved problem. Switching the HeadingBlot to inherit from the `AbstractLineBlot` instead of the `TextBlot` directly solves the issue.

|Before|After|
|---|---|
|image|image|

I've added a parser test for this particular scenario as well.

### DIFF
--- a/library/Vanilla/Formatting/Quill/Blots/Lines/HeadingBlot.php
+++ b/library/Vanilla/Formatting/Quill/Blots/Lines/HeadingBlot.php
@@ -5,14 +5,15 @@
  * @license https://opensource.org/licenses/GPL-2.0 GPL-2.0
  */
 
-namespace Vanilla\Formatting\Quill\Blots;
+namespace Vanilla\Formatting\Quill\Blots\Lines;
+use Vanilla\Formatting\Quill\BlotGroup;
 
 /**
  * Blot to represent headings.
  *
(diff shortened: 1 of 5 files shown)


library/Vanilla/Formatting/Quill/Blots/Lines/HeadingBlot.php
```diff
@@ -41,12 +42,38 @@ public function getGroupClosingTag(): string {
...
 
+    /**
+     * The heading blot can be the ONLY overriding blot in a group. Even other headings.
+     *
+     * @param BlotGroup $group
+     * @return bool
+     */
+    public function shouldClearCurrentGroup(BlotGroup $group): bool {
+        $overridingBlot = $group->getPrimaryBlot();
+        return !!$overridingBlot;
```
initvector: This is technically fine, but I think you could get a more explicit result out of using something like [`boolval`](http://php.net/boolval) or just [casting the type to a boolean](http://php.net/manual/en/language.types.type-juggling.php#language.types.typecasting).

PR #7547, opened 2018-07-30
Add basic support for feature flags

### LINKED ISSUE #7488: Remove /authenticate pages from 2.7 release

initvector:
Vanilla will be using feature flags, instead of feature branches, for many of its upcoming larger features that are expected to be in-progress for several weeks. To make working with these flags easier, this update includes a `FeatureFlagHelper` class. This class has two methods:

1. `featureEnabled`, to check if a feature is enabled in the config and return a boolean value reflecting that state. The "Feature" namespace is currently being enforced for these feature flags.
1. `throwIfNotEnabled`, to throw an exception if the feature isn't enabled. The type of exception and its constructor parameters are fully customizable.

An example implementation has been performed on the authentication API. If "Feature.AuthenticationAPI" is a truthy value in a site's config, the /authentication pages, authenticators and authentication API endpoints will be available. Otherwise, they will throw a fatal error, stating the feature is not enabled. Whole controllers are being blocked, so `throwIfNotEnabled` is the only method utilized there.

There are tests for the authenticators API endpoint, so enabling the feature has been added to those tests.

Closes #7488


library/Vanilla/FeatureFlagHelper.php
```diff
@@ -0,0 +1,45 @@
...
+     * Is a feature enabled?
+     *
+     * @param string $feature The config-friendly name of the feature.
+     * @return bool
+     * @throws \Garden\Container\NotFoundException If unable to find the config class in the container.
+     * @throws \Garden\Container\ContainerException If unable to get the config class instance from the container.
+     */
+    public static function featureEnabled(string $feature): bool {
+        // We're going to enforce the root "Feature" namespace.
+        $configValue = Gdn::getContainer()->get(\Gdn_Configuration::class)->get("Feature.{$feature}");
```
charrondev: Should we be getting the config directly through DI in the constructor? My understanding was that we were trying to move away from needing to touch the `Gdn` object in new classes.

initvector: The class is just a collection of static "helper" methods, similar to `ApiUtils`. That means that, ideally, it doesn't need to be instantiated (i.e. no constructor call for DI). If we were going to pass the config instance via instantiating and DI, whatever was using these methods would need to ensure the instance passed through the container too. Those classes would need their own separate calls to `Gdn::getContainer` to retrieve an instance or would need to add this helper class as a dependency on each of their individual constructors.

I can replace this with a call to `c` to avoid directly referencing the `Gdn` class.

PR #7549, opened 2018-07-31
Fix Drafts double escaping special characters

### LINKED ISSUE #7474: RichEditor draft double escaping
### Description
Draft content is double escaped

### Issue severity
Low

### Steps to reproduce

1. Put any html in your content. ex: `<﻿😛﻿`
2. Save draft
3. Go check drafts

### Screenshot (if applicable)

screen shot 2018-07-17 at 2 14 53 pm

cchabilall83:
This fixes special characters being displayed incorrectly on the drafts index page because they were being displayed incorrectly.

Closes #7474

### DIFF
--- a/applications/vanilla/views/drafts/drafts.php
+++ b/applications/vanilla/views/drafts/drafts.php
@@ -10,7 +10,7 @@ foreach ($this->DraftData->resultArray() as $Draft) {
 
     $draftID = val('DraftID', $Draft);
     $discussionID = val('DiscussionID', $Draft);
-    $excerpt = htmlspecialchars(sliceString(Gdn_Format::plainText(val('Body', $Draft), val('Format', $Draft)), 200));
+    $excerpt = sliceString(Gdn_Format::plainText(val('Body', $Draft), val('Format', $Draft)), 200);
 
     $isDiscussion = (!is_numeric($discussionID) || $discussionID <= 0);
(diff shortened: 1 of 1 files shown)


charrondev:
I've approved this but before you merge it I'd like you to add some of the extra context from what we discussed. Something along the lines of how plaintText sufficiently sanitized the content coming through already. Otherwise looking back at this PR historically I'd wonder, isn't this just adding XSS to the draft page?

cchabilall83:
We are sanitizing the text in the Gdn_Format::plainText, so the htmlspecialchars on the view is not necessary in this case.
https://github.com/vanilla/vanilla/blob/master/library/core/class.format.php#L1105

linc:
> This fixes special characters being displayed incorrectly on the drafts index page because they were being displayed incorrectly.

(╯°□°)╯︵ ┻━┻ 

> Special characters were being incorrectly double-encoded when viewing drafts, garbling user-generated content. We can safely remove the second encoding by using text sanitizing instead, via our `plainText()` formatter. This steps back the overzealous changes made in https://github.com/vanilla/vanilla/pull/7181 without reintroducing its original issue.

┬──┬◡ﾉ(° -°ﾉ)

Also note that it is better practice to use sanitizing on the final string, rather than slicing after the sanitizing, lest someone figure out a way to use the additional processing in their favor to generate a new vulnerability.

PR #7556, opened 2018-08-02
Style tweaks for rich editor

slafleche:
Final design tweaks for first iteration of Rich Editor. Note that compatibility styles will be in another PR.

Related: https://github.com/vanilla/rich-editor/pull/100

Closes: https://github.com/vanilla/rich-editor/issues/88


charrondev:
The only thing I have for this is that the editor no longer aligns properly in the comment box.

image

It's centered well on the post page but not here.

slafleche:
@charrondev good catch. Fixed!

PR #7559, opened 2018-08-04
Fix Twitter callback URL for profile connections

bleistivt:
<strike>This was never correct</strike>(see below), but twitter did not check for a correct callback_url until recently:
https://twittercommunity.com/t/action-required-sign-in-with-twitter-users-must-whitelist-callback-urls/105342

related:
https://github.com/vanilla/docs/pull/299

### DIFF
--- a/plugins/Twitter/class.twitter.plugin.php
+++ b/plugins/Twitter/class.twitter.plugin.php
@@ -274,7 +274,7 @@ class TwitterPlugin extends Gdn_Plugin {
             $redirectUri .= (strpos($redirectUri, '?') === false ? '?' : '&').$query;
         }
 
-        $params = ['oauth_callback' => $redirectUri];
+        $params = ['callback_url' => $redirectUri];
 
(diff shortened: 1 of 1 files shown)


bleistivt:
Ok, Twitter is kind of ambiguous there, mentioning the `callback_url` parameter here:
https://developer.twitter.com/en/docs/basics/developer-portal/guides/callback-urls.html
... and calling it `oauth_callback` there:
https://developer.twitter.com/en/docs/basics/authentication/api-reference/request_token

So now both are added. /profile/connections requests only work on my site if I have the first one set.
The same happens, when you try to connect from your profile on the open source forum, check it out: https://open.vanillaforums.com/profile/connections

    Callback URL not approved for this client application. Approved callback URLs can be adjusted in your application settings

charrondev:
@nervoustwit Did we get anywhere with this?

nervoustwit:
I tested this. It works.

nervoustwit:
Tested locally, the changes check out.

PR #7602, opened 2018-08-09
Prefix the default analytics subdomain with https in vanillastats

topcheese13:
Implementing the fix Ryan suggested. Regardless of the current protocol (http or https), the analytics subdomain will always be prefixed by https to avoid these preflight/CORS errors.

Closes https://github.com/vanilla/support/issues/79

### DIFF
--- a/plugins/VanillaStats/class.vanillastats.plugin.php
+++ b/plugins/VanillaStats/class.vanillastats.plugin.php
@@ -40,7 +40,7 @@ class VanillaStatsPlugin extends Gdn_Plugin {
      * VanillaStatsPlugin constructor.
      */
     public function __construct() {
-        $this->AnalyticsServer = c('Garden.Analytics.Remote', 'analytics.vanillaforums.com');
+        $this->AnalyticsServer = c('Garden.Analytics.Remote', 'https://analytics.vanillaforums.com');
         $this->VanillaID = Gdn::installationID();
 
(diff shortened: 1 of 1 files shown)


initvector:
Do you think the `Gdn_Statistics` class needs updating, as well? It also uses the "Garden.Analytics.Remote" config.

https://github.com/vanilla/vanilla/blob/5a61714b0b3e7eeff7b0d40e5f44da581e546409/library/core/class.statistics.php#L33-L35

library/core/class.statistics.php
```diff
@@ -30,10 +30,15 @@ class Gdn_Statistics extends Gdn_Pluggable {
...
 
-        $analyticsServer = c('Garden.Analytics.Remote', 'analytics.vanillaforums.com');
-        $analyticsServer = str_replace(['http://', 'https://'], '', $analyticsServer);
+        $analyticsServer = c('Garden.Analytics.Remote', 'https://analytics.vanillaforums.com');
         $this->AnalyticsServer = $analyticsServer;
 
+        $scheme = parse_url($this->AnalyticsServer, PHP_URL_SCHEME); // Will give you the protocol (http, https) or null/false on error.
+
+        if ($scheme === null) {
+            $this->AnalyticsServer = 'https://'.$this->AnalyticsServer();
```
linc: AnalyticsServer is not a method on this class.

PR #7608, opened 2018-08-10
Add classes to pager module to improve ability to style

slafleche:
The pager module made it difficult to style on mobile. I've added some helper class for the themer:

- How many links are there in the pager
- Common class for next/prev
- Common class for page links

### DIFF
--- a/applications/dashboard/modules/class.pagermodule.php
+++ b/applications/dashboard/modules/class.pagermodule.php
@@ -376,9 +376,9 @@ class PagerModule extends Gdn_Module {
 
         // Previous
         if ($currentPage == 1) {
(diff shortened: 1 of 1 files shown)


applications/dashboard/modules/class.pagermodule.php
```diff
@@ -422,14 +425,15 @@ public function toString($type = 'more') {
 
             $pager .= '<span class="Ellipsis">'.$separator.'</span>';
             $pager .= anchor($pageCount, $this->pageUrl($pageCount), $this->_GetCssClass($pageCount, $currentPage));
+            $linkCount = $linkCount + 4;
```
initvector: There's nothing "wrong" with this, but it's equivalent to:

```php
$linkCount += 4;
```

PR #7612, opened 2018-08-13
Place frontend build tooling directly in the vanilla/vanilla repo

charrondev:
Closes https://github.com/vanilla/vanilla-cli/issues/98

I've brought over a few of the logging utility functions from the core build. Everything else ended up getting rewritten.

## The build

A docs PR is up https://github.com/vanilla/docs/pull/302. It is primarily usage instructions.

### Whats different from the vanilla-cli core build.

- Setup consists only of a `yarn install` in the top of the vanilla repo.
- Builds all addons at once. This is required. An individual addon cannot be built without everything else knowing about it. This allows us to simplify immensely while enabling powerful features like bundle splitting. This is consistent with us moving towards running this build on our data servers. As a result of this we can do a few things:
  - No longer need to specify anything in `addon.json`. 
  - No longer need to list out `exports` of an addon.
  - No longer need an intermediary build step.
  - We can do tree-shaking for smaller bundles.
  - We can do chunk splitting from dynamic imports.
- All build commands are yarn scripts. `yarn build`, `yarn build:dev` `yarn build:analyze` `yarn build:polyfills`
- Includes webpack bundle analyzer. `yarn build:analyze`.
- React Hot Loader integrated.
- Dynamic imports work.
- Outputted files have moved slightly. They are still in the JS folder however, and I've updated the AssetModel to load these files instead.
- No longer needs a PHP entry-point to hand of to a the JS build process.
- Developer build makes sure the HotReload config flag is set.

### What is the same

- Outputted files get built into the JS folder of their respective repos.
- Source files live in `src/scripts`
- Entry files live in `src/scripts/entries`
- Hot build still builds all addons (with react-hot-loader now too!)
- Typescript is still compiled properly.
- Polyfills still work and build properly (although the build separately with the `yarn build:polyfills` command).  _Tested in IE11_.

## Other change

- Build properties removed from addon.json. They are no longer necessary.
- The `app` section has been renamed to `forum`. This is in preparation of another frontend `app` section that will have different bundles.
- The rich editor is now mounted through a dynamic import.


applications/dashboard/models/class.usermodel.php
```diff
@@ -4060,7 +4060,7 @@ public function saveAuthentication($data) {
     /**
      * Set fields that need additional manipulation after retrieval.
      *
-     * @param array|object &$user
+     * @param array|object &$usera
```
slafleche: typo?

PR #7614, opened 2018-08-13
Allow the Memcached layer to use the container

tburry:
This is the first PR in an incremental attempt to remove cache config from `Gdn_Cache` and instead move towards config in the bootstrap.

## Why remove cache config from `Gdn_Cache`?

- I want to start moving to a PSR cache requirement for most objects rather than our own built in cache object. Injecting the cache driver will allow us to reuse the same driver for `Gdn_Cache` backwards compatibility and the PSR cache.
- Being able to inject the underlying cache driver allows the server to more easily roll its own caching rather than have our cache object handle 100% of the config. Once we've done the injection we'll be able to gut most of the config logic from `Gdn_Cache` and simplify our code base.
- Longer term, I want to remove `Gdn_Cache`'s dependency on `Gdn_Configuration`.
- Once we've moved to a PSR cache we will be able to use cache chaining objects for custom cache strategies (ex. Try APC, then go to Memcached). Doing so will allow us to remove this kind of bespoke implementation from `Gdn_Configuration`.

## Changes

- The `Gdn_Memcached` class can now take its underlying `Memcached` object as a constructor arg.
- The `Gdn_Cache` container rule no longer inherits. Since its initialize method acts as a factory itself it should only be used by the base class.
- Fixed some doc blocks on `Gdn_Cache`.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -49,6 +49,7 @@ $dic->setInstance('Garden\Container\Container', $dic)
     ->rule('Gdn_Cache')
     ->setShared(true)
     ->setFactory(['Gdn_Cache', 'initialize'])
+    ->setInherit(false)
     ->addAlias('Cache')
 
     // Configuration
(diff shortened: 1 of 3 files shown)


tburry:
The big question mark on these changes is infrastructure. I've tried this branch out on infrastructure and it seems to work okay. I'm not sure where else infrastructure uses `Gdn_Cache` though.

PR #7625, opened 2018-08-14
Enforce coding standards on changed code in build

### LINKED ISSUE #7393: Enforce PHP coding standards on changed code
Our current PHPCS config in https://github.com/vanilla/standards and CodeSniffer along with https://github.com/exussum12/coverageChecker should allow us to enforce our PHP coding standards on only changed code.

This should then be run in travis.

initvector:
### Overview
This update adds enforcement of [vanilla/standards](https://github.com/vanilla/standards) rules on changed code in a pull request. It accomplishes it in the following way:

1. Run a `git diff` on the current branch's code versus the pull request's target branch's code. Travis creates a shallow clone with only refs for the current branch, so resetting the refs and grabbing branches is required to perform this step. Capture the output.
1. Perform a [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) scan, using the rules from Vanilla's coding standards repo, on all files with changes. Capture the output.
1. Use [coverageChecker](https://github.com/exussum12/coverageChecker), along with the output from the previous two steps, to filter out the PHP_CodeSniffer messages not specifically related to lines with changed code. If any messages make it through, the build will fail.

The aforementioned steps are only performed if the build was triggered by a pull request. This is different from a build being triggered by a code push, alone. The branch used for the build is actually a test merge with the base branch, generated by GitHub. If GitHub can't create this branch (i.e. there are merge conflicts), the pull request build cannot be run.

### Testing

1. Make a change to a line in a file that will trigger a coding standard error. For example, add a new parameter to a function, but don't include any spacing (e.g. `public function __construct($argA, $argB, $argC,$newArg=false)`).
1. ~~Commit the change.~~
1. Run the diff-standards.sh script. The script takes two parameters: branch and directory. The branch parameter is required. It will be the used for the diff. The second parameter is optional and defaults to the current working directory. From the root of the repo, you may execute the following command: ./tests/travis/diff-standards.sh master
    1. Alternatively, you can push the commit and wait for the build to complete. View the build log to verify success. Afterwards, undo the commit (`git reset --hard HEAD~1`) and overwrite the remote branch (`git push --force`).

### Notes

Folding in the Travis build log is implemented here. A successful coding standards scan should result in everything be collapsed under the first line. The line can be clicked on to expand all output. You can search a completed build's log for the "coding_standards" slug or just scroll down to find "Verify changed code against coding standards".

Closes #7393

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,5 @@ npm-debug.log
 /cgi-bin/
 phpunit.xml
 yarn-error.log
+git-diff.txt
+phpcs-diff.json
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,9 @@
         "vanilla/vanilla-connect": "~0.0"
     },
     "require-dev": {
+        "exussum12/coverage-checker": "~0.10",
         "phpunit/phpunit": "~6.0",
+        "vanilla/standards": "~1.0",
         "voku/html-min": "~3.0"
     },
     "provide": {
--- a/composer.json
+++ b/composer.json
@@ -76,7 +76,8 @@
     },
     "scripts": {
         "pre-install-cmd": "Vanilla\\Setup\\ComposerHelper::preUpdate",
-        "pre-update-cmd": "Vanilla\\Setup\\ComposerHelper::preUpdate"
+        "pre-update-cmd": "Vanilla\\Setup\\ComposerHelper::preUpdate",
+        "lint-diff": "./tests/travis/diff-standards.sh"
     },
     "autoload-dev": {
         "psr-4": {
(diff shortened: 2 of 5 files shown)


charrondev:
I pushed a few tweak here but overall its looking pretty good.

- Only output the travis line breaks if the travis env variable is set.
- Add some extra whitespace in the output.
- Default the branch to check against as `master`. The override can still be set and the branch its checking will be outputted on the first line.
- Have the git diffs include unstaged files. This makes things easier to work with locally (not needing to commit to see if you passed the linter or not).
- Git ignoring the diff files.

Issue #7576, opened 2018-08-08
Rich Editor Feedback


slafleche:
- [x] Focussable embeds shouldn't get a border when not focussed (move shadow inside)
- [x] Make overlay white for embeds and only show on :active state
- [x] Not Enough Margins between blocks
  - [x] Typography tweaks
- [x] Quote embed -> Remove overlay
- [x] Quote embed needs a title 
- [x] Quote embed move chevron up on title line
- [x] Quotes needs a redesign for the toggle button
- [x] Get rid of ugly yellow for spoiler
- [x] Styles for code blocks to match Dan's designs
	- [x] Looks like there's extra padding on bottom of code blocks
- [x] Increase margins between elements in editor
- [x] Set typography line height to 1.5
- [x] Spoiler icon inconsistencies
- [x] Add missing viewbox to SVGs
- [x] Opacity -> change hover
- [x] Focus -> change bg
- [x] In the case of having both hover and focus, hover styles take over

tburry:
Here are my general list of changes:

## Changes for style.css

- [ ] Get rid of the borders, border radius on code blocks, make their backgrounds `rgba(127, 127, 127, .15)`.
- [ ] Remove the `font-size: 110%` from `.ItemDiscussion .Message`.

## Variables

I'm not sure what variables you are using so please feel free to rename stuff to what you are already using.

```scss
$font-size: 14px; // This is our font size. We don't need to set a rule for it long term, but I want to test.
$lh-condensed: 1.25; // line height for non-body text
$type-margin: $font-size;
$type-heading-margin: $type-margin * 1.5;
```

## User content changes

- [ ] All headings should have a `line-height: $lh-condensed`.
- [ ] All block-level elements should have a `margin-bottom: $type-margin`.
- [ ] All headings should have a `margin-top: $type-heading-margin`.
- [ ] All block elements with a border or background color should have a `padding: $type-margin`.
- [ ] Inline code should be `display: inline; padding: .2em .4em`.
- [ ] Lists should have a small margin between items: `li + li { margin-top: .25em }`.
- [ ] Lists should have a left margin of `1.5em` instead of 3em.
- [ ] Remove the left border and font color from embed quotes. It will use regular colors so it doesn't look weird when inline styles change color.
- [ ] Get rid of borders, border radius on spoilers. Reduce the font size on the heading to the regular font size. This probably also means the icon should be smaller to work with the smaller text.
[...]

PR #7634, opened 2018-08-15
Rich Editor Feedback and Bug Fixes

slafleche:
This is the final v1 feedback for the styles of the Rich Editor.

Note that we are doing a little bit of refactoring for core styles, but it'll only be applied to clients with the rich editor.

Closes: https://github.com/vanilla/vanilla/issues/7576


applications/dashboard/scss/editor/_richEditor.scss
```diff
@@ -214,6 +213,11 @@ $richEditor-mobileEmbedBreakPoint                 : 550px !default;
     user-select: none;
     cursor: pointer;
 
+    //.Close-x,
```
charrondev: Could we clear this out?

charrondev:
A few things some of which you already aware of:

## QuoteEmbed

- the top margin on the title doesn't seem to be even with the sides.
- The negative margin on the first item in the content is causing misalignment of the 
- The userContent class seems to be getting applied twice in the quote excerpt.
- We are console loggin the data every time we render an quote embed.

## General

- We need to do a production build on every PR until we start building on a build server.

charrondev:
## Hover focus states

image

The contrast here feels very low. Could we bring this up a little bit? This feels like a pretty significant regression from the previous hover/active/focus states.

PR #7637, opened 2018-08-16
Category-Following: add translation for "all"

dannimajdalani:
Added missing translation for "All" in Category-Following.


vrjrobitaille:
If this can get into the 2.7 VIP release I would be forever grateful. A VIP reported these issues after 2.6 and has held off on enabling Category Following until the menu is completely localized. Thanks guys!

PR #7639, opened 2018-08-16
Add Integration tests for comment drafts and modifying drafts

cchabilall83:
Added Integration test for the following

1. Alter the category of a saved draft discussion before posting.
2. Save a draft from the New Comment form on a discussion.
3. Post a saved draft comment to a discussion.

Closes Vanilla#7503

### DIFF
--- a/tests/APIv0/SmokeTest.php
+++ b/tests/APIv0/SmokeTest.php
@@ -503,6 +503,179 @@ class SmokeTest extends BaseTest {
     }
 
     /**
+     * Test modifying a category of a draft.
+     *
+     * @depends testRegisterBasic
+     * @large
+     */
+    public function testModifyDraftCategory() {
+        $api = $this->api();
+        $api->setUser($this->getTestUser());
+        $category = $this->createCategory();
+
+        $draft = [
+            'DiscussionID' => '',
+            'DraftD' => 0,
+            'CategoryID' => 1,
+            'Name' => 'testModifyDraftCategory',
+            'Format' => 'Markdown',
+            'Body' => 'Test posting a new draft',
+            'DeliveryType' => 'VIEW',
+            'DeliveryMethod' => 'JSON',
+            'Save Draft' => 'Save Draft',
+
+        ];
+
+        $r = $api->post(
+            '/post/discussion.json',
+            $draft
+        );
+        $responseBody = $r->getBody();
+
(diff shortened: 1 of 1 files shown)


tests/APIv0/SmokeTest.php
```diff
@@ -468,40 +448,152 @@ public function testPostDiscussionFromDraft() {
...
+        $this->assertEquals(200, $responseCode2);
+
+        $commentModel = new \CommentModel();
+        $dbComment = $commentModel->getWhere(['CommentID' => $commentID ])->firstRow(DATASET_TYPE_ARRAY);
+        $this->assertEquals($dbComment['DiscussionID'], $postComment['DiscussionID']);
+        $this->assertEquals($dbComment['Body'], $postComment['Body']);
+    }
+
+    /**
+     * Test that a category with restricted permissions can be created.
```
initvector: I don't think this description is accurate.

tests/APIv0/SmokeTest.php
```diff
@@ -468,40 +448,152 @@ public function testPostDiscussionFromDraft() {
...
+     * @large
+     */
+    private function createCategory() {
+        $api = $this->api();
+        $admin = $api->querySystemUser(true);
+        $api->setUser($admin);
+
+        $r = $this->api()->post('/vanilla/settings/addcategory.json', [
+            'Name' => 'Test Category ',
+            'UrlCode' => 'test',
```
initvector: If you call this function more than once in this test suite, I'm pretty sure it's going to fail those subsequent times, because the URL code is not going to be unique. Is there anything you can do here to ensure categories created with this function at least have a chance at unique names/URL codes?

cchabilall83: I added a name and url param to the method, this should any possible collisions.

tests/APIv0/SmokeTest.php
```diff
@@ -468,40 +448,153 @@ public function testPostDiscussionFromDraft() {
...
+    }
+
+    /**
+     * Create a category for testing.
+     *
+     * @param string $name Category name.
+     * @param  string $url Category url.
+     * @return array $category
+     */
+    private function createCategory($name = null, $url = null) {
```
initvector: I would've also accepted a dumb static int var in here that's incremented whenever the method is called and uses that value to create new names/URL codes.

Issue #7641, opened 2018-08-16
discussions/tagged/{tag} not displaying latest discussions


dannimajdalani:
### Details

Tagged discussions in discussions/tagged/{tag} is not displaying latest discussions.

Recent change in TagModel::getDiscussions did not introduce an orderBy when selecting tagged discussions.

[class.tagmodel.php#L527](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/models/class.tagmodel.php#L527)

PR #7642, opened 2018-08-16
Fix tagged discussions page not including the most recent discussions

dannimajdalani:
Closes [#7641](https://github.com/vanilla/vanilla/issues/7641)

discussions/tagged/{tag} was not displaying latest discussions.

### DIFF
--- a/applications/dashboard/models/class.tagmodel.php
+++ b/applications/dashboard/models/class.tagmodel.php
@@ -519,7 +519,7 @@ class TagModel extends Gdn_Model {
      * @return Gdn_DataSet
      * @throws Exception
      */
-    public function getDiscussions($tag, $limit, $offset) {
+    public function getDiscussions($tag, $limit, $offset, $sortField = 'DiscussionID', $sortDirection = 'desc') {
         if (!is_array($tag)) {
             $tags = array_map('trim', explode(',', $tag));
         }
@@ -530,7 +530,7 @@ class TagModel extends Gdn_Model {
             ->join('Tag t', 't.TagID = td.TagID')
             ->whereIn('t.Name', $tags)
             ->limit($limit, $offset)
-            ->orderBy('td.DiscussionID', 'desc')
+            ->orderBy($sortField, $sortDirection)
             ->get()->resultArray();
 
         $taggedDiscussionIDs = array_column($taggedDiscussionIDs, 'DiscussionID');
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.tagmodel.php
```diff
@@ -530,6 +530,7 @@ public function getDiscussions($tag, $limit, $offset) {
             ->join('Tag t', 't.TagID = td.TagID')
             ->whereIn('t.Name', $tags)
             ->limit($limit, $offset)
+            ->orderBy('td.DiscussionID', 'desc')
```
initvector: What about adding sort order and direction as parameters on the function, with these values as the defaults, instead of forcing a sort order? Maybe one day we'll want an alphabetized list, etc.

dannimajdalani: That would be a better practice.

applications/dashboard/models/class.tagmodel.php
```diff
@@ -530,6 +532,7 @@ public function getDiscussions($tag, $limit, $offset) {
             ->join('Tag t', 't.TagID = td.TagID')
             ->whereIn('t.Name', $tags)
             ->limit($limit, $offset)
+            ->orderBy($sortField, $sortDirection)
```
initvector: What about using these parameters when getting discussions, a couple lines down? If I were to pass in "DateInserted" for `$sortField`, I would expect it to be applied to the discussion, but as it stands, it'll be applied to when the discussion was tagged. I don't know if `DiscussionModel::get` would be flexible enough to allow you to set sorting options. You might need to switch it up to `DiscussionModel::getWhere`.

initvector: Now that we're using the sort for the discussions, we probably don't need to sort the TagDiscussion results too.

applications/dashboard/models/class.tagmodel.php
```diff
@@ -530,18 +532,22 @@ public function getDiscussions($tag, $limit, $offset) {
...
             [
                 'Announce' => 'all',
                 'd.DiscussionID' => $taggedDiscussionIDs,
-            ]
+            ],
+            $sortField,
+            $sortDirection,
+            '',
+            0
+
```
initvector: Do we need these last two parameters? You have an empty string for `$limit` and a zero for `$offset. If you omit both of those values, does it still function the same way?

initvector: An empty line has snuck into your parameters!

dannimajdalani: Those can be omitted.

linc:
This PR did not have adequate QA steps listed, nor does it describe the intended outcome. The issue is also vague about the desired functionality ("latest discussion" is imprecise).

Issue #7520, opened 2018-07-25
Mention menu issues


charrondev:
From @slafleche https://github.com/vanilla/vanilla/issues/7162#issuecomment-407062961

### Description

@mentions screw up with selection leaves editor and comes back.

There is likely some edge case not being handled in the rich editor `instanceReducer`.

### Issue severity
Low

### Steps to reproduce

1. Start @mention so we have suggested results
2. Click outside editor (suggestion disappear)
3. Click back to end of @mention (suggestions come back)
4. Hit "enter" we get a new line instead of accepting the selected selection. We instead get a new line and the @mention stays open.

We should complete the @mention instead of inserting a new line.

charrondev:
- Double check interactions with the paragraph menu.
- Both menus can be open at once.
- Applying a line format while the mention is active creates an error that crashes the editor.
- @mentions can be created inside of code blocks.
- @mentions cannot be created inside of lists (list keyboard shortcut is taking priority).

PR #7643, opened 2018-08-16
Fix various rich mention issues

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7520

This PR fixes multiple issues with the mention menu:
- [x] @mentions screw up when selection leaves editor and comes back mid-mention
- [x] Opening on of the other flyout menus or parapraph menu closes the mention menu and stops the mention.
- [x] Code block format can be applied while a mention is in the selection or line. This actually resolves the issue for all formats when applying code. Simply put, inline embeds are converted into plaintext versions of themselves when converted to a codeBlock.
- [x] The mention menu does not appear inside of a code block.
- [x] Mentions can now be create inside of all line formats including lists.

## Implementation details

- The MentionToolbar has been updated to use the selections provided in `instanceState` from redux, similar to the `InlineToolbar` and `ParagraphToolbar`.
- There is an additional computed property on the instanceState `mentionSelection`. This is now calculated during the `SET_SELECTION` action.
  - Since this is not calculated in the store, the selection listener inside of the `MentionToolbar` was able to be removed.
- The mention suggestions are now truncated to their limit in the outer `MentionToolbar` component. This fixes the arrowing behaviour as the outer toolbar and inner suggestions lists now work with the same data.
- The `getMentionSelection()` utility function has been updated to return a null selection  more often (such as in a code block).
- The `MentionSuggestionList` no longer needs a quill instance.
- The code blot formatter function now properly strips all values inside of it, including inline embeds. Additional unit tests have been added to ensure this.


charrondev:
I've pushed changes resolving your feedback. I added a few additional unit tests as well, mostly around getting the mention selection (that's where most of your reported issues came from). The list item was missing from this list:
https://github.com/vanilla/vanilla/blob/38c33221d5f17296a1d970267d757fd82c8be5bc/plugins/rich-editor/src/scripts/quill/registerQuill.ts#L99-L106

PR #7644, opened 2018-08-17
Fix discussion CSS classes

tburry:
- Only try and apply the discussion CSS classes if a discussion field is detected.
- Pass $inList = true in category lists.


tburry:
This really needs to go in as without it the styles are noticeably ugly when in a discussion.

PR #7651, opened 2018-08-20
Add recorded twitch embeds

topcheese13:
We currently handle live channels twitch embeds, this addition will also handle twitch embeds of previously recorded videos.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1434,13 +1434,10 @@ class Gdn_Format {
                 'regex' => ['/https?:\/\/embed.gettyimages\.com\/([\w=?&;+-_]*)\/([\d]*)\/([\d]*)/i']
             ],
             'Twitch' => [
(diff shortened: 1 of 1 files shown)


library/core/class.format.php
```diff
@@ -1436,6 +1436,12 @@ private static function embedReplacement($url) {
             'Twitch' => [
                 'regex' => ['/https?:\/\/(?:www\.)?twitch\.tv\/([\w]+)/i']
             ],
+            'TwitchRecorded' => [
+                'regex' => ['/https?:\/\/(?:www\.)?twitch\.tv\/videos\/(\w+)$/i']
+            ],
+            'Twitch' => [
+                'regex' => ['/https?:\/\/(?:www\.)?twitch\.tv\/([\w]+)$/i']
```
dannimajdalani: Why do we need to regex entries for Twitch?
line 1436 & 1443

topcheese13: We don't! :P

PR #7652, opened 2018-08-21
memcache type error when posting comments

dannimajdalani:
Update array $finalOptions type casting.

We experienced errors on cluster 40011
`
Aug 21 09:29:30 app2.cl40011.vanilladev.com php-error:  2018-08-21T12:29:30Z	php-error	{"level":"Fatal","error":"Uncaught TypeError: Argument 2 passed to Gdn_Cache::makeKey() must be of the type array, null given, called in /var/www/frontend-b755a7995be475/library/core/class.memcached.php on line 443 and defined in /var/www/frontend-b755a7995be475/library/core/class.cache.php:644","program":"php-error","source":"app2.cl40011.vanilladev.com"}`
 to fix it here is the patch.

### DIFF
--- a/library/core/class.memcached.php
+++ b/library/core/class.memcached.php
@@ -440,7 +440,7 @@ class Gdn_Memcached extends Gdn_Cache {
         $expiry = (int)val(Gdn_Cache::FEATURE_EXPIRY, $finalOptions, 0);
         $useLocal = (bool)$finalOptions[Gdn_Cache::FEATURE_LOCAL];
 
-        $realKey = $this->makeKey($key, $finalOptions);
+        $realKey = $this->makeKey($key, $finalOptions ?? []);
 
         // Should auto sharding be enabled?
         $keySize = strlen(serialize($value));


initvector:
This fix should probably include pulling type declarations off those existing public method parameters. A particular occurrence of this behavior has been addressed, but it seems like a backwards-compatibility-breaking change was introduced.

alex-mtl:
@initvector Yes it was. But I prefer to push project forward to good standards. I suspect we have not that much code which calls this makeKey() and strict typing will easily detect them all and to detect where we use it wrong. So this patch s temporary, I'll review that code to understand where that null is coming from, because on the first look it should never be null.

initvector:
> I suspect we have not that much code which calls this makeKey() and strict typing will easily detect them all and to detect where we use it wrong.

With Vanilla being an open-source project, it's important to keep in mind that third-party addons exist that we could potentially be breaking with these updates. This could easily break someone's caching adapter addon. They could update their code to be compliant, but in that case, I think the responsibility is more on Vanilla to not introduce breaking changes like this without sufficient warning.

alex-mtl:
@initvector I would agree if that was a theoretical discussion, but looking at our cache class, I have big doubt anybody would reinvent their own cache adapter. 
Yes, I am new to the project, please discuss with Todd and Linc and feel free to make any decision you feel comfortable. But my opinion - that is better to push strict typing as much as possible.

initvector:
> that is better to push strict typing as much as possible.

For new stuff, absolutely. But you're increasing the strictness of existing public functionality. I don't think that's a good practice, without advanced planning and warnings for users. Generally speaking, for publicly-accessible members, you can only make them more permissive, not less. If you increase the potential for errors, you're going to get more errors. I don't think this is discussion is specific to caching adapters. If it's a public method, existing parameters shouldn't suddenly stop handling values they previously accepted.

alex-mtl:
@initvector  

>   I don't think this is discussion is specific to caching adapters
To me it is not even about "caching adapters" it is about class.memcached.php and class.cache.php files. Can you explain  me how library/core/class.memcached.php:443 $finalOptions get null value? I can see 
`        $this->StoreDefaults = [
            Gdn_Cache::FEATURE_COMPRESS => false,
            Gdn_Cache::FEATURE_TIMEOUT => false,
            Gdn_Cache::FEATURE_EXPIRY => false,
            Gdn_Cache::FEATURE_NOPREFIX => false,
            Gdn_Cache::FEATURE_FORCEPREFIX => null,
            Gdn_Cache::FEATURE_SHARD => false,
            Gdn_Cache::FEATURE_LOCAL => true
        ];`
in __construct() of memcached and I see

`$finalOptions = array_merge($this->StoreDefaults, $options);`
in library/core/class.memcached.php:438

So to me something very wrong happens on production to get `array_merge` returning null.
And I prefer to find and fix it than to keep weird backward compatibility for scenarios that should not exist ever.

If you have any idea why makeKey() should accept null instead of array - please share.
And again in my opinion it is not about patterns and open source it is about particular class, method, arguments implementation.

initvector:
> If you have any idea why makeKey() should accept null instead of array - please share.

Because it did. I'm not saying it should forever support `null` in perpetuity, but a *public method* shouldn't go from supporting non-array values to only supporting array values, overnight. If you want to only support arrays, that's fine, but I am of the opinion you should use a phased approach.

PR #7653, opened 2018-08-21
Fix spam logging & log restoring by packing IPs before insertions

topcheese13:
What this PR does
---
We recently updated our handling of IPv6. We found 2 cases where IPs aren't encoded/packed prior to insertion.

We need to encode IPs before inserting them into the Log table to make sure we aren't busting the column limit. This PR does this. We also need to encode IPs before inserting them when restoring from the log, our current process decodes them on fetch. This PR does this.

Strict Mode & IPv6
---
FYI, With our current IP columns set to `VARBINARY 16`, without MySQL strict mode it will actually truncate the value for IPv6, hence, storing a erroneous IP, but with MySQL strict mode it will fail entirely, throwing a `Data too long for column ...` error.

### DIFF
--- a/applications/dashboard/models/class.logmodel.php
+++ b/applications/dashboard/models/class.logmodel.php
@@ -866,7 +866,7 @@ class LogModel extends Gdn_Pluggable {
             $tableName = $log['RecordType'];
         }
 
-        $data = $log['Data'];
+        $data = ipEncode($log['Data']);
 
         if (isset($data['Attributes'])) {
             $attr = 'Attributes';
--- a/applications/dashboard/models/class.spammodel.php
+++ b/applications/dashboard/models/class.spammodel.php
@@ -103,6 +103,9 @@ class SpamModel extends Gdn_Pluggable {
 
         // Log the spam entry.
         if ($spam && val('Log', $options, true)) {
+            // Make sure all IP addresses are packed before insertion
+            $data = ipEncodeRecursive($data);
+
             $logOptions = [];
             switch ($recordType) {
                 case 'Registration':


linc:
Are we in strict mode on cloud?

linc:
This should ideally be referencing an issue it's fixing. At the very least, it should describe the error report we received that it is fixing in the description.

dannimajdalani:
Once a comment is marked as spam, if you flag it as not spam from the dahsboard you'll get the below error:

`Data too long for column 'InsertIPAddress' at row 1|Gdn_ Database|Query|replace `GDN_Comment` (`DiscussionID`, `InsertUserID`, `Body`, `Format`, `DateInserted`, `InsertIPAddress`) values (:DiscussionID, :InsertUserID, :Body, :Format, :DateInserted, :InsertIPAddress)`

topcheese13:
Like I said with strict mode it will fail `Data too long for column ...` without strict mode it won't fail but it will truncate the value, hence, storing a faulty IP address for IPv6.

Regardless of this all of our IPs are stored encoded, not only is it preventing errors and truncating it's also following what we do everywhere else.

topcheese13:
There is nothing wrong with my PR.

It's a result of us storing the IPv6 correctly, it probably gets decoded and never encoded again before restoring it. We need to also encode those IPs before the restore goes through.

dannimajdalani:
And we should create another PR for that?

linc:
Let's get the whole workflow fixed here, please. Update the description of this PR to indicate the scope of the problem we're solving.

Adding a patch and shrugging about MySQL strict mode compliance isn't OK.

topcheese13:
> Let's get the whole workflow fixed here, please. Update the description of this PR to indicate the scope of the problem we're solving.
> 
> Adding a patch and shrugging about MySQL strict mode compliance isn't OK.

I have updated the PR with an extra fix, the title and the description to reflect those changes. MySQL strict mode was just extra information on this PR to explain how this process behaved under strict mode or not, it was never related to the subsequent fix that was needed to restore records. This fix was needed because storing full IPv6 in the log table was than causing the same bug we had on the actual insertion of logs when we were restoring them to their rightful table.

PR #7654, opened 2018-08-21
Add a few more classes for theming with category headings

slafleche:
For theming, it's useful to know if you're not on a category heading and the last category before the heading starts.

### DIFF
--- a/applications/vanilla/views/categories/helper_functions.php
+++ b/applications/vanilla/views/categories/helper_functions.php
@@ -161,6 +161,16 @@ if (!function_exists('writeListItem')):
                 </div>
             </li>
         <?php else: ?>
+            <?php
+                $cssClass .= ' Item-Category';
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/categories/helper_functions.php
```diff
@@ -161,6 +161,16 @@ function writeListItem($category, $depth) {
                 </div>
             </li>
         <?php else: ?>
+            <?php
+                $cssClass .= ' Item-Category';
+                if (isset($category['TreeRight'])) {
+                    $nextCategoryID = $category['TreeRight'];
+                    $nextCategoryDisplay = 'Heading'; // should be dynamic
+                    if ($nextCategoryDisplay === 'Heading') {
```
charrondev: I'm a little confused here. This if statement will always be true.

Issue #7662, opened 2018-08-22
Add ability to bust cache on dynamically imported JavaScript modules


initvector:
#7612 introduced dynamic importing of a module, but no way to bust the cache on new revisions. #7661 attempted a temporary workaround for one specific instance, but does not offer a general purpose caching solution for dynamically-imported modules. Vanilla needs such a solution if it is to continue utilizing dynamic imports.

PR #7670, opened 2018-08-23
Fix cache busting of dynamically imported modules

charrondev:
## Chunk-hash as a cache buster
Closes https://github.com/vanilla/vanilla/issues/7662

- Reverts the changes from #7661
- Adds the hash of a dynamic chunks contents as a query parameter at the end of its url.

### To verify

- Checkout this branch
- Navigate to a page where the rich editor mounts.
- Inspect the head of the document.
- Notice that the dynamic chunk `plugins/rich-editor/js/webpack/chunks/mountEditor.min.js` has this query parameter on the end of it.

image

## Webpack jsonp function name

Closes https://github.com/vanillaforums/immediate-media/issues/4

A conflict was reported by a client with us both using the same webpackJsonP global name. Webpack stores it's list of modules in a global variable for all chunks to reference, but we were using the default `window.webpackJsonp`. A client was also using the default and having both scripts creates a conflict preventing ours from loading.

The remedy for this is setting the `library` name or the the `jsonpfunction` https://webpack.js.org/configuration/output/#output-jsonpfunction.


js/webpack/runtime-admin.min.js
```diff
@@ -1,2 +1,2 @@
...
...e,r=0;r<u.length;r++){for(var n=u[r],t=!0,l=1;l<n.length;l++){var a=n[l];0!==o[a]&&(t=!1)}t&&(u.splice(r--,1),e=i(i.s=n[0]))}return e}var t={},o={1:0},u=[];function i(r){if(t[r])return t[r].exports;var n=t[r]={i:r,l:!1,exports:{}};return e[r].call(n.exports,n,n.exports,i),n.l=!0,n.exports}i.m=e,i.c=t,i.d=function(e,r,n){i.o(e,r)||Object.defineProperty(e,r,{enumerable:!0,get:n})},i.r=function(e){"undefined"!=typeof Symbol&&Symbol.toStringTag&&Object.defineProperty(e,Symbol.toStringTag,{value:"Module"}),Object.defineProperty(e,"__esModule",{value:!0})},i.t=function(e,r){if(1&r&&(e=i(e)),8&r)return e;if(4&r&&"object"==typeof e&&e&&e.__esModule)return e;var n=Object.create(null);if(i.r(n),Object.defineProperty(n,"default",{enumerable:!0,value:e}),2&r&&"string"!=typeof e)for(var t in e)i.d(n,t,function(r){return e[r]}.bind(null,t));return n},i.n=function(e){var r=e&&e.__esModule?function(){return e.default}:function(){return e};return i.d(r,"a",r),r},i.o=function(e,r){return Object.prototype.hasOwnProperty.call(e,r)},i.p="/";var l=window.webpackJsonpvanillaadmin=window.webpackJsonpvanillaadmin||[],a=l.push.bind(l);l.push=r,l=l.slice();for(var f=0;f<l.length;f++)r(l[f]);var p=a;n()}([]);
```
charrondev: Notice the runtime chunk has a different global variable for it's jsonP function.

`window.webpackJsonp` -> `window.webpackJsonpvanillaadmin`
`window.webpackJsonp` -> `window.webpackJsonpvanillaforum`

Issue #87, opened 2009-08-17
PHPMailer default charset


ghost:
Change default charset in PHPMailer from "iso-8859-1" to "utf-8"

PR #7673, opened 2018-08-24
Tagged Discussions: Update tagged discussions sort order

dannimajdalani:
closes [#87](https://github.com/vanilla/support/issues/87)
related [#7642](https://github.com/vanilla/vanilla/pull/7642)

### Description
PR [#7642](https://github.com/vanilla/vanilla/pull/7642) was ordering discussion by page on DiscussionID. I modified the query to order tagged discussion by DateLastComment.

### To Test

1. Tag enough discussions so that you can have multiple pages on /discussions/tagged/{Tag.Name}.
1. Get the full list of discussions for a tag, sorted by the `DateLastComment` field. You can use a database query for this.
1. View /discussions/tagged/{Tag.Name}, using your chosen tag. Verify the results are the same as your list from step two for the first page.
1. Advance to the second page. Verify your list from step two continues in the same order on this page.

### DIFF
--- a/applications/dashboard/models/class.tagmodel.php
+++ b/applications/dashboard/models/class.tagmodel.php
@@ -521,7 +521,7 @@ class TagModel extends Gdn_Model {
      * @return Gdn_DataSet
      * @throws Exception
      */
-    public function getDiscussions($tag, $limit, $offset, $sortField = 'DiscussionID', $sortDirection = 'desc') {
+    public function getDiscussions($tag, $limit, $offset, $sortField = 'd.DateLastComment', $sortDirection = 'desc') {
         if (!is_array($tag)) {
             $tags = array_map('trim', explode(',', $tag));
         }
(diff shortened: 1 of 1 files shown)


dannimajdalani:
I ran a test on a large database with tagged discussions, query took ~70 ms with the JOIN  and ~30ms without it.

applications/dashboard/models/class.tagmodel.php
```diff
@@ -530,8 +530,10 @@ public function getDiscussions($tag, $limit, $offset, $sortField = 'DiscussionID
             ->select('td.DiscussionID')
             ->from('TagDiscussion td')
             ->join('Tag t', 't.TagID = td.TagID')
+            ->join('Discussion d', 'd.DiscussionID = td.DiscussionID')
             ->whereIn('t.Name', $tags)
             ->limit($limit, $offset)
+            ->orderBy($sortField, $sortDirection)
             ->get()->resultArray();
```
cchabilall83: should we validate the $sortField and $sortDirection?

PR #7677, opened 2018-08-24
Fix new username validation being incorrectly applied to SSO connections by re-applying code lost in bad merge

topcheese13:
Why we need this
---
Users using SSO couldn't register with an "invalid" username. After investigation we realized that pr #7250 which introduced our new username validation logic was missing an important part of its implementation. At this point we are not exactly sure why (some merging voodoo perhaps).

Locally re-introducing these changes allowed invalid usernames (spaces and dots) to register properly if we are linking by email or if it's a complete new user without any linking to make.

A little history of what happened and when
---

1. A feature branch, feature/usermodel-name-validation, was created for work on updating username validation.
1. PR #7148, targeting feature/usermodel-name-validation,  was created on May 1st to update username validation.
1. PR #7163, also targeting feature/usermodel-name-validation, was created on May 2nd, spinning off portions of the first PR into a smaller diff for review. This PR then became a blocker for the first PR.
1. Code not relevant to #7163 made it into the branch during the split. It was reverted with 57de5ff8217ed50143a21347575492575e13ade8. This code still existed in #7148.
1. Both PRs were merged around the same time. According to GitHub, they were merged within the same minute.
1. PR #7250 was created on May 24 to merge feature/usermodel-name-validation into master. It was merged shortly thereafter.

The revert commit from #7163 seems to have targeted the original code from #7148 during the merge into the feature branch.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -499,8 +499,10 @@ class EntryController extends Gdn_Controller {
             return;
         }
 
+        $isTrustedProvider = $this->data('Trusted');
+
         // Check if we need to sync roles
-        if (($this->data('Trusted') || c('Garden.SSO.SyncRoles')) && $this->Form->getFormValue('Roles', null) !== null) {
+        if (($isTrustedProvider || c('Garden.SSO.SyncRoles')) && $this->Form->getFormValue('Roles', null) !== null) {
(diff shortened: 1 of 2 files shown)


initvector:
The revert commit (57de5ff8217ed50143a21347575492575e13ade8) was done because those changes were supposed to be part of #7148. I'm assuming that since the revert commit in branch A actually followed the real commits from branch B, it took priority during the merge, even though branch B was merged after branch A.

Issue #7682, opened 2018-08-27
Add basic new controller class


initvector:
Reflect to https://github.com/vanilla/knowledge/issues/80 Add "empty" clean Controller class
Add a child class PageController (which supports: meta tags, js and css resources to add)
KnowledgePageController should extend PageController

PR #7699, opened 2018-08-30
New base controller empty class and TwigTrait initializer

alex-mtl:
Garden\Controller and Garden\TwigTrait are set (for Knowledgebase)
Fixes: #7682

### DIFF
--- a/library/Garden/TwigTrait.php
+++ b/library/Garden/TwigTrait.php
@@ -19,7 +19,7 @@ trait TwigTrait {
      * @return \Twig_Environment
      */
     protected static function twigInit() {
-        $loader = new \Twig_Loader_Filesystem(self ::$twigDefaultFolder);
+        $loader = new \Twig_Loader_Filesystem(self::$twigDefaultFolder);
         return new \Twig_Environment($loader);
     }
 }
--- a/library/Garden/TwigTrait.php
+++ b/library/Garden/TwigTrait.php
@@ -12,14 +12,14 @@ namespace Garden;
  *
  */
 trait TwigTrait {
-    protected static $defaultFolder = PATH_ROOT.'/plugins/knowledge/views';
+    protected static $twigDefaultFolder = PATH_ROOT;
     /**
      * Twig templating environment initialize
      *
      * @return \Twig_Environment
      */
     protected static function twigInit() {
-        $loader = new \Twig_Loader_Filesystem(self ::$defaultFolder);
+        $loader = new \Twig_Loader_Filesystem(self ::$twigDefaultFolder);
         return new \Twig_Environment($loader);
     }
 }
--- a/library/Garden/TwigTrait.php
+++ b/library/Garden/TwigTrait.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * @author Alexander Kim <alexander.k@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+namespace Garden;
+
+/**
+ * For classes that need to cache some static values and configs.
+ *
+ */
+trait TwigTrait {
+    protected static $defaultFolder = PATH_ROOT.'/plugins/knowledge/views';
+    /**
+     * Twig templating environment initialize
+     *
+     * @return \Twig_Environment
+     */
+    protected static function twigInit() {
+        $loader = new \Twig_Loader_Filesystem(self ::$defaultFolder);
+        return new \Twig_Environment($loader);
+    }
+}
--- a/library/Garden/Controller.php
+++ b/library/Garden/Controller.php
@@ -1,0 +1,15 @@
+<?php
+/**
+ * @author Alexander Kim <alexander.k@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL v2
+ */
+
+namespace Garden;
+
+/**
+ * Base Class Controller.
+ */
+class Controller {
+
+}


library/Garden/TwigTrait.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace Garden;
+
+/**
+ * For classes that need to cache some static values and configs.
+ *
+ */
+trait TwigTrait {
+    protected static $defaultFolder = PATH_ROOT.'/plugins/knowledge/views';
```
initvector: Is there any way you can avoid hardcoding a reference to an addon here?

library/Garden/TwigTrait.php
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Twig templating environment initialize
+     *
+     * @return \Twig_Environment
+     */
+    protected static function twigInit() {
+        $loader = new \Twig_Loader_Filesystem(self ::$defaultFolder);
```
initvector: There shouldn't be a space before the scope resolution operator.

library/Garden/TwigTrait.php
charrondev: We definitely shouldn't be hardcoding paths to external projects here.

library/Garden/TwigTrait.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ */
+trait TwigTrait {
+    protected static $defaultFolder = PATH_ROOT.'/plugins/knowledge/views';
+    /**
+     * Twig templating environment initialize
+     *
+     * @return \Twig_Environment
+     */
+    protected static function twigInit() {
```
initvector: Is it possible to make this something that doesn't need to be instantiated and configured by every controller that uses this trait? It'd be great if we had a single Twig instance that could be used throughout Vanilla and not one per class.

alex-mtl: That is something I feel we have opposite points of view.
I see that practice everywhere in old Vanilla code sending same objects/containers everywhere.
Why Don't you like to have very well isolated areas? 
In this case whenever I use Twig in my controller class I will be absolutely confident that it is my fresh object.
Why should we reuse this object in any other class? what is the profit/pros? I see only cons.
By the way, we can implement singleton to have shared instance using same trait in future. I just can't get why do we need it.

initvector: It makes a lot more sense to me to configure the view layer one time, and allow the application to use it in as many areas as it needs. For example, if an addon (or Vanilla) wanted to extend Twig, it shouldn't need to be done more than once. We could use a shared instance to achieve that. With a per-class instance, it becomes more difficult. Are there any real "pros" to potentially having multiple instances of Twig, each with their own separate config ("because they're isolated" isn't specific enough)? The only real "con" I can see from a shared instance would be possible data pollution, but I don't think that's even a concern here.

library/Garden/TwigTrait.php
```diff
@@ -12,14 +12,14 @@
...
-    protected static $defaultFolder = PATH_ROOT.'/plugins/knowledge/views';
+    protected static $twigDefaultFolder = PATH_ROOT;
     /**
      * Twig templating environment initialize
      *
      * @return \Twig_Environment
      */
     protected static function twigInit() {
-        $loader = new \Twig_Loader_Filesystem(self ::$defaultFolder);
+        $loader = new \Twig_Loader_Filesystem(self ::$twigDefaultFolder);
```
initvector: Don't forget: no space before the scope resolution operator.

library/Garden/TwigTrait.php
alex-mtl: >For example, if an addon (or Vanilla) wanted to extend Twig, it shouldn't need to be done more than once.
I can't get this idea. If we want to allow to extend - that means to extend as many time as many plugins we have. That is kind definition of OOP "extends" term I have in my mind. So when I see "it shouldn't need to be done more than once" following after "wanted to extend" I lose the idea completely. 

>With a per-class instance, it becomes more difficult
IMHO it becomes more clear, inheritance is the thing. Whenever I am in my child class - I can easily use any parent methods and I can redefine any or create something new when I need. So I can't get what exactly you worry about. probably we need to discuss some example to illustrate the issue/problem you want to avoid

charrondev: @initvector If we wanted to do some global configuration of twig configuration couldn't we do it in this trait?

I prefer taking the magic out of the view rendering and making it more explicit in this way. We currently do not have any requirement or plan or intention to require any weird lookup chains for where the template lives, so:

If the declaration of what view we need to render is explicit, why can't the declaration of your templating system?

initvector: > If we wanted to do some global configuration of twig configuration couldn't we do it in this trait?

Sure! I'm not opposed to that. Ultimately, my primary concern is that Vanilla's usage of Twig won't be this simple for very long. I anticipate needing to configure filters or tags or extensions or helpers or whatever else. I don't see the real benefit in having to create a new instance and potentially reconfigure it with these items, every time something needs to be rendered. I don't think the controller necessarily needs to be concerned with defining the details of the view layer. There doesn't seem to be a persistent state for a Twig instance, outside of its configuration, so I'm not seeing the benefit of having multiple instances with separate configurations for the sake of potentially shaving off some bits or ticks in certain scenarios. **However**, this is a team decision. If I'm in the minority on this, that's fine. I won't consider this to be a blocker.

library/Garden/TwigTrait.php
```diff
@@ -0,0 +1,25 @@
...
+ * @license GPLv2
+ */
+
+namespace Garden;
+
+/**
+ * For classes that need to cache some static values and configs.
+ *
+ */
+trait TwigTrait {
```
initvector: Could name of the trait and its properties be less specific to Twig? I'm not altogether opposed to hardcoding the usage of Twig in the body of these methods, but the next time Vanilla switches templating engines, it'd be nice if we we could swap out the guts of these methods without the need to also rename traits and methods for every class using this trait.

alex-mtl: IMHO  that is the only reason for this trait to exist. The idea is to have multiple traits specific to any particular library we need and to have common entry point for other classes to use it. So whenever any class need Twig to get - it uses TwigTrait, if class needs Mustache - we can have MustacheTrait to implement if we need Smarty - we can have SmartyTrait.

In this particular scenario my idea is to have method MustacheInit, SmartyInit etc. And our new knowledge base controller 
https://github.com/vanilla/knowledge/blob/19610933abbdc208b78a2b072882ffe82644c0a6/plugins/knowledge/Controllers/PageController.php#L40
will call particular templating system/library to instantiate it for particular Controller. Where every controller will define itself which templating it need by `use TwigTrait` or `use SmartyTrait` etc.

PR #7692, opened 2018-08-29
Ongoing removal of val from codebase

alex-mtl:
Refactor joinUsers function of class.usermodel to avoid val() calls
Fixes: vanilla/vanilla-patches#350

Refactor _formatStringCallback function of functions.general to avoid val() calls
Fixes: vanilla/vanilla-patches#361

Refactor getVisibleCategories function of class.categorymodel to avoid val() calls
Fixes: vanilla/vanilla-patches#360

Refactor buildCssClass function of class.nestedcollection to avoid val() calls
Fixes: vanilla/vanilla-patches#359

Refactor toString function of class.headmodule to avoid val() calls
Fixes: vanilla/vanilla-patches#364

Refactor flattenTreeInternal function of class.categorycollection to avoid val() calls
Fixes: vanilla/vanilla-patches#365

Refactor smarty_function_asset function of function.asset.php to avoid val() calls
Fixes: vanilla/vanilla-patches#366

PR #7705, opened 2018-08-31
Add array access to object rows in DataSet results

bleistivt:
Idea by @charrondev here: https://github.com/vanilla/vanilla/pull/7692#pullrequestreview-150776737

However, Gdn_DataSet does not benefit from array access as the result set is an array anyway.
But the rows it contains can be either objects or arrays.

If these objects implement the `ArrayAccess` interface, result rows can be accessed as both `$row['DiscussionID']` and `$row->DiscussionID` interchangeably. This would allow refactoring all views to use array notation _right now_, independently of model changes, and would ease the transition to array only return types.

Conveniently, there is already a PHP standard class that implements all array functionality:
http://php.net/manual/en/class.arrayobject.php

Using the `ARRAY_AS_PROPS` flag makes this "backwards compatible" as it allows changing getting/setting properties using both notations.

This is for discussion only as I haven't done any extensive checks. That said, I didn't notice any changes using this on my live forum and no errors popped up.

---

Notes:

1. This would seem to be the elegant way to do this, but it does not work:

    $this->_PDOStatement->fetchAll(PDO::FETCH_CLASS, ArrayObject::class)

...as just setting object properties via fetchAll does not populate the internal `Arrayobject::storage` array that manages the data.

2. All checks for `is_object` work, but raw array to `(object)` casts should be avoided using this (probably a good idea anyway).

### DIFF
--- a/library/database/class.dataset.php
+++ b/library/database/class.dataset.php
@@ -123,7 +123,7 @@ class Gdn_DataSet implements IteratorAggregate, Countable, JsonSerializable {
                             //$this->_Result[$Index] = (array)$this->_Result[$Index];
                             break;
                         case DATASET_TYPE_OBJECT:
-                            $row = (object)$row;
+                            $row = new ArrayObject($row, ArrayObject::ARRAY_AS_PROPS);
                             //$this->_Result[$Index] = (object)$this->_Result[$Index];
                             break;
                     }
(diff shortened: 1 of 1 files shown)


bleistivt:
Fetching directly into objects using `PDO::FETCH_PROPS_LATE` works as well:

```php
$result = $this->_PDOStatement->fetchAll(PDO::FETCH_CLASS|PDO::FETCH_PROPS_LATE, ArrayObject::class, [[], ArrayObject::ARRAY_AS_PROPS]);
```
http://php.net/manual/de/pdostatement.setfetchmode.php#98876

But in my tests, it did not make a difference at all (in fact it was slightly slower).

PHP 7.0.3 calling get()->result() 10000 times (DataSet of 3660 items, table of 14 columns), mean of 10 requests:
<table>
<tr><td>PDO::FETCH_CLASS (not compatible)</td><td>27.47s</td></tr>
<tr><td>PDO::FETCH_CLASS|PDO::FETCH_PROPS_LATE</td><td>27.90s</td></tr>
<tr><td>PDO::FETCH_ASSOC + object creation in foreach</td><td>27.45s</td></tr>
</table>

PR #7707, opened 2018-08-31
Add new basic base model

initvector:
In this update, we're adding a new base model. This will be an ongoing process, but this is the first step necessary for removing dependency on the existing `Gdn_Model` base model.

### Changes

1. New base model.
    1. This is intended to be a very slim model. Currently, it only supports three operations: insert, update and get. More methods will be added in future iterations.
    1. The model lives in the `Vanilla` namespace. This was chosen over `Garden`, because it has a dependency on `Gdn_Database`.
    1. The database object is injected into the model, which should be beneficial for testing.
    1. Validation is now done with Garden Schema, much like Vanilla's API. `Gdn_Validation` is out.
    1. Similar to `Gdn_Model`, a model's table schema is fetched from the database to build the data schema. Whereas `Gdn_Model` lazy loaded the schema, `Vanilla\Model` will load the schema when the instance is built by the container, since all of its operations require a schema. Everything is private, so the option to switch to lazy loading should be trivial.
1. My IDE complained whenever I used PHP's JSON functions, because the project's stated requirements in its Composer config did not cite JSON. Now it does.
1. A new static class for general-purpose field filters has been added. These filters are intended for use with Garden Schema instances that Vanilla uses. Currently, there are only two: one for unpacking a JSON/serialized field and one for packing (encoding as JSON) a value. Currently, this is functionality targeted at attributes fields.
1. There is now a "string utilities" class. It has one method: `jsonEncodeChecked`. This was previously a global function. It's guts have been moved into the utility class. The function now calls the utility method. It had no outside dependencies, so it seemed like a good candidate for transplanting.
1. `Gdn_Database` has two new methods: `simpleDataType` (private) and `simpleSchema` (public). These methods are used to translate a database schema into a Garden Schema instance. This instance can then be used by models for basic data validation, directly against their database table.
1. A small change has been made to `Gdn_SQLDriver::set`. If it receives a value that is an instance of `DateTimeImmutable`, it will convert that value into a datetime string for insertion into the database. This was primarily done to avoid the need to revert Garden Schema's translation of datetime strings to `DateTimeImmutable` objects when such fields are validated.

### DIFF
--- a/library/Vanilla/Model.php
+++ b/library/Vanilla/Model.php
@@ -13,7 +13,7 @@ use Garden\Schema\ValidationField;
 /**
  * Basic model class.
  */
-class Model {
+class Model implements InjectableInterface {
 
     /** @var \Gdn_Database */
     private $database;
@@ -30,60 +30,10 @@ class Model {
     /**
      * Basic model constructor.
      *
-     * @param string $table
-     * @param \Gdn_Database $database
+     * @param string $table Database table associated with this resource.
      */
-    public function __construct(string $table, \Gdn_Database $database) {
+    public function __construct(string $table) {
         $this->table = $table;
-        $this->database = $database;
-
-        $schema = $this->database->simpleSchema($table);
-        $this->setReadSchema(clone $schema)
-            ->setWriteSchema(clone $schema);
-    }
-
-    /**
-     * Attempt to decode a encoded string value into a more complex type.
-     *
-     * @param mixed $value Raw database value.
-     * @param ValidationField $field An object representing the Garden Schema field.
-     * @return mixed Unpacked attributes field value.
-     */
-    private function filterValueDecode($value, ValidationField $field) {
-        if ($value === null || $value === '') {
-            $value = null;
-        } elseif (is_string($value)) {
-            $value = json_decode($value, true);
-            if ($value === null) {
-                $value = unserialize($value);
-                if ($value === false) {
-                    $value = null;
-                }
-            }
-        }
-
-        return $value;
-    }
-
-    /**
-     * Attempt to decode a encoded string value into a more complex type.
-     *
-     * @param mixed $value Raw database value.
-     * @param ValidationField $field An object representing the Garden Schema field.
-     * @return mixed Unpacked attributes field value.
-     * @throws \Exception If there was an error encoding the value.
-     */
-    private function filterValueEncode($value, ValidationField $field) {
-        if ($value === null || $value === '') {
-            $value = null;
-        } else {
-            $value = json_encode($value, JSON_UNESCAPED_SLASHES);
-            $jsonError = json_last_error();
-            if ($jsonError !== JSON_ERROR_NONE) {
-                throw new \Exception("An error was encountered while encoding the JSON value ({$jsonError}).");
-            }
-        }
-        return $value;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


library/Vanilla/Model.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPLv2
+ */
+
+namespace Vanilla;
```
charrondev: Could we move this to a `Vanilla\DB` or `Vanilla\Database` namespace? I don't want to overcrowd the top level here.

library/core/functions.general.php
```diff
@@ -2329,37 +2329,11 @@ function joinRecords(&$data, $column = '', $unset = false, $checkCategoryPermiss
...
-                $errorMessage = 'A value of a type that cannot be encoded was given.';
-                break;
-            default:
-                $errorMessage = 'An unknown error has occurred.';
-        }
-        if ($errorMessage !== null) {
-            throw new Exception("JSON encoding error: {$errorMessage}", 500);
-        }
-        return $encoded;
+        return \Vanilla\Utility\StringUtils::jsonEncodeChecked($value, $options);
```
charrondev: Could we get a `deprecated()` call in here?

charrondev:
Overall seems fine. Just fix the lint error + feedback and it'll have an approval from me

library/core/functions.general.php
initvector: I'd be hesitant to add a call to `deprecated` there, until we actually took steps to remove it from the code, which should probably be a separate PR. There's currently a <span>@</span>deprecated tag on the function as a hint to the IDE.

library/Vanilla/Model.php
initvector: I moved this into `Vanilla\Models`. It was an existing namespace and seemed to fit.

library/Vanilla/Utility/SchemaFilters.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Attempt to decode a encoded string value into a more complex type.
+     *
+     * @param mixed $value Raw database value.
+     * @param ValidationField $field An object representing the Garden Schema field.
+     * @return mixed Unpacked attributes field value.
+     */
+    public static function encodeValue($value, ValidationField $field) {
+        if ($value === null || $value === '') {
+            $value = null;
```
linc: Why reset a `null` to `null`?

linc: I guess there's 2 schools of thought here - either we value a single return more, or not branching our logic more. Over the years, I've moved from the former to the latter. My feeling is this would do better as `if null return null` and then just return the json encoding as a matter of course otherwise. I'm unbothered if you have reason to keep an else here.

library/Vanilla/Models/Model.php
```diff
@@ -0,0 +1,139 @@
...
+     */
+    public function __construct(string $table) {
+        $this->table = $table;
+    }
+
+    /**
+     * Get resource rows from a database table.
+     *
+     * @param array $where Conditions for the select query.
+     * @param array $options Options for the select query.
```
linc: It's nice to list our options in the docs when we have an unknown set of keys allowed.

library/Vanilla/Utility/SchemaFilters.php
```diff
@@ -0,0 +1,55 @@
...
+     * @return mixed Unpacked attributes field value.
+     */
+    public static function decodeValue($value, ValidationField $field) {
+        if ($value === null || $value === '') {
+            $value = null;
+        } elseif (is_string($value)) {
+            $value = json_decode($value, true);
+            if ($value === null) {
+                try {
+                    $value = unserialize($value);
```
linc: I'm curious why we're allowing for serialization in our pristine new model. Without this, doesn't this basically work out to an elaborate alias for json_decode? Are we planning to point this model at old, busted data? Again, not super bothered, but curious what the plans are here.

initvector: > Are we planning to point this model at old, busted data?

Yep! Kinda, anyway. The aim here is to allow some kinda upgrade path for old models. For those, we'll allow reading of the existing serialized data, but force writing of JSON-encoded data.

alex-mtl:
Just a note.
To me this is not a model implementation but new hybrid of sql adapter and base model.
I mean 

```
public function get(array $where = [], array $options = []): array {
        $orderFields = $options['orderFields'] ?? '';
        $orderDirection = $options['orderDirection'] ?? 'asc';
        $limit = $options['limit'] ?? false;
        $offset = $options['offset'] ?? 0;
         $result = $this->sql()
            ->getWhere($this->table, $where, $orderFields, $orderDirection, $limit, $offset)
            ->resultArray();
         $schema = Schema::parse([':a' => $this->readSchema]);
        $result = $schema->validate($result);
         return $result;
    }
```
That is definitely SQL implementation of model. There is no way to transform it to use NoSQL or XML or TXT as source of data. IMHO that is not what model should have. 

I think we already have all these functions : get(), update(), etc... already somewhere - what is the reason to have a new set of those? What is wrong with old ones?

initvector:
> That is definitely SQL implementation of model. There is no way to transform it to use NoSQL or XML or TXT as source of data. IMHO that is not what model should have.

That is what injecting the database object into the model is for. The abstraction necessary for the type of flexibility you're describing currently isn't in Vanilla, but when it is, it can be utilized here. If the DBA supported it, and we had adapters for the NoSQL, XML, etc., you could pass two difference database class instances to models and have them perform without issue (e.g. MySQL for the website, XML for tests).

> what is the reason to have a new set of those? What is wrong with old ones?

This is the beginning of moving away from `Gdn_Model` to a new, slimmed down base model. It is an iterative process. The first step is introducing a very basic model for very basic database operations. This step also uses Garden Schema for input and output validation. This step avoids tacking on the capabilities of `Gdn_Pluggable`. There will be additional steps. One such step will be the introduction of a middleware-style pattern, which will allow extending and modifying these operations and their results.

library/Vanilla/Models/Model.php
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * @param array $set Field values to set.
+     * @param array $where Conditions to restrict the update.
+     * @throws \Exception If an error is encountered while performing the query.
+     * @return bool True if no errors were encountered when performing the query.
+     */
+    public function update(array $set, array $where): bool {
+        $set = $this->writeSchema->validate($set, true);
+        $result = $this->sql()->put($this->table, $set, $where);
+        if (!($result instanceof \Gdn_DataSet)) {
```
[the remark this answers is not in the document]
initvector: Vanilla's current base model, `Gdn_Model`, is a thousand-line class with lots of methods and functionality that most data models just don't need. It also extends `Gdn_Pluggable`, which brings along its own methods and dependencies. We are trying to slim down the base model to something that is still useful, but does not include the "code bloat" associated with the current model.

library/Vanilla/Utility/SchemaFilters.php
initvector: This was copied over from the `dbencode` function, but modified to have one return value instead of multiple. In general, I prefer that pattern for debuggging (can just break on the one return statement). In this case, the function is so small that I don't think adding another return statement would hinder its readability or debug-ability.

PR #7712, opened 2018-08-31
Fix Travis not considering all command results to determine a successful build

initvector:


charrondev:
Some manual breaking fo the build to test the fix https://travis-ci.com/vanilla/vanilla/jobs/143257720

PR #7715, opened 2018-09-02
Fix Gdn_SQLDriver::whereNotIn for empty sets

### LINKED ISSUE #2969: SQL->whereNotIn() broken for empty arrays
`Gdn_SqlDriver::whereNotIn('field', array())` produces the following SQL where clause:
`where field not in (null)`

Unfortunately, this condition always evaluates to false. This is contrary to what a developer might expect when writing e.g. `SQL->whereNotIn('d.DiscussionID', array());`. 

This is due to `Gdn_SqlDriver::_whereIn('field', array(), 'not in')` handling the special case of empty arrays with '(null)' which makes sense for the 'in' operator, but not the 'not in' operator.

See line 2045,Gdn_SqlDriver::_whereIn():

```
        if (count($In) > 0) {
            $InExpr = '('.implode(', ', $In).')';
        } else {
            $InExpr = '(null)';
        }
```

Solution: return without producing any SQL if the operator is 'not in' and count($In) = 0. Better handle this in a new method Gdn_SqlDriver::_whereNotIn().

bleistivt:
fixes #2969

### DIFF
--- a/library/database/class.sqldriver.php
+++ b/library/database/class.sqldriver.php
@@ -2139,6 +2139,10 @@ abstract class Gdn_SQLDriver {
         if (count($in) > 0) {
             $inExpr = '('.implode(', ', $in).')';
         } else {
+            // WHERE NOT IN (NULL) would yield an empty result set.
+            if ($op == 'not in') {
+                return $this;
+            }
             $inExpr = '(null)';
         }
 


charrondev:
I see travis hiccuped during the build here. I'm re-running it.

alex-mtl:
I will not approve or refuse this "improvement".

Personally I disagree with initial issue statement:

> This is contrary to what a developer might expect when writing e.g. SQL->whereNotIn('d.DiscussionID', array());

I would be very surprised if ever meet such fragment of code. This particular one should kind never exist. What I would expect to meet:

`SQL->whereNotIn('d.DiscussionID', $filterIds);`

But  that is another case. My opinion we need to check array before we call `whereNotIn` . 

That method should bring Exception when gets an empty array as argument.
But that is my personal opinion. 

However, I think that such "patch" when result output of the method changes data to opposite ("empty result" before, and "all records" after) is pretty dangerous. I would prefer to keep that as is. It is very difficult or even impossible to test how often empty array is passed to `whereNotIn`.

bleistivt:
There are only 2 uses of `whereNotIn` in core:
https://github.com/vanilla/vanilla/search?q=wherenotin

Both cases **explicitly handle empty arrays**, because of whereNotIn's failure to do so:

1. `getMessagesForLocation` makes sure MessageID 0 is excluded, because otherwise no messages would be shown for users who have never dismissed any:
https://github.com/vanilla/vanilla/blob/d82c73f3e3f6d6a59468b8704af3188cc993f484/applications/dashboard/models/class.messagemodel.php#L139-L142

2. `getJunctionPermissions` just adds the where statement if there actually are any exclusions (which is what this PR does), because otherwise every table name would be excluded.

https://github.com/vanilla/vanilla/blob/d82c73f3e3f6d6a59468b8704af3188cc993f484/applications/dashboard/models/class.permissionmodel.php#L583-L585

The current behaviour does not make any sense. WhereNotIn means "exclude this". If you exclude _nothing_, you should get _everything_.

Sure, this could break a private plugin I am not aware of, but I would argue that depending on buggy behaviour is a bug in itself, especially since a "get nothing" query would never yield anything in the first place.

alex-mtl:
@bleistivt 
I will try to do some research about `whereNotIn` usage.

But meanwhile your 2nd example:
```
            if (count($disabledWhere) > 0) {
                $sQL->whereNotIn('JunctionTable', $disabledWhere);
            }
```
is exactly what I suggested before. 

If we check array size before we call `whereNotIn` - that is not the case, since it should never call the method. And to me that makes perfect sense.

initvector:
> I would be very surprised if ever meet such fragment of code. This particular one should kind never exist. What I would expect to meet:

I don't think the original issue is meant to suggest someone would necessarily program a call to that method with a hardcoded empty array, but it might be possible for an empty array to be passed to that method through other means (e.g. an empty query result). For example, if you had a list of users and you wanted to filter out moderators using `whereNotIn`, but had none available, your "moderators" filter value would be empty and thus kill all results to your query. You might expect to get all available users, but you'd actually get none.

> But that is another case. My opinion we need to check array before we call whereNotIn .

A call to `whereNotIn`, like most of the methods in `Gdn_SQLDriver`, was designed to be fluent. Forcing developers to wrap a call to that method in a conditional isn't inline with that design.

> Sure, this could break a private plugin I am not aware of, but I would argue that depending on buggy behaviour is a bug in itself, especially since a "get nothing" query would never yield anything in the first place.

There are definitely additional calls to this method on internal projects. Some of them do not utilize the is-empty check, so they are relying on things to function as they currently do. Those calls need to be vetted before we can apply this update.

I believe we're primarily thinking of this in the context of select statements, but it's possible for this condition to be used in (potentially destructive) write operations. I don't think it would be prudent to alter a condition that currently affects no rows into one that affects "all rows", when it comes to things like deletes or updates. A more careful, measured approach is probably more advisable. Maybe we can begin with only triggering a deprecation notice when an empty array is used with this method.

charrondev:
I'm closing this out due to the disagreement over the implementation and how it has stalled.

I would recommend some discussion back on #2969 and far more detail in the pull request description before trying again. Backwards compatibility is key in APIs that have existed for as long as this one.

PR #7716, opened 2018-09-02
Use CategoryModel::checkPermission to validate user's permission to view discussions in categories.

cchabilall83:
The DiscussionModel::checkPermission calls userModel_getCategoryViewPermission_created hook  to validate if a user has permission to view a discussion in a certain category.  The permission check in this hook is done against the user.  The Ranks plugin sets the user permission on the session, so the the permissions in the session don't match those in the database.  Certain actions(warn, report) are denied as a result.  This was because 

1) only Administrator and Moderators were given privileges on that category.  
2) This permission check was done against the user's role permissions, not the rank permissions which are set in the session.

To avoid this CategoryModel::checkPermission is called if there's a valid session matching the userid  and we fall back to userModel::getCategoryViewPermission.

To test you need to the following

Based off master
1) Enable Ranks, Reporting and Warnings
2) Import customer's Ranks table
3) Create a category
4) Set custom privileges  where Admins and Moderators only haver permissions (see areanet's reported post category)
5) Create user with Role: Member - Rank:Administrator
6) Sign in as that user
7) Move in a discussion into the newly created category
8) Attempt to Flag the discussion (ie. Report\Warn)
- You should receive a permission error
9) Checkout fix branch
10) repeat step 8
- It should be successful

Closes vanilla/support#92

### DIFF
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -584,7 +584,6 @@ class VanillaHooks implements Gdn_IPlugin {
 
             $options = ['ForeignID' => $permissionCategoryID];
             $result = Gdn::userModel()->checkPermission($userID, $permission, $options);
-
             return $result;
         }
         return false;
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -583,7 +583,11 @@ class VanillaHooks implements Gdn_IPlugin {
             }
 
             $options = ['ForeignID' => $permissionCategoryID];
-            $result = Gdn::userModel()->checkPermission($userID, $permission, $options);
+
+            $result = (Gdn::session()->isValid())
+                ? Gdn::session()->checkPermission($permission)
+                : Gdn::userModel()->checkPermission($userID, $permission, $options);
+
             return $result;
         }
         return false;
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -583,10 +583,7 @@ class VanillaHooks implements Gdn_IPlugin {
             }
 
             $options = ['ForeignID' => $permissionCategoryID];
-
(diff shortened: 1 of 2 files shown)


linc:
Great find. I'd like to know how this interacts with the goals of https://github.com/vanilla/vanilla/pull/6003 before we merge it. We'll probably need @initvector's input for that since it descends into the nitty gritty of ongoing permissions management modernization.

applications/vanilla/settings/class.hooks.php
```diff
@@ -583,7 +583,11 @@ public function userModel_getCategoryViewPermission_create($sender) {
             }
 
             $options = ['ForeignID' => $permissionCategoryID];
-            $result = Gdn::userModel()->checkPermission($userID, $permission, $options);
+
+            $result = (Gdn::session()->isValid())
+                ? Gdn::session()->checkPermission($permission)
```
initvector: Some addons (e.g. Category Roles) use the "loadPermissions" event to augment a user's permissions. This event is fired in `UserModel::getPermissions`, which is called by `UserModel::checkPermission`. I don't believe that event would be fired from `Gdn_Session`. Could Ranks be updated to use the "loadPermissions" event, at least for category permissions?

cchabilall83: Using the "loadPermissions" event i'm able to merge the sessions permissions with the users permission in a similar way that Category Roles did.  Only possible issue I could see is if caching isn't enabled, we default back to the permissions set in the db.  Do you think this could be an issue?
```   
 } elseif (!Gdn::cache()->activeEnabled() && is_array($user->Permissions)) {
            // Only attempt to use the DB field value if permissions aren't being cached elsewhere.
            $permissions = new Vanilla\Permissions($user->Permissions);
```
https://github.com/vanilla/vanilla/blob/master/applications/dashboard/models/class.usermodel.php#L286

Issue #3474, opened 2016-01-20
Discussion titles should not parse to text


linc:
PostController:

```
            // Make sure that the title will not be invisible after rendering
            $Name = trim($this->Form->getFormValue('Name', ''));
            if ($Name != '' && Gdn_Format::text($Name) == '') {
                $this->Form->addError(t('You have entered an invalid discussion title'), 'Name');
            }
```

It then filters with the `text()` formatter before saving to the database. This is the opposite of just about everything else in Vanilla and is unexpected behavior. Why let me post something like `<WOW>!` and have the discussion name render as a single exclamation mark? It doesn't make a lick of sense.

Issue #2881, opened 2015-07-04
Enhance mentions in discussion titles?


linc:
What do we think of having mentions autocomplete in discussion titles? Should they link on the discussion view?

<img width="500" alt="screen shot 2015-07-04 at 11 32 33 am" src="https://cloud.githubusercontent.com/assets/117672/8508481/82a8f806-2240-11e5-86f4-38c52bf7d28b.png">

PR #7717, opened 2018-09-03
Escape discussion titles in views instead of model

bleistivt:
Discussion titles are escaped in the discussion model instead of the view.
https://github.com/vanilla/vanilla/blob/8a485b2e536e78d40c9e4cc9bb6a14444fa8cb2e/applications/vanilla/models/class.discussionmodel.php#L979

This is not quite as bad as saving filtered text to the DB but it is really annoying for several reasons:

- HTML tags in titles get stripped #3474
- It blocks #2881
- Titles loaded back into the editor for editing have broken ampersand signs, showing `&amp;` literally to the user 

**This needs to be tested for possible XSS exploits.** Plugins may expect discussion titles from the model to be escaped.

There are still various places where Gdn_Format::text() is used instead of htmlspecialchars which would be more appropriate. This PR just adds `htmlspecialchars` where raw titles would be displayed otherwise.

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -976,7 +976,6 @@ class DiscussionModel extends Gdn_Model {
         $archiveTimestamp = Gdn_Format::toTimestamp(Gdn::config('Vanilla.Archive.Date', 0));
 
         // Fix up output
-        $discussion->Name = Gdn_Format::text($discussion->Name);
         $discussion->Attributes = dbdecode($discussion->Attributes);
         $discussion->Url = discussionUrl($discussion);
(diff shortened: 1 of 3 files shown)


alex-mtl:
I like the idea and approach. But I have doubt it will pass since there are some vulnerabilities this modification opens. Since Vanilla does not use any templating library by default that means nobody can guarantee that `htmlspecialchars` or any other escaping function will be applied when output. We have dozen themes and plugins which can have their own outputs. We can't revise them all.

But I do vote for this issue to be resolved. We need to split formatting part from data layer.

bleistivt:
Closing in favour of a safer alternative #7724

Issue #3218, opened 2015-10-22
Not filtering UTF-8 characters disallowed in XML


linc:
We should clean up our output when generating XML to ensure the error `Char 0x0 out of allowed range` (or whatever character) is never generated by XML readers.

See: http://stackoverflow.com/questions/3466035/how-to-skip-invalid-characters-in-xml-file-using-php

charrondev:
Closed by https://github.com/vanilla/vanilla/pull/7727

PR #7727, opened 2018-09-06
Filter invalid characters from discussion titles in feeds

### LINKED ISSUE #4490: Discussions RSS breaks if discussion title contains '\x04'
How to reproduce: `UPDATE GDN_Discussion SET Name = CHAR(4,68,69,65,84,72,32,84,79,32,82,83,83,33) WHERE DiscussionID = ?;`

screen shot 2016-09-16 at 12 05 37 pm

Let's be wise and find out what else could reproduce the same issue.

bleistivt:
HTMLawed already filters all invalid characters.

Fixes #4490 and #3218

### DIFF
--- a/applications/vanilla/views/discussions/index_rss.php
+++ b/applications/vanilla/views/discussions/index_rss.php
@@ -6,7 +6,7 @@
 foreach ($this->DiscussionData->result() as $Discussion) {
     ?>
     <item>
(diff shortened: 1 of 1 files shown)


linc:
A two-for-one using our own built-in methods we forgot about! Damn! 💃

PR #7729, opened 2018-09-06
Add CSS class to comments with same author as the discussion

### LINKED ISSUE #5325: Add CSS class to comments with same author as the discussion
Having an optional designator for styling "OP" comments differently would be a good feature enhancement.

bleistivt:
Add CSS class to comments with same author as the discussion

fixes #5325

### DIFF
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -85,10 +85,16 @@ if (!function_exists('writeComment')) :
         $sender->EventArguments['Permalink'] = $permalink;
 
         // Needed in writeCommentOptions()
-        if ($sender->data('Discussion', null) === null) {
+        $discussion = $sender->data('Discussion', null);
+        
+        if ($discussion === null) {
             $discussionModel = new DiscussionModel();
             $discussion = $discussionModel->getID($comment->DiscussionID);
             $sender->setData('Discussion', $discussion);
+        }
+        
+        if (($discussion->InsertUserID ?? false) == $comment->InsertUserID) {
+            $cssClass .= ' original-poster';
         }
 
         // DEPRECATED ARGUMENTS (as of 2.1)


applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -85,11 +85,17 @@ function writeComment($comment, $sender, $session, $currentOffset) {
...
-        if ($sender->data('Discussion', null) === null) {
+        $discussion = $sender->data('Discussion', null);
+        
+        if ($discussion === null) {
             $discussionModel = new DiscussionModel();
             $discussion = $discussionModel->getID($comment->DiscussionID);
             $sender->setData('Discussion', $discussion);
         }
+        
+        if (($discussion->InsertUserID ?? false) == $comment->InsertUserID) {
```
charrondev: A few things on this line in particular. Are you sure that the discussion coming in is always an object and not an array? We might need a check here due to the inconsistency data is sometimes passed around as, especially since the the discussion itself is in a data array of any sender that could be passed in.

Also `===` instead of `==` please. Comparing numbers that could be 0 with `==` is a recipe for disaster.

applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -85,11 +85,17 @@ function writeComment($comment, $sender, $session, $currentOffset) {
...
+            $cssClass .= ' original-poster';
```
charrondev: Could we change this class name up to be `isOriginalPoster` or `isOP`? We are trying to move to a convention of `is{Something}` for our modifier CSS classes like this.

applications/vanilla/views/discussion/helper_functions.php
bleistivt: I guess that's a rare case `val`.

applications/vanilla/views/discussion/helper_functions.php
bleistivt: 👍  `isOriginalPoster` is less ambiguous

applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -85,11 +85,17 @@ function writeComment($comment, $sender, $session, $currentOffset) {
...
-        if ($sender->data('Discussion', null) === null) {
+        $discussion = $sender->data('Discussion', null);
+        
+        if ($discussion === null) {
             $discussionModel = new DiscussionModel();
             $discussion = $discussionModel->getID($comment->DiscussionID);
             $sender->setData('Discussion', $discussion);
         }
+        
+        if (val('InsertUserID', $discussion) === $comment->InsertUserID) {
```
charrondev: I know we haven't messaged it publicly right now, but at the moment we are trying to deprecate and remove many usages of the val function and put the onus on code to know what it's dealing with. Something like the following would align with that better:

```php
$insertUserID = is_array($discussion) ? $discussion['InsertUserID'] : $discussion->InsertUserID;
if ($insertUserID === $comment->InsertUserID) {

```

bleistivt: How about:

    $originalPosterID = $discussion->InsertUserID ?? $discussion['InsertUserID'] ?? false;

While there shouldn't ever be a comment without a discussion, this is just a render function that should work independently of outside data existing or not.

charrondev: This function is certainly poorly architected.

As for you suggested solution:

```php
$thing = new \stdClass();
$thing->test = "test";
$other = $thing['test'] ?? $things->test ?? false;
// Cannot use object of type stdClass as array

$thing = ['test' => 'test'];
$other = $things->test ?? $thing['test'] ?? false;
// Works??
```

It seems to work the second way but not the first, so I guess that is a bit cleaner than the `is_array` or `is_object` check. Please be sure to accompany it with a comment of why we need this check in the first place.

bleistivt: It does work, but I couldn't find anything in the PHP documentation about it.
Is is evaluated in this order `a ?? (b ?? c)`. So clearly there is the possibility of object access on an array, which works, but there is no guarantee this will keep working in the future.

So i settled on just using Controller::data() (which uses valr internally, but avoids array/object confusion here, once it has been refactored).

PR #7731, opened 2018-09-06
Change announcement dismiss animation to highlight

### LINKED ISSUE #5572: 'Dismiss' announcement animation hides the OP
Currently, 'Dismiss' triggers a rollup animation on the discussion being dismissed, even in the single discussion view, thus hiding it.

I don't think any change at all should be triggered. No animation, no refresh - just nice and steady as you go.

bleistivt:
fixes #5572 and is more in line with other actions that hightlight the item such as closing.

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -402,7 +402,7 @@ class DiscussionController extends VanillaController {
             redirectTo('discussions');
         }
 
-        $this->jsonTarget("#Discussion_$discussionID", null, 'SlideUp');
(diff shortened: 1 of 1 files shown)


charrondev:
A little bit of context for anyone else looking into this that is not familiar with how our old JSON target works:

- An ID and the name of a CSS animation can sent down in the response. Some global listener in `global.js` (eww) listens for this result and applies the animation to the first element with that ID. In this case it is the comment.
- The previous animation `SlideUp`, actually animates the first comment up and removes it. `Highlight` transitions a background colour on the element then returns it to its normal state.

cchabilall83:
Based on the acceptance criteria this looks good to me

linc:
I dig it.

Issue #7734, opened 2018-09-06
Implement static cache optimization 


alex-mtl:
Relates to #7689

As a part of continuous optimization initiative we need to implement static cache mechanizm on some classes to avoid unnecessary loops and calculations of the same "static data"

First candidates to apply this are our t() and c() global functions which are used very widely but seem in many cases they don't save any results which brings unnecessary calculations we can easily avoid.

PR #7736, opened 2018-09-07
Add static cache traits to the project

alex-mtl:
Implement static cache optimization

Like we discussed, I'll split optimization PR into pieces.
This is 1st part just to get Traits in

Relates to #7734 and to #7689

### DIFF
--- a/library/Garden/StaticCacheTrait.php
+++ b/library/Garden/StaticCacheTrait.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * @author Alexander Kim <alexander.k@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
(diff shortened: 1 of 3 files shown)


library/Garden/StaticCacheTrait.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Returns of cached value or calls self::f() function if key not found yet
+     *
+     * @param string $key Key to get
+     * @param mixed $default Default value for key to set if not set up
+     *
+     * @return mixed
+     */
+    public static function sc(string $key, $default = false) {
```
linc: The `c()` and `t()` functions in core are very special cases. Please don't establish a new precedent of abbreviation method names all over the place.

Issue #7735, opened 2018-09-07
Remove unnecessary fireEvent('BeforeTranslate')


alex-mtl:
Working on optimization epic I've found out this line

https://github.com/vanilla/vanilla/blob/336bdcac94498f057a1381e17286751cd6e54cda/library/core/class.locale.php#L360

That means on every t() or translate() call we do fire the event (even if there are no any listener)

I have a question: is there any good reason to fire this event that many times for each item/word/phrase we need to translate on the page.

After some additional research I've found just one potential listener plugin 'Localligator' plugin.

I suspect that this plugin is just dev tool and even if we need it time to time it should not be active in 99% of uptime on production.

My main question/suggestion for this issue:
can we just comment/remove  this fireEvent line in repo?
when anybody need to use Localligator - he can manually uncomment this line and collect info, but not ever push uncommented line back to repository?

linc:
Yes, let's get that event turned off by default.

I suggest only firing the event when in debug mode (there's a config for that), or modify Localigator to enable a generic-named config setting (`Garden.Locales.CollectEvents` or something) and then the event only fires if that config setting is true.

I don't like a "uncomment code to make a tool work" scenario very much because it's secret knowledge that's hard to communicate. That said, if the config solution I describe is too expensive, I'd be OK with it in this 1 instance as long as you update the Localigator README to clearly explain that, and document the line in core so we don't accidentally delete it later.

DaazKu:
No matter the decision, never forget that anyone could have used an event that we remove.
We have no control over what is used and what is not. Scanning our own plugins is not enough. Good example: https://open.vanillaforums.com/addons

Deprecating the event and putting that in our releases note should be the first step.
If we decide to be more agressive about it the config route is a good idea.

linc:
> No matter the decision, never forget that anyone could have used an event that we remove.

In the broad case, I agree. They should get deprecated before removal.

In this specific case, I think it's safe to be more aggressive. We added it for a very narrow use case, I've never heard of it being used elsewhere, it hasn't been around very long, and the benefit to removal is very big.

Also, wrapping it with a config-based conditional would let folks sidestep that concern fairly easily.

R-J:
I have more than 100 (even ancient) plugins installed and uncountable snippets where I've tested even absurd things. The only place where I can find that event being used is in an old version of vanillas own Addons application. In the current version of the Addons applications class.hooks.php this method is no longer included.

alex-mtl:
[...]

PR #7737, opened 2018-09-07
Remove unnecessary event fire 'beforeTranslate'

alex-mtl:
Fixes: https://github.com/vanilla/vanilla/issues/7735

Add condition to fire event `beforeTranslate`.

## Additionally
Fixes some issue with original StaticCacheConfigTrait and StaticCacheTranslationTrait (was not calling correct method and because of that was not caching the response)

### DIFF
--- a/library/core/class.locale.php
+++ b/library/core/class.locale.php
@@ -357,7 +357,7 @@ class Gdn_Locale extends Gdn_Pluggable {
 
         $this->EventArguments['Code'] = $code;
         $this->EventArguments['Default'] = $default;
-        if (Gdn::config('EnabledPlugins.localligator', false)) {
+        if (Gdn::config('Debug', false)) {
             $this->fireEvent('BeforeTranslate');
         }
 
--- a/library/core/class.locale.php
+++ b/library/core/class.locale.php
@@ -355,9 +355,9 @@ class Gdn_Locale extends Gdn_Pluggable {
             }
         }
 
-        $this->EventArguments['Code'] = $code;
-        $this->EventArguments['Default'] = $default;
         if (Gdn::config('Debug', false)) {
+            $this->EventArguments['Code'] = $code;
+            $this->EventArguments['Default'] = $default;
             $this->fireEvent('BeforeTranslate');
         }
 
--- a/library/core/class.locale.php
+++ b/library/core/class.locale.php
@@ -15,6 +15,7 @@ use Vanilla\AddonManager;
  * for different locales. It is a singleton class.
  */
 class Gdn_Locale extends Gdn_Pluggable {
+    use \Garden\StaticCacheConfigTrait;
 
     /**  @var string The name of the currently loaded Locale. */
     public $Locale = '';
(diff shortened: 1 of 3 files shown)


alex-mtl:
I see your point. But imho there is no good solution. 
Your approach to enable when "debug" mode - kind resolves production issue. 
But why all developers need to have 'beforeTranslate' enabled when debug mode but they don't use localligator?
As I said at the very beginning, in general this event should not exist at all.
None of any plugin should ever be bind to this event. That is just a tool. like phpunit or profiler or xdebug tool. We should not have any code for those tools in codebase.
It is kind environment variables that should handle that.
For sure I can set another config value like 'BeforeTranslate.Enabled' - to have no cross references, but that is worse approach to me, since next developer need to investigate that again - why and when anybody can have it enabled or disabled. Same time `if (Gdn::config('EnabledPlugins.localligator', false))` is very explicit itself.

R-J:
I think references from core to plugins in whatever aspect and for whatever purpose are ugly and should be avoided. Eliminating the event without first deprecated it might be okay in such a case, but if I were the Admin of a forum that uses a plugin which relies on that event

library/core/class.locale.php
```diff
@@ -357,7 +357,9 @@ public function translate($code, $default = false) {
 
         $this->EventArguments['Code'] = $code;
```
charrondev: We should move event argument setting inside of the conditional as well

library/core/class.locale.php
```diff
@@ -357,7 +357,9 @@ public function translate($code, $default = false) {
...
         $this->EventArguments['Default'] = $default;
-        $this->fireEvent('BeforeTranslate');
+        if (Gdn::config('Debug', false)) {
```
charrondev: Since this is so commonly called, my feeling is that this continuous lookup of this config value would be particularly expensive.

This should probably be use the static config cache trait you've written to prevent that.

I don't have a particular opinion about the key here, but Debug seems better than a brand new config key.

library/core/class.locale.php
alex-mtl: Done.

charrondev:
Let's get https://github.com/vanilla/vanilla/pull/7736 merged in first so that we don't reduce performance with a config lookup here.

library/core/class.locale.php
alex-mtl: I've tested - there is performance improvement when config() is checked. 
But in general I agree that config() call kind same "weight" of complexity.

linc:
> For sure I can set another config value like 'BeforeTranslate.Enabled' - to have no cross references, but that is worse approach to me, since next developer need to investigate that again - why and when anybody can have it enabled or disabled. Same time `if (Gdn::config('EnabledPlugins.localligator', false))` is very explicit itself.

The underlying assumption was that you'd then toggle that new config value automatically in Localigator when it's enabled, not leave it to be discovered by a developer. The point was to de-couple the reference so each config has 1 unique function, not make it a manual task to turn it on.

linc:
I'm fine with coupling it to debug mode, but Localigator needs a note to say that the site must be in debug mode for it to work, now.

alex-mtl:
@linc https://github.com/vanilla/internal/pull/1726

linc:
> if I were the Admin of a forum that uses a plugin which relies on that event, I would be thankful if there is a way to be able to continue running my forum without paying a developer. Therefore I would support using a dedicated config setting

I agree, generally. In this particular case, because it was only ever added to support a developer tool AND because it's significantly detrimental to use it in production, I'm acquiescing to it being tied to the `Debug` config setting. Pretty much any other scenario dictates a separate config.

Issue #7691, opened 2018-08-29
Update frontend build tools to support continuous building


charrondev:
Pending the finalization of https://github.com/vanilla/vanillainfrastructure/issues/149

This issue is to outline how we could going to adjust the build files.

## Strict requirements

- Build all files into a single top level `build/dist` or `dist` or `assets` directory.
- Build is triggered in the same composer hook that clears the vanilla cache.
- All built files are removed from git.
- Remove all knowledge references in core.
  - AssetModel -> going away when move place assets are built and re-write of them.
  - Sections -> Get rid of all of the section and heavily leverage webpack dynamic chunk to ensure per-page bundles are small. Or we can do more filesystem lookups on build script startup to determine what we need to build.
  - tsconfig -> look into extending a parent config with typescript 3.
- Update documentation on this repo to indicate the change in requirements for building from git.
- Update phing build

## Other notes

- Coordinate with @DaazKu to test this on infrastructure in a deployment.

Issue #7708, opened 2018-08-31
Pull framework/library/editor js/css of out of dashboard application


charrondev:
## Dashboard

A few pieces in particular here are "global"/framework related and not really part of the dashboard specifically. In my mind these are:

### JS
- Top level JS utilities live `application`, `dom`, `utility`, `embeds`, `apiv2`, `permissions`.
- `bootstrap` file.
- some parts of the `components` folder.
- Everything in `user-content`

### Styles
- Variable sheets (primary one and editor)
- embed styles
- User related styles from authenticate page
- Editor styles that are not forum specific. These should be moved to the rich-editor plugin.

slafleche:
Is this part of this sprint? I won't have time this sprint.

charrondev:
Closed by #7751

PR #7738, opened 2018-09-07
Implement continuous building of frontend dependencies

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7691

_Do __NOT__ merge this pull request yet. I requires new dependencies on our infrastructure https://github.com/vanilla/vanillainfrastructure/issues/159. I'm coordinating with @DaazKu to get this running and using our test infra cluster to verify it works._

This prod and dev builds should still run locally without issue.

## Required testing

- [x] Localhost
- [x]  Inf test cluster
- [x] Staff test cluster
- [x] Some inf site that is served out of a subdirectory such as hub/node.

## Things that were not touched

There are a few things I specifically steered clear of here because we've had issues in the past:

- The way files are cached.
- The way those caches are busted.

### Added a `frontend-build.sh` script

This gets run in a few places:
  - The travis build script.
  - A composer post-install hook. This is how infra and phing will be triggering the builds.

This script runs a yarn installation and the build. It returns a non-zero value if either of those steps do.

### Updated travis script

- Travis now runs the build script.
- Travis now lets the build run the linter instead of running it directly.

### Build changes

- The build now __cleans up__ files from previous production builds before starting a new one.
- Happypack has been removed. It was over-aggressively spawning new process which actually was slowing down the build and choking CI servers. I've noticed build times decreasing with it removed even on my localhost.
- All references to Kb were removed.
- This was accomplished by replacing addonUtils with an `EntryModel` class that does some additional lookups to find what sections are available.
  - Adding a file of a particular name in the `src/scripts/entries` creates a new build "section".
  - `.ts` and `.tsx` entries are treated as being the same entry name.
  - `bootstrap` is a special case and is excluded. I have plans to move this up with https://github.com/vanilla/vanilla/issues/7708.
- A lot the logic of `build.ts` has been pulled into a separate class `Builder.ts`.
- The production build now returns a non-zero error code if errors are thrown during the build.

### New Build file Locations

- Build files now live in a top level `dist` directory, with each section getting its own subdirectory. 
- This directory is git ignored.
- Old built files in `/js/webpack` directories have been removed
- Entry points from addons now live in the `/dist/{sectionName}/addons` and are named according to the addon key that provides them.
- Dynamically imported entry-points have been given a little bit more leeway to allow webpack to optimize and split them up.
  - As a result of these changes the webpackChunkName provided no longer needs the addon directory. Instead of `webpackChunkName: "/plugins/rich-editor/js/webpack/chunks/mountEditor"` we can do `webpackChunkName: "mountEditor"`
- I've updated the `AssetModel` to load these files from the new location.

Example:

**/dist**
- `/dist/forum/runtime.min.js`
- `/dist/forum/vendors.min.js`
- `/dist/forum/shared.min.js`
- `/dist/forum/bootstrap.min.js`
- `/dist/forum/addons/dashboard.min.js`
- `/dist/forum/addons/some-plugin.min.js`
- `/dist/forum/addons/rich-editor.min.js`


library/setup/ComposerHelper.php
```diff
@@ -33,6 +38,61 @@ private static function clearAddonManagerCache() {
...
+
+        // Build bootstrap can be used to configure this build if env variables are not available.
+        $buildBootstrap = realpath($vanillaRoot . "/conf/build-bootstrap.php");
+        if (file_exists($buildBootstrap)) {
+            include $buildBootstrap;
+        }
+
+        $nodeArgs = getenv(self::NODE_ARGS_ENV) ?: "";
+        $disableValidationFlag = getenv(self::DISABLE_VALIDATION_ENV) ? "--low-memory" : "";
+        $buildCommand = "TS_NODE_PROJECT=$tsConfig node $nodeArgs -r $tsNodeRegister $buildScript $disableValidationFlag";
```
initvector: I realize the chance of this happening are slim, but does this command accommodate arguments/paths with spaces in them? If not, could it?

charrondev: We're shelling out to execute this command, so you can provide a path with spaces, but you would need to escape the whitespace the way you would for any shell.

I haven't tested that scenario particularly though.

charrondev:
I've gotten word from @DaazKu that all of our infrastructure is now updated to be compatible with this PR.

PR #7739, opened 2018-09-08
Optimize tweak

tburry:


charrondev:
Already done in another PR https://github.com/vanilla/vanilla/blob/master/library/core/functions.general.php#L778

PR #7756, opened 2018-09-11
Roles: userRole table not getting updated after role is deleted, leaving orphaned records

### LINKED ISSUE #7758: Roles: Deleting roles and updating UserRole table
### Issue

When deleting a role and selecting a replacement role, the role is removed from GDN_Role but not GDN_UserRole.
[class.rolemodel.php#L750](https://github.com/vanilla/vanilla/blob/fix/delete-orphaned-roles/applications/dashboard/models/class.rolemodel.php#L750) is not evaluated in that case.

### API Call

This was noticed during an api call to users [api/v2/users ](https://docs.vanillaforums.com/help/apiv2/swagger/#/Users/get_users).
screen shot 2018-09-11 at 3 08 09 pm

### To Replicate
screen shot 2018-09-11 at 3 06 14 pm

dannimajdalani:
### Details
closes #7758 
When a role is deleted with a replacement role selected , table GDN_UserRole is getting orphaned role records.

### To test this

- Delete a role 
- Select a replacement role
- Check GDN_UserRole table (role is not deleted)

### DIFF
--- a/applications/dashboard/models/class.rolemodel.php
+++ b/applications/dashboard/models/class.rolemodel.php
@@ -746,15 +746,14 @@ class RoleModel extends Gdn_Model {
                 ->set('UserRole.RoleID', $newRoleID)
                 ->where(['UserRole.RoleID' => $roleID])
                 ->put();
-        } else {
-            $this->SQL->delete('UserRole', ['RoleID' => $roleID]);
         }
-
+        
         // Remove permissions for this role.
         $permissionModel = Gdn::permissionModel();
         $permissionModel->delete($roleID);
 
         // Remove the role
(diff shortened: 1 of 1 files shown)


topcheese13:
I'm ok with having no issue linked to the PR, but, like I said the description becomes even more important to explain the context and problem.

>When a role is deleted with a replacement role selected , table GDN_UserRole is getting orphaned role records.

This doesn't explain the problem to me correctly. I feel like the description should be more extensive/explicit.

dannimajdalani:
I created an issue and referenced the task.

cchabilall83:
This looks good to me.  Maybe @initvector might be able see any possible repercussions with the is change.

PR #7757, opened 2018-09-11
Decouple editor and library styles from the dashboard

charrondev:
Closes https://github.com/vanilla/vanilla#7708

- [x] @slafleche and I need to go through this with a fine tooth comb, but I think I've left it in a pretty good place.

## Library SCSS
The following has been moved to `library/src/scss` and are accessible for import through `~library-scss/*`.
- Core variables
- Core mixins
- Shared user content styles
- Shared embed styles

## Editor SCSS

The styles for the frame of the editor were decoupled from the content styles by @slafleche. These now live in the `rich-editor` plugin and need are generate there own stylesheet.

## Style compat

**This still contains compatibility styles and mixins for:**
  - secret groups
  - group search
  - category following
  - rich editor
  - authenticate pages (we might do something different with this in the future)

After talking with @slafleche we've determined that since this is our "old" system it is better to have it be aware of `rich-editor` than have `rich-editor` be aware of it, so the styles have imports into the `rich-editor` plugin.

## Additionally

A few more mixins have been broken out of giant mixin files and broken into mixin partials.


library/src/scss/mixins/_mixin.breakpoints.scss
```diff
@@ -0,0 +1,28 @@
+/*!
```
slafleche: These are specific to the rich editor, they should be with the rich editor styles. Also prefix the name of the file with richEditor

plugins/rich-editor/src/scss/mixins/_mixin.mobileStyles.scss
```diff
@@ -0,0 +1,44 @@
+/*!
```
slafleche: These are legacy styles

PR #7771, opened 2018-09-13
Improve javascript test setup

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7772
Blocking https://github.com/vanilla/knowledge/pull/131

- Uses the `ForkTsCheckerPlugin` for typechecking, similar to our main build. This resolves the primary type-checking issue from https://github.com/vanilla/vanilla/issues/7772.
- Drops the extra babel step in loader chain for typescript files. Our current typescript setup does not require this step.
- Adds the `TsConfigWebpackPlugin` to allow us to use our path aliases from our `tsconfig.json` file instead of redefining those aliases.
- Fixes a typo in the tsconfig file making it invalid JSON to some parsers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
         "karma-mocha-reporter": "2.2.5",
         "karma-sourcemap-loader": "0.3.7",
         "karma-spec-reporter": "0.0.32",
-        "karma-webpack": "4.0.0-rc.2",
+        "karma-webpack": "3.0.4",
         "mini-css-extract-plugin": "0.4.2",
         "mocha": "5.2.0",
         "node-sass": "4.9.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -79,7 +79,7 @@
         "karma-mocha-reporter": "2.2.5",
         "karma-sourcemap-loader": "0.3.7",
         "karma-spec-reporter": "0.0.32",
-        "karma-webpack": "3.0.0",
+        "karma-webpack": "4.0.0-rc.2",
```
alex-mtl: rc -  means release candidate?

charrondev: This was incorrect. I've rolled back this version change. It can be revisited once the final is released.

PR #7776, opened 2018-09-13
Deprecation class implemented to log when deprecated function is called.

alex-mtl:
Deprecation class implemented to log when deprecated function is called.

This implementation should allow to track if deprecated function is in use on production

First candidate to deprecate: translateContent()

### DIFF
--- a/library/Vanilla/Utility/Deprecation.php
+++ b/library/Vanilla/Utility/Deprecation.php
@@ -5,7 +5,7 @@
  * @license GPL v2
  */
 
-namespace Garden;
+namespace Vanilla\Utility;
 
 /**
  * Class Deprecation.
--- a/library/Vanilla/Utility/Deprecation.php
+++ b/library/Vanilla/Utility/Deprecation.php
@@ -40,6 +40,7 @@ class Deprecation {
         if (!key_exists($info['function'], self::$calls)) {
             $fileName = str_replace(PATH_ROOT, '', $info['file']);
             $message = 'Deprecated function '.$info['function'].' called from '.$fileName.' at line : '.$info['line'];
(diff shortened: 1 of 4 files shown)


library/Vanilla/Utility/Deprecation.php
```diff
@@ -0,0 +1,47 @@
...
+     * This function outputs 1 message per function per request
+     *
+     * @return void
+     */
+    public static function log() {
+        $info = debug_backtrace()[1];
+        if (!key_exists($info['function'], self::$calls)) {
+            $fileName = str_replace(PATH_ROOT, '', $info['file']);
+            $message = 'Deprecated function '.$info['function'].' called from '.$fileName.' at line : '.$info['line'];
+            error_log($message, E_USER_ERROR);
```
tburry: If this also triggered a deprecate error then developers would see it too. I'm not sure if the debug bar can grab stuff from error_log.

tburry: You could also use `Vanilla\Logger::error(...)`.

alex-mtl: trigger_error added

library/core/functions.general.php
```diff
@@ -3571,6 +3571,7 @@ function t($code, $default = false) {
      * @see Gdn::translate()
      */
     function translateContent($code, $default = false) {
```
initvector: If we're going to deprecate a function, we should probably slap a [<span>@</span>deprecated tag](http://docs.phpdoc.org/references/phpdoc/tags/deprecated.html) in its doc block.

alex-mtl: Done. Check here:
https://github.com/vanilla/vanilla/pull/7784/files

PR #7784, opened 2018-09-14
Add config toggle to Vanilla\Utility\Deprecation::log

alex-mtl:
Add check for config 'Vanilla.Deprecation.Enabled' if false - to not output any deprecation error

This will allow to disable log flooding if ever deprecated function detected on particular client instance.

### DIFF
--- a/library/Vanilla/Utility/Deprecation.php
+++ b/library/Vanilla/Utility/Deprecation.php
@@ -40,7 +40,7 @@ class Deprecation {
      */
     public static function log() {
         // Set Vanilla.Deprecation.Enabled to false to disable log output
-        if (self::c('Vanilla.Deprecation.Enabled', true)) {
+        if (self::c('Garden.Log.Deprecation', true)) {
             $info = debug_backtrace()[1];
             if (!key_exists($info['function'], self::$calls)) {
                 $fileName = str_replace(PATH_ROOT, '', $info['file']);
--- a/library/Vanilla/Utility/Deprecation.php
+++ b/library/Vanilla/Utility/Deprecation.php
@@ -7,10 +7,13 @@
 
 namespace Vanilla\Utility;
 
+use Garden\StaticCacheConfigTrait;
+
 /**
  * Class Deprecation.
  */
 class Deprecation {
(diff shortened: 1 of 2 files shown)


library/Vanilla/Utility/Deprecation.php
```diff
@@ -36,13 +39,16 @@ class Deprecation {
...
     public static function log() {
-        $info = debug_backtrace()[1];
-        if (!key_exists($info['function'], self::$calls)) {
-            $fileName = str_replace(PATH_ROOT, '', $info['file']);
-            $message = 'Deprecated function '.$info['function'].' called from '.$fileName.' at line : '.$info['line'];
-            trigger_error($message, E_USER_DEPRECATED);
-            error_log($message, E_USER_ERROR);
-            self::$calls[$info['function']] = true;
+        // Set Vanilla.Deprecation.Enabled to false to disable log output
+        if (self::c('Vanilla.Deprecation.Enabled', true)) {
```
initvector: Stuff under the Vanilla config namespace is *usually* reserved for functionality specific to the Vanilla application. Since this is probably going to be used throughout the project, what about renaming the config key to something like Garden.Log.Deprecation?

alex-mtl: Since @tburry asked me to move this class under Vanilla namespace. I don't feel guilty to put config value for this class under Vanilla.* as well.

initvector: I think you're going to find that most components under the `Vanilla` code namespace don't use the "Vanilla" config namespace. The naming convention typically dictates that if we're configuring an addon, it gets the top-level key. In this case, you're using Vanilla as the top-level key, which would indicate this is a configuration for the Vanilla application. However, that is not the case.

alex-mtl: Yes. I already discovered that naming convention is not obvious and transparent. Finally I don't care. I'll put your variant `Garden.Log.Deprecation`.

alex-mtl: Done.

Issue #7774, opened 2018-09-13
Implement AliasAutoloader


charrondev:
## Why not just declare `class_alias`?

`class_alias`  autoloads the aliased class when the alias is declared. This means any class we have any alias for would immediately autoload a lot of classes that may not actually be in use. An autoloader for these would remove that inefficiency.

## Requirements

- Maintains a static array of old class names to their new class name.
- Provides an autoload function that will create an alias from the old class name to the new one if the old class name is used.
- Will autoload the new class.
- Has unit tests for different behaviour and some interactions with the container.

An initial implementation was made in https://github.com/vanilla/vanilla/pull/7471 That should be pulled out.

PR #7786, opened 2018-09-14
Add AliasProviderTrait and AliasLoader

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7774

Provides an autoloader for class aliases.

## Usage
The final process of adding an alias will look like:

1. Add an alias (or aliases) in `AliasLoader::provideAliases()`. See example in `TestAliasLoader`
2. In a class that you are adding an alias for say `NewClass`, add a line of `AliasLoader::createAliases(NewClass::class)` after your class declaration.

## Implementation
My initial strategy of only needing ___just___ the autoloader failed when checking class inheritance for some the of the permutations in the tests here so I needed to add step 2.
 
I was reluctant to add step 2,  but to ensure things like `is_subclass_of` work properly we need to ensure aliases are loaded from the initial class because they do not trigger an autoload. If you have any other suggestions that would not require this step I'm open to them, right now the process of loading an alias can happen to ways.

1. New class name is used -> load class -> class uses `AliasLoader::createAliases()` to create it's aliases.
2. Old class name is used -> `AliasLoader::autoload()` -> `class_exists()` check triggers step 1.

To ensure the aliases only need to be defined in one place I've added the `AliasLoader::createAliases()` method.

## Tests

On the testing side in order to ensure no classes are autoloaded I needed to set `$runTestInSeparateProcess`. This caused issue with container pollution when the globals were carried over to the new process so I needed to set `preserveGlobalState` to false as well. I put both of these properties in a new `IsolatedTestCase`. @tburry had the idea of using this in the future for our integration tests so we could drop the need to route out to nginx for a fresh PHP process.


tests/Library/Vanilla/AliasProviderTraitTest.php
```diff
@@ -0,0 +1,138 @@
...
+        spl_autoload_register([TestAliasLoader::class, 'autoload']);
+        self::assertFreshAutoload(NewClass::class);
+        self::assertFreshAutoload(ExtendsNewClass::class);
+    }
+
+    /**
+     * Verify that loading the old class loads the new class through the autoloader.
+     */
+    public function testSimpleAliasAutoload() {
+        new \VanillaTests\OldClass();
```
initvector: Would it be worth capturing this instance in a variable and asserting it's actually an instance of `NewClass`, instead of only making sure the class was available after this object was instantiated?

library/Vanilla/AliasLoader.php
```diff
@@ -0,0 +1,28 @@
...
+ * which will then autoload their new classes if they are not loaded yet.
+ */
+class AliasLoader {
+
+    use AliasProviderTrait;
+
+    /**
+     * @inheritdoc
+     */
+    protected static function provideAliases(): array {
```
initvector: It looks like this is intentionally hardcoded. Is the idea that only core Vanilla will be able to add class aliases and that those aliases will require manually updating the value returned from `AliasLoader::provideAliases`?

tests/Library/Vanilla/AliasProviderTraitTest.php
charrondev: The fact that is actually an instance of the class is the one line of description the class_alias call has.

http://docs.php.net/manual/da/function.class-alias.php
> Creates an alias named alias based on the defined class original. The aliased class is exactly the same as the original class.

This test function is specifically to test the autoloader. The inheritance checks below kind of cover then case your describing, or the trickier edge case at least.

library/Vanilla/AliasLoader.php
charrondev: That is correct.

This is the `Vanilla\AliasLoader` which is aliases for Vanilla. If another one of our projects, or someone else wants to provide aliases in a similar way the can use the `AliasProviderTrait` and register their own autoloader.

That is what the `TestAliasLoader` does as well.

initvector: I am *slightly* concerned that other projects might not be able to register their autoloader early enough, but I don't think there's a high enough probability of that occurring to re-engineer anything here.

tests/Library/Vanilla/AliasProviderTraitTest.php
initvector: It pains me to see an object instantiated, but not referenced. That's my own personal hangup, though. You are correct in that `assertClassExtendsClass` isn't exactly doing what I'm asking, but it's probably close enough to be practical.

Issue #7788, opened 2018-09-14
Optimize core Gdn_Cache to avoid unnecessary calculations


alex-mtl:
We need to optimize Cache class to be extremely fast and slim.

At the moment almost on every call to get or store any value we call makeKey() method
which is pretty complex and that method calls getPrefix() method which again is not something flat and simple and calls Gdn::config()

I suggest to use static cache for this class just to store prefix at least. 
I've already made research and performance test - that small change bring few percents performance increase.

After discussions with @charrondev  and @tburry: need to create unit tests to cover getPrefix method

Relates to: https://github.com/vanilla/vanilla/pull/7689

PR #7789, opened 2018-09-14
Optimization for Gdn_Cache class to store statically prefix value.

alex-mtl:
Optimization for Gdn_Cache class to store statically prefix value.

Relates to : #7689 and #7734
Fixes: #7788

### DIFF
--- a/library/core/class.cache.php
+++ b/library/core/class.cache.php
@@ -130,6 +130,12 @@ abstract class Gdn_Cache {
 
     /** @var int  */
     public static $trackTime = 0;
+    /**
+     * All possible prefixes to cache as static array and avoid recalculations
+     *
+     * @var array
+     */
+    protected static $cConfig = [];
 
     /**
      *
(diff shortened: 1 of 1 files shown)


library/core/class.cache.php
```diff
@@ -564,9 +570,14 @@ protected function fallback($key, $options) {
 
     public function getPrefix($forcePrefix = null, $withRevision = true) {
         static $configPrefix = false;
+        $cKey = ($forcePrefix ?? 'null:').($withRevision ? 'true' : 'false');
+        if (key_exists($cKey, self::$cConfig)) {
```
charrondev: Shouldn't this be `array_key_exists()`? Looks like this an alias from PHP <5 that's still around.

alex-mtl: Done.

initvector:
I'm putting a hold tag on this PR until the internal discussion has been resolved. Please report back here with whatever is decided.

kaecyra:
In my opinion we should not even think about merging a change to something this deep in Vanilla without first implementing unit tests for how the methods USED to behave. Otherwise we're no better than animals and just rolling the dice on a regression. With the potential for really screwing things up on deploy.

charrondev:
Closing until those unit tests are present.

Issue #6845, opened 2018-03-23
Fetch the current user in API v2


chrisvanpatten:
In APIv1, you can easily get information about the current user via `/api/v1/users/get.json` without any parameters.

As far as I can tell from Swagger and some brief experimentation, no similar capability exists in APIv2.

Two possible endpoints could be `/api/v2/users/me` or `/api/v2/users/current`. I've also seen APIs where ID `0` returns the current user.

If you're going through the auth flow over the API you'll have the current user's ID, but for cookie-based auth (we're doing some custom JS widgets that get inserted into a hosted platform theme) or access token auth, there doesn't appear to be an APIv2-native way of getting that info.

PR #7794, opened 2018-09-17
Add ability to get fragment for the current user to users API v2 endpoint

### LINKED ISSUE #7733: Add ability to get current user fragment from the users API endpoint
The users API should allow a query that returns a fragment of information about the currently-signed-in user. The response should have a clear indicator of whether or not the user is actually signed-in (i.e. not a guest).

initvector:
This update adds the ability to get information about the current user from the users API v2 endpoint. Requesting /api/v2/users/me should respond with values reflecting the current session. The existing "user fragment" schema is used for the output schema.

### Example "member" response
```json
{
    "userID": 1,
    "name": "Vanilla",
    "photoUrl": "//we.vanillicon.com/v2/b58996c504c5638798eb6b511e6f49af.svg",
    "dateLastActive": "2018-09-01T01:00:00+00:00"
}
```

### Example "guest" response
```json
{
    "userID": 0,
    "name": "Guest",
    "photoUrl": "http://example.com/applications/dashboard/design/images/defaulticon.png",
    "dateLastActive": null
}
```

Closes #7733

### DIFF
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -24,6 +24,9 @@ class UsersApiController extends AbstractApiController {
     /** @var Gdn_Configuration */
     private $configuration;
 
+    /** @var array */
+    private $guestFragment;
+
     /** @var Schema */
     private $idParamSchema;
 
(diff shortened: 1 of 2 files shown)


chrisvanpatten:
Also closes #6845.

Thanks! ❤️

PR #7801, opened 2018-09-17
Update garden-password and Smarty versions in composer.lock file

cchabilall83:
PR to update vanilla's composer.lock file because vanilla/garden-password was recently updated. 

**Note**. When I ran composer update, smart was updated as well.  Not sure if we should update that as well.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -389,16 +389,16 @@
         },
         {
             "name": "smarty/smarty",
-            "version": "v3.1.32",
+            "version": "v3.1.33",
             "source": {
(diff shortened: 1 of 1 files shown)


initvector:
> When I ran composer update, smart was updated as well. Not sure if we should update that as well.

The update here moves from Smarty v3.1.32 to v3.1.33. This should only be a bug fix release and, according to [Smarty's change log](https://github.com/smarty-php/smarty/blob/master/change_log.txt), that's what it is.

Issue #7797, opened 2018-09-17
FeatureFlagHelper should throw its own exception


tburry:
This is a topic for some discussion.

Currently, `FeatureFlagHelper::throwIfNotEnabled()` takes an exception class string and arguments to throw a custom feature. This is something that we should be getting away from. Instead we should be throwing structured exceptions, possibly inherited from our general `ServerException` or `ClientException` classes.

What I recommend is something like:

```php
public static function ensureFeature(string $feature, string $message = '', $code = 403) {
    // Check for feature.
   throw new FeatureNotEnabledException($message, $code);
}
```

Something like that.

Another observation I made is that while defining feature constants on a class is nice, it does require us to load that class just to check for the feature. This may end up having us include a bunch of classes when building menus in the dashboard for example.

initvector:
Forcing a standard exception is a good idea. The original intent was to give a lot of freedom to the classes calling the method, but in hindsight it does that to a fault. Allowing callers to dictate what exception is thrown makes it a lot harder to catch and handle in a standard way. Using a standard exception should remedy that.

I do want to keep the feature flags/configs as constants, but I'm not insistent they necessarily remain in any particular class. Addons using constants that aren't class-specific, particularly these feature flags, could be moved into a minimal (abstract) class, where the values could be referenced with bare minimum loading of additional class code (and no state). Keeping the values as constants allows for IDE hints/autocompletion, avoids errors in entering the actual flag/config string and makes it easier to update the flag/config value without having to update it across several areas of the application.

charrondev:
Closed by #7804

PR #7804, opened 2018-09-18
Add explicit exception for accessing disabled features

initvector:
It was proposed in #7797 that Vanilla throw a specific, explicit exception for whenever access to disabled feature is detected with the `FeatureFlagHelper` class. This update implements that exception.

### Summary of Changes
1. Add `FeatureNotEnabledException` class.
1. Add `FeatureFlagHelper::ensureFeature` method. This method does the same thing `FeatureFlagHelper::throwIfNotEnabled` did, only it forces a standard exception.
1. Deprecated `FeatureFlagHelper::throwIfNotEnabled` in favor of `FeatureFlagHelper::ensureFeature`.
1. Remove call to global `c` function in favor of `Gdn::config` in `FeatureFlagHelper::featureEnabled`.
1. Update all calls to `FeatureFlagHelper::throwIfNotEnabled`, including the one added in #7798.

### DIFF
--- a/applications/dashboard/controllers/class.importcontroller.php
+++ b/applications/dashboard/controllers/class.importcontroller.php
@@ -96,13 +96,9 @@ class ImportController extends DashboardController {
      */
     private function checkAccess() {
         $this->permission('Garden.Import'); // This permission doesn't exist, so only users with Admin == '1' will succeed.
-        \Vanilla\FeatureFlagHelper::throwIfNotEnabled(
+        \Vanilla\FeatureFlagHelper::ensureFeature(
             'Import',
-            Gdn_UserException::class,
-            [
-                t('Imports are not enabled.', 'Imports are not enabled. Set the config Feature.Import.Enabled = true to enable imports.'),
-                403
-            ]
+            t('Imports are not enabled.', 'Imports are not enabled. Set the config Feature.Import.Enabled = true to enable imports.')
         );
     }
 
--- a/library/Vanilla/FeatureFlagHelper.php
+++ b/library/Vanilla/FeatureFlagHelper.php
@@ -2,13 +2,15 @@
 
 namespace Vanilla;
 
-use Gdn;
 use Vanilla\Exception\FeatureNotEnabledException;
+use Garden\StaticCacheConfigTrait;
 
 /**
  * A helper class for gatekeeping code behind feature flags.
  */
 class FeatureFlagHelper {
+
+    use StaticCacheConfigTrait;
 
     /**
      * Is a feature enabled?
@@ -18,7 +20,7 @@ class FeatureFlagHelper {
      */
     public static function featureEnabled(string $feature): bool {
         // We're going to enforce the root "Feature" namespace.
-        $configValue = Gdn::config("Feature.{$feature}.Enabled");
+        $configValue = self::c("Feature.{$feature}.Enabled");
         // Force a true boolean.
         $result = filter_var($configValue, FILTER_VALIDATE_BOOLEAN);
         return $result;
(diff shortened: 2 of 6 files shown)


library/Vanilla/FeatureFlagHelper.php
```diff
@@ -19,22 +18,40 @@ class FeatureFlagHelper {
...
      * If the feature is not enabled, throw the specified exception.
      *
      * @param string $feature The config-friendly name of the feature.
      * @param string $exceptionClass The fully-qualified class name of the exception to throw.
      * @param array $exceptionArguments Any parameters to be passed to the exception class's constructor.
+     * @deprecated 2.7 Use FeatureFlagHelper::ensureFeature instead.
      */
-    public static function throwIfNotEnabled(string $feature, string $exceptionClass = Exception::class, array $exceptionArguments = []) {
+    public static function throwIfNotEnabled(string $feature, string $exceptionClass = \Exception::class, array $exceptionArguments = []) {
         if (self::featureEnabled($feature) === false) {
```
charrondev: Can we get a `deprecated` call in here?

library/Vanilla/FeatureFlagHelper.php
```diff
@@ -19,22 +18,40 @@ class FeatureFlagHelper {
      */
     public static function featureEnabled(string $feature): bool {
         // We're going to enforce the root "Feature" namespace.
-        $configValue = c("Feature.{$feature}.Enabled");
+        $configValue = Gdn::config("Feature.{$feature}.Enabled");
```
charrondev: Especially if we're going to be using these feature flags more often, we should update this config check to use `StaticCacheConfigTrait`.

applications/dashboard/controllers/class.importcontroller.php
```diff
@@ -96,13 +96,9 @@ public function go($transientKey = '') {
...
         $this->permission('Garden.Import'); // This permission doesn't exist, so only users with Admin == '1' will succeed.
-        \Vanilla\FeatureFlagHelper::throwIfNotEnabled(
+        \Vanilla\FeatureFlagHelper::ensureFeature(
             'Import',
-            Gdn_UserException::class,
-            [
-                t('Imports are not enabled.', 'Imports are not enabled. Set the config Feature.Import.Enabled = true to enable imports.'),
-                403
-            ]
+            t('Imports are not enabled.', 'Imports are not enabled. Set the config Feature.Import.Enabled = true to enable imports.')
```
charrondev: This should use the `StaticCacheTranslationTrait` so feature checks are more efficient.

initvector: I'm not sure how that might apply in the context of this line.

library/Vanilla/FeatureFlagHelper.php
initvector: I don't think feature flags are necessarily going to get a lot of mileage out of static config caching, since they probably won't be checked multiple times in the same request (or at least enough to see a real return here). However, I don't think it'll hurt anything either, so it's been implemented.

applications/dashboard/controllers/class.importcontroller.php
charrondev: There is a is a translation call here. It will forget the result of its translation check and lookup again if it's just using the `t()` function. Instead if you use `StaticCacheTranslationTrait` and `self::t()` instead of `t` the translation result will be cached. This way adding a feature check into a loop somewhere will not regress our performance.

initvector:
That's two approvals. I'm assuming everything is good, so merge incoming...

PR #7818, opened 2018-09-19
Simplify Gdn_Configuration::get()

tburry:
- We no longer need specific unserialize() sugar because the entire config is properly saved.
- The Gdn_ConfigurationSource class has been moved to its own file.
- The arrayMergeRecursiceDistinct() function has been deprecated and moved to the deprecated file.


tburry:
I didn't make a new class. I just moved it to its own file because it was always a confusing edge case. I'd rather any renaming there not be the purview of this PR, especially since this particular code is used so differently on infrastructure.

initvector:
Looks good to me. Tested a bit. Need another sign-off to merge, though.

PR #7831, opened 2018-09-19
Update composer (2.7)

dannimajdalani:
Related #7801


charrondev:
I'll note that twig is getting added to the lock file for this time on 2.7 with this.

@alex-mtl I should have caught that here: https://github.com/vanilla/vanilla/pull/7674/files but we should always be updating the lock file when a new dependency is added.

PR #7832, opened 2018-09-19
Update composer (2.6)

dannimajdalani:
Related #7801

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -629,20 +629,23 @@
         },
         {
             "name": "vanilla/garden-password",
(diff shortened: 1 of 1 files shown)


charrondev:
The travis failure here seems to be because of the backwards incompatible changes in `garden-schema` 1.9 release. If I recall it required updating unit tests.

You should roll back your changes and update just the dependency you are touching here then.

Eg. `composer update "vanilla/garden-password"` instead of `composer update`. This is probably a better strategy for backport patches.

PR #7840, opened 2018-09-20
Enforce add-on types and keys

tburry:
Enforce what add-on types and keys can be passed to the `AddonManager` so that appropriate error messages are surfaced to the user.

### DIFF
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -515,7 +515,7 @@ class AddonManager {
      * @return Addon|null Returns an addon object or null if one isn't found.
      */
     private function lookupSingleCachedAddon($addonDirName, $type) {
-        if (empty($addonDirName)) {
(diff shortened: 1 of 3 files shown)


library/Vanilla/AddonManager.php
```diff
@@ -1328,4 +1336,28 @@ public function add(Addon $addon, $start = true) {
...
+        }
+    }
+
+    /**
+     * Validate an addon key.
+     *
+     * @param string $key The key to validate.
+     */
+    private static function validateKey(string $key) {
+        if (!preg_match('`^[a-z0-9_-]*$`i', $key)) {
```
initvector: This would match on an empty addon key. Is that intentional?

initvector: It is intentional! The example that was cited:

> When no theme is selected its an empty addon.

linc:
Backported to `release/2.6`.

PR #7843, opened 2018-09-21
Fix memory exhaustion while moving a discussion

### LINKED ISSUE #7105: ModerationController::confirmDiscussionMoves is exhausting allowed memory size
### Details

Moving a discussion from one category to another is exhausting the allowed memory size and is causing a "500 Internal server error".
php-error {"level":"Fatal","error":"Allowed memory size of *** bytes exhausted..

**Error when moving a discussion**
tt22

alex-mtl:
Closes #7105 

Avoid cache recalculations for CategoryModel during ajax response - just update DB and reset cache. 
Cache should be auto recovered/recalculated by any next call coming

### Note:

Main memory consumption was on setCache() of CategoryModel which has a call to buildCache(). Technically caching layer made request 2-3x times slower with abnormal memory usage during execution.

Before (request takes fro 1,5-3 s and php allocates memory from 2Mb to >100Mb during execution):
screen shot 2018-09-24 at 4 50 39 pm

After (request takes less than 0.5s memory allocation does not change from one request to another):
screen shot 2018-09-24 at 4 51 02 pm

As you can see there is a huge performance increase with new "no cache update" solution. 

Fixes: #7105

### DIFF
--- a/applications/vanilla/controllers/class.moderationcontroller.php
+++ b/applications/vanilla/controllers/class.moderationcontroller.php
@@ -446,9 +446,8 @@ class ModerationController extends VanillaController {
 
             // Update recent posts and counts on all affected categories.
             foreach ($AffectedCategories as $categoryID => $counts) {
-                $DiscussionModel->updateDiscussionCount($categoryID);
+                CategoryModel::recalculateCounts($categoryID);
             }
-            (new CategoryModel())->counts('CountAllDiscussions');
             CategoryModel::clearCache();
             // Clear selections.
             if ($ClearSelection) {
--- a/applications/vanilla/controllers/class.moderationcontroller.php
+++ b/applications/vanilla/controllers/class.moderationcontroller.php
@@ -345,7 +345,6 @@ class ModerationController extends VanillaController {
         $Session = Gdn::session();
         $this->Form = new Gdn_Form();
         $DiscussionModel = new DiscussionModel();
-        $CategoryModel = new CategoryModel();
 
         $this->title(t('Confirm'));
 
@@ -394,8 +393,7 @@ class ModerationController extends VanillaController {
 
             // Iterate and move.
             foreach ($AllowedDiscussions as $DiscussionID) {
-                $Discussion = val($DiscussionID, $DiscussionData);
-
+                $Discussion = ($DiscussionData[$DiscussionID] ?? false);
                 // Create the shadow redirect.
                 if ($RedirectLink) {
                     $DiscussionModel->defineSchema();
@@ -448,26 +446,10 @@ class ModerationController extends VanillaController {
 
             // Update recent posts and counts on all affected categories.
             foreach ($AffectedCategories as $categoryID => $counts) {
-                $CategoryModel->refreshAggregateRecentPost($categoryID, true);
-
-                // Prepare to adjust post counts for this category and its ancestors.
-                list($discussionOffset, $commentOffset) = $counts;
-
-                // Offset the discussion count for this category and its parents.
-                if ($discussionOffset < 0) {
-                    CategoryModel::decrementAggregateCount($categoryID, CategoryModel::AGGREGATE_DISCUSSION, $discussionOffset);
-                } else {
-                    CategoryModel::incrementAggregateCount($categoryID, CategoryModel::AGGREGATE_DISCUSSION, $discussionOffset);
-                }
-
-                // Offset the comment count for this category and its parents.
-                if ($commentOffset < 0) {
-                    CategoryModel::decrementAggregateCount($categoryID, CategoryModel::AGGREGATE_COMMENT, $commentOffset);
-                } else {
-                    CategoryModel::incrementAggregateCount($categoryID, CategoryModel::AGGREGATE_COMMENT, $commentOffset);
-                }
+                $DiscussionModel->updateDiscussionCount($categoryID);
             }
-
+            CategoryModel::recalculateAggregateCounts();
+            CategoryModel::clearCache();
             // Clear selections.
             if ($ClearSelection) {
                 Gdn::userModel()->saveAttribute($Session->UserID, 'CheckedDiscussions', false);
--- a/applications/vanilla/controllers/class.moderationcontroller.php
+++ b/applications/vanilla/controllers/class.moderationcontroller.php
@@ -446,7 +446,8 @@ class ModerationController extends VanillaController {
 
             // Update recent posts and counts on all affected categories.
             foreach ($AffectedCategories as $categoryID => $counts) {
-                CategoryModel::recalculateCounts($categoryID);
+                CategoryModel::recalculateRecentFields($categoryID);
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/class.moderationcontroller.php
```diff
@@ -448,26 +446,9 @@ public function confirmDiscussionMoves($DiscussionID = null) {
 
             // Update recent posts and counts on all affected categories.
             foreach ($AffectedCategories as $categoryID => $counts) {
-                $CategoryModel->refreshAggregateRecentPost($categoryID, true);
```
initvector: Don't we still need to update the `LastDiscussionID` fields for the categories in the path we're moving to, as well as the ones we're moving away from?

applications/vanilla/models/class.categorymodel.php
```diff
@@ -3601,6 +3601,103 @@ private static function recalculateAggregateCounts() {
...
+     * @return void
+     */
+    public static function recalculateCategoryCounts(int $categoryId) {
+        $sql = Gdn::sql()
+            ->select('DiscussionID', 'count', 'DiscussionCount')
+            ->select('CountComments', 'sum', 'CountComments')
+            ->from('Discussion')
+            ->where('CategoryID', $categoryId);
+
+        $exclude = Gdn::config('Vanilla.Archive.Exclude');
```
initvector: Isn't this chunk essentially a duplicate of `DiscussionModel::addArchiveWhere`?

applications/vanilla/models/class.categorymodel.php
```diff
@@ -3601,6 +3601,103 @@ private static function recalculateAggregateCounts() {
         self::instance()->clearCache();
     }
 
+    /**
+     * Recalculate  CountDiscussions and CountComments columns for category.
+     *
+     * @param int $categoryId Category ID to update
+     * @return void
+     */
+    public static function recalculateCategoryCounts(int $categoryId) {
```
initvector: Since we're already in `CategoryModel`, we could probably make this a a more descriptive name without making it any longer. How about `recalculatePostCounts`?

applications/vanilla/models/class.categorymodel.php
```diff
@@ -3601,6 +3601,103 @@ private static function recalculateAggregateCounts() {
...
+            ->put();
+    }
+
+    /**
+     * Recalculate  CountAllDiscussions count column for category.
+     *
+     * @param int $categoryId Category ID to update
+     * @return void
+     */
+    public static function recalculateCategoryCountsAll(int $categoryId) {
```
initvector: Similar to the other recalculation function, could we rename this to something with a more descriptive name, like `recalculateAggregatePostCounts`? Anytime I see "recalculate" and "all" in a method name, I think it's going to recalculate *all* of something. I know the column names are "CountAllDiscussions" and "CountAllComments", but I think including "all" in this method name could be a little misleading.

applications/vanilla/models/class.categorymodel.php
```diff
@@ -3601,6 +3601,103 @@ private static function recalculateAggregateCounts() {
...
+
+            $parent = Gdn::sql()
+                ->select('ParentCategoryID')
+                ->from('Category')
+                ->where('CategoryID', $categoryId)
+                ->get()
+                ->firstRow(DATASET_TYPE_ARRAY);
+            $categoryId = $parent['ParentCategoryID'] ?? 0;
+        }
+    }
```
initvector: I think you might be able to condense most of this method down into a simpler do-while. Something like:

```php
self::recalculateCategoryCounts($categoryId);
do {
    self::recalculateCategoryCountsAll($categoryId);
    $category = Gdn::sql()
       ->select("ParentCategoryID")
       ->from("Category")
       ->where("CategoryID", $categoryId)
       ->get()
       ->firstRow(DATASET_TYPE_ARRAY);
    $categoryId = $category["ParentCategoryID"] ?? null;
} while ($categoryId && $categoryId > 0);
```

applications/vanilla/controllers/class.moderationcontroller.php
alex-mtl: From what I see in old code - that was not there.

applications/vanilla/models/class.categorymodel.php
alex-mtl: to me `Post` is very confusing in this context. But I do agree that method names are not very descriptive.

applications/vanilla/models/class.categorymodel.php
alex-mtl: :-)

applications/vanilla/models/class.categorymodel.php
linc: the elusive do-while appears in the wild

alex-mtl: Ok. On it.

alex-mtl: Done.

applications/vanilla/controllers/class.moderationcontroller.php
alex-mtl: Good point. I need to recheck this stuff.

initvector:
Looks like the build is broken, but it might just be coding standards.

applications/vanilla/models/class.categorymodel.php
initvector: For the sake of minimizing the scope of this PR, I'm okay with this being dupe for now.

initvector:
Since this update is introducing updates that could potentially adversely affect the category structure, we'll need some tests to ensure those risks are minimized. Since this test introduces a lot of new methods, and only really alters existing code in the "Move Discussion" action of the moderation controller, I think the tests should at least start off with:

1. Create a dozen or so categories. Split them up into two separate branches, each with half the categories. Each category having one child until you get out to the leaf node. You should have two branches, each about six levels deep.
1. Create a discussion at the deepest node of one branch.
1. Add a comment to the discussion you just created.
1. Verify the counts for all categories in the discussion's branch have been properly updated.
1. Using a request to the moderation controller, move the discussion from the tip of the original branch, to the tip of the other branch.
1. Verify the discussion's category was properly updated.
1. Verify the counts for all of the categories in the origin and destination branches were properly updated.
1. Repeat the above steps on new data. This time, create and move the discussion between the middle of each branch, not the tip. Verify counts for the full tree, including the unaffected categories, are correct. Categories lower in each branch should not be altered.

Ideally, this test would be written to work with the current revision of master, then applied to this PR with the intention of ensuring identical functionality.

linc:
> would it be better to add a way to opt out of caching?

I think that'd be a good idea to pursue.

initvector:
Looks like we're rolling with #7899 as the primary memory-exhaustion solution for now, and there haven't been any real updates to this branch in a few weeks, so I'm closing 'er out.

alex-mtl:
#7105 was fixed by #7899

There were open few other issues related to this:  #7883, #7885, #7897

PR #7846, opened 2018-09-21
Fix embedding comments on local builds

nervoustwit:
We have not supported non-pretty URLs for a few years now (`/index.php?p=discussions` instead of `/discussions`). There remains one URL in the embedding Javascript that still routs using the `p` variable that breaks local builds and impedes local testing/development.

This PR will remove that line from embed.js.

This has been tested on infrastructure, and has no side-effects.

For testing: https://github.com/vanilla/stub-embed-providers/pull/3

### DIFF
--- a/js/embed.js
+++ b/js/embed.js
@@ -153,9 +153,8 @@ window.vanilla.embed = function(host) {
                 if (currentPath != message[1]) {
                     currentPath = message[1];
                     // Strip off the values that this script added
-                    currentPath = currentPath.replace('/index.php?p=', ''); // 1
-                    currentPath = stripParam(currentPath, 'remote='); // 2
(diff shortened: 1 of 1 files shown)


charrondev:
I noticed there's actually another reference to `index.php?p` in that file. Could you nuke that one too?

nervoustwit:
D’oh!

nervoustwit:
I have looked at 
`currentPath = currentPath.replace('/index.php?p=', ''); // 1`
Since this only removes the `/index.php?=` maybe, just in case, we should leave it in?

charrondev:
The comment above it indicates its supposed to be some cleanup for the part we're removing now. If we don't dump it now it will likely persist for eternity. I'd say remove it. In my local testing it doesn't seem to affect anything and in any event we don't support that URL scheme.

PR #7858, opened 2018-09-25
Refactored PopovrerController and added generic dropdown component

slafleche:
This PR is to refactor some components we used in the rich editor and make them more generic and reusable. 

Also did a minor fix to the loader for accessibility. I change the page title on load and put a hidden H1 in the page too.

Added generic dropdown component


library/src/scripts/utility.ts
```diff
@@ -325,3 +325,28 @@ export function sanitizeUrl(url: string) {
...
 }
+
+/**
+ * Register an keyboard listener for the escape key.
+ *
+ * @param root - The element to watch for the escape listener in.
+ * @param returnElement - The element to return to when escape is pressed.
+ * @param callback
+ */
+export function addEscapeListener(
```
charrondev: This is thoroughly a DOM based utility method and should go into the `dom.ts` file as a result. `utility.ts` is for pure utility methods with no external dependencies outside of the javascript language.

library/src/scripts/components/Sentence.tsx
```diff
@@ -0,0 +1,103 @@
...
+export interface ISentence {
+    className?: string;
+    contents: IWord[] | string;
+    counter?: number;
+}
+
+/**
+ * Combines multiple inline elements together. Usually for translated text with links, datetimes and text.
+ *
+ * @param contents
```
charrondev: The extra line after a docblock prevents the IDE from picking up it's documenation.

image

Also what's up with the param tag here?

library/src/scripts/components/dropdown/items/DropDownContents.tsx
```diff
@@ -0,0 +1,45 @@
...
+    id: string;
+    parentID: string;
+    className?: string;
+    children: React.ReactNode;
+    isVisible?: boolean;
+    isPositionedFromTop: boolean;
+    isPositionedFromRight: boolean;
+}
+
+export default class DropDownContents extends React.Component<IProps> {
```
charrondev: A doc block is missing for this component

library/src/scripts/components/dropdown/items/DropDownItemRadio.tsx
```diff
@@ -0,0 +1,91 @@
...
+    private hasOptions: boolean;
+
+    public constructor(props) {
+        super(props);
+        this.hasOptions = props.options && props.options.length > 0;
+        let selectedIndex = 0;
+
+        if (this.hasOptions && !props.selectedOption) {
+            props.selectedOption = props.options[0].value;
+            props.selectedOption.some((option, index) => {
```
charrondev: This could be a forEach loop because we don't do anything with the result.

Issue #7564, opened 2018-08-06
Fatal Error when trying to view Ban Rules filtered by users


LauraMakesArt:
![image](https://user-images.githubusercontent.com/7501633/43740855-1d8af768-9992-11e8-9068-750a10061cbe.png)
Trying to view the users hit by a ban rule is causing a fatal error:
Fatal Error in PHP.gdn_ErrorHandler();
Unknown column 'inet6_ntoa(LastIPAddress)' in 'where clause'
![image](https://user-images.githubusercontent.com/7501633/43740912-4bbb1c08-9992-11e8-9bd4-114aba5a8a3e.png)
 
can be replicated here: https://laura.vanillademo.com/dashboard/settings/bans

PR #7872, opened 2018-09-27
Fix ban rules fatal error on search by not escaping field in SQL query

topcheese13:
Closes https://github.com/vanilla/vanilla/issues/7564

Simply following the call stack from the issue above, when creating one of the where clauses, escaping the field results in something like this:
```
`inet6_ntoa(LastIPAddress)` like :inet6_ntoaLastIPAddress
```
We are including the MySQL native function in the column name when escaping the field.

We are doing that in 2 places that are resulting in a fatal error when searching for users through ban rules on a IP Address rule.

By setting the `$escapeFieldSql` parameter of the where method of the SQL object we will prevent adding the back ticks as illustrated above, users cannot alter that field and therefore changing this to not be escaped should not open us to any SQL exploit.

This is a result of this https://github.com/vanilla/vanilla-patches/pull/249. We already fixed a few of these occurrences in the past and as per Todd we are not interested in a lower SQL fix for the moment.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2597,7 +2597,7 @@ class UserModel extends Gdn_Model {
         $this->fireEvent('AfterUserQuery');
 
         if (isset($where)) {
-            $this->SQL->where($where);
+            $this->SQL->where($where, null, false);
         }
 
         if (!empty($roleID)) {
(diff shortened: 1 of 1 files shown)


topcheese13:
Alex and I talked in person recapping the changes and he then approved the PR. Simply to answer his previous comment.

PR #7875, opened 2018-09-27
Fix addon scanning when an addon has a bad key

### LINKED ISSUE #7873: Improve handling for bad addons when looking up all of a type
Since #7840, Vanilla has been more strict with what keys an addon can have. However, the error messaging has some issues.

[`AddonManager::lookupAllByType` builds an array of addons, indexed by the array key.](https://github.com/vanilla/vanilla/blob/6f07f44ef639f4cbea5e5d62cc37faf725c7ef39/library/Vanilla/AddonManager.php#L297) The `AddonManager::lookupSingleCachedAddon` method is used to perform the individual addon lookup. However, this method [can return a `null` if the addon key is invalid](https://github.com/vanilla/vanilla/blob/6f07f44ef639f4cbea5e5d62cc37faf725c7ef39/library/Vanilla/AddonManager.php#L518). `AddonManager::lookupAllByType` does not consider this condition and, in the event of an addon with a bad key, can attempt to invoke the `getKey` method on a `null`. This would trigger a PHP fatal error.

`AddonManager::lookupAllByType` should be able to gracefully handle encountering addons with bad keys.

tburry:
Add-ons with invalid keys should not be included in scans, but should also not cause an exception.

Closes #7873.

### DIFF
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -346,7 +346,9 @@ class AddonManager {
             try {
                 $addon = new Addon($subdir);
                 $key = $addon->getKey();
-                if (!array_key_exists($key, $addons)) {
+                if (!static::validateKey($key)) {
+                    trigger_error("The $type in $subdir has an invalid key: $key.", E_USER_WARNING);
+                } elseif (!array_key_exists($key, $addons)) {
                     $addons[$key] = $addon;
                 } else {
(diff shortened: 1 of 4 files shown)


tburry:
I goofed a bit with the reproduce errors commit because I forgot to add the new fixtures. I reverted the fix to make the errors show up before doing the fix.

linc:
All hail King Frank! That guy loves cobb salad.

PR #7882, opened 2018-10-01
Add event-firing trait

### LINKED ISSUE #7826: Add event-firing trait
The new base model does not extend `Gdn_Pluggable`, so it does not have event-firing capabilities. A new trait could be added to Vanilla to allow models to opt into these abilities. The trait should be configurable and require the shared instance of `Garden\EventManager` to fire its events.

initvector:
This update adds a trait for models to opt into some event-firing abilities. It is intended to replace `Gdn_Pluggable` for newer models. Only basic support for event firing has been added. `Gdn_Pluggable` had the ability to "fire as", as well as automatically firing events when PHP's magic [`__call` method](http://php.net/manual/en/language.oop5.overloading.php#language.oop5.overloading.methods) was invoked on an instance. These are both functionalities newer models should probably avoid, so they are not included here. In its current state, the trait is largely just an alias for `EventManager::fire`. However, as more model-centric event capabilities are needed, they can easily be added.

Closes #7826

### DIFF
--- a/library/Garden/EventManager/FireEventTrait.php
+++ b/library/Garden/EventManager/FireEventTrait.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace Garden\EventManager;
+
+use RuntimeException;
+use Garden\EventManager;
+
+/**
+ * Trait for adding the ability to fire events with EventManager.
+ */
+trait FireEventTrait {
+
+    /** @var EventManager */
+    private $eventManager;
+
+    /**
+     * Fire an event.
+     *
+     * @param string $event The name of the event.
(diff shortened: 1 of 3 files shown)


initvector:
> In mind this feels like an abstraction for with no use at the moment.

I agree. As stated in the PR description, in its current form, this is basically an alias to an existing method at the moment. The idea was originally to do away with `Gdn_Pluggable` for future models, but still allow models to opt into those types of event behavior. However, as you've already pointed out, current development practices are already moving away from most of the facilities provided by `Gdn_Pluggable`. What we're left with is a pretty anemic trait. It doesn't do much more than call a method on an injected dependency that could already be available to the model. I believe there's the potential to expand on its abilities in the future for model-oriented event capabilities, but we don't have any concrete plans for any such abilities at the moment.

I still think we'll see some need for this trait in the future, but until there are real plans for fleshing out its capabilities, I'm agreeable with waiting on the implementation.

PR #7888, opened 2018-10-02
Improve modal, dropdown, and editor components

charrondev:
There a few changes rolled into this PR.

## Modals blocking body scrolling

Our modal components which were recently added here have been improved in the following ways.

- Modals no longer naively prevent body scrolling. They now bring in a well maintained npm library for blocking body scrolling [body-scroll-lock](https://www.npmjs.com/package/body-scroll-lock). Actually cross-browser implementation of this is not minimal. A quick rundown of issues with more naive approaches:

- the approach `document.body.ontouchmove = (e) => { e.preventDefault; return false; };` locks the body scroll, but ALSO locks the scroll of a target element (eg. modal).
- the approach `overflow: hidden` on the body or html elements doesn't work for all browsers
- the `position: fixed`  approach (_What I was using before_) causes the body scroll to reset. this is noticeable if the modal is not fullscreen.
- some approaches break inertia/momentum/rubber-band scrolling on iOS

## Other modal improvements

- The `<ModalRouter />` has been extracted out of our KB router and given a more generic interface. It allows for 2 different routes to be loaded on the screen by creating links with the `<ModalLink />`
- A barrel file has been added for the modal components.
- The modal link now uses a `<NavLink />` instead of a `<Link />` for greater flexibility.
- the `<BackLink />` component actually does a back navigation action when possible. This preserves scroll state of the previous page.

## Dropdown

- `<DropdownItemLink />` has been updated with an option to open in a modal. This will have it use a `<ModalLink />` instead of a `<NavLink />`.
- A barrel file has been added for the dropdown components.

## Editor changes

- `<Editor />` can now accept a callback that runs on text change events.
- `<Editor />` now has a public method for retrieving the text content (as opposed to the delta content). This is useful for checking the length of editor content.
- `Editor.prototype.setEditorContent()` now clears the history module. This prevents issues where a user can undo the initial content of the editor being set. That is the only usage of this method. I've also begun using this method when initializing content from the old legacy text area syncing.


charrondev:
I've circled back and fix the issues you've described.

- Modals now exit only 1 at a time.
- The picker has loading indicators.
- The state editor has loading indicators, through a new prop `isLoading`.

Issue #7869, opened 2018-09-27
Create unit tests to cover confirmDiscussionMoves 


alex-mtl:
Relates to: #7843

Need to create tests to cover confirmDiscussionMoves  of ModerationController 
(applications/vanilla/controllers/class.moderationcontroller.php)

https://github.com/vanilla/vanilla/pull/7843#issuecomment-424797134

Tests should cover multiple scenarios (use cases):

1. Source and Destination category are brand new Lvl1 (root).
Discussion is new and has no comments.
Initial Category set to Source
Discussion moved to Destination category.
Note: Lvl2 skipped to keep it empty but to check aggregated fields to be updated correctly
Source and Destination category are brand new Lvl3-4-5-6-7.
Discussion is new and has no comments.
Discussion moved to Destination category same level.

2. Source categories are existing Lvl1-3-4-5 destination categories from previous step.
Destination category is brand new Lvl3.
Discussions are existing created on previous step.
4 Discussions from 4 different Source categories are moved to the same Destination category.

3. Source category is existing Lvl3 category created on previous step 2.
Destination category is existing Lvl2 category created on step 1.
Discussions are existing moved on previous step.
2 Discussions from the same Source category are moved to the same Destination category.

4. Source category is existing Lvl7 category created on step 1.
Destination category is existing Lvl3 category created on step 1.
Discussion is existing most recent created on step 1.
4.1 Source category is existing Lvl6 category created on step 1.
Destination category is existing Lvl3 category created on step 2.
Discussion is existing created on step 1.
4.2 Source category is existing Lvl3 category just used as Destination in 4.1.
Destination category is existing Lvl2 category created on step 1 and used as Destination in step 2.
Discussion is the existing one just moved in 4.1.

alex-mtl:
Closed by: #7898

Issue #7897, opened 2018-10-03
LastCommentID of Category table not properly updated on DB when discussion moves


alex-mtl:
## Issue:

Columns LastCommentID of Category table for parents of src category is not properly updated when move discussion from one category to another using confirmDiscussionMoves of ModerationController.
```
/moderation/confirmdiscussionmoves?discussionid=...
```
Relates to: #7869 and #7885

## Steps to reproduce the bug:

1. Create 2 root categories with 1 child Lvl2 + 1 child Lvl3 category in each
<img width="662" alt="screen shot 2018-10-03 at 9 22 43 am" src="https://user-images.githubusercontent.com/15682507/46424828-29d08f80-c708-11e8-80ef-83f0acec9dda.png">

2. Create new discussion in any child category. `Category` table got updated like:
<img width="662" alt="screen shot 2018-10-03 at 9 23 22 am" src="https://user-images.githubusercontent.com/15682507/46424862-48cf2180-c708-11e8-8116-5bdc38fdb8c6.png">

3. Create new comment for just created discussion.  Category table should get `LastCommentID` and `LastDateInserted` changes:
<img width="664" alt="screen shot 2018-10-03 at 9 23 53 am" src="https://user-images.githubusercontent.com/15682507/46424952-85028200-c708-11e8-8be1-81b8db8448cc.png">

4. Move the discussion just created to another child category. 
<img width="415" alt="screen shot 2018-10-03 at 9 25 07 am" src="https://user-images.githubusercontent.com/15682507/46425070-d579df80-c708-11e8-868f-9ae2ad1eba47.png">

Check database changes in `Category` table:
<img width="667" alt="screen shot 2018-10-03 at 9 25 30 am" src="https://user-images.githubusercontent.com/15682507/46425125-f17d8100-c708-11e8-9a7e-633d65383bf3.png">

As you can see `LastCommentID` and `LastDateInserted` data for parent categories of previous discussion category stayed the same as if discussion still belongs to old category, which is wrong.

 <img width="642" alt="screen shot 2018-10-03 at 9 26 36 am" src="https://user-images.githubusercontent.com/15682507/46425923-00653300-c70b-11e8-8a2b-7a399d8e1914.png">

Another bug related to this one: if we delete discussion completely initial category parents will still hold outdated data
[...]

PR #7898, opened 2018-10-03
Add tests for moving posts with the moderation controller

alex-mtl:
### covers `confirmdiscussionmoves` action for several cases:

1. Source and Destination category are brand new **Lvl1** (root).
Discussion is **new and has no comments.**
Initial Category set to Source
Discussion moved to Destination category.
**_Note:_** **Lvl2** skipped to keep it empty but to check aggregated fields to be updated correctly
Source and Destination category are brand new **Lvl3-4-5-6-7**.
Discussion is new and has no comments.
Discussion moved to Destination category same level.

2. Source categories are **existing** **Lvl1-3-4-5** destination categories from previous step.
Destination category is brand **new** **Lvl3**.
Discussions are **existing** created on previous step.
4 Discussions from 4 **different Source** categories are moved to the **same Destination** category.

3. Source category is **existing Lvl3** category created on previous step 2.
Destination category is **existing Lvl2** category created on step 1.
Discussions are **existing** moved on previous step.
2 Discussions from the **same Source** category are moved to the **same Destination** category.

4. Source category is **existing Lvl7** category created on step 1.
Destination category is **existing Lvl3** category created on step 1.
Discussion is **existing most recent** created on step 1.
**4.1**  Source category is **existing Lvl6** category created on step 1.
Destination category is **existing Lvl3** category created on step 2.
Discussion is **existing** created on step 1.
**4.2** Source category is **existing Lvl3** category just used as Destination in 4.1.
Destination category is **existing Lvl2** category created on step 1 and used as Destination in step 2.
Discussion is the **existing** one just moved in 4.1.

5. Source category is **existing Lvl2** category created on step 1.
Destination category is **existing Lvl3** category created on step 3.
Discussion is the **existing** the 1st one created on step 1
but with fresh **new comment** just added.

6. Source category is **existing Lvl2** category created on step 1.
Destination category is **existing Lvl3** category created on step 1.
Discussion is the **existing** one created on step 1
but with fresh **new comment** just added.
**6.1** Source category is **existing Lvl3** category created on step 3.
Destination category is **existing Lvl3** category created on step 1.
Discussion is the **existing with comment**  just created on step 5.

Relates to #7869

### DIFF
--- a/tests/APIv0/Controllers/ModerationControllerTest.php
+++ b/tests/APIv0/Controllers/ModerationControllerTest.php
@@ -1,15 +1,17 @@
 <?php
 /**
- * Created by PhpStorm.
- * User: alexander
- * Date: 2018-09-27
- * Time: 11:31 AM
+ * @author Alexander Kim <alexander.k@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL-2.0-only
  */
 
 namespace VanillaTests\Controllers;
 
 use VanillaTests\APIv0\BaseTest;
 
+/**
+ * Class ModerationControllerTest: tests confirmdiscussionmoves action
+ */
 class ModerationControllerTest extends BaseTest {
     /**
      * @var array Array of categories holding initial and continually updated valid values for tests
@@ -20,7 +22,9 @@ class ModerationControllerTest extends BaseTest {
      */
     protected static $discussions = [];
 
-
+    /**
+     * Initial config set up for test
+     */
     public function testModerationDiscussionMoveInintDB() {
         $this->api()->saveToConfig([
             'Garden.Registration.Method' => 'Basic',
@@ -32,6 +36,14 @@ class ModerationControllerTest extends BaseTest {
         $this->api()->setUser($this->addAdminUser());
     }
 
+    /**
+     * Helper function updates valid data array for tests
+     *
+     * @param string $catKey
+     * @param string $fieldToUpdate
+     * @param mixed $newValue
+     * @param bool $recursively
+     */
     protected static function updateValidValues(string $catKey, string $fieldToUpdate, $newValue, bool $recursively = true) {
         do {
             $continue = false;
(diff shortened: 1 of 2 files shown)


tests/APIv0/Controllers/ModerationControllerTest.php
```diff
@@ -0,0 +1,484 @@
...
+     * We expect detCategory recent field to not be updated
+     *
+     * @large
+     *
+     * @depends testInitDB
+     * @depends testMoveNewDiscussionsEmptyCategories
+     * @depends testMoveExistingDiscussionToEmptyCategory
+     * @depends testMoveExistingDiscussionToNotEmptyCategory
+     * @depends testMoveMostRecentDiscussionToNotEmptyCategory
+     * @depends testAddCommentMoveDiscussion
```
initvector: I'm concerned with how tightly these tests seem to be coupled. It looks like you're relying on data being set in one test in order to run another. There's a heavy dependence on the "depends" tag, which would typically be used to skip a test because functionality it relies on failed its tests (e.g. "don't test revoking access tokens, if we can't issue them"), but the usage here appears to be to ensure the referenced category and discussion data has been set from other tests.

A better way to handle your site data might be to create your category structure in a `setupBeforeClass` (or reset it in a `setup` method) and add your relevant post data only in each test. I believe this would allow you to run specific tests without the current "dependencies" that require almost all tests to be run, in some cases.

tests/APIv0/Controllers/ModerationControllerTest.php
```diff
@@ -0,0 +1,484 @@
...
+                    break;
+                case '--':
+                    self::$categories[$catKey][$fieldToUpdate]--;
+                    break;
+                default:
+                    self::$categories[$catKey][$fieldToUpdate] = $newValue;
+            }
+
+            if ($recursively) {
+                if (($pos = strrpos($catKey, '_')) > 0) {
```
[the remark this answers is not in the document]
alex-mtl: No in this case I don't need to gat a parent. I need to get an array key.

tests/APIv0/Controllers/ModerationControllerTest.php
alex-mtl: As usually, I can agree on theory. But in reality, if we set "good answer" (expected) against "unit test" that should work. But in reality we have integration tests. And all these scenarios appeared only because I did run them in different order and that helped me to discover more and more weird behavior  and weird values. Right now it "seems" or "looks like" we know "expected" good values. But in fact we don't. As Lincoln mentioned during one of our discussion, probably best approach is to do step back and to set clear technical requirements document and then we can set up clear unit tests even before we start an implementation. But in our particular case we have code running without any description of what that code should do and what it shouldn't.

If you want, I can split this big "integration" test into pieces (I can kind do snapshot after each step and store intermediate results), but I have strong opinion that it is wrong approach which will lead to tests which can detect much less issues.

tests/APIv0/Controllers/ModerationControllerTest.php
```diff
@@ -0,0 +1,484 @@
...
+        // Check if everything is updated correctly
+        $this->recheckCategories();
+    }
+
+    /**
+     * Helper function:
+     * automatically creates source and destination categories with its parents recursively
+     * create discussion under source category and move it to destination
+     *
+     * @large
```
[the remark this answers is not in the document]
alex-mtl: done

tests/APIv0/Controllers/ModerationControllerTest.php
alex-mtl: Another option: we can just transform several tests into one big test with consecutive actions and assertions

tests/APIv0/Controllers/ModerationControllerTest.php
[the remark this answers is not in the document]
initvector: Aren't you using these array keys as alternate parent category IDs? It looks like you're making the following structure:

- cat1
    - cat1_1
        - cat1_1_1

You're then chopping off the trailing _1 to get the key for the parent. Why would this be better than using the parent category ID column that category rows already have?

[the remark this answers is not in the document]
alex-mtl: I don't want to depend on any data returned from api(). So I don't know what is ParentCategoryID, but i know what is cat1. To me the array $categories is just dataset with values that should match something I get from api(). So I am trying to keep it "independent". so if for some reason `$categories['cat1_1']` has no ParentCategoryID - that is fine. I don't want to implement  `if ParentCategoryID== -1 break;` I am not implementing categoryModel. I do implement test. 

my `$categories['cat1_1']`  can become `[]` at some point and I don't want to implement new `val()` to check what is going on. Those elements are not Categories they are just data sets.

tests/APIv0/Controllers/ModerationControllerTest.php
initvector: If you want to make a larger test, that's fine, but my basic point is that I should have the option of running one test or every test in the "test case". Each test should be verifying specific behavior, so it shouldn't require an unnecessarily long list of dependencies. Here, attempting to run `testCommentDestinationLastCommentFresher` shouldn't require half a dozen other tests to run that don't functionally preclude this one test from running. In this particular case, where we're testing to make sure the last comment was updated when a discussion is moved, it *might* make sense to have `testAddCommentMoveDiscussion` as a dependency, but that's about it. Even then, it's only a dependency because if we aren't moving discussions properly, there's no point in checking the comments. The contents of `testCommentDestinationLastCommentFresher` shouldn't care about what was written to the database in other tests. It's self-contained. If it's needs data, it generates it, unless that data was made available by one of the setup methods.

alex-mtl: I am 100% sure the problem is the language. I just see things and can not express myself.

>  there's no point in checking the comments.
That would be true if we had an application created with respect of some technical requirements. But we don't. My way to test helped me to discover weird side effects in data changes. That is something you did care from the beginning. You wanted this test to test flow to assure that next implementation will not break the current data changes even if those are weird and unexpected. 

If I would follow your logic I would never probably find this bug:
https://github.com/vanilla/vanilla/issues/7897

Because according to general logic that should never exist. So why should we test that case?
But I prefer to "play" with data and top "play" with application. I do use them as "black box" and send same requests or very similar just to confirm that  result remains the same. I will work to transform this sequence of tests to set of independent tests. But I disagree that it is something good. The whole story of why that became necessary to be made is against this approach.

initvector: I said "there's no point in checking the comments", **if the `testAddCommentMoveDiscussion` failed**. I did not say you should not test for that behavior, at all. I was illustrating a *good* usage of test dependencies. With the current test setup, you could add `testAddCommentMoveDiscussion` (and only `testAddCommentMoveDiscussion`) as a dependency to `testCommentDestinationLastCommentFresher`. If we were failing to move a discussion and it's comments, it would make sense to skip testing something that depends on discussions/comments moving. Dependencies shouldn't be used as a way to order your tests so you'll have test data by the time you execute the target.

I'm all for you testing whatever scenarios and data you can think of. I just want them to be structured in a sane way.

PR #7899, opened 2018-10-04
Bypass cascade cache update when move discussion

### LINKED ISSUE #7105: ModerationController::confirmDiscussionMoves is exhausting allowed memory size

alex-mtl:
This tricky fix should temporary disable cascade cache update of CategoryModel when **confirmdiscussionmoves** action of ModerationController and reset Category cache.

This should increase performance and decrease memory usage on this action.
Any next incoming request should get fresh data and reinitialize CategoryModel cache keys and values 

Relates to #7843 

Fixes #7105

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -3514,6 +3514,7 @@ SQL;
         }
 
         // Update the cache.
+/*
         $categoriesToUpdate = self::instance()->getWhere(['CategoryID' => $updatedCategories]);
         foreach ($categoriesToUpdate as $current) {
             $currentID = val('CategoryID', $current);
@@ -3524,6 +3525,7 @@ SQL;
                 ['CountAllDiscussions' => $countAllDiscussions, 'CountAllComments' => $countAllComments]
             );
         }
+*/
     }
 
     /**
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -3483,8 +3483,9 @@ SQL;
      * @param int $categoryID
      * @param string $type
      * @param int $offset A value, positive or negative, to offset a category's current aggregate post counts.
+     * @param bool $skipCache Flag to not update cache
      */
-    private static function adjustAggregateCounts($categoryID, $type, $offset) {
+    private static function adjustAggregateCounts($categoryID, $type, $offset, bool $skipCache = false) {
         $offset = intval($offset);
 
         if (empty($categoryID)) {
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -3483,8 +3483,9 @@ public static function flattenTree($categories) {
      * @param int $categoryID
      * @param string $type
      * @param int $offset A value, positive or negative, to offset a category's current aggregate post counts.
+     * @param bool $skipCache Flag to not update cache
      */
-    private static function adjustAggregateCounts($categoryID, $type, $offset) {
+    private static function adjustAggregateCounts($categoryID, $type, $offset, bool $skipCache = false) {
```
initvector: This isn't "wrong", but I'm curious: did you have a reason for naming this `$skipCache` and defaulting it to `false` versus naming it something like `$cache` and defaulting it to `true`? This is purely personal preference, but I probably would've gone with the latter. My thinking is that we're essentially turning a behavior on/off that is on by default, so the flag would reflect the desired state of the behavior.

Again, this is mostly a personal preference thing. I just want to get some insight into your thought process. 😄

alex-mtl: I would agree to name param `$cache` if I would work on something regular from scratch and do it correct way. `$cache` sounds and looks like **cache enabler** but in this particular case it is not. In this case I had to add new parameter because I don't want to break the system and flow which I don't want to touch. When I name parameter `$skipCache` - I send a signal to developer what is that parameter is for. Whenever developer on IDE will try to call this method it will "suggest" to add or not "skipCache" - which should give a clear idea to developer if he wants or not to SKIP CACHE but not to enable or disable cache layer. 
screen shot 2018-10-09 at 8 59 36 am
To me that is more explicit and/or lets say "semantically correct" than to get
screen shot 2018-10-09 at 9 01 29 am

`$cache` sounds very general and too much "obvious" when `$skipCache` sounds like something we should think twice before use. And that was my intention.

To me the whole this patch is very tricky and should not exist. So whenever I add more to existing mess, I don't want to make an impression on next developer that this is something good and correct, but to highlight that that is something weird and need to be refactored and redone.

That is my way to "protest" to things like `t()` or `c()` which make an impression that they are good, lightweight and correct, and everybody are welcome to use them. In fact for sure it was never written. But names of method or variable always have the spirit that developer can feel. So I don't want any developer to think that my new parameter is something that they can easily use and switch cache on/off using that parameter. No that is not for that. It was not designed for that. It was not tested for that. This parameter was created for weird case to maintain weird behavior and next developer should think twice if he can reuse this parameter when he feels it needed.

initvector: It sounds like you're hoping developers will make some assumptions about usage, based on the chosen parameter name. If you want to warn developers about using a parameter, you're going to be *much* better served by using comments/documentation.

alex-mtl: I don't like the idea of writing huge documentation/comment over the code that should never exist in my opinion. And in this case yes, I prefer to send a signal to developer but not overdocument weird code fragments. The fact your eye caught the parameter name - just confirms that I've got my target.

initvector: > The fact your eye caught the parameter name - just confirms that I've got my target.

But it caught my eye because it seemed like the opposite of what I would've done. It doesn't instill any sense of caution from me. I see a lot of "that's different" things in PRs, but they don't usually make me pause to consider if someone is intentionally being different for the sake of making the developer question whether or not they should be using the functionality. Relying on developer assumptions is usually not a good plan. Developers of all skill levels contribute to Vanilla, so I think the only real way for you to ensure they're going to *know* something is by documenting it.

> I don't like the idea of writing huge documentation/comment over the code that should never exist in my opinion.

I think that's an especially good reason to document this code. I'm not saying you should write a novel, but you're adding code you don't think should be necessary, that you don't want anyone to use, and relying on your innocuous name to be a hint to developers they shouldn't be using it. You could add a sentence or two on the parameter tag description to warn developers of potential issues. In that case, you wouldn't have to be concerned with whether or not they're going to understand "skip cache" also means "use this parameter sparingly, because it can make category data inaccurate".

Issue #7901, opened 2018-10-04
Category-Following: Category Unfollow option missing when viewing "Following" categories


dannimajdalani:
### Details

When following a category, then selecting view "Following", clicking the cogwheel next to a category, would not display the "Unfollow" option.
<img width="177" alt="screen shot 2018-10-04 at 10 42 06 am" src="https://user-images.githubusercontent.com/31856281/46481840-3ca79a80-c7c2-11e8-80e8-3c88f64d4c69.png">
### Current behaviour

<img width="141" alt="screen shot 2018-10-04 at 10 42 26 am" src="https://user-images.githubusercontent.com/31856281/46481848-403b2180-c7c2-11e8-8034-8be82ef23d3a.png">

### Correct behaviour

<img width="157" alt="screen shot 2018-10-04 at 10 44 35 am" src="https://user-images.githubusercontent.com/31856281/46481962-842e2680-c7c2-11e8-8cd7-dbce0127d016.png">

### Cause

In [class.categorymodel.php#L1739](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1739) we are getting array_keys of categories then in [class.categorymodel.php#L1742](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1742) we get userData as an array with CategoryIDs as the key. (the problem)
 in [class.categorymodel.php#L1748](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1748) when we do a $row = ($userData[$iD] ?? false);

PR #7902, opened 2018-10-04
Category-Following: Displaying Unfollow option when filtering by followed categories

dannimajdalani:
closes [7901](https://github.com/vanilla/vanilla/issues/7901)
details in the referenced ticket.

### Current behaviour
Ex.  a user is following 2 categories with CategroyIDs[1,3]
When debugging this, if you put a break point in [class.categorymodel.php#L1742](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1742) we had [$iDs](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1739) 
screen shot 2018-10-04 at 11 11 18 am
screen shot 2018-10-04 at 11 22 14 am

userData would be

screen shot 2018-10-04 at 11 11 33 am
[ $row = ($userData[$iD] ?? false);](https://github.com/vanilla/vanilla/blob/master/applications/vanilla/models/class.categorymodel.php#L1748) would be false , and when in view "Following" . instead of having the option to Unfollow a category, you would still see the "Follow" option.

### Changes
I updated $categories array_keys to be the value of each category id.

screen shot 2018-10-04 at 11 20 18 am
screen shot 2018-10-04 at 11 21 25 am

### To Test

- Enable category-following
- Follow categories
- select view "Following", click the cogwheel. you should have the option to Unfollow the categories.

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -1736,7 +1736,7 @@ class CategoryModel extends Gdn_Model {
      * @param bool $addUserCategory
      */
     public static function joinUserData(&$categories, $addUserCategory = true) {
-        $iDs = array_column($categories,'CategoryID', 'CategoryID');
+        $iDs = array_column($categories, 'CategoryID', 'CategoryID');
         $categoriesKeys = array_column($categories, 'CategoryID');
         $categories = array_combine($categoriesKeys, $categories);
 
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -1736,7 +1736,9 @@ class CategoryModel extends Gdn_Model {
      * @param bool $addUserCategory
      */
     public static function joinUserData(&$categories, $addUserCategory = true) {
-        $iDs = array_keys($categories);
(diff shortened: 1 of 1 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -1736,7 +1736,9 @@ public static function joinRecentChildPosts(&$category = null, &$categories = nu
      * @param bool $addUserCategory
      */
     public static function joinUserData(&$categories, $addUserCategory = true) {
-        $iDs = array_keys($categories);
+        $iDs = array_column($categories, 'CategoryID', 'CategoryID');
+        $categoriesKeys = array_column($categories, 'CategoryID');
```
topcheese13: This looks like an extra line of code you can avoid.

```
$iDs = array_column($categories, 'CategoryID', 'CategoryID');
$categories = array_combine($iDs, $categories);
```

cchabilall83: I don't really follow why we extract the ID's.  But that being said, I don't I think doing a refactor on the code increases the risk of breaking something else.  I'm not sure the if there is any value doing a refactor in the code.

PR #7903, opened 2018-10-04
Add different fields frequently requested to /utility/exportprofiles

topcheese13:
Context
---
A customer requested we add the "Location" field to the endpoint. I thought it would be good to ask around and see what fields aren't a part of the current data this endpoint generate and add them to reduce the number of export request we receive.

Added fields
---
Location (customer request)
LastIPAddress
Roles (concatenated names)
Rank (name) (only if plugin is enabled)

Optimizations
---
I also made a small optimization by calling the `select()` function only once with an array instead of once for every field in the basic query.

Other info
---
This endpoint is capped at 100,000 users through the `$UserMegaThreshold` variable of the `UserModel`. If you have more than that it returns the message "you have too many users" before even trying to fetch them. Joining on roles shouldn't be so bad as the RoleID column are indexed. The RankID column however isn't and we do join on it to fetch the name. I tested it locally with 99,000 users and there was no performance issue whatsoever.

Testing
---
I recommend testing users with multiple roles, testing with and without ranks enabled.

### DIFF
--- a/plugins/ProfileExtender/class.profileextender.plugin.php
+++ b/plugins/ProfileExtender/class.profileextender.plugin.php
@@ -538,7 +538,7 @@ class ProfileExtenderPlugin extends Gdn_Plugin {
                 'u.InviteUserID',
                 'u2.Name as InvitedByName',
                 'u.Location',
-                'group_concat(r.Name)',
+                'group_concat(r.Name) as Roles',
             ])
             ->from('User u')
             ->leftJoin('User u2', 'u.InviteUserID = u2.UserID and u.InviteUserID is not null')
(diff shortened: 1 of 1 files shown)


plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -522,30 +522,45 @@ public function utilityController_exportProfiles_create($sender) {
 
         // Determine profile fields we need to add.
         $fields = $this->getProfileFields();
-        $columnNames = ['Name', 'Email', 'Joined', 'Last Seen', 'Discussions', 'Comments', 'Points', 'InviteUserID', 'InvitedByName'];
+        $columnNames = ['Name', 'Email', 'Joined', 'Last Seen', 'LastIpAddress', 'Discussions', 'Comments', 'Points', 'InviteUserID', 'InvitedByName', 'Location', 'Roles'];
```
initvector: Can we keep the existing casing for `LastIPAddress`?

topcheese13: Done!

PR #7908, opened 2018-10-10
Updates for KB - Paragraph, Icons, Sentence

slafleche:
Minor refactoring for Paragraph, Icons and Sentence components for KB PR.


library/src/scripts/utility.ts
```diff
@@ -325,3 +325,13 @@ export function sanitizeUrl(url: string) {
...
 }
+
+/**
+ * Check if valid array
+ *
+ * @param items The array to validate
+ * @param minItems Minimum expected items, defaults to 1;
+ */
+export function loopableArray(items?: any[], minItems: number = 0) {
+    return !!items && items.length && items.length > minItems;
```
charrondev: Do we really need a utility for this? The intermediary items.length seems unnessary because we have what is ensured to be an array type which always has that property. In that case this utility function would allow us to use the following as equivalents:

```ts
if (items && items.length > 1) {}
// with
import {loopableArray} from "@library/utility";
if (loopableArray(items, 1) {}
```

It doesn't seem like a gain to me. If you still want to keep this function though I think it needs a better name because "loopable" doesn't mean much to me. A 0 length array is still loopable. You can iterate on it. Maybe something like `arrayHasMinLength()` but then you're longer in length than the original check.

applications/dashboard/src/scripts/pages/recoverPassword/RecoverPasswordPage.tsx
```diff
@@ -70,13 +70,13 @@ export class RecoverPasswordPage extends React.Component<IProps, IState> {
...
                     <Paragraph
-                        content={t("RecoverPasswordLabelCode", "Enter your email to continue.")}
+                        children={t("RecoverPasswordLabelCode", "Enter your email to continue.")}
                         className="authenticateUser-paragraph"
                     />
                     <form id={this.id} onSubmit={this.handleSubmit} aria-labelledby={this.pageTitleID} noValidate>
                         <Paragraph
                             className="authenticateUser-paragraph"
-                            content={getGlobalErrorMessage(this.props.requestPasswordState, ["email"])}
+                            children={getGlobalErrorMessage(this.props.requestPasswordState, ["email"])}
```
charrondev: Children is typically used the contents between the tags. If we're going to rename this property as children, we should use it that way. Otherwise, content is a better name if we're going to call it as a normal prop.

slafleche: I do like being able to pass data either through children. 
I didn't want to do too much refactoring in authenticate pages in this PR.

Issue #1750, opened 2013-12-10
Disable plugin fails when folder does not match case of Plugin Name in array


peregrine-web:
in library/core/pluginmanager.php

change around line 1035 in vanilla 2.1b2

from

  // 2. Perform necessary hook action
      $this->_PluginHook($PluginName, self::ACTION_DISABLE, TRUE);

to

   
       $RealRoot = GetValue('RealRoot', $Plugin);
       $PluginName = preg_replace('/.*plugins\//', '', $RealRoot); 
      // 2. Perform necessary hook action
      $this->_PluginHook($PluginName, self::ACTION_DISABLE, TRUE);

this will retrieve the actual folder and set the config correctly and disable the plugin.

unfortunately my connection is to slow to use github pull push, etc.

bleistivt:
The documentation now says:
> The array key (FancyPlugin above) must exactly match the folder name, including capitalization.
http://docs.vanillaforums.com/developers/plugins/quickstart/

So I wouldn't allow this - a wrong folder name voids your warranty.

Issue #1748, opened 2013-12-05
Access to crossrider.com happens always via HTTP://


skaag:
We use SSL for our forum, and get this warning:

The page at 'https://talk.mastercoin.org/index.php?p=/vanilla/categories/general' was loaded over HTTPS, but displayed insecure content from 'http://resources.crossrider.com/apps/30846/resources/meta/0?r=0.2908627854194492': this content should also be loaded over HTTPS.

This causes an annoying browser warning about mixed assets (http + https). Instead, the JS should check if the site is being accessed via HTTP or HTTPS and use the correct protocol with Crossrider.com.

PR #7914, opened 2018-10-11
Fix IPV6 compatibility when viewing eventlog

dannimajdalani:
**Closes** [#1748](https://github.com/vanilla/internal/issues/1748)
**Related** [#1750](https://github.com/vanilla/internal/pull/1750)

I changed IP to IPAddress because of the issue mentioned in the referenced task 1748 and after the change in 1750.


dannimajdalani:
https://github.com/vanilla/vanilla/pull/7923

PR #7917, opened 2018-10-11
Fix post counters on profile pages exceeding allotted area

### LINKED ISSUE #7916: Counts are not treated before displaying on profile page
**Describe the bug**
On the profile page panel, Discussions and Comments counts are not treated properly and might get huge for such a small space.

**Vanilla info**
**1. Are you sure this is a core Vanilla problem and not caused by any addon you are using?** Yes.
**3. What Vanilla version are you using?** `master` branch
**4. When did the issue start?** Always been like that?

**To Reproduce**
1. Visit any profile page on any theme with a user that has more than 9000 comments or discussions.
2. Check the panel to see some ugly count bubbles.

**Expected behavior**

igraziatto:
Closes #7916

### DIFF
--- a/applications/vanilla/settings/class.hooks.php
+++ b/applications/vanilla/settings/class.hooks.php
@@ -697,8 +697,11 @@ class VanillaHooks implements Gdn_IPlugin {
             $discussionsLabel = sprite('SpDiscussions').' '.t('Discussions');
             $commentsLabel = sprite('SpComments').' '.t('Comments');
             if (c('Vanilla.Profile.ShowCounts', true)) {
-                $discussionsLabel .= '<span class="Aside">'.countString(getValueR('User.CountDiscussions', $sender, null), "/profile/count/discussions?userid=$userID").'</span>';
-                $commentsLabel .= '<span class="Aside">'.countString(getValueR('User.CountComments', $sender, null), "/profile/count/comments?userid=$userID").'</span>';
+                $discussionsCount = getValueR('User.CountDiscussions', $sender, null);
+                $commentsCount = getValueR('User.CountComments', $sender, null);
+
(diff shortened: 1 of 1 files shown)


slafleche:
@initvector I can't think of any edge cases where this would be a problem, can you?

initvector:
> I can't think of any edge cases where this would be a problem, can you?

I think it's fine. The `bigPlural` function is used for counting discussions in a category, so it seems logical that it would be safe to use for counting posts associated with other record types (e.g. users).

applications/vanilla/settings/class.hooks.php
```diff
@@ -697,8 +697,11 @@ public function profileController_addProfileTabs_handler($sender) {
             $discussionsLabel = sprite('SpDiscussions').' '.t('Discussions');
             $commentsLabel = sprite('SpComments').' '.t('Comments');
             if (c('Vanilla.Profile.ShowCounts', true)) {
-                $discussionsLabel .= '<span class="Aside">'.countString(getValueR('User.CountDiscussions', $sender, null), "/profile/count/discussions?userid=$userID").'</span>';
-                $commentsLabel .= '<span class="Aside">'.countString(getValueR('User.CountComments', $sender, null), "/profile/count/comments?userid=$userID").'</span>';
+                $discussionsCount = getValueR('User.CountDiscussions', $sender, null);
+                $commentsCount = getValueR('User.CountComments', $sender, null);
+
+                $discussionsLabel .= '<span class="Aside">'.countString(bigPlural($discussionsCount, '%s discussions'), "/profile/count/discussions?userid=$userID").'</span>';
```
alex-mtl: If I remember correctly `bigPlural` function expects to get "single" variant of string `'%s discussions'` -> `'%s discussion'`

PR #7919, opened 2018-10-12
Add asset model method for getting JS locale script

charrondev:
In pursuit of https://github.com/vanilla/knowledge/issues/197

## Changes

Refactors the generation of a script URL for a given asset by moving it to `AssetModel:: getJSLocalePath()` so it can be used outside of a `Gdn_Controller`.

Functionality should be unchanged.

### DIFF
--- a/applications/dashboard/models/class.assetmodel.php
+++ b/applications/dashboard/models/class.assetmodel.php
@@ -254,14 +254,14 @@ class AssetModel extends Gdn_Model {
     /**
      * Get the resource path for a javascript bundle of a particular locale bundle.
      *
-     * @param string $localeKey The key of the locale to lookup
+     * @param string $localeKey The key of the locale to lookup.
      *
      * @return string The path the locale javascript file.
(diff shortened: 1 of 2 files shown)


applications/dashboard/models/class.assetmodel.php
```diff
@@ -251,6 +251,18 @@ public function getInlinePolyfillJSContent(): string {
...
+    /**
+     * Get the resource path for a javascript bundle of a particular locale bundle.
+     *
+     * @param string $localeKey The key of the locale to lookup
+     *
+     * @return string The path the locale javascript file.
+     */
+    public function getJSLocalePath(string $localeKey): string {
+        $cacheBuster = $this->cacheBuster();
+        return '/api/v2/locales/' . rawurlencode($localeKey) . "/translations.js?etag=$cacheBuster";
```
initvector: Does this still need the `url` treatment to handle subdirectory/node installs?

PR #7922, opened 2018-10-12
Add SiteNav component

slafleche:
Added new Site Nav component.

Notes: 
* For full accessibility spec see: https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html
* Some click handlers are on the SiteNav and some are on the SiteNavNode.

### DIFF
--- a/library/src/scripts/components/siteNav/SiteNav.tsx
+++ b/library/src/scripts/components/siteNav/SiteNav.tsx
@@ -13,6 +13,7 @@ import SiteNavNode from "@library/components/siteNav/SiteNavNode";
 import TabHandler from "@library/TabHandler";
 
 interface IProps extends RouteComponentProps<{}> {
+    id?: string;
     className?: string;
     children: any[];
 }
@@ -22,19 +23,18 @@ export interface IState {
 }
 
 /**
- * Recursive component to generate site nav
- * No need to set "counter". It will be set automatically. Kept optional to not need to call it on the top level. Used for React's "key" values
+ * Implementation of SiteNav component
  */
 export class SiteNav extends React.Component<IProps, IState> {
+    private id;
+
     public constructor(props) {
         super(props);
-        this.state = {
-            id: getRequiredID(props, "siteNav"),
-        };
+        this.id = getRequiredID(props, "siteNav");
     }
 
     public get titleID() {
-        return this.state.id + "-title";
+        return this.id + "-title";
     }
 
     public render() {
@@ -55,7 +55,7 @@ export class SiteNav extends React.Component<IProps, IState> {
                   })
                 : null;
         return (
-            <nav onKeyDownCapture={this.handleKeyDown as any} className={classNames("siteNav", this.props.className)}>
+            <nav onKeyDownCapture={this.handleKeyDown} className={classNames("siteNav", this.props.className)}>
                 <h2 id={this.titleID} className="sr-only">
                     {t("Site Navigation")}
                 </h2>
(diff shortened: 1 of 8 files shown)


library/src/scripts/components/siteNav/SiteNav.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Recursive component to generate site nav
+ * No need to set "counter". It will be set automatically. Kept optional to not need to call it on the top level. Used for React's "key" values
+ */
+export class SiteNav extends React.Component<IProps, IState> {
+    public constructor(props) {
+        super(props);
+        this.state = {
+            id: getRequiredID(props, "siteNav"),
```
charrondev: Why our we using this instead of `uniqueIDFromPrefix()` as a property initializer? Since none if this actually comes from props it could doesn't need `getRequiredId`. In fact if you added the type `IProps` on your constructor this would fail the type checker because there is no ID there.

We could replace this constructor override with

```ts
class SiteNav extends React.Component {
    private id = uniqueIDFromPrefix("siteNav");
}
```

and access it with `this.id`.

library/src/scripts/components/siteNav/SiteNav.tsx
```diff
@@ -0,0 +1,146 @@
...
+                              titleID={this.titleID}
+                              visible={true}
+                              location={this.props.location}
+                              depth={0}
+                          />
+                      );
+                  })
+                : null;
+        return (
+            <nav onKeyDownCapture={this.handleKeyDown as any} className={classNames("siteNav", this.props.className)}>
```
charrondev: Can we remove this `any` assertion? Changing the type of the event in our handler to `React.KeyboardEvent` will fix the type-checking.

library/src/scripts/components/siteNav/SiteNav.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    public firstVisibleOfType = (container, selector: string = ".siteNavNode") => {
+        return container.querySelector(selector + ":not(.isHidden):first-child") || null;
+    };
+
+    public lastVisibleOfType = (container, selector: string = ".siteNavNode") => {
+        return container.querySelector(selector + ":not(.isHidden):last-child") || null;
+    };
+
+    // https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html
```
charrondev: Let's convert this into a proper doc block.

library/src/scripts/components/siteNav/SiteNav.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    this.props.name
+                }\"`}</h2>
+                <ul className="siteNav-children" role="tree" aria-labelledby={this.titleID}>
+                    {content}
+                </ul>
+            </nav>
+        );
+    }
+
+    public firstVisibleOfType = (container, selector: string = ".siteNavNode") => {
```
charrondev: This needs a doc block. The container could use a type here as well. `Element`  should be the correct type.

library/src/scripts/components/siteNav/SiteNavNode.tsx
```diff
@@ -0,0 +1,240 @@
...
+    location: any;
+    depth: number;
+}
+
+interface IState {
+    open: boolean;
+    current: boolean;
+}
+
+export default class SiteNavNode extends React.Component<IProps, IState> {
```
charrondev: Class description missing.

library/src/scripts/components/siteNav/SiteNavNode.tsx
```diff
@@ -0,0 +1,240 @@
...
+    public constructor(props) {
+        super(props);
+        this.state = {
+            open: false,
+            current: false,
+        };
+    }
+
+    public open = (callbackIfAlreadyOpen?: any) => {
```
charrondev: Everything in this class is missing doc blocks.

library/src/scripts/components/siteNav/SiteNavNode.tsx
```diff
@@ -0,0 +1,240 @@
...
+    };
+
+    public handleClick = e => {
+        e.preventDefault();
+        this.toggle();
+    };
+
+    public currentPage(): boolean {
+        if (this.props.location && this.props.location.pathname) {
+            return this.props.location.pathname === this.props.url;
```
charrondev: I don't know how much we care about non-canonical URLs, but this won't work with those. Maybe just a note of that.

If it ends up being an issue we'll have to do some parsing of URL and match the record type + record ID on the navigation items here.

library/src/scripts/components/siteNav/SiteNavNode.tsx
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    public componentDidMount() {
+        this.openRecursive();
+    }
+
+    public render() {
+        const hasChildren = this.props.children && this.props.children.length > 0;
+        const topLevel = this.props.counter === 1;
```
charrondev: This variable is unused.

library/src/scripts/components/siteNav/SiteNavNode.tsx
```diff
@@ -0,0 +1,240 @@
...
+                        <ul className={classNames("siteNavNode-children", { isHidden: !this.state.open })} role="group">
+                            {childrenContents}
+                        </ul>
+                    )}
+                </div>
+            </li>
+        );
+    }
+
+    private next = (tabHandler, currentLink) => {
```
charrondev: `tabHandler` should be typed as `tabHandler: TabHandler` here. `currentLink` seems to be an `Element`.

library/src/scripts/components/siteNav/SiteNavNode.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    />
+                );
+            });
+        const space = `&nbsp;`;
+        return (
+            <li
+                className={classNames("siteNavNode", this.props.className, { isCurrent: this.state.current })}
+                role="treeitem"
+                aria-expanded={this.state.open}
+                style={{ marginLeft: `${(this.props.depth - 1) * 18}px` }}
```
charrondev: I feel like this styling should be done through CSS so we can use our margin variables.

Classes like `isDepth-x` would probably do the trick right?

PR #7923, opened 2018-10-15
Fix IPV6 compatibility when viewing eventlog

dannimajdalani:
**Closes** [#1748](https://github.com/vanilla/internal/issues/1748)

### Details
When viewing the eventlog in settings/eventlog.json, calling ipDecodeRecursive with $key == 'IP' in [class.controller.php#L1580](https://github.com/vanilla/vanilla/blob/master/library/core/class.controller.php#L1580) is not getting decoded in [functions.general.php#L4219](https://github.com/vanilla/vanilla/blob/master/library/core/functions.general.php#L4219).

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -4216,7 +4216,7 @@ if (!function_exists('ipDecodeRecursive')) {
     function ipDecodeRecursive($input) {
         walkAllRecursive($input, function(&$val, $key = null, $parent = null) {
             if (is_string($val)) {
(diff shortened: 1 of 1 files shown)


topcheese13:
This is the solution that was agreed upon with Todd/Linc ?

dannimajdalani:
> This is the solution that was agreed upon with Todd/Linc? Because I remember this is not what we talked about in a previous meeting.

Yes. I mentioned this in the main issue.

topcheese13:
Indeed, my bad! There was 3 IPv6 PRs I got confused.

PR #7926, opened 2018-10-15
Updates to dropdown / popover controllers for KB

slafleche:
Updates to dropdown and popover controller components. These changes are needed for the KB's selectbox component.

### DIFF
--- a/library/src/scripts/components/SelectBox.tsx
+++ b/library/src/scripts/components/SelectBox.tsx
@@ -1,0 +1,140 @@
+/*
+ * @author Stéphane LaFlèche <stephane.l@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+import * as React from "react";
+import classNames from "classnames";
+
+import { getRequiredID } from "../componentIDs";
+import { t } from "../application";
+import DropDownItemButton from "@library/components/dropdown/items/DropDownItemButton";
+import { checkCompact, downTriangle } from "@library/components/Icons";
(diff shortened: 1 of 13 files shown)


library/src/scripts/components/SelectBox.tsx
```diff
@@ -0,0 +1,140 @@
...
+                              disabled={i === this.state.selectedIndex}
+                              clickData={child}
+                              index={i}
+                              current={selected}
+                              buttonClassName="dropDownItem-button selectBox-buttonItem"
+                          >
+                              <span className="selectBox-checkContainer sc-only">
+                                  {selected && checkCompact("selectBox-selectedIcon")}
+                                  {!selected && (
+                                      <span className="selectBox-spacer" dangerouslySetInnerHTML={{ __html: space }} />
```
charrondev: This is not the correct way to set a space inside of the element. We should almost _never_ use dangerouslySetInnerHTML.
Here `<span className="selectBox-spacer">{` `}</span>}` will work. Gave it a quick test to double check. You can't specify the encoded form of HTML entities though. You should specify the plain form, and react will handle converting to HTML entities for you.

library/src/scripts/components/SelectBox.tsx
```diff
@@ -0,0 +1,140 @@
...
+            selectedIndex: index,
+            selectedItem,
+        });
+    };
+
+    public render() {
+        const space = `&nbsp;`;
+
+        const selectItems =
+            this.props.children && this.props.children.length > 0
```
charrondev: This if check here is largely unnecessary. Mapping over an empty array will return an empty array, which react will render the same way that it renders `null` or `undefined`.

PR #7929, opened 2018-10-16
Fix users not appearing logged-in in embedded comments by breaking the cache.

### LINKED ISSUE #7927: Embedded comments being cached, logged-in user does not appear logged in.
Aggressive browser caching means that the embed comments URL in the iFrame is cached. If a user is logged in, he may not appear to be logged in. Some sort of cache-breaking string needs to be added to the URL of the Vanilla URL. This may also be caused by the fact that the URL does not end with a PHP file it ends with /discussions/embed/.

nervoustwit:
**The Problem**:
When embedding comments in an articling site or blog the embedded document can be cached by browsers. This means that a user who was not logged-in will be served the same embedded document after he logs in, making it impossible to add comments.

**The Fix**:
In the embed.js file that generates the iframe and injects the URL into it, add a time stamp to the URL.

**Test**:
I found that Firefox has the most aggressive browser caching. Create an embed comments page. 

 - Before checking out this fix, visit the embedded comments page not logged in. 
 - Click on the "Comment as..." button to log in.
 - When you come back you will still see the "Comment as.." button.
 - Checkout this fix and repeat the above steps. You will be see the message "You are logged in as.." next to the Comment button.

This will close #7927

### DIFF
--- a/js/embed.js
+++ b/js/embed.js
@@ -237,6 +237,8 @@ window.vanilla.embed = function(host) {
 
         if (embed_type == 'comments') {
             result = '//' + host + '/discussion/embed/'
+            // Break the cache.
+            + '&c=' + new Date().getTime()
             + '&vanilla_identifier=' + encodeURIComponent(foreign_id)
             + '&vanilla_url=' + encodeURIComponent(foreign_url);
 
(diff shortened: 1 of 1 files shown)


charrondev:
I gave this an approval because I understand the need it serves. I'm just not really sure what the affect of this on our servers ends up being. This means essentially that every single page load will miss the cache.

---

So this really a "blind" approval on my side. Maybe we can find a second reviewer that could chime in a bit better. @initvector @tburry

linc:
I'm fine with this, but 1 #yolo approval is enough.

Issue #7942, opened 2018-10-17
Refactor, move, and decompose the AssetModel


charrondev:
Related https://github.com/vanilla/vanilla/issues/3048

## Steps

1. Move class into `\Vanilla\Web\Assets\LegacyAssetModel` https://github.com/vanilla/vanilla/pull/7943
- Move file
- Fix root namespace calls to classes to be prefixed with `\` or have a use statement.
- Fix filename.
- Adjust classname.
- Ensure aliases and class names are properly configured so that events fired with automatic names still work.
- Create alias.
2. Update usages in the Vanilla\Vanilla repo. https://github.com/vanilla/vanilla/pull/7944
3. Refactor out some separate pieces to be used independently of the main asset model (I'm thinking having dedicated asset classes, and separating our new webpack js stuff from the nasty Gdn::pluginManager stuff happening in the static `AssetModel` methods). https://github.com/vanilla/vanilla/pull/7947

PR #7947, opened 2018-10-18
Refactor AssetModel into multiple classes

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7942
Closes https://github.com/vanilla/vanilla/issues/7998

To be merged at the same time as https://github.com/vanilla/knowledge/pull/307 This PR contains breaking changes for that project.

## Pieces

- [x] Create new asset classes
- [x] Pull new testable, DIed AssetProvider out of the mostly static `LegacyAssetProvider`
- [x] Add ability to load CSS assets
- [x] Unit tests

## Testing

I've created automated tests for the new asset provider and cache busting, and have performed manual testing of the KB project and forum to verify the cache busting and loading is working properly.

## Core project updates

I've added a few new rules for the container to facilitate unit testing a bit easier.

- Added `AddonInterface` and `AddonProviderInterface` interfaces that are implemented by `Addon` and `AddonManager`. There is a now a `MockAddonProvider` that simplifies mocking of addons for testing purposes.
- Added `ConfigurationInterface` interface and set it as an alias of `Gdn_Configuration`. Currently it has `get()` but this is the 99% use case for configuration use.
- Adds a new utility class `ContainerUtils` with static methods for returning references to configuration values and the current locale key. This allows container config to inject just 1 or 2 config values instead of the whole config object.
- Added a `DateTimeInterface` to the container rules. This defaults to an ImmutableDateTime, but gives some flexibility in providing times through DI. The null construct args needed to be added as the defaults because our containers reflection doesn't work properly on language built ins.

### `Gdn_Request` changes.

I was on the fence about doing this but in order to properly facilitate testing these components I needed to add the `AssetRoot` I needed to get the asset root into the `ArticleInterface`. I was considering just adding the dual use `assetRoot($newAssetRoot)` to the interface but it didn't feel right (didn't match the rest of the interface). So instead I've done the following:

- Add 2 new methods `getAssetRoot` and `setAssetRoot` to the `RequestInterface` similar to `getRoot` and `setRoot`.
- Implement these 2 methods on `Gdn_Request` and our `Fixtures\Request`.
- Update the dual purpose `Gdn_Request::assetRoot()` to use the new methods.
- Deprecate `Gdn_Request::assetRoot()`
- Replace our usages of `Gdn_Request::assetRoot()` with the 2 new methods.

## Decomposing `LegacyAssetModel`

- `LegacyAssetModel::cacheBuster()` has been deprecated. Its logic has been moved into a new class `DeploymentCacheBuster`. This method was quite re-usable and there is some likely-hood that it may have been used in 3rd party addons, which is why I've opted for a deprecation rather than a removal.

### New Assets

Our new Webpack built assets had their own methods glommed onto the `AssetModel` before it was renamed to `LegacyAssetModel`. `LegacyAssetModel` is mostly static methods that are not dependency injected properly, and has dependencies on things like the deprecated `PluginManager` and event system.

I've broken all of the stuff used for our built code (excluding dashboards old build process which we still don't have a migration strategy for) in the following hierachy:

- `AssetInterface` -> Simple interface for a class based asset.
  - `SiteAsset` -> An asset that uses the request and a deployment time based CacheBuster.
     - `WebpackAsset` and `WebpackAddonAsset` - Assets for core and addon related JS and stylesheet assets.
     - `LocaleAsset` -> Asset representing the JS file that can be produced by the locales endpoint. I were to do more of these API -> js file endpoints we might consider breaking out an `ApiAsset` but at this point I don't see any point adding that level of hierarchy.
  - `HotBuildAsset` - > Asset for hot. Eg. `HotReload.Enabled === true`.

Providing these assets are the `WebpackAssetProvider`. This is the way to gather the assets with its methods replacing providing all scripts and stylesheets needed for vanilla's new stack. I'm open to renaming this to something else, because it does provider more than _just_ the webpack assets, but I still feel like just `AssetProvider` might not be specific enough. I'm open to suggestions here.

## Removed `LegacyAssetModel` methods

I've straight up removed the our previous methods for providing:

- Inline polyfill script
- Locale script
- Webpack built scripts.

I'm reasonably confident we were the only ones using these methods (introduced in vanilla 2.7) and they only had to usages:

1. `Gdn_Controller` - Usages updated in this PR.
2. Our new base controller for KB. - Usages updated in this companion PR https://github.com/vanilla/knowledge/pull/307

## New functionality

I've removed a kludge where the rich editor manually loaded its built stylesheet files. There is now a dedicated method for this `getStylesheets()` which returns stylesheet assets in the same way as `getScripts` returns script assets.


library/Garden/Web/RequestInterface.php
```diff
@@ -56,6 +56,21 @@ public function getRoot();
...
+    public function getAssetRoot();
+
+    /**
+     * Set the root folder for static files.
+     *
+     * @param string $root The new root path of files.
+     * @return $this
+     */
+    public function setAssetRoot(string $root);
+
```
alex-mtl: I can not get the logic why  getAssetRoot and setAssetRoot should belong to this interface?

tests/Library/Vanilla/Web/Asset/WebpackAssetTest.php
```diff
@@ -0,0 +1,103 @@
...
+                "http://example.com/someRoot/dist/testSec/test.min.js",
+            ],
+            [
+                (new Request())->setHost("me.com"),
+                "cacheBuster",
+                "http://me.com/dist/testSec/test.min.js?h=cacheBuster",
+            ],
+        ];
+    }
+}
```
alex-mtl: Probably we can add case with both: -> setAssetRoot and ->setHost ?

charrondev: Done

library/Garden/Web/RequestInterface.php
charrondev: It's there because `->getRoot()` and `setRoot()` are already on that interface. Right now `Gdn_Request/RequestInterface` is where we get information from the server about how to make our URLs. I'd support a proposal for trying to pull a bunch of that stuff out of our request object but at the moment I consider that outside of the scope of the PR.

With the idea that this PR should not require decomposing our Request object (and it cannot without a lot of coordination with infrastructure), than that leaves me with 4 options if I want to write tests for this:

1. Leave the existing `assetRoot` as is and add it to `RequestInterface`. I disagree with this because having a dual-purpose getter/setter does not align with the rest of the methods on `RequestInterface`.
2. The implementation in this PR (deprecate assetRoot in favor of get/set assetRoot).
3. Not write tests involving the assetRoot.
4. Move the asset root setting/getting somewhere else, but still need the `Request` object anyways for the base path.

library/core/class.request.php
```diff
@@ -100,16 +100,37 @@ public function __construct() {
...
+     */
+    public function getAssetRoot() {
+        return $this->_parsedRequestElement('AssetRoot');
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function setAssetRoot(string $assetRoot) {
+        $this->_parsedRequestElement('AssetRoot', rtrim('/'.trim($assetRoot, '/'), '/'));
```
initvector: I know you're copying forward existing code, but let's give it a little more breathing room.

```suggestion
        $this->_parsedRequestElement('AssetRoot', rtrim('/' . trim($assetRoot, '/'), '/'));
```

initvector: I'm possibly overlooking something (and maybe this was discussed before), but the trimming seems a little redundant to me.

1. Provide an asset root (/path/to/assets/).
1. Trim off forward slashes from either end (path/to/assets).
1. Prepend a forward slash (/path/to/assets).
1. Trim any trailing forward slashes. But these should've been removed in step two, right?

tests/InternalRequest.php
```diff
@@ -216,6 +216,27 @@ public function setScheme($scheme) {
...
+     */
+    public function setAssetRoot(string $root) {
+        $this->setRoot($root);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function urlDomain($withDomain = true): string {
+        return "/";
```
initvector: A forward slash doesn't appear to be a valid return value for a method documented to "conditionally get the domain of the request". The return values appear to be based on `$withDomain`:

https://github.com/vanilla/vanilla/blob/2e0109e3acef1d750c1ec926defa1fe69b104e57/library/core/class.request.php#L1697-L1701

tests/Library/Vanilla/Web/Asset/WebpackAssetProviderTest.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+class WebpackAssetProviderTest extends TestCase {
+
+    private $fs;
+
+    /**
+     * @inheritdoc
+     */
+    public function setUp() {
+        $this->fs = vfsStream::setup();
```
initvector: Is this property necessary? It doesn't look like it's ever referenced.

library/Vanilla/Web/Asset/LegacyAssetModel.php
```diff
@@ -21,33 +21,25 @@
...
     /**
-     * @var \Vanilla\AddonManager
+     * LegacyAssetModel constructor.
+     *
+     * @param string $cacheBuster
      */
-    private $addonManager;
-
-    public function __construct(\Vanilla\AddonManager $addonManager) {
+    public function __construct(string $cacheBuster) {
```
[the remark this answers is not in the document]
charrondev: The AddonManager even being added to it was a pretty recent development (2.7) and should only have ever been instantiated through the container. Most usage of the this class was it's static methods.

Theoretically someone could have been using this in open source if they were working against master, but we never publicly documented any of the public methods. I don't have particular concern as a result.

library/Vanilla/Web/Asset/SiteAsset.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Create a web-root url, not an asset URL. This is useful for application level resources like ones created from
+     * API endpoints.
+     *
+     * @param string[] $pieces The pieces to join together.
+     *
+     * @return string
+     */
+    protected function makeWebPath(string ...$pieces): string {
```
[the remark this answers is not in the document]
charrondev: Done.

library/Vanilla/Web/Asset/WebpackAssetProvider.php
```diff
@@ -0,0 +1,243 @@
...
+        $this->addonProvider = $addonProvider;
+    }
+
+    /**
+     * Enable loading of hot reloading assets in place of the normal ones.
+     *
+     * @param bool $enabled The enable value.
+     * @param string $ip Optionally override the ip address the hot bundle is served from.
+     */
+    public function setHotReloadEnabled(bool $enabled, string $ip) {
```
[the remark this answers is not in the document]
charrondev: Done. I've defaulted it to "" though because I use `?:` and PHP doesn't have union types.

library/Vanilla/Web/Asset/LegacyAssetModel.php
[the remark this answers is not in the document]
initvector: `AddonManager` has been injected here since 05519ce8f2617a70ca19841ce411f39da1d613e9 (v2.6), but it sounds like it didn't see any real usage in Vanilla until 2.7. That, coupled with you not perceiving any real risk here, I'm gonna call this "resolved".

library/core/class.configuration.php
```diff
@@ -335,7 +335,7 @@ public static function format($data, $options = []) {
      * @param mixed $defaultValue If the parameter is not found in the group, this value will be returned.
      * @return mixed The configuration value.
      */
-    public function get($name, $defaultValue = false) {
+    public function get(string $name, $defaultValue = false) {
```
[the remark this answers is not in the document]
charrondev: I've removed it, which means it also needed to be removed from the interface until we target PHP7.2 and get type widening.

tests/Library/Vanilla/Web/Asset/WebpackAssetProviderTest.php
charrondev: Removed

library/core/class.request.php
charrondev: There's definitely a difference in the output https://repl.it/repls/OrchidThistleTerabyte.

Even if "not" having the extra trim is easier to follow it does adjust the way the paths are put together a good bit, and this type of thing affects all of the legacy asset system.

Changing that in such a big pull request like this would worry me a lot more than the configuration signature change.

initvector: To recap: it looks like this is to force never ending a path with a slash. I'm assuming this was to make subsequent path concatenation easier. If the result of your path calculation ends up being "/", Vanilla normalizes that to "no path" (an empty string). Otherwise, you could end up with a path that looks like "//path/to/asset", where the first "/" is the "asset root".

tests/InternalRequest.php
```diff
@@ -216,6 +216,27 @@ public function setScheme($scheme) {
...
+     */
+    public function setAssetRoot(string $root) {
+        $this->setRoot($root);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function urlDomain($withDomain = true): string {
+        return string;
```
initvector:
```suggestion
        return "";
```

PR #7963, opened 2018-10-23
Updates for KB Revision list - heading, icons, page loader and core variables

slafleche:
Modifications for KB's revision list:

- Allowed heading component to render a different heading than the style
- Added status icons
- Modified page loader to accept a hard coded status
- Fixed some errors in variable names (@igraziatto) in core variables, removed semi bold (as many fonts do not have it) and decreased spacing between columns in panel layout

Note that the icons are not be final for the statuses.


library/src/scripts/components/Heading.tsx
```diff
@@ -31,6 +32,8 @@ export default class Heading extends React.Component<IHeadingProps> {
         depth: 2,
     };
 
+    private renderAsDepth = this.props.renderAsDepth ? this.props.renderAsDepth : this.props.depth;
```
charrondev: This assignment works and in most cases would be fine, unless one of these 2 props were changed.

In this instance a computer getter is more flexible, as it will be re-computed on subsequent updates.

```suggestion
    private get renderAsDepth(): number {
        return this.props.renderAsDepth ? this.props.renderAsDepth : this.props.depth!;
    }
```

PR #7964, opened 2018-10-23
Add global permissions field to /users/me API v2 endpoint response

### LINKED ISSUE #7949: Add ability to get current user's permissions to API v2
There needs to be a way for the API to retrieve permissions for the current user. This could be appended to the /users/me endpoint with an expand parameter.

initvector:
In an ongoing attempt to make Vanilla's API more accommodating to SPAs, this update introduces the ability to get the current user's permissions as an optional parameter on the /users/me API v2 endpoint. If the "expand" parameter is passed to this endpoint includes "permissions" (or "all"), the signed-in user or guest will have their permissions included as part of the response.

### Changes

1. ~~`getPermissionFragment` has been added as a protected method on `AbstractApiController`. It is intended to represent a permission row in the "new" format.~~
1. ~~`getUserFragment` in `AbstractApiController` has been updated to include an optional "permissions" field, which uses the result of `getPermissionFragment`.~~
1. Rename `index_me` method on `UsersApiController` to `get_me`. This is a matter of semantics. An index in Vanilla's API means a list. A get is a single row. This change has no impact on functionality.
1. ~~Support for an "expand" query string parameter has been added to the /users/me endpoint. It currently supports one field: permissions. If a request is received with this value, the current user's permissions will be included in the response.~~ Add the current user's global permissions to the response of a request to the /users/me API v2 endpoint.
1. The `getExpandDefinition` method of `ApiUtils` has been updated to always include the "all" value in expand definitions.
1. `massageSchema` in `ReflectAction` now checks to see if "all" is in an expand field's options, before adding it.
1. ~~`PermissionsTranslationTrait` has been updated with the `formatPermissions`, which will take a `Vanilla\Permissions` object and return an array of permissions in the new scope-centric format. This format can currently be seen on the roles API endpoint. Essentially: each permission "row" has at least a type key, either "global" or a particular resource, and a permissions key, which contains an array of permissions relevant to that scope. There is also an optional id key, if a resource is specified (e.g. type equals "category" and id is a particular category ID). The primary difference to what will be seen in the users resource response, versus the roles response, is the users response will only return permissions a user has. The roles endpoint includes all permissions and indicates whether or not a particular role has them with a boolean flag. Since users only returns permissions the user has, the result has all permissions with a `true` for their boolean flag.~~

Closes #7949


library/Vanilla/PermissionsTranslationTrait.php
```diff
@@ -100,6 +101,85 @@ private function consolidatePermissions(array $permissions) {
...
+        $global = [];
+        $categories = [];
+
+        /**
+         * Format an array of permission names.
+         *
+         * @param array $perms
+         * @return array
+         */
+        $format = function (array $perms) {
```
charrondev: I feel like this deserves to be pulled out into its own function

library/Vanilla/PermissionsTranslationTrait.php
```diff
@@ -100,6 +101,85 @@ private function consolidatePermissions(array $permissions) {
         return $result;
     }
 
+    /**
+     * Given a permissions object, return its data array in the explicit-scope format.
+     *
+     * @param \Vanilla\Permissions $permissions
+     * @return array
+     */
+    private function formatPermissions(Permissions $permissions) {
```
charrondev: Did this scope exist somewhere else previously? Or is it just being defined for the first time here?

charrondev: Where is the "explicit-scope format" defined? If the definition is here I'd like to see an example of what it's supposed to be and/or unit tests about how it's derived.

library/Vanilla/PermissionsTranslationTrait.php
```diff
@@ -100,6 +101,85 @@ private function consolidatePermissions(array $permissions) {
...
+                'id' => $categoryID == -1 ? 0 : $categoryID,
+                'type' => 'category'
+            ];
+
+            $categoryPermissions['permissions'] = $format($permissionSlugs);
+            $result[] = $categoryPermissions;
+        }
+
+        /**
+         * Step four: return the result. We should have an array of permission rows. Each row containing a
```
charrondev: Can we move the description of the return type up to the return description of the doc block?

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -218,7 +224,7 @@ public function getGuestFragment() {
      * Get a list of users, filtered by username.
      *
      * @param array $query
-     * @return array
+     * @return Data
      */
     public function index_byNames(array $query) {
```
charrondev: If we're updating this we might as well make it and explicit function return type. I'm still quite confused though about why we sometimes return Data and other times arrays. It seems inconsistent.
```suggestion
    public function index_byNames(array $query): Data {
```

applications/dashboard/controllers/api/AbstractApiController.php
```diff
@@ -76,11 +79,28 @@ public function getUserFragmentSchema() {
...
 
+    /**
+     * Get a fragment representing a single permission row.
+     *
+     * @return Schema
+     */
+    protected function getPermissionFragment() {
+        if ($this->permissionsFragment === null) {
+            $this->permissionsFragment = $this->schema([
+                "id:i?",
```
charrondev: Can we add descriptions for these? The "permissions" one in particular seems relevant.

applications/dashboard/controllers/api/AbstractApiController.php
```diff
@@ -76,11 +79,28 @@ public function getUserFragmentSchema() {
                 'name:s' => 'The username of the user.',
                 'photoUrl:s' => 'The URL of the user\'s avatar picture.',
                 'dateLastActive:dt|n' => 'Time the user was last active.',
+                "permissions:a?" => $this->getPermissionFragment(),
```
charrondev: Was there a particular use case you were imagining where someone would have need to the current user and _not_ have need of their permissions?

If not than this should be an always present field instead of an expand field.

initvector: > Was there a particular use case you were imagining where someone would have need to the current user and not have need of their permissions?

Definitely. Maybe if it were going to *only* be global permissions, including them by default would probably be fine. However, some sites have notoriously complicated permissions setups, which can quickly ballon the size of the response. If I have 1,000 categories and custom permissions for each, my "permissions" entry is going to be huge. If all I need to do is build a simple "me" box or get the current user's ID or name, a call to this endpoint will suffice. I don't want to unnecessarily bloat that response with an excessive amount of unnecessary data. Having the consumer explicitly request permissions makes more sense to me.

applications/dashboard/controllers/api/AbstractApiController.php
initvector: Unfortunately not. There seems to be an issue with garden-schema, where expanding on this schema beyond its current state results in erroneous validation errors. I believe we'll need to file an issue against that endpoint, but additional discovery is required.

applications/dashboard/controllers/api/UsersApiController.php
initvector: While I'm okay with this particular change, I don't think this should be standard. I made a very minor update to an existing method's documentation, because my IDE was complaining about it. You're now proposing I change the function signature and increase the potential for a fatal error. Again, in this particular case, it shouldn't be a big deal, but this isn't a normal "while we're in there" type of change.

> I'm still quite confused though about why we sometimes return Data and other times arrays. It seems inconsistent.

You only return a `Data` object when you want to modify more of the response than just the body. If you return an array, that's the body of the response. If you return a `Data` object, you can modify the response codes and other "meta'" information about that response. In this particular case, we're adding "Link" headers for pagination.

library/Vanilla/PermissionsTranslationTrait.php
initvector: I don't know that it'll have any real use outside of `formatPermissions` and with this being a trait, the class implementing it will get all the private methods. I'd rather limit the visibility by keeping it as an anonymous function, until we have to use it elsewhere.

library/Vanilla/PermissionsTranslationTrait.php
initvector: I mentioned this in the PR description, but this particular format was introduced back when the roles API controller was originally built. It never had an official name, so to differentiate it, I'm calling it the "explicit-scope format". This is in contrast to the current permission array format, where global is the implied scope on indexed array keys and per-resource permissions are associative array values.

applications/dashboard/controllers/api/UsersApiController.php
charrondev: I get what you saying but seeing a short method returning a `new Data()` object it makes sense to me. It's not like this is some ancient legacy code here. The whole thing was written in the last 2 years. We should be able to know what are methods return confidently on recent code.

applications/dashboard/controllers/api/AbstractApiController.php
charrondev: So is this giving me "all" permissions? Because that's going to be an immediate issue. My understanding is that were going to put __only__ global permissions here, and that resource specific permissions would come down with the resource, or in a separate API call on the resource.

So if we end up needing category specific permissions we would expand permissions in a request to the permissions.

---

Maybe I put this comment on the wrong line. My thought process is:

- `/users/me` will __always__ have permissions
- The permissions on `/users/me` will __only__ be global permissions.

applications/dashboard/controllers/api/AbstractApiController.php
charrondev: Ah that's what you were talking about! Just update this with that issue once it's created.

library/Vanilla/PermissionsTranslationTrait.php
```diff
@@ -100,6 +101,85 @@ private function consolidatePermissions(array $permissions) {
...
+                'type' => 'global',
+                'permissions' => $format($global)
+            ]
+        ];
+
+        // Step three: add in per-resource permissions, also formatted.
+        foreach ($categories as $categoryID => $permissionSlugs) {
+            // Default category (-1) permissions now fall under an ID of zero (0).
+            $categoryPermissions = [
+                'id' => $categoryID == -1 ? 0 : $categoryID,
```
charrondev: What's the reasoning for treating the -1 here like a 0?

applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -281,16 +306,31 @@ public function index_byNames(array $query) {
...
 
-        $this->schema([], "in")->setDescription("Get information about the current user.");
-        $out = $this->schema($this->getUserFragmentSchema(), "out");
+        $in = $this->schema([], "in")->setDescription("Get information about the current user.");
+        $out = $this->schema(Schema::parse([
+            "userID",
+            "name",
+            "photoUrl",
+            "dateLastActive",
+            "permissions?" => [
```
charrondev:
```suggestion
            "permissions" => [
```

PR #7968, opened 2018-10-24
Add event in theme options view to assist with customizing the page

### LINKED ISSUE #7900: Add fireEvent on themeoptions.php view
Reference: https://github.com/vanilla/vanilla/blob/master/applications/dashboard/views/settings/themeoptions.php#L103

This view has no hooks, I'm currently overwriting it on [Keystone theme](https://github.com/vanilla/themes/blob/feature/keystone/keystone/views/settings/themeoptions.php) to add more options besides color schemes.

It would be great to have a hook on this page.

igraziatto:
Closes #7900 

This view has no hooks, the new core theme [Keystone](https://github.com/vanilla/themes/blob/feature/keystone/keystone/class.keystone.themehooks.php#L76) depends on this event to add more options besides color schemes.

### DIFF
--- a/library/Vanilla/AliasLoader.php
+++ b/library/Vanilla/AliasLoader.php
@@ -23,6 +23,8 @@ class AliasLoader {
      * @inheritdoc
      */
     protected static function provideAliases(): array {
-        return [];
+        return [
+            \Vanilla\Web\Assets\LegacyAssetModel::class => ['AssetModel'],
+        ];
     }
 }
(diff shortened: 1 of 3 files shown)


cache/.htaccess
```diff
@@ -1 +0,0 @@
-Deny from all
```
initvector: This isn't intentional, is it?

charrondev: Good catch. This is why it's best to clear your cache by doing a `composer install`. It will leave behind the .htaccess file.

PR #7972, opened 2018-10-26
Added "react search" to Vanilla, styles for "suggestedTextInput" and "big search" input

slafleche:
These updates are for the KB. I've added "react search" as well as some reusable overwrites for that component and some styles used for the big search we have on the search page of the KB.

Note that this style/layout/functionality is not final, I just wanted to break up my PR in multiple pieces.


library/src/scripts/components/forms/Button.tsx
```diff
@@ -16,6 +16,7 @@ interface IProps extends IOptionalComponentID {
     prefix?: string;
     legacyMode?: boolean;
     onClick?: () => void;
+    onKeyDown?: () => void;
```
charrondev: We seem to have added this to the props here but never do anything with it.

library/src/scripts/components/forms/select/BigSearch.tsx
```diff
@@ -0,0 +1,156 @@
...
+    public static defaultProps = {
+        disabled: false,
+    };
+
+    private id;
+    private prefix = "bigSearch";
+    private searchButtonID;
+    private searchInputID;
+
+    constructor(props) {
```
charrondev: Can we get a type on this?

library/src/scripts/components/forms/select/BigSearch.tsx
```diff
@@ -0,0 +1,156 @@
...
+                backspaceRemovesValue={true}
+            />
+        );
+    }
+
+    public getValue = value => {
+        return value;
+    };
+
+    private BigSearchControl = props => {
```
charrondev: This should be it's own class component. Props deserves an annotation as well, here even if its any. The class comment for it should probably indicate that it for extending ReactSelect.

The primary reason to pull it into its own class component is because it has a callback method that is being defined on every render (`preventFormSubmission`) of this subcomponent.

The other reason is to make the stuff we are giving this ourselves (setQuery) and the its own props coming from react select distinct.

library/src/scripts/components/forms/select/overwrites/ClearIndicator.tsx
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import * as React from "react";
+
+import classNames from "classnames";
+import { ButtonBaseClass } from "@library/components/forms/Button";
+import { t } from "@library/application";
+import { clear } from "@library/components/Icons";
+
+export function ClearIndicator({ children, ...props }) {
```
charrondev: This should also be a class component. The `handleKeyDown` function is being recreated on every render in this setup.

library/src/scripts/components/forms/select/overwrites/DoNotRender.tsx
```diff
@@ -0,0 +1,9 @@
+/*
+ * @author Stéphane LaFlèche <stephane.l@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+export default function ClearIndicator() {
```
charrondev: This className here does not match up with the file name.

library/src/scripts/components/forms/select/overwrites/Option.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+import * as React from "react";
+import classNames from "classnames";
+import DateTime from "@library/components/DateTime";
+import BreadCrumbString from "@library/components/BreadCrumbString";
+
+export default function Option(props: any) {
```
charrondev: Could we give this a more explicit name? Make `SelectOption`?

PR #7979, opened 2018-10-29
Profile Extender - add modify user fields event

cchabilall83:
This even is added to profile extender to allow for modules or plugins to hook in and modify the a users extended fields data or the extended fields themselves.

The use case for this is vanillaforums/advancedbionics#5

### DIFF
--- a/plugins/ProfileExtender/class.profileextender.plugin.php
+++ b/plugins/ProfileExtender/class.profileextender.plugin.php
@@ -236,6 +236,8 @@ class ProfileExtenderPlugin extends Gdn_Plugin {
 
         // Get user-specific data
         $this->UserFields = Gdn::userModel()->getMeta($Sender->Form->getValue('UserID'), 'Profile.%', 'Profile.');
+
+        $this->fireEvent('beforeGetProfileFields');
         // Fill in user data on form
         foreach ($this->UserFields as $Field => $Value) {
             $Sender->Form->setValue($Field, $Value);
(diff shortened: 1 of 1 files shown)


cache/.htaccess
```diff
@@ -1 +0,0 @@
-Deny from all
```
topcheese13: This file should be removed from the PR. It came up for me to today, looks like something new.

plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -236,6 +236,8 @@ private function profileFields($Sender) {
 
         // Get user-specific data
         $this->UserFields = Gdn::userModel()->getMeta($Sender->Form->getValue('UserID'), 'Profile.%', 'Profile.');
+        $this->EventArguments['userFields'] = &$this->UserFields ?? '';
```
topcheese13: You could use the $sender object instead of passing an event argument by reference, objects are always passed by reference as opposed to arrays.

plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -236,6 +236,8 @@ private function profileFields($Sender) {
...
+        $this->fireEvent('modifyUserFields');
```
topcheese13: The event is normally named on the current "event" happening, by that I mean what we are currently doing when firing it. This event name sounds more like what we are doing in the hook.

cchabilall83: beforeGetProfileFields, better? @topcheese13

linc:
Deleting `cache/.htaccess` seems problematic.

plugins/ProfileExtender/class.profileextender.plugin.php
topcheese13: Personally I do like it better.

Issue #7950, opened 2018-10-18
APIv2: Token based authentication doesn't apply permissions granted through ranks


JReko:
When you setup a rank you have the option to attribute the permission of a role through the Rank.

>Users with this rank will gain the permissions of this role.

My setup
---
I have 1 Test user.
He has and only has a standard Member role.
He has a Test rank (manually applicable) that grants the permission of the Administration role.

The problem 
---
If I browse to the API endpoint on the site, i.e. https://dev.vanilla.localhost/api/v2/users.
* It works as expected and returns the expected json.

If I make a curl request using this user's access token I receive this permission error.
```
{
    "message": "You need one of Garden.Users.Add, Garden.Users.Edit, Garden.Users.Delete permissions to do that.",
    "status": 403,
    "permissions": [
        "Garden.Users.Add",
        "Garden.Users.Edit",
        "Garden.Users.Delete"
    ],
    "description": null
* Connection #0 to host dev.vanilla.localhost left intact
}
```

The cause
---
To put it very simply the problem comes from the moment the session is created.

When browsing the site the session is created through:
```
class.session.php:442, Gdn_Session->start()
class.auth.php:59, Gdn_Auth->startAuthenticator()
bootstrap.php:396, require_once()
index.php:22, {main}()
```

When making an API call and using the access token to authenticate yourself, the request goes through the same stack above but since we have no UserID the session is started and we load the permissions of a guest.
```
public function start($userID = false, $setIdentity = true, $persist = false) {
        if (!c('Garden.Installed', false)) {
            return;
        }

        $this->permissions = new Permissions();

        // Retrieve the authenticated UserID from the Authenticator module.
        $userModel = Gdn::authenticator()->getUserModel();
        $this->UserID = $userID !== false ? $userID : Gdn::authenticator()->getIdentity();
        $this->User = false;
        $this->loadTransientKey();

        // Now retrieve user information.
        if ($this->UserID > 0) {
```
```
// Load guest permissions if necessary
        if ($this->UserID == 0) {
            $guestPermissions = $userModel->getPermissions(0);
            $this->permissions->setPermissions($guestPermissions->getPermissions());
        }
```

To assign the permissions through a token based authentication we use a hook

https://github.com/vanilla/vanilla/blob/f841b0f4dbedad56160694884324640acf994afe/applications/dashboard/settings/class.hooks.php#L624
[...]

PR #7981, opened 2018-10-29
Fix token & standard authentication not happening at same time during request

topcheese13:
Issue: https://github.com/vanilla/vanilla/issues/7950

After talking a little with both Ryan and Todd, the goal here was to trigger the token based authentication at the same "time" than standard auth to prevent different behaviours depending on your auth method as explained in the issue linked above. After taking both of your inputs into consideration this is what I came up with.

We add an event in the startAuthenticator method of the Auth class. We then change on what even the checkAccessToken method of the dashboard hooks is being called. Before `gdn_dispatcher_appStartup_handler` ,now `gdn_auth_startAuthenticator_handler`.

I have tested the changes in this PR and the API now works as expected. Let me know what you guys  think @initvector @tburry.

There is also a small casening fix in there.

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -670,11 +670,8 @@ class DashboardHooks extends Gdn_Plugin {
      *
      * @param Gdn_Auth $sender
      */
-    public function gdn_auth_startAuthenticator_handler($sender) {
+    public function gdn_auth_startAuthenticator_handler() {
         $this->checkAccessToken();
-        if (Gdn::session()->isValid()) {
-            $sender->authHandled = true;
-        }
     }
 
     /**
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -666,9 +666,7 @@ class DashboardHooks extends Gdn_Plugin {
     }
 
     /**
-     * Check is we have a valid token associated with the request.
-     *
-     * @param Gdn_Auth $sender
+     * Check if we have a valid token associated with the request.
      */
     public function gdn_auth_startAuthenticator_handler() {
         $this->checkAccessToken();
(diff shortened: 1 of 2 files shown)


applications/dashboard/settings/class.hooks.php
```diff
@@ -663,7 +663,13 @@ public function gdn_dispatcher_appStartup_handler($sender) {
                 redirectTo($url);
             }
         }
+    }
+
+    public function gdn_auth_startAuthenticator_handler($sender) {
         $this->checkAccessToken();
+        if (Gdn::session()->isValid()) {
+            $sender->authHandled = true;
```
initvector: This event property kinda seems redundant. Couldn't I check the result of `Gdn::session()->isValid()`, here and in other areas, to see if authentication had already been handled? It's a very lightweight method.

topcheese13: Maybe I misunderstood this but when I proposed using the isValid method of the session, @tburry wanted a way to invalidate a previous auth if we need it in the future. He can probably provide deeper info.

Once this is settled I will definitely add a doc bloc for this.

applications/dashboard/settings/class.hooks.php
```diff
@@ -663,7 +663,18 @@ public function gdn_dispatcher_appStartup_handler($sender) {
                 redirectTo($url);
             }
         }
+    }
+
+    /**
+     * Hooking on the startAuthenticator event of the auth class.
+     *
+     * @param $sender Auth
```
initvector:
```suggestion
     * @param Gdn_Auth $sender
```

applications/dashboard/settings/class.hooks.php
tburry: I'm going to go back on what I said and agree with @initvector to make things simpler. We already have an improved way of doing this kind of thing in the future so we can just use `Gdn::session()->isValid()` on this for now. Sorry for the doublespeak.

PR #7983, opened 2018-10-29
Updates for KB's Advanced Search

slafleche:
The following updates do not affect Vanilla, they are needed for the KB's advanced search.

- Updated InputTextBlock with a base class, optional label , note and more flexible type for label
- Added help icon
- Added "SelectOne" (which is a react select configured to be like a regular `<select/>`
- Added "Token" (which is a react select configured to have "tokens" as inputs
- Added more overwrites for react select
- Added "radioButtonAsTab" component. 
- Added missing checkbox styles


library/src/scripts/components/forms/select/SelectOne.tsx
```diff
@@ -0,0 +1,127 @@
...
+        this.props.setData(newValue);
+    };
+
+    public render() {
+        const { className, disabled, options, searchable } = this.props;
+
+        /** The children to be rendered inside the indicator. */
+        const componentOverwrites = {
+            IndicatorsContainer: doNotRender,
+            SelectContainer: selectContainer,
```
charrondev: Let's pull this out into a class variable. It doesn't seem to change between renders.

If it did then we could make it a getter, but generally I think it's best to try and keep the render methods as small and straightforward as possible.

library/src/scripts/components/forms/select/SelectOne.tsx
```diff
@@ -0,0 +1,127 @@
...
+            Menu: menu,
+            MenuList: menuList,
+            Option: SelectOption,
+            ValueContainer: valueContainer,
+            Control: controlContainer,
+        };
+
+        const getTheme = theme => {
```
charrondev: This can be pulled out into a method on class.

library/src/scripts/components/forms/select/SelectOne.tsx
```diff
@@ -0,0 +1,127 @@
...
+            return {
+                ...theme,
+                borderRadius: {},
+                borderWidth: 0,
+                colors: {},
+                spacing: {},
+            };
+        };
+
+        const customStyles = {
```
charrondev: This can be pulled out into a property or getter on the class.

library/src/scripts/components/forms/select/SelectOne.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { t } from "@library/application";
+import menuList from "@library/components/forms/select/overwrites/menuList";
+import menu from "@library/components/forms/select/overwrites/menu";
+import selectContainer from "@library/components/forms/select/overwrites/selectContainer";
+import doNotRender from "@library/components/forms/select/overwrites/doNotRender";
+import Paragraph from "@library/components/Paragraph";
+import { IComboBoxOption } from "./BigSearch";
+import SelectOption from "@library/components/forms/select/overwrites/selectOption";
+import { IFieldError } from "@library/@types/api";
+import ErrorMessages from "@dashboard/components/forms/ErrorMessages";
```
charrondev: We're importing something from dashboard into library here.

library/src/scripts/components/forms/select/Tokens.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    private handleOnChange = (newValue: any, actionMeta: any) => {
+        this.props.setAuthor(newValue);
+    };
+
+    public render() {
+        const { className, disabled, options } = this.props;
+
+        /** The children to be rendered inside the indicator. */
+        const componentOverwrites = {
```
charrondev: I'd like to see the same thing as mentioned in the `SelectOne` feedback. Let's pull out everything we can that is not JSX here into class getters, properties, or methods.

library/src/scripts/components/radioButtonsAsTabs/RadioButtonTab.tsx
```diff
@@ -0,0 +1,55 @@
...
+    label: string;
+    className?: string;
+    data: any;
+    defaultTab: any;
+}
+
+/**
+ * Implement what looks like a tab, but what is semantically radio button. To be used in the RadioButtonsAsTabs component
+ */
+// export default class RadioButtonTab extends React.Component<IRadioButtonTab> {
```
charrondev: What's up with this line?

library/src/scripts/contexts/TabContext.tsx
```diff
@@ -0,0 +1,40 @@
...
+    groupID: string;
+    defaultTab: any;
+    childClass: string;
+}
+
+const TabContext = React.createContext<ITabProps>({} as any);
+export default TabContext;
+
+/**
+ * HOC to inject DeviceContext as props.
```
charrondev: This comment seems off.

library/src/scripts/components/forms/ErrorMessages.tsx
```diff
@@ -1,11 +1,12 @@
...
+/*
+ * @author Stéphane LaFlèche <stephane.l@vanillaforums.com>
  * @copyright 2009-2018 Vanilla Forums Inc.
  * @license GPL-2.0-only
  */
 
 import React from "react";
 import classNames from "classnames";
-import { getRequiredID, IRequiredComponentID } from "@library/componentIDs";
+import { getRequiredID, IRequiredComponentID } from "../../componentIDs";
```
alex-mtl: I am not sure if that is best practice to have relative paths.
We can lose it during build and deploy phases

slafleche: Good catch!

PR #7989, opened 2018-10-30
Improve generated quill outline, add smooth scrolling in user content

charrondev:
Implementation for articles https://github.com/vanilla/knowledge/pull/288

## Quill outlines

- Generates "unique" IDs for heading elements every time their value is saved. There can be clashes, but can  improve on this in the future. I try and do a little bit fo parsing to get some nicer looking slugs in the URL (dash-casing, transliteration, lowercasing.) This is done by extending the built in Heading Blot.
- Updates formatting unit tests to reflect this new data/outputted ID.

## Smooth scroll

- Adds a "scroll-behaviour" smooth polyfill.


alex-mtl:
Lets do it on backend when POST initial data. And lets store it in DB: `outline` as well as `excerpt`. We should not generate them on the fly when GET. We should do all those calculations when PUT, POST, PACTH and store it in DB.

slafleche:
@charrondev is this to get reviewed/merged as is, or there will be a scope change for next sprint?

charrondev:
It is to be reviewed as is. In my mind it is complete with these PRs

PR #8004, opened 2018-11-05
Updated KB's editor styles

slafleche:
For non legacy editor (i.e. KB):
- Moved the menu up top and on mobile, we hide the emoji menu and add the paragraph menu up top. 
- Style updates, more minimalistic
- Added positioning props for flyouts
- Updated location picker styles
- Fixed styles for code blocks
- 

For legacy and non legacy editor:
- The icons of the editor have been updated.

Other fixes/improvements;
- Fixed copyright in several SASS partials
- Reorganized icons to be in library and made separate files for groups of icons


plugins/rich-editor/src/scripts/components/toolbars/ParagraphDropDown.tsx
```diff
@@ -0,0 +1,139 @@
...
+    hasFocus: boolean;
+}
+
+/**
+ * Implemented ParagraphDropDown component, this is for mobile
+ */
+export class ParagraphDropDown extends React.PureComponent<IProps, IState> {
+    private quill: Quill;
+    private ID: string;
+    private componentID: string;
```
charrondev: `componentID`, `menuID`, and `buttonID` are all set in the constructor, but are not not used anywhere in the component.

plugins/rich-editor/src/scripts/components/toolbars/ParagraphDropDown.tsx
```diff
@@ -0,0 +1,139 @@
...
+                />
+            </DropDown>
+        );
+    }
+
+    /**
+     * Get the active format for the current line.
+     */
+    private get activeFormatIcon(): JSX.Element {
+        const { activeFormats } = this.props;
```
charrondev: This seem to be duplicated now across the 2 different versions of this. Could we:

- Make it it's own component. Maybe `ParagraphActiveIcon` and it just takes `activeFormats` as a prop. It __shouldn't__ need to be directly connect with the `withEditor` HOC.
- Replace this and the equivalent `activeFormatIcon` in the toolbar version of this.

slafleche: Good call

PR #8005, opened 2018-11-05
Add database support for generic drafts

### LINKED ISSUE #8000: Add new draft table and model

initvector:
Vanilla's [drafts API v2 resource](https://docs.vanillaforums.com/help/apiv2/swagger/#/Drafts/get_drafts) uses a generic format, but performs request translations to map fields to the existing, Vanilla-specific (i.e. discussions and comments) drafts table. The next step in moving towards revising drafts is introduced here: adding a new table for truly generic drafts and a model for managing them. The new database table is based on the API schema, which itself is based off #5535 (and its comments).

### Overview
1. Add the new drafts table. Drafts are being overhauled, slowly, one step at a time. The new drafts table will need to exist alongside the existing Vanilla drafts table, until the next phase of refactoring sees everything moving to the new table structure.
1. Add `Vanilla\Models\DraftModel` to manage data in the new drafts table. Thanks to pipelines, this is a very basic model. However, it does perform some on-the-fly JSON packing/unpacking, which leads us to...
1. Add `Vanilla\Database\Operation\JsonFieldProcessor` for automatically packing and unpacking JSON fields.
1. Add tests for the `JsonFieldProcessor` class.
1. Add ability to simulate select operations in `BasicPipelineModel`, a fixture used in tests.
1. Update comment block on the new base model's `selectSingle` method.

Closes #8000

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -854,7 +854,8 @@ $Construct
 $Construct
     ->table("contentDraft")
     ->primaryKey("contentDraftID")
-    ->column("recordType", "varchar(64)", false, ["index", "index.parentRecord"])
+    ->column("recordType", "varchar(64)", false, ["index", "index.record", "index.parentRecord"])
+    ->column("recordID", "int", true, "index.record")
     ->column("parentRecordID", "int", true, "index.parentRecord")
     ->column("attributes", "text")
     ->column("insertUserID", "int", false, "index")
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -851,6 +851,18 @@ $Construct
     ->column('UpdateIPAddress', 'ipaddress', true)
     ->set($Explicit, $Drop);
 
+$Construct
+    ->table("contentDraft")
+    ->primaryKey("contentDraftID")
+    ->column("recordType", "varchar(64)", false, ["index", "index.parentRecord"])
+    ->column("parentRecordID", "int", true, "index.parentRecord")
+    ->column("attributes", "text")
+    ->column("insertUserID", "int", false, "index")
+    ->column("dateInserted", "datetime")
+    ->column("updateUserID", "int")
+    ->column("dateUpdated", "datetime")
+    ->set($Explicit, $Drop);
+
 // If the AllIPAddresses column exists, attempt to migrate legacy IP data to the UserIP table.
 if (!$captureOnly && $AllIPAddressesExists) {
     $limit = 10000;
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -853,7 +853,7 @@ $Construct
 
 $Construct
     ->table("contentDraft")
-    ->primaryKey("contentDraftID")
+    ->primaryKey("draftID")
     ->column("recordType", "varchar(64)", false, ["index", "index.record", "index.parentRecord"])
     ->column("recordID", "int", true, "index.record")
     ->column("parentRecordID", "int", true, "index.parentRecord")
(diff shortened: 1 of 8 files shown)


library/Vanilla/Models/DraftModel.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * DraftModel constructor.
+     *
+     * @param Gdn_Session $session
+     */
+    public function __construct(Gdn_Session $session) {
+        parent::__construct("contentDraft");
+        $this->session = $session;
+
+        $dateProcessor = new CurrentDateFieldProcessor();
```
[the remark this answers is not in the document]
initvector: > Is there a particular reason we are instantiating these and not DIing them? 

We *could* use DI for them, but it doesn't seem necessary for these. They're very simple objects. If they had more complex dependencies of their own (or a complex chain of dependencies), or advanced setup routines (all the setup here is class-specific), utilizing DI would make a lot more sense to me.

> CurrentDateFieldProcess (time related), feels like something you might want to stub out particularly in a test.

The fields/values generated by these classes are already unit tested, separately. We shouldn't need to validate them again as part of other component tests.

applications/dashboard/settings/structure.php
```diff
@@ -851,6 +851,19 @@
...
+    ->column("recordType", "varchar(64)", false, ["index", "index.record", "index.parentRecord"])
+    ->column("recordID", "int", true, "index.record")
+    ->column("parentRecordID", "int", true, "index.parentRecord")
+    ->column("attributes", "text")
+    ->column("insertUserID", "int", false, "index")
+    ->column("dateInserted", "datetime")
+    ->column("updateUserID", "int")
+    ->column("dateUpdated", "datetime")
+    ->set($Explicit, $Drop);
+
```
linc: So:

* I understand our API-first methodology expects camelCase parameter names.
* I understand we're currently shimming every API endpoint to transform our CapitalCase database structure into a camelCase schema.

But:

* We have no plan or priority for actually migrating our database to the new standard.
* There is little-to-no relevant performance overhead incurred by the shimming.
* There is little technical debt to putting the shim in new endpoints. (They can just be deleted when we move with the rest of them.)
* There are no extensive plans for additional core tables beyond this one presently.

Those are the technical facts on the ground. Now, enter opinion:

* You think it's more efficient to move to a new standard as soon as it's settled to minimize future code changes.
* I think deviating from established norms has real communication & training costs, both internally and in the open source community. I think those costs outweigh the benefits of halfway messing with database capitalization, especially when we have no plans to update our developer docs to reflect these changes or their rationale.

This change isn't free, and I'm not keen to absorb the cost. So the toll is gonna be a developer doc explaining our database naming conventions & why they're migrating in the public dev docs.

linc: For the purposes of this sprint, I will accept an IOU in the form of a docs repo issue, provided it gets in before end of year.

initvector: Vanilla's code is *supposed* to be camel-cased. The odd man out is currently our database, because changing that is a tremendous endeavor. Not only would the database need to be updated, but every piece of code that referenced these fields would also need to be updated to avoid using the capital-case names. Column names aren't case-sensitive in MySQL, but referencing them in PHP would be.

I don't agree that we should continue perpetuating that standard, purely for the sake of consistency. It requires us to continue using mixed casing schemes in our code, even for new components. The API is a good example, where we're having to adjust casing to present the desired output, but this applies to just about anything that uses these tables. If we know this is the direction we want to go in, and apply it now, then it's one less thing to have to deal with when we make the switch. Again, this includes API endpoints, as well as any models, controllers, etc. that reference these fields with the legacy capital-case format.

linc: PAY THE TAX

PR #8006, opened 2018-11-05
Implement Mutual exclusion functionality on the Vanilla's Base Model

cchabilall83:
This adds the ability to lock resources by assigning a cache key to the process id trying to access it.  If there are competing processes for the same resources, the first one in will receive a master key that is cached.  If the cache key exists, an attempt to access the resource will fail.  Once the the process is complete, it can release the resource by calling Gdn_Model::releaseCacheLock.

Something similar was done on the categories model 
https://github.com/vanilla/vanilla/blob/9e495f412581c19aa21a2eccd7b0f4279f09b7c6/applications/vanilla/models/class.categorymodel.php#L549

This implementation will be a standard way to implement mutual exclusion across Vanilla.

### DIFF
--- a/library/core/class.model.php
+++ b/library/core/class.model.php
@@ -991,12 +991,12 @@ class Gdn_Model extends Gdn_Pluggable {
      * @return bool $keyHasBeenRelease Whether a master key has been released.
      */
     protected function releaseCacheLock($lockKey = ''):bool {
-        $keyHasBeenRelease = false;
+        $keyReleased = false;
 
         if (isset($lockKey)) {
-            $keyHasBeenRelease = Gdn::cache()->remove($lockKey);
+            $keyReleased = Gdn::cache()->remove($lockKey);
         }
 
-        return $keyHasBeenRelease;
+        return $keyReleased;
     }
 }
--- a/library/core/class.model.php
+++ b/library/core/class.model.php
@@ -967,17 +967,22 @@ class Gdn_Model extends Gdn_Pluggable {
      *
      * @param string $lockKey Cache key to be assigned.
      * @param int $gracePeriod Period of time the key will stay valid.
-     * @return bool $hasMasterKey Whether a master key has been assigned.
+     * @return bool Whether a master key has been assigned.
      */
-    protected static function buildCacheLock(string $lockKey, int $gracePeriod = 60):bool {
-        $hasMasterKey = null;
+    protected static function buildCacheLock(string $lockKey, int $gracePeriod = 60): bool {
+        static $hasMasterKey = null;
+        $masterKey = null;
 
         if (is_null($hasMasterKey)) {
             // First instance will receive the $master key and lock the record.
             $instanceKey = getmypid();
-            $masterKey = Gdn::cache()->add($lockKey, $instanceKey, [
+            $keyAdded = Gdn::cache()->add($lockKey, $instanceKey, [
                 Gdn_Cache::FEATURE_EXPIRY => $gracePeriod
             ]);
+
+            if ($keyAdded) {
+                $masterKey = Gdn::cache()->get($lockKey);
+            }
             $hasMasterKey = ($instanceKey == $masterKey);
         }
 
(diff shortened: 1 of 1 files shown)


library/core/class.model.php
```diff
@@ -961,4 +961,42 @@ public static function editContentTimeout($data, &$timeLeft = 0) {
...
+    /**
+     * Releases a locked resource so that it can be used again.
+     *
+     * @param string $lockKey Cache key to be assigned.
+     * @return bool $keyHasBeenRelease Whether a master key has been released.
+     */
+    protected function releaseCacheLock($lockKey = ''):bool {
+        $keyReleased = false;
+
+        if (isset($lockKey)) {
```
initvector: You're checking if a parameter to your function has been set? I think the only time this would ever resolve to `false` is if you're passing in a `null` for `$lockKey`, but your parameter docs don't say that's supported (and I don't think you want it to be).

library/core/class.model.php
```diff
@@ -961,4 +961,42 @@ public static function editContentTimeout($data, &$timeLeft = 0) {
...
+     * @param int $gracePeriod Period of time the key will stay valid.
+     * @return bool $hasMasterKey Whether a master key has been assigned.
+     */
+    protected static function buildCacheLock(string $lockKey, int $gracePeriod = 60):bool {
+        $hasMasterKey = null;
+
+        if (is_null($hasMasterKey)) {
+            // First instance will receive the $master key and lock the record.
+            $instanceKey = getmypid();
+            $masterKey = Gdn::cache()->add($lockKey, $instanceKey, [
```
initvector: Are you sure it'll work like this? It looks like you're expecting it to return the cached value, but I don't think that'll be the case. For example, the `Gdn_Memcached` caching adapter's `add` method returns boolean values. You're comparing that result, stored as `$masterKey`, with the result of `getmypid`, which would be an integer. In PHP, this might coincidentally work on the first attempt, because `1234 == true` evaluates to `true` in PHP. Subsequent calls, even in the same process, would see the value of `$masterKey` being `false`, which would then lead to `$hasMasterKey` also being `false`. This is because the caching adapter's `add` method adds new cached values under a key __if it doesn't already exist__ and returns a success state *or does nothing and responds with a failure state if the cache key already exists.*

cchabilall83:
Should we do some like this

$hasMasterKey = Gdn::cache()->exists

if (!$hadMasterKey) {
// add cachekey
}

initvector:
> Should we do some like this

The point of the mutex here is to lock other processes (threads) out of accessing a particular resource. The thread that locked it should still be able to access it. I think your current implementation is *close*, but it looks like it might be based on an errant assumption. When you create the lock, you're caching the ID of the current process. When you check it, you'll want to see if the cached value (if there is one) matches your current process's ID. If it does, that process has the "master key". If it doesn't, it's a different process than what "locked" the resource, so it does not have that "master key". Only testing whether or not a cached value exists is not enough.

library/core/class.model.php
```diff
@@ -961,4 +961,42 @@ public static function editContentTimeout($data, &$timeLeft = 0) {
...
+     * Locks a resource so that it can only be accessed one at a time.
+     *
+     * @param string $lockKey Cache key to be assigned.
+     * @param int $gracePeriod Period of time the key will stay valid.
+     * @return bool $hasMasterKey Whether a master key has been assigned.
+     */
+    protected static function buildCacheLock(string $lockKey, int $gracePeriod = 60):bool {
+        $hasMasterKey = null;
+
+        if (is_null($hasMasterKey)) {
```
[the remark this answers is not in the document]
linc: `$hasMasterKey` needs to be static. I said a dumb during one of our reviews.

library/core/class.model.php
linc: This is an error in our original example code. Thus proving you need to double-check assumptions when copying code.

library/core/class.model.php
```diff
@@ -967,17 +967,22 @@ public static function editContentTimeout($data, &$timeLeft = 0) {
...
         if (is_null($hasMasterKey)) {
             // First instance will receive the $master key and lock the record.
             $instanceKey = getmypid();
-            $masterKey = Gdn::cache()->add($lockKey, $instanceKey, [
+            $keyAdded = Gdn::cache()->add($lockKey, $instanceKey, [
                 Gdn_Cache::FEATURE_EXPIRY => $gracePeriod
             ]);
+
+            if ($keyAdded) {
+                $masterKey = Gdn::cache()->get($lockKey);
```
initvector: If the key was successfully added, meaning it didn't already exist, then we already know we have the current "master key", right?

PR #8007, opened 2018-11-06
Add DateFilterSphinxSchema to help with generating date range values for Sphinx

alex-mtl:
Relates to: https://github.com/vanilla/knowledge/issues/297

Blocker for: https://github.com/vanilla/knowledge/pull/310


library/Vanilla/DateFilterSchema.php
```diff
@@ -314,4 +314,67 @@ public static function dateFilterField($field, array $dateData) {
...
+                        }
+                        break;
+                    case '=':
+                        $result['startDate'] = $dates[0];
+                        $result['endDate'] = $dates[0];
+                        break;
+                }
+            }
+        } else {
+            throw new Exception('Invalid data supplied to dateFilterField');
```
initvector:
```suggestion
            throw new Exception('Invalid data supplied to dateFilterRange');
```

tests/Library/Vanilla/DateFilterSchemaTest.php
```diff
@@ -288,4 +298,95 @@ public function provideDateFilterFields() {
             ],
         ];
     }
+
+    /**
+     * Provider data for testDateFilterRange.
+     */
+    public function provideDateFilterRange() {
+        $schema = new DateFilterSchema();
```
initvector: If you're going to use an object like this, it's probably better to move it into your test, instead of your data provider. You can then pass whatever parameters to your test function and hand them down to the `validate` method while keeping everything local to `testDateFilterRange` (avoiding the potential for object pollution).

library/Vanilla/DateFilterSchema.php
```diff
@@ -314,4 +314,67 @@ public static function dateFilterField($field, array $dateData) {
...
+     *
+     * @throws Exception If dateData structure is wrong throw exception.
+     *
+     * @return array Structure: DateTime startDate, DateTime endDate, Boolean inclusive
+     */
+    public static function dateFilterRange(array $dateData) {
+        $validOperators = ['=', '>', '<', '>=', '<=', '[]', '()'];
+        $result = [
+            'startDate' => (new \DateTime())->setDate(1970, 1, 1),
+            'endDate' => (new \DateTime())->setDate(2100, 12, 31), // '2100-12-12'
```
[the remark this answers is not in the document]
alex-mtl: Because Sphinx setFilterRange method accepts only integers

http://sphinxsearch.com/docs/current/api-func-setfilterrange.html

so we need to send min and max value anyway.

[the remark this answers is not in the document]
initvector: If this reasoning is specific to Sphinx, shouldn't it be implemented in the code where Sphinx is being used? This looks like a general-purpose method.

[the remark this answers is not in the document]
alex-mtl: This whole class pretend to be "general-purpose". Adam did ask to use it since it is general.
And this class has `dateFilterFiled` method
https://github.com/vanilla/vanilla/blob/ad52ea99664bea5add6dc004804e38275a4dd45e/library/Vanilla/DateFilterSchema.php#L270?

which looks like general-purpose method, but it is not. It is **SQL WHERE part builder**

IMHO I don't break any OOP logic here.

library/Vanilla/DateFilterSchema.php
```diff
@@ -314,4 +314,67 @@ public static function dateFilterField($field, array $dateData) {
...
+     * return a structure: startDate, endDate, exclude flag.
+     * Useful for Sphinx setFilterRange
+     *
+     * @param mixed $dateData The decoded and validated date data.
+     *
+     * @throws Exception If dateData structure is wrong throw exception.
+     *
+     * @return array Structure: DateTime startDate, DateTime endDate, Boolean inclusive
+     */
+    public static function dateFilterRange(array $dateData) {
```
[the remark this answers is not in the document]
alex-mtl: Done.

library/Vanilla/DateFilterSchema.php
alex-mtl: Done.

tests/Library/Vanilla/DateFilterSchemaTest.php
alex-mtl: I've found `new DateFilterSchema();` almost in every method in this file. I don't see how my new method is different from others.

library/Vanilla/DateFilterSchema.php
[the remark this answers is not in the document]
initvector: The doc block on `dateFilterField` says:

> If the parameter value is a valid date filter value, return an array of query conditions.

To me, that says this function was built for SQL queries. Conversely, the doc block on this method says:

> If the parameter value is a valid date filter value, return a structure: startDate, endDate, exclude flag. Useful for Sphinx setFilterRange

This does not indicate, to me, that this method has behavior specific for Sphinx. It says the result could be used for Sphinx requests. My suggestion here is to pull out the stuff specific to Sphinx (e.g. the arbitrary default start/end dates) and add those where you actually need them (e.g. when making a Sphinx request). Keeping them in here also means I can't provide my own defaults, wherever I'm invoking this method, because the result will always have a valid start/end date.

tests/Library/Vanilla/DateFilterSchemaTest.php
initvector: It isn't about not instantiating a new `DateFilterSchema` instance. I'm saying you definitely should, but it should be moved to your test, instead of your data provider. Right now, you have one object instance being shared across all your tests, because you're putting it in your data provider. They do two very different things. If you put it in your test, you wouldn't have the potential for object pollution that you do now, since the object will be new with each test.

For example, an array row might be changed to:

```php
'Equal (Date)' => [
                '2017-01-01',
                [
                    'startDate' => new DateTimeImmutable('2017-01-01 00:00:00'),
                    'endDate' => new DateTimeImmutable('2017-01-01 23:59:59'),
                    'exclude' => false
                ]
            ]
```

And the test method would become:

```php
    public function testDateFilterRange(string $dateString, array $expectedResult) {
        $schema = new DateFilterSchema();
        $data = $schema->validate($dateString);
        $result = DateFilterSchema::dateFilterRange($data);
        $this->assertEquals($expectedResult, $result);
    }
```

library/Vanilla/DateFilterSphinxSchema.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @throws Exception If dateData structure is wrong throw exception.
+     *
+     * @return array Structure: DateTime startDate, DateTime endDate, Boolean inclusive
+     */
+    public static function dateFilterRange(array $dateData): array {
+        $validOperators = ['=', '>', '<', '>=', '<=', '[]', '()'];
+        $result = [
+            'startDate' => null,
+            'endDate' => null,
```
initvector: When I'd asked about nulling these values out, when they were in `DateFilterSchema`, you said: 

> Because Sphinx setFilterRange method accepts only integers [...] so we need to send min and max value anyway.

Is that not the case? My only real objection to this method before was that it had Sphinx-specific stuff in it, but it looks like you've removed that by nulling these fields. If you still want to keep this method in a separate class, that's cool too, but I want to make sure these are valid values.

alex-mtl: `null`s are not valid values, but I handle them on Knowledge side

PR #8012, opened 2018-11-09
Updates for KB Drafts

slafleche:
These changes are for the KB Drafts page. The only impact on Vanilla should be some minor margin fixes on embeded urls.

To see drafts, go to /kb/drafts. Note the links won't work, and you can't really delete, but the modal shows up.

- Fixes for focus management of modals
- Fixed class prop for PopoverController
- Disabled all inputs in editor while we're in a loading state
- Moved all embed styles to Library (there were a few varaibles in coreVaraibles that didn't belong there and when I moved them into the embed varaibles, i noticed I had to move them


charrondev:
One weird thing I'm noticing with the confirmation modal is that it is not fixed to the viewport. It seems to be positioned relative to the document body. Eg. If I open a confirmation modal a bit down the screen it will be offset by however much the body was scrolled.

slafleche:
I don't see that bug @charrondev. Anyways, that view is in a full screen modal now, so it shouldn't be a problem anymore.

PR #8015, opened 2018-11-12
Extend upload media endpoint to accept different upload types

alex-mtl:
Extend upload media endpoint to accept different extensions different than images

Fixes: https://github.com/vanilla/knowledge/issues/303

**Note:** Does not cover "external resource" functionality

### DIFF
--- a/applications/dashboard/controllers/api/MediaApiController.php
+++ b/applications/dashboard/controllers/api/MediaApiController.php
@@ -17,7 +17,7 @@ use Vanilla\UploadedFileSchema;
  */
 class MediaApiController extends AbstractApiController {
     const TYPE_IMAGE = 'image';
-    const TYPE_ANY = 'any';
+    const TYPE_OTHER = 'other';
 
     /** @var Schema */
     private $idParamSchema;
@@ -302,7 +302,7 @@ class MediaApiController extends AbstractApiController {
     public function post(array $body) {
         $this->permission('Garden.Uploads.Add');
         switch ($body['type']) {
-            case self::TYPE_ANY:
+            case self::TYPE_OTHER:
                 $uploadExtensions = $this->config->get('Garden.Upload.AllowedFileExtensions', []);
                 break;
             case self::TYPE_IMAGE:
@@ -319,7 +319,7 @@ class MediaApiController extends AbstractApiController {
                 'description' => 'The upload type.',
                 'enum' => [
                     self::TYPE_IMAGE,
-                    self::TYPE_ANY,
+                    self::TYPE_OTHER,
                 ]
             ]
         ],'in')->setDescription('Add a media item.');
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/api/MediaApiController.php
```diff
@@ -158,6 +165,11 @@ public function get($id) {
         }
 
         $row = $this->normalizeOutput($row);
+        $ext = pathinfo($row['url'], PATHINFO_EXTENSION);
+        if (!in_array($ext, array_keys(ImageResizer::getExtType()))) {
```
initvector: Removing the width and height fields should probably go in the `normalizeOutput` method. Right now, these fields are still present in post responses.

applications/dashboard/controllers/api/MediaApiController.php
```diff
@@ -292,15 +304,36 @@ public function normalizeOutput(array $row) {
...
+                $uploadExtensions = self::EXTENSIONS_ARCHIVE;
+                break;
+            case self::TYPE_IMAGE:
+            default:
+                $uploadExtensions = array_keys(ImageResizer::getExtType());
+        }
+
         $uploadSchema = new UploadedFileSchema([
-            'allowedExtensions' => array_keys(ImageResizer::getExtType())
+            'allowedExtensions' => $uploadExtensions
```
initvector: Vanilla has a `Garden.Upload.AllowedFileExtensions` config that should hold an array of all extensions a site allows for uploads. Now that we're adding more extensions to this endpoint, it might make sense to get an intersection of the two arrays to ensure only the configured extensions are allowed.

applications/dashboard/controllers/api/MediaApiController.php
```diff
@@ -292,15 +304,36 @@ public function normalizeOutput(array $row) {
...
         ]);
 
         $in = $this->schema([
             'file' => $uploadSchema,
             'type:s' => [
                 'description' => 'The upload type.',
-                'enum' => ['image']
+                'default' => self::TYPE_IMAGE,
```
initvector: It would make more sense to me for the client to be required to tell us what type they're uploading, instead of making an assumption if they don't specify.

applications/dashboard/controllers/api/MediaApiController.php
initvector: Wouldn't you still want to perform an intersection of allowed extensions? If a site doesn't allow GIFs, and a client submits a request to this endpoint with a type of "image", wouldn't they still be able to upload a GIF?

applications/dashboard/controllers/api/MediaApiController.php
```diff
@@ -16,6 +16,8 @@
  * API Controller for `/media`.
  */
 class MediaApiController extends AbstractApiController {
+    const TYPE_IMAGE = 'image';
+    const TYPE_ANY = 'any';
```
initvector: What do you think about changing "any" to "other"? The word "any" would, literally, include images. The word "other" would (hopefully) convey the user is choosing to upload something that doesn't match our existing types.

Additionally, I liked your previous implementation of specific type-based "buckets" of extensions. It would let us to grow specific types, allowing for custom input/output handling, while making users be more explicit in what they're sending. It'd also be easier to expand on those types, instead of having to limit the scope of an umbrella "any"/"other" type, later.

Ultimately, it's your decision!

PR #8019, opened 2018-11-12
Add additional input components

charrondev:
### MultiUserInput

- I've taken what was the `mentionReducer`, `mentionActions`, and `MentionTrie` and refactored them into `UserSuggestionModel` and `UserSuggestionActions`, `SuggestionTrie`. The rich editor's mentions have been updated to use the new model.
- Our `Tokens` input needed updates to make this possible. It was not used anywhere else.

### DateInput

- Attempts to use native input on iOS and Android (the browsers there have nice built in date pickers).
- Everywhere else gets a custom date picker.

**New dependencies**

Adds 'react-day-picker' and `moment` as dependencies.

**Accessibility**
Accessibility is limited on the custom date picker. It appears as a plain text input. I attempted to still put a type="date" on the input but it is impossible to fully unstyle the browsers extra paddings and controls when that is placed, even with prefixes.

Instead the picker appears as a plain text input, with a label indicating the format requested.

### QueryString

This component automatically persists its props into the query string.

### SmartLink

Component that uses a context to determine whether to make a normal link or a react router link.

## Other Changes

- Pulled a more generic `InputBlock` component out of `InputTextBlock`.
- Changed the DeviceProvider's event listener to be throttled instead of debounced.
- Add query string serialization on get requests with `ReduxActions.dispatchApi`.
- Adds utility function for guessing the user's OS.


library/src/scripts/components/forms/Checkbox.tsx
```diff
@@ -75,4 +75,12 @@ export default class Checkbox extends React.Component<IProps, IState> {
             </label>
         );
     }
+
+    private get checked(): boolean {
```
slafleche: Is this necessary? It's simple enough to pass `{true}` instead of `{"true"}`.

library/src/scripts/components/navigation/SmartLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * Eg.
+ * Context = https://test.com/root
+ * To = https://test.com/root/someUrl/deeper/nested
+ * Result = /root/someUrl/deeper/nested (react router navigation)
+ *
+ * Context = https://test.com/otherRoot
+ * To = https://test.com/root/someUrl/deeper/nested
+ * Result = https://test.com/root/someUrl/deeper/nested (full refresh)
+ */
+export default function SmartLink(props: IProps) {
```
slafleche: Hah... I have a similar component in my PR too. We'll use this one.

PR #8023, opened 2018-11-13
KB Header

slafleche:
Short version: Adds components for KB header. Note that the dropdown components and responsive styles will be separate.

Long version:

- Added ConditionalWrap component to make it easier to conditionallyC wrap a component
- Created FlexSpacer component for aligning asymmetric flexed elements
- Updated Popover controller to pass up the open state
- Updated language dropdown to add "lang" on options
- Added Vanilla Header Component
- Added MeBox component
- Fixed bad classname in SearchBar and added some extra props to make it more flexible
- Updated FrameFooter to be more flexible with children type
- Updated FrameHeader to allow for non modal use
- Added CompactMenu (for mobile, WIP)
- Added CompactSearch (for mobile, WIP)
- Added HeaderLogo
- Added MessagesDropdown (for MeBox)
- Added NotificationsDropdown (for MeBox)
- Added PageTitleDropdown (for mobile, WIP)
- Added UserDropdown Component
- Added UserPhoto Component
- Added VanillaHeaderNav
- Added VanillaHeaderNavItem
- Fixed some class and variable names


library/src/scripts/components/mebox/pieces/CompactSearch.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { search } from "@library/components/icons/header";
+import { uniqueIDFromPrefix } from "@library/componentIDs";
+import AsyncCreatableSelect from "react-select/lib/AsyncCreatable";
+import SearchOption from "@library/components/search/SearchOption";
+
+export interface ICompactSearchProps {
+    className?: string;
+    placeholder?: string;
+    open: boolean;
+    openSearch: () => void;
```
charrondev: We should name these more similarly to how normal event listener props are named. Eg. `onOpenSearch`, `onCloseSearch()`.

library/src/scripts/components/mebox/pieces/CompactSearch.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * Implements Compact Search component for header
+ */
+export default class CompactSearch extends React.Component<ICompactSearchProps> {
+    private id;
+    private ref = React.createRef<AsyncCreatableSelect<any>>();
+
+    public constructor(props) {
```
charrondev: You can remove this constructor and do just do the property initialization. Eg.

```ts
private id = uniqueIDFromPrefix("compactSearch");
```

library/src/scripts/components/mebox/pieces/DropDownMessageList.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Implements Navigation component for header
+ */
+export default class DropDownMessageList extends React.Component<IVanillaHeaderNavProps> {
+    public render() {
+        const count = this.props.children.length;
+        const content = this.props.children.map((item, key) => {
+            return (
+                <React.Fragment key={`dropDownMessageList-${key}`}>
```
charrondev: With a single item these fragments should be unnecessary. You can just put the key on `DropDownMessage` directly.

library/src/scripts/components/mebox/pieces/VanillaHeaderNav.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/**
+ * Implements Navigation component for header
+ */
+export default class VanillaHeaderNav extends React.Component<IVanillaHeaderNavProps> {
+    public render() {
+        const content = this.props.children.map((item, key) => {
+            return (
+                <React.Fragment key={`headerNavItem-${key}`}>
```
charrondev: This fragment can be removed and the key put directly on the `VanillaHeaderNavItem`. Key's also don't need readable names or to be unique across the whole application. They're purely for react's internal use and only need to be unique within their own iteration.

library/src/scripts/components/mebox/pieces/VanillaHeaderNav.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * @license GPL-2.0-only
+ */
+
+import * as React from "react";
+import classNames from "classnames";
+import VanillaHeaderNavItem, { IVanillaHeaderNavItemProps } from "./VanillaHeaderNavItem";
+
+export interface IVanillaHeaderNavProps {
+    className?: string;
+    children: IVanillaHeaderNavItemProps[];
```
charrondev: Same deal with naming children. `navItems` maybe?

library/src/scripts/components/mebox/pieces/VanillaHeaderNavItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+    linkClassName?: string;
+    linkContentClassName?: string;
+}
+
+interface IProps extends IVanillaHeaderNavItemProps, RouteComponentProps<{}> {}
+
+/**
+ * Implements Navigation item component for header
+ */
+export class VanillaHeaderNavItem extends React.Component<IProps> {
```
charrondev: Can we get an issue for the spec on when these are supposed to marked as active? It's a bit of a mess in vanilla's current nav.

I'm thinking sub-path matching should work. Eg. If this link is to `/kb` and I'm on `/kb/articles/something` it should be active? Maybe we need to get more granular and offer a mode of what marks it as active.

library/src/scripts/components/search/SearchOption.tsx
```diff
@@ -0,0 +1,60 @@
...
+import * as React from "react";
+import DateTime from "../DateTime";
+import BreadCrumbString from "../BreadCrumbString";
+import { OptionProps } from "react-select/lib/components/Option";
+import { SelectOption } from "../forms/select/overwrites";
+import classNames from "classnames";
+import { IComboBoxOption } from "@library/components/forms/select/SearchBar";
+
+interface IProps extends OptionProps<any> {
+    data: IComboBoxOption<any>;
```
charrondev: Just removing the interface does not improve this being moved to library. It's actually worse now because we have no idea what the data is. We'll need to either:

1. Make the extra data for this a more generic shape and map our multiple possible search endpoints into it.
2. Or Make the search option component pluggable so that KB can supply a different item than forum.

library/src/scripts/components/forms/select/SearchBar.tsx
```diff
@@ -39,6 +41,9 @@ interface IProps extends IOptionalComponentID {
     isLoading?: boolean;
     onSearch: () => void;
     optionComponent?: React.ComponentType<OptionProps<any>>;
+    getRef?: any;
```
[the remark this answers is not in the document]
slafleche: That ref is for the "compact search" for the header... Why don't we meet next sprint about that component. It might be best to refactor a bit next sprint.

library/src/scripts/components/mebox/pieces/VanillaHeaderNav.tsx
slafleche: I want to differentiate them from the panel nav, but opted for a shorter: "IHeaderNav"

library/src/scripts/components/mebox/pieces/VanillaHeaderNavItem.tsx
slafleche: https://github.com/vanilla/knowledge/issues/329

library/src/scripts/components/search/SearchOption.tsx
slafleche: https://github.com/vanilla/knowledge/issues/330

PR #8024, opened 2018-11-14
Add attachment endpoint to media API v2 resource

### LINKED ISSUE #8001: Add ability to attach an upload to a row under the media API v2 resource

initvector:
Uploads made using the media API v2 resource are automatically given a "foreign type" of "embed" and a "foreign ID" value set to the current user's ID. There is currently no way to manually update a media item's attachment status using the API.

This update introduces the media "attachment" endpoint, which allows users the ability to edit the attachment status of media items. A patch request made to /api/v2/media/1/attachment will allow you to update the "foreign type" and "foreign ID" of a media item, thus updating its attachment status.

### Testing

1. Create a new upload using the media API v2 endpoint.
1. Grab the ID of the new upload.
1. Make a patch request to /api/v2/media/{id}/attachment using the `foreignType` and `foreignID` fields in the body of your request, where "{id}" is the upload ID grabbed in step two.
1. Get the media item using the same ID from step two (a get request to /api/v2/media/{id}). The `foreignType` and `foreignID` fields should be updated.

An integration test has been added to `MediaTest`.

Closes #8001

### DIFF
--- a/applications/dashboard/controllers/api/MediaApiController.php
+++ b/applications/dashboard/controllers/api/MediaApiController.php
@@ -118,6 +118,21 @@ class MediaApiController extends AbstractApiController {
     }
 
     /**
+     * Given a media row, verify the current user has permissions to edit it.
+     *
+     * @param array $row
+     */
+    private function editPermission(array $row) {
+        $insertUserID = $row["InsertUserID"] ?? null;
+        if ($this->getSession()->UserID === $insertUserID) {
+            // Make sure we can still perform uploads.
+            $this->permission("Garden.Uploads.Add");
+        } else {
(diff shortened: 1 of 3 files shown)


alex-mtl:
Not sure if that is ok `foreignType` is not validated. I think it should be `enum` of allowed foreign tables/objects whatever, but we need to limit it somehow.

Another concern is attachment implementation. We did discuss few times and I remember the idea is to have possibility to attach same media to many "attachment" objects. For example if we have many revisions its content can have many references to the same media object. and authors do not need to upload same media file many times. Current implementation just make it 1-1 media-attachment relationship.

PR #8036, opened 2018-11-15
Add interface for validating media attachment updates

initvector:
This update adds an interface to standardize validation of a media item's attachment to a resource by the current user. This is currently event-driven, so this interface attempts to ensure all classes that need to validate media attachments do so with a standardized, consistent, strongly-typed hook method.

### DIFF
--- a/applications/dashboard/controllers/api/MediaApiController.php
+++ b/applications/dashboard/controllers/api/MediaApiController.php
@@ -4,6 +4,7 @@
  * @license GPL-2.0-only
  */
 
+use Garden\EventManager;
 use Garden\Schema\Schema;
 use Garden\Web\Exception\ClientException;
 use Garden\Web\Exception\NotFoundException;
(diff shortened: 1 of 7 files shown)


library/Vanilla/Utility/AttachmentValidatorInterface.php
```diff
@@ -0,0 +1,24 @@
...
+interface AttachmentValidatorInterface {
+    /**
+     * Verify the current user can attach a media item to a specific resource.
+     *
+     * @param bool $canAttach Current state of whether or not a user can attach to this resource.
+     * @param string $foreignType Target resource type/table (e.g. comment, discussion).
+     * @param int $foreignID Unique numeric ID of the resource being attached to.
+     * @return bool Whether or not the current user should be able to attach a media item to this specific resource.
+     */
+    public function canAttachMedia_handler(bool $canAttach, string $foreignType, int $foreignID): bool;
```
alex-mtl: >  * Verify the current user can attach a media item to a specific resource.
>  @param bool $canAttach Current state of whether or not a user can attach to this resource.
I can't get the meaning of this parameter.
If this function is to **_Verify the current user can attach a media item to a specific resource._** - why do we need that flag to be passed as a parameter.
I suspect that we need it for some reason but I think doc block should have more clear  method description.

library/Vanilla/Utility/AttachmentValidatorInterface.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * An interface for classes intending to validate a media's attachment status.
+ *
+ * @package Vanilla\Utility
+ */
+interface AttachmentValidatorInterface {
+    /**
+     * Verify the current user can attach a media item to a specific resource.
+     *
+     * @param bool $canAttach Current state of whether or not a user can attach to this resource.
```
charrondev:
```suggestion
      * @param bool $canAttach Current state of whether or not a user can attach to this resource. For a handler to have no effect, it can return the current `$canAttach` state. This result will be fired through all handlers in sequence by the event manager.
```

charrondev:
This PR is now defunct.

PR #8037, opened 2018-11-15
Improvements to MeBox Notifications and Messages drop dows.

slafleche:
Reworked the Notifications and Messages dropdown a bit. I noticed there was some missing data for the messages. 

Also fixed some bad capitalization in a SASS partial.


slafleche:
@charrondev I renamed some of my classes and interfaces for clarity. It was a bit confusing.

I implemented most of your changes, but I can't get rid of the `as any` in VanillaHeader. it gives me a typescript error. Did I miss anything or do we need to keep it?

charrondev:
> I implemented most of your changes, but I can't get rid of the as any in VanillaHeader. it gives me a typescript error. Did I miss anything or do we need to keep it?

We can keep it for now, and Ryan can resolve it when he puts in the actual data.

PR #8039, opened 2018-11-19
Add validation for maximum post length

### LINKED ISSUE #8041: Posting settings - Max Comment Length has no validation
**Issue**
A user can save the max comment to any number they wish.

**Vanilla info**
The body field of a discussion is limited to 65,000 characters.  So any post greater than will be cut off, or and error will thrown.  This leads to a poor user experience.

cchabilall83:
We aren't currently setting a limit for the amount of characters a post can contain.  The body field in the discussions table is of type text which holds a limit of approximately 65,000 characters (64kb).  A user can enter any amount in the _Max Comment Length_ field.  This could lead to some confusion as to why posts are being cut off or the user is receiving an error.  A check is also done in the utility update to see if the current limit is too high, if so we set it to an agreed limit of 50,000 characters.

Note:  We selected 50,000 characters as the limit for now, but utf8 characters will have to be considered.  A utf8 character could be up to 4bytes.  If a post consists of all utf8 characters, it could be limited to approximately 16,000 characters in a text field.  A separate issue is being created to open a discussion on how to address this (#8042)

To test:

Posting page

1) Navigate to /vanilla/settings/posting
2) In the Max Comment Length field enter a number > 50k
3) An error should appear
4) Repeat step with a number < 50k
5) Form should be saved

Utility Update

1) In you config. php file set $Configuration['Vanilla']['Comment']['MaxLength'] = 51000;
2) run utility update ie. **domain/**/dashboard/utility/update
3) Check your config, the setting should have changed to 50,000

Closes #8041

### DIFF
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -22,6 +22,9 @@ class UtilityController extends DashboardController {
     /** @var  Gdn_Form $Form */
     public $Form;
 
+    /** @var int The maximum length a post can be. */
+    const MAX_POST_LENGTH = 50000;
+
     /**
      * @var array Special-case HTTP headers that are otherwise unidentifiable as HTTP headers.
      * Typically, HTTP headers in the $_SERVER array will be prefixed with
@@ -310,13 +313,12 @@ class UtilityController extends DashboardController {
 
         try {
             $maxCommentLength = Gdn::config('Vanilla.Comment.MaxLength');
-            if ($maxCommentLength > 30000) {
-                saveToConfig('Vanilla.Comment.MaxLength', 30000);
+            if ($maxCommentLength > self::MAX_POST_LENGTH) {
+                saveToConfig('Vanilla.Comment.MaxLength', self::MAX_POST_LENGTH);
             }
         } catch (Exception $ex) {
             $this->setData('Success', false);
             $this->setData('Error', $ex->getMessage());
-
         }
 
         if (Gdn::session()->checkPermission('Garden.Settings.Manage')) {
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -22,7 +22,6 @@ class UtilityController extends DashboardController {
     /** @var  Gdn_Form $Form */
     public $Form;
 
-
     /**
      * @var array Special-case HTTP headers that are otherwise unidentifiable as HTTP headers.
      * Typically, HTTP headers in the $_SERVER array will be prefixed with
--- a/applications/dashboard/controllers/class.utilitycontroller.php
+++ b/applications/dashboard/controllers/class.utilitycontroller.php
@@ -308,6 +308,17 @@ class UtilityController extends DashboardController {
             }
         }
 
+        try {
+            $maxCommentLength = Gdn::config('Vanilla.Comment.MaxLength');
+            if ($maxCommentLength > 30000) {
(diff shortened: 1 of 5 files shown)


applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -308,6 +311,16 @@ public function update() {
...
+            $maxCommentLength = Gdn::config('Vanilla.Comment.MaxLength');
+            if ($maxCommentLength > self::MAX_POST_LENGTH) {
+                saveToConfig('Vanilla.Comment.MaxLength', self::MAX_POST_LENGTH);
+            }
+        } catch (Exception $ex) {
+            $this->setData('Success', false);
+            $this->setData('Error', $ex->getMessage());
+        }
+
         if (Gdn::session()->checkPermission('Garden.Settings.Manage')) {
```
linc: This is not where this belongs. You've added it to the meta stuff that controls the running of utility/update, instead of adding it to the to-do list. You probably want the `structure()` method in the hooks file of the Vanilla app.

applications/vanilla/controllers/class.vanillasettingscontroller.php
```diff
@@ -83,15 +86,21 @@ public function posting() {
...
         } else {
             // This is a "reverse" field on the form. Disabling URL embeds is associated with a toggle that enables them.
             $disableUrlEmbeds = $this->Form->getFormValue('Garden.Format.DisableUrlEmbeds', true);
             $this->Form->setFormValue('Garden.Format.DisableUrlEmbeds', !$disableUrlEmbeds);
 
+            if ($this->Form->_FormValues['Vanilla.Comment.MaxLength'] > self::MAX_POST_LENGTH) {
+                $this->Form->addError(
+                    'Max comment length of' .
+                    $this->Form->_FormValues['Vanilla.Comment.MaxLength'] .
+                    ' value of is greater than the allowed size of' . self::MAX_POST_LENGTH . '.'
```
linc: "value of is greater than the allowed size of" isn't even grammatically correct, and it's generally ill-advisable to echo back user values to them (it's self-apparent when they typed in the form, it makes translations more difficult, and it can open a reflected XSS vector).

linc: "The highest allowed limit is X." would be a fine message.

applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -22,6 +22,9 @@ class UtilityController extends DashboardController {
     /** @var  Gdn_Form $Form */
     public $Form;
 
+    /** @var int The maximum length a post can be. */
+    const MAX_POST_LENGTH = 50000;
```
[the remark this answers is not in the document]
linc: Generally I would expect to find something like this on a model. In this case, DiscussionModel.

applications/dashboard/controllers/class.utilitycontroller.php
linc: This try/catch block isn't necessary, and I'd say the catch condition is just plain wrong. Did you understand what you were doing here and/or test this failure condition?

cchabilall83: it's been moved.

applications/vanilla/settings/structure.php
```diff
@@ -441,6 +441,12 @@
     }
 }
 
+// Check the if the max post and adjust it appropriately.
```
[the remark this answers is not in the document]
cchabilall83: done

PR #8040, opened 2018-11-19
KB User drop down

slafleche:
Added components for UserDropDown.

Styles updates to Drop Downs.

Added NumberFormatted component for counts. Note that it's not currently multilingual.

Added Plural component to mimic functionality we have in Vanilla.

Added GuestNavLinks (Currently only sign in link).


library/src/scripts/components/Plural.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * @license GPL-2.0-only
+ */
+
+import * as React from "react";
+import classNames from "classnames";
+import Translate from "@library/components/translation/Translate";
+
+interface IProps {
+    value: number;
+    singular?: string;
```
charrondev: I feel like you have this backwards. `plural` should be the optional prop and singular the required, because that's what you checking in the render method.

library/src/scripts/components/Plural.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * Implements helper plural components for displaying singular or plural translation strings based on the count and locale.
+ */
+export default class Plural extends React.Component<IProps> {
+    public render() {
+        const { value, singular, className } = this.props;
+        const plural = this.props.plural ? this.props.plural : `${singular}s`;
+        const title = `${<Translate source={value === 1 ? singular : plural} c0={value} />}`;
```
charrondev: I think this is an improper use of the `<Translate />` tag here. There is no `<0/>` to replace in the source string.

charrondev:
A few UI things:

## Search bar takes up the whole header

Is this intentional? 

image

## DropDownItemMetas are broken

image

## Inconsistencies with margins around hovered items

Since we're moving to having vertical margins on the menu items I think we might have missed something. Not sure what the spec was but I've heard Todd chime in about this before. I'm guessing it's supposed to be similar to a native dropdown right?

If that's the case I think we're missing margins on the top and bottom of the list.

|Us|Native|
|---|---|
|image|image|

## Profile link

Also could we make the profile image/name be links to the profile page? It seems odd to take up so much space, but not have it go anywhere.

PR #8045, opened 2018-11-20
Update size limitations of generic drafts to allow larger content

initvector:
The new drafts table stores its attributes as a MySQL text field, which have a limit of approximately sixty-five thousand characters. We're changing the field type to a "medium text" field, which has a limit of approximately sixteen million characters.

### DIFF
--- a/applications/dashboard/settings/structure.php
+++ b/applications/dashboard/settings/structure.php
@@ -857,7 +857,7 @@ $Construct
     ->column("recordType", "varchar(64)", false, ["index", "index.record", "index.parentRecord"])
     ->column("recordID", "int", true, "index.record")
     ->column("parentRecordID", "int", true, "index.parentRecord")
-    ->column("attributes", "text")
+    ->column("attributes", "mediumtext")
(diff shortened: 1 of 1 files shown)


linc:
<s>characters</s> bytes

initvector:
> ~characters~ bytes

https://dev.mysql.com/doc/refman/5.7/en/string-type-overview.html

> TEXT[(M)] [CHARACTER SET charset_name] [COLLATE collation_name]
> A TEXT column with a maximum length of 65,535 (216 − 1) characters. [...]

> MEDIUMTEXT [CHARACTER SET charset_name] [COLLATE collation_name]
> A TEXT column with a maximum length of 16,777,215 (224 − 1) characters. [...]

linc:
The VERY NEXT LINE in those docs says "The effective maximum length is less if the value contains multibyte characters." which means... it's actually in bytes. MySQL docs assumes they are equivalent as far as I can tell, which isn't really the case, as their own explanation demonstrates.

PR #8048, opened 2018-11-20
Fix generating NavLink in SmartLink when search string present

initvector:
The `SmartLink` component generates a child `NavLink` component by massaging an incoming URL. However, if the incoming URL contains a search/query string, it can be passed through as the `pathname` property. This can adversely affect routing in the application. The search/query string should be separated and passed through as the `search` property. This update performs that separation when using the `SmartLink` component.


charrondev:
I've also jumped in and added a few unit tests as I reviewed this.

PR #8061, opened 2018-11-23
Compact MeBox content

slafleche:
Added content to compact me box content. Added new tab component.

@charrondev Note that I had another PR approved today for a lot of changes to the header/mebox. Let me know if you have feedback on that.

Closes: https://github.com/vanilla/knowledge/issues/340

### DIFF
--- a/library/src/scripts/components/mebox/pieces/CompactSearch.tsx
+++ b/library/src/scripts/components/mebox/pieces/CompactSearch.tsx
@@ -24,6 +24,7 @@ export interface ICompactSearchProps extends IApiProps {
     onCloseSearch: () => void;
     cancelButtonClassName?: string;
     buttonClass?: string;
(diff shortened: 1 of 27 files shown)


library/src/scripts/components/mebox/pieces/CompactSearch.tsx
```diff
@@ -65,34 +68,40 @@ export class CompactSearch extends React.Component<ICompactSearchProps, IState>
...
-                            title={t("Search")}
-                            aria-expanded={true}
-                            aria-haspopup="true"
-                            aria-controls={this.id}
-                            baseClass={ButtonBaseClass.CUSTOM}
-                        >
-                            {t("Cancel")}
-                        </Button>
-                    </div>
+                    <React.Fragment>
```
charrondev: Do we need this fragment here still? It only has 1 child.

Issue #8058, opened 2018-11-22
Categories - cache not calculating CountAllDiscussions correctly


cchabilall83:
**Issue**

The categories cache isn't storing CountAllDiscussions correctly.  This behaviour appears to be happening only with categories that have child categories.  When a discussion is commented on, the cache is being rebuilt and CountAllDiscussions is being incremented and then stored in the cache.  The DB value is correct however.

The calculateData method appears to be the culprit, it's called every time the category cache is rebuilt.

https://github.com/vanilla/vanilla/blob/d022d069cee29aba408439f6d5d33b2981bbc75f/applications/vanilla/models/class.categorymodel.php#L666

This what's in my local cache:
![screen shot 2018-11-22 at 3 41 29 pm](https://user-images.githubusercontent.com/36996270/48921840-86cb0500-ee70-11e8-9d16-368ee4378500.png)

This is the DB:
![screen shot 2018-11-22 at 4 09 10 pm](https://user-images.githubusercontent.com/36996270/48921895-fa6d1200-ee70-11e8-89a4-65ac4a0561af.png)

**Solution:**

TBD

alex-mtl:
There are some related issues:

https://github.com/vanilla/vanilla/issues/7897
https://github.com/vanilla/vanilla/issues/7885
https://github.com/vanilla/vanilla/issues/7883

And some research was done but not merged here:

https://github.com/vanilla/vanilla/pull/7843/files

PR #8062, opened 2018-11-26
Categories - Opt out of caching when recalculating discussion totals

cchabilall83:
After discussion with @alex-mtl, we came up with this patch to opt out of recalculating the record totals when a comment or discussion are made.  There were attempts to improve caching functionality made with pr #7843, but it was never merged.   So based on the complexity of how category caching works in Vanilla, this patch was made as a bandaid until a better solution comes up.

Addresses #8058

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -1345,7 +1345,8 @@ class CategoryModel extends Gdn_Model {
         self::updateLastPost($discussion);
 
         // Update the aggregate discussion count for this category and all its parents.
-        self::incrementAggregateCount($categoryID, self::AGGREGATE_DISCUSSION);
+        self::incrementAggregateCount($categoryID, self::AGGREGATE_DISCUSSION, 1, false);
+        CategoryModel::clearCache();
 
         // Set the new LastCategoryID.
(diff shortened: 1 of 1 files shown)


linc:
This solution appears to drop the category cache on every comment, which is a non-starter due to the impact that would have on performance.

cchabilall83:
So we looked into how we display discussion counts elsewhere.
It looks like we are using the CategoryCollection class to cache categories by id and slug

This is my stack call
class.categorycollection.php:425, CategoryCollection->getTree()
class.categorymodel.php:949, CategoryModel->getChildTree()
class.categoriescontroller.php:153, CategoriesController->getCategoryTree()
class.categoriescontroller.php:570, CategoriesController->all()
class.categoriescontroller.php:299, CategoriesController->index()
class.dispatcher.php:861, Gdn_Dispatcher->dispatchController()
class.dispatcher.php:278, Gdn_Dispatcher->dispatch()
index.php:29, {main}()

https://github.com/vanilla/vanilla/blob/master/applications/vanilla/controllers/class.categoriescontroller.php#L149

https://github.com/vanilla/vanilla/blob/110de1837505b79806e6e9bd9076283aa6ddb4d3/applications/vanilla/library/class.categorycollection.php#L409

PR #8067, opened 2018-11-26
Allow resource routes to specify a root controller

tburry:
Resource routes can now specify a root controller that will be looked at after specific controllers fail, allowing the root of a path to be specifically routed to.


charrondev:
I've come in and also removed the root level resource route. It's no longer needed and that is what was causing the issues in the first place.

PR #8069, opened 2018-11-26
KB home mobile header styles

slafleche:
Updates for home page mobile header styles. I also made it so that we can render flyouts in a modal instead of a flyout. Currently only used for locale change on home page.


library/src/scripts/components/DeviceChecker.tsx
```diff
@@ -60,10 +62,15 @@ export default class DeviceChecker extends React.Component<IDeviceCheckerProps>
     }
 
     /**
-     * @inheritDoc
+     * There's a bug in webpack and there's no way to know the styles have loaded from webpack. In debug mode,
      */
     public componentDidMount() {
         window.addEventListener("resize", this.throttledUpdateOnResize);
+        if (get(window, "gdn.meta.context.debug", false)) {
```
charrondev:
~~~You can replace this with `getMeta` from `@library/application`. `getMeta('context.debug')`.~~~

On second thought, since this is a hot reload issue only, we could do a more accurate check. `if (module.hot) {}`  would be more appropriate.

library/src/scripts/components/PopoverController.tsx
```diff
@@ -94,15 +111,23 @@ export default class PopoverController extends React.PureComponent<
...
-                        id: this.contentID,
-                        initialFocusRef: this.initalFocusRef,
-                        isVisible: this.state.isVisible,
-                        closeMenuHandler: this.closeMenuHandler,
-                        renderAbove: this.props.renderAbove,
-                        renderLeft: this.props.renderLeft,
-                    })}
+                    this.state.isVisible && (
+                        <React.Fragment>
+                            {!this.props.openAsModal && this.props.children(childrenData)}
```
charrondev: Instead of doing the check twice here could we use a ternary? `openAsModel ? <Modal /> : this.props.children()`.

library/src/scripts/components/SelectBox.tsx
```diff
@@ -95,7 +96,9 @@ export default class SelectBox extends React.Component<ISelfLabelledProps | IExt
                     index={i}
                     current={selected}
                     lang={child.lang}
-                    buttonClassName="dropDownItem-button selectBox-buttonItem"
+                    buttonClassName={classNames("dropDownItem-button", "selectBox-buttonItem", {
+                        isInModal: this.props.openAsModal,
```
charrondev: Do we actually use this new class anywhere here? And if we do wouldn't it be better to use a CSS selector like `.modal .selectBox-buttonItem`?

slafleche:
This PR is still a WIP. I'll have more changes to come. I don't want to optimize the header too early.

library/src/scripts/components/SelectBox.tsx
slafleche: No, I want to cut down as much as is possible and reasonable to nest styles for nothing. It cuts down a LOT on CSS in the long run. Good example is the panel layout, I cut by about half the CSS by setting classes in React.

slafleche:
> Additionally I don't the creating component should have to be aware of the how a dropdown renders. It should just be based off the device size (all of the dropdowns I've seen in the mockups are modals on mobile).

I do think it's better to leave a door open to opt out of it for special cases. I could make it opt out instead of opt in though.

PR #8074, opened 2018-11-27
Backport: Add event in theme options

igraziatto:
Backporting https://github.com/vanilla/vanilla/pull/7969
Original Issue https://github.com/vanilla/vanilla/issues/7900


igraziatto:
As it was defined and I was not aware, this sort of backporting doesn't happen anymore.

PR #8078, opened 2018-11-28
KB - Mobile Header and Desktop Search

slafleche:
Updates to header for responsive styles and desktop search. Only affects KB.


charrondev:
I can't access items on the left of the mobile header.

image

--- 

I feel like there is a missing margin here:

image

charrondev:
The mobile the menu here seems to be off spec:

|Actual|Spec|
|---|---|
|image|image|

---

A separate seems to be missing in the bottom of the mobile menu now:

image

PR #8080, opened 2018-11-28
Fix DateFilterSchema or-equal-to comparisons on date-only values not including full date

### LINKED ISSUE #8047: DateFilterSchema and DateFilterSphinxSchema are incorrect for <= Date
All other operators seem to be working properly.

These 2 unit tests (mostly duplicated between the 2 schema's tests should not have the same expected result. When pass than Less-Than or Equal (Date) we should be starting from the end of the day.

```php
'Less-Than or Equal (Date)' => [
    'dateInserted',
    $schema->validate('<=2017-01-01'),
    ['dateInserted <=' => new DateTimeImmutable('2017-01-01 00:00:00')]
],
'Less-Than (Date)' => [
    'dateInserted',
    $schema->validate('<2017-01-01'),
    ['dateInserted <' => new DateTimeImmutable('2017-01-01 00:00:00')]
],
```

initvector:
`DateFilterSchema` (and by extension `DateFilterSphinxSchema`) do not include the full date when performing a simple comparison on a date-only value. For example, You might expect "<=2018-01-01" to match on anything less-than-but-including January 1, 2018. What you would actually get is just less-than January 1, 2018. The comparison is exclusive.

This update tweaks the logic in `DateFilterSchema::parseSimple` to include the whole day of a date-only value when using an or-equal-to comparison.

Closes #8047

### DIFF
--- a/library/Vanilla/DateFilterSchema.php
+++ b/library/Vanilla/DateFilterSchema.php
@@ -187,7 +187,6 @@ class DateFilterSchema extends Schema {
                     ];
                     break;
                 case "<=":
-                case ">=":
                     $dateTimes = [$dateTimes[0]->modify('+1 day')->modify('-1 second')];
                     break;
             }
--- a/library/Vanilla/DateFilterSchema.php
+++ b/library/Vanilla/DateFilterSchema.php
@@ -178,11 +178,19 @@ class DateFilterSchema extends Schema {
         }
 
         // If all we have is a date, give us a range in that date.
-        if ($operator == '=' && !preg_match('/\d\d:\d\d:\d\d/', $date)) {
-            $dateTimes = [
-                $dateTimes[0],
-                $dateTimes[0]->modify('+1 day')->modify('-1 second'),
-            ];
+        if (!preg_match('/\d\d:\d\d:\d\d/', $date)) {
+            switch ($operator) {
+                case "=":
+                    $dateTimes = [
+                        $dateTimes[0],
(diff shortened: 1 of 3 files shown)


library/Vanilla/DateFilterSchema.php
```diff
@@ -178,11 +178,19 @@ private function parseSimple($date, $operator, ValidationField $field) {
...
+        if (!preg_match('/\d\d:\d\d:\d\d/', $date)) {
+            switch ($operator) {
+                case "=":
+                    $dateTimes = [
+                        $dateTimes[0],
+                        $dateTimes[0]->modify('+1 day')->modify('-1 second'),
+                    ];
+                    break;
+                case "<=":
+                case ">=":
```
charrondev: This greater than or equals case doesn't seem correct to me.

tests/Library/Vanilla/DateFilterSphinxSchemaTest.php
```diff
@@ -69,7 +69,7 @@ public function provideDateFilterRange() {
             'Greater-Than or Equal (Date)' => [
                 '>=2017-01-01',
                 [
-                    'startDate' => new DateTimeImmutable('2017-01-01 00:00:00'),
+                    'startDate' => new DateTimeImmutable('2017-01-01 23:59:59'),
```
charrondev: Greater than or equals should be at the start of the day like it was.

PR #8100, opened 2018-12-04
Navigation Management - Part 1

slafleche:
I'm going to split my PRs for the navigation management to make it easier to review. 

This first PR is mostly styles for the new Navigation Manager (KB)

- Added new NavigationManager icons
- Added Navigation Manager styles

Other details: 
- Pass event in modal's close frame function
- Moved icon to separate partial
- Minor updates to Modal styles


library/src/scripts/components/PageHeading.tsx
```diff
@@ -12,26 +12,24 @@ import Heading from "@library/components/Heading";
...
     actions?: React.ReactNode;
+    includeBackLink?: boolean;
 }
 
 /**
  * A component representing a top level page heading.
  * Can be configured with an options menu and a backlink.
  */
 export default class PageHeading extends React.Component<IPageHeading> {
+    public defaultProps = {
```
charrondev: This needs to be `public static defaultProps: Partial<IPageHeading>`

The type is __necessary__ or it will break higher order components types.

slafleche:
@charrondev still having issues focussing the input text... any ideas?

Issue #8104, opened 2018-12-06
Write proper unit tests and story for SiteNav and SiteNavNode


charrondev:
The SiteNav needs some proper unit tests. A few particular scenarios are deserving of tests:

- [x] Initializing with an active record.
- [x] Changing the active record.
- [x] Passing a bad active record.
- [ ] ~A user manually collapsing and uncollapsing nodes. (What happens to these when the active record changes. Should they be preserved?).~ (should be part of `SiteNavContext` tests)

PR #8105, opened 2018-12-06
Update SiteNav and SiteNavNode to handle real data

charrondev:
This PR updates the `<SiteNav>` and `<SiteNavNode>` components to be able to better handle real data. A few changes:

- Rather than using a location and comparing urls the `SiteNav` now takes an `IActiveRecord`. This is more resilient when
  - The user is visiting a non-canonical URL
  - Comparing relative/absolute paths.
  - Is more explicit.
- Updates the links to use `<SmartLink />`
- The existing "unit tests" weren't really unit tests but home-grown snapshot tests, making it painful to update the component without really testing anything useful. I've deleted them and opened a followup issue for @initvector and I to write proper unit tests: https://github.com/vanilla/vanilla/issues/8104
- Make the expected record type and interface instead of `any`.
- Some state was removed to protect against infinite recursion. Instead of `componentDidUpdate` has been removed and a strong key is used instead.

## Other
- `<Breadcrumbs />` has been updated to take accept a prop to determine if it should render with a single item or not.


charrondev:
https://github.com/vanilla/knowledge/issues/428

PR #8112, opened 2018-12-11
Remove the easyXDM cors file

tburry:
We don’t use easyXDM for CORS so this file presents a threat vector.


tburry:
I should say am 99% sure this file is not needed for our easyXDM embed so I think it's safe to try removing it.

PR #8138, opened 2018-12-19
Refactor legacy flyouts

charrondev:
Required for/blocking https://github.com/vanilla/themes/issues/253

## Context

Our latest iteration of flyouts is done in react and switches between rendering the flyout as a dropdown on desktop and a modal on mobile. There were some really nasty javascript/stylesheet hacks to make mobile/responsive flyouts possible in the various client themes we've done in the past 2 years. 

The goal of this PR is to:
- Allow a new base theme to have nice responsive flyouts.
- Not break compatibility with any flyouts in our forum.
- Not break compatibility with any existing themes, including the numerous responsive ones that are unsetting/recreating the flyout event listeners themselves.
 make it possible for these to be removed for the new base theme.

## Plumbing

- Adds a feature flag `NewFlyouts`. I was debating calling this `NewLegacyFlyouts`.
- Updates `Gdn_Controller` to pass along this value to the frontend.
- Takes all flyout related code in `global.js` and moves it into `flyouts.js`. I did not put it on the new build because all of the themes using the workaround or unsetting event listeners will expect these event listeners to be set synchronously before their own javascript is loaded. The loading of built javascript is `deferred` and I intend to keep it that way.
- Updates `Gdn_Controller` to be able to load additional javascript files when certain ones are loaded. This will allows us to break down `global.js` without needing to worry about the 30-40 controllers spread across allow the repos that expect it to be loaded immediately after `global.js` (also if it was left for the controllers to manually add these broken out pieces, this would be a breaking change for open source at least). **TL;DR; if you include `global.js`, `flyouts.js` will be included immediately after it.** 
- Deleted ~50 lines of commented out code in `global.js`.

## `flyouts.js`

This file is 90% the same logic as it was in `global.js`. I've made a few changes though:

- Moved the duplicated logic of opening/closing a flyout into dedicated functions.
- Moved all event handler callbacks into named functions.
- Added docblocks to everything.
- Made Button Dropdowns and Flyouts aware of each other. Previously the opening a flyout would close the dropdown but not the other way around. This no works both ways. Eg. Only 1 dropdown/flyout may be open at a time.

When the `useNewFlyouts` config is set the following changes will take affect in order to offer more flexibility to themes:

- `$.hide()` and `$.show()` will no longer be called and `display == 'none'` will no longer be used to determine if a flyout is open or not. Instead only the CSS class will determine the open/closed state.
- On content load will wrap any `Flyout` or `Dropdown` in a `<span class='mobileFlyoutOverlay'></span>`. This is to allow easier styling of a background overlay from a theme.

## Testing

I've manually tested this with our core themes as well as some of our custom ones that do event overriding/setting. The current vanilla theme boilerplate uses this method. https://github.com/vanilla/themes

Note: Flyouts won't properly display in any existing theme. Themes will need to update their stylesheets to enable this the new feature flag.

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -365,6 +365,17 @@ class Gdn_Controller extends Gdn_Pluggable {
     }
 
     /**
+     * Mapping of how certain legacy javascript files have been split up.
+     * 
+     * If you include the key, all of the files in it's value will be included as well.
+     */
+    const SPLIT_JS_MAPPINGS = [
+        'global.js' => [
+            'flyouts.js',
+        ],
+    ];
+
+    /**
      * Adds a JS file to search for in the application or global js folder(s).
      *
      * @param string $fileName The js file to search for.
@@ -384,6 +395,13 @@ class Gdn_Controller extends Gdn_Pluggable {
         }
 
         $this->_JsFiles[] = $jsInfo;
+
+        if ($appFolder === '' && array_key_exists($fileName, self::SPLIT_JS_MAPPINGS)) {
+            $items = self::SPLIT_JS_MAPPINGS[$fileName];
+            foreach ($items as $item) {
+                $this->addJsFile($item, $appFolder, $options);
+            }
+        }
     }
 
     /**
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -649,6 +649,8 @@ class Gdn_Controller extends Gdn_Pluggable {
             'context' => [], 'ui' => []
         ];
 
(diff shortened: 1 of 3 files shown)


library/core/class.controller.php
```diff
@@ -364,6 +364,17 @@ public function addInternalMethod($methodName) {
         $this->internalMethods[] = strtolower($methodName);
     }
 
+    /**
+     * Mapping of how certain legacy javascript files have been split up.
+     *
```
initvector:
```suggestion
     *
```

Issue #3456, opened 2016-01-11
Profile page shows a thread as un-viewed to guest. 


LauraMakesArt:
https://vanillaforums.org/profile/discussions/10878/Linc 

When logged in, this page adjusts to show you what threads you've viewed, but as a guest this thread: https://vanillaforums.org/discussion/30856/hq-news-vanilla-has-a-new-r-d-lead-spoilers-its-me is highlighted as being unread inexplicably.

bleistivt:
Can confirm for vforg. This seems to be a cache or cloud only issue, I can't reproduce it on self hosted.

charrondev:
I tried to do a skim through the various involved methods:

- `DiscussionModel::getByUser()`
- `DiscussionModel::calculate()`
- `cssClass()` in the discussions `helper_methods.php`

The code is quite convoluted and I don't see an obvious issue. I'd need a debugger and to be able to reproduce it locally to figure this out.

charrondev:
Pretty sure I've figured out the issue. The `$watchUserID` is getting set to user who's profile you are viewing as a guest. So you are seeing the read/unread status of the whoever you are viewing rather than   the guest user.

PR #8161, opened 2019-01-04
Fix unread status on discussions for guests

charrondev:
Closes https://github.com/vanilla/vanilla/issues/3456

## What was wrong

- When calling `Discussion::getByUser` with the guest or no user ID we would set use the userID of the user whose profile was being viewed (`empty(0) = empty(null) = false`). This was clearly incorrect. I've updated this to initialize the default to be the guest user ID.
- Our calculations for unread calculations are stupidly complex. Rather than dive into the nastiness of it (with no tests mind you) I've opted to just set the `Read` status to be `false` for guest users for all discussions.

## Housekeeping

- I've added a constant for the guest user ID and used it in a couple of places. I'm quite surprised that we didn't have one.

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -1290,7 +1290,7 @@ class DiscussionModel extends Gdn_Model {
         }
 
         // If no user was provided, view from the perspective of a guest.
-        if (!filter_var($userID, FILTER_VALIDATE_INT)) {
+        if (!filter_var($watchUserID, FILTER_VALIDATE_INT)) {
             $watchUserID = UserModel::GUEST_USER_ID;
         }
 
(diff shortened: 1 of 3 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1284,9 +1289,9 @@ public function getByUser($userID, $limit, $offset, $lastDiscussionID = false, $
             return new Gdn_DataSet([]);
         }
 
-        // Allow us to set perspective of a different user.
-        if (empty($watchUserID)) {
-            $watchUserID = $userID;
+        // If no user was provided, view from the perspective of a guest.
+        if (!filter_var($userID, FILTER_VALIDATE_INT)) {
+            $watchUserID = UserModel::GUEST_USER_ID;
```
alex-mtl: Not sure about new code equal to previous one.
Previous check was done against `$watchUserID` and set new value of it only if it was empty.
New code checks another variable `$userID` and set `$watchUserID` only if `$userID` is not valid integer.
hmmmm... Not clear refactoring to me.

Issue #3848, opened 2016-04-30
Remove nested set fields from the database


tburry:

initvector:
`TreeLeft` and `TreeRight` are still used pretty heavily in Vanilla.  Does this issue depend on another being closed first?

### Nested set fields used in Vanilla
1. [applications/vanilla/views/categories/all.php:81](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/views/categories/all.php#L81)
1. [applications/vanilla/controllers/class.vanillasettingscontroller.php:674](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/controllers/class.vanillasettingscontroller.php#L674)
1. [applications/vanilla/models/class.categorymodel.php:150](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L150)
1. [applications/vanilla/models/class.categorymodel.php:1329](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L1329)
1. [applications/vanilla/models/class.categorymodel.php:1335](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L1335)
1. [applications/vanilla/models/class.categorymodel.php:1376](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L1376)
1. [applications/vanilla/models/class.categorymodel.php:1389](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L1389)
1. [applications/vanilla/models/class.categorymodel.php:1458](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L1458)
1. [applications/vanilla/models/class.categorymodel.php:1460](https://github.com/vanilla/vanilla/blob/63a21b8ca60429a208bccaa37451c5c9588d438b/applications/vanilla/models/class.categorymodel.php#L1460)
[...]

PR #8163, opened 2019-01-04
Fix CategoryModel::getAll references to nonexistent parameters

charrondev:
These parameters were removed some time ago, but the doc block was never updated. I've additionally removed the 1 instance where a parameter was still being passed.

Related https://github.com/vanilla/vanilla/issues/3848

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -2045,12 +2045,7 @@ class CategoryModel extends Gdn_Model {
      * Get list of categories (disregarding user permission for admins).
      *
      * @since 2.0.0
-     * @access public
-     *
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -2045,12 +2045,7 @@ private function getOne($id) {
      * Get list of categories (disregarding user permission for admins).
      *
      * @since 2.0.0
-     * @access public
-     *
-     * @param string $OrderFields Ignored.
-     * @param string $OrderDirection Ignored.
-     * @param int $Limit Ignored.
-     * @param int $Offset Ignored.
+     *
```
initvector:
```suggestion
     *
```

Coding standards complaint

Issue #4264, opened 2016-08-03
Different classes for "Forgot" and "I remember now"


slafleche:
When you log in, the "Forgot" link has the class "ForgotPassword" and the "I remember now!" link has the same class. We should have a unique class for each.

PR #8164, opened 2019-01-04
Add extra CSS class for remember password

charrondev:
Closes https://github.com/vanilla/vanilla/issues/4264


applications/dashboard/views/entry/signin.php
```diff
@@ -96,7 +96,7 @@
         <li class="Buttons">
             <?php
             echo $this->Form->button('Request a new password', ['class' => 'Button Primary']);
-            echo anchor(t('I remember now!'), '/entry/signin', 'ForgotPassword');
+            echo anchor(t('I remember now!'), '/entry/signin', 'ForgotPassword isRemember');
```
slafleche: "Remember" is not a state. You can use something like "ForgotPassword-remember"

PR #8174, opened 2019-01-06
Fix Leaving page links that are double encoded

cchabilall83:
**Issue:**
When _Garden.Format.WarnLeaving_ is enabled and posted links that contained '&' characters in the query string, the link in the leaving page is double encoded and will not bring the user to the proper address.

**Solution:**
If the link is valid html it should be decoded before the link is urlencoded.

**To Test:**

1) Enable _Garden.Format.WarnLeaving_
2) Create a post with the following link

https://glumobile.helpshift.com/a/kim-kardashian-hollywood/?p=web&s=transferring-your-game-to-a-new-device&f=how-can-i-transfer-my-game-from-an-ios-device-to-an-android-device-or-from-an-android-device-to-an-ios-device&l=en

3) Click on link, it will bring you to https://glumobile.helpshift.com and won't resolve the query string
4) Checkout branch and repeat steps.

Closes Vanilla/Support#118

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -1950,10 +1950,11 @@ EOT;
                 }
                 $parts[$i] =
                     anchor(
-                    '@' . $mention,
-                    url(str_replace('{name}', rawurlencode($mention), self::$MentionsUrlFormat), true),
-                    '',
-                    $attributes
+                        '@' . $mention,
+                        url(str_replace('{name}', rawurlencode($mention), self::$MentionsUrlFormat),
+                            true),
+                        '', 
+                        $attributes
                     ) . $suffix;
             } else {
                 $parts[$i] = '@' . $parts[$i];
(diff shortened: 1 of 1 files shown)


library/core/class.format.php
```diff
@@ -1939,7 +1948,7 @@ protected static function formatMentionsCallback($str) {
                 if (self::$DisplayNoFollow) {
                     $attributes['rel'] = 'nofollow';
                 }
-                $parts[$i] = anchor('@'.$mention, url(str_replace('{name}', rawurlencode($mention), self::$MentionsUrlFormat), true), 'atMention', $attributes).$suffix;
+                $parts[$i] = anchor('@'.$mention, url(str_replace('{name}', rawurlencode($mention), self::$MentionsUrlFormat), true), '', $attributes).$suffix;
```
charrondev: While we're in here, could expand this onto multiple lines. It's really difficult to follow what parameters is what with so much inline logic.

Example:
```php
anchor(
    '@' . $mention,
    url(str_replace(/***/, true),
    '',
    $attributes
) . $suffix;
```

PR #8176, opened 2019-01-07
Fix mounting rich editor on any content load events

charrondev:
Closes https://github.com/vanilla/internal/issues/1705

This PR fixes the mounting of the editor anytime new HTML is added to the page. We were handling the specific event of a comment edit form being added to the page but there are many other places that might expect the editor to be mounted, see the linked issue.

After the creation of the editor we added 1 common event to handle all of these scenarios including comment edit boxes, so I've removed our specific handler and just used the `onContent()` handler form `@library/application`.

I also added some protected from duplicated mounting by settings a JS class once mounting has started.


initvector:
I enjoy PRs where the project is improved *and* the codebase is slimmed down a little.

PR #8178, opened 2019-01-08
Add twig support to the existing Gdn_Controller view infrastructure

charrondev:
Blocking https://github.com/vanilla/knowledge/issues/487

While working on some new dashboard pages it became apparent that we were going to want support for twig in our old controllers as well.

This PR:

- Adds an interface to for our legacy ViewHandlers (Gdn_Smarty) was the only implementation.
- Adds a new view handler using the `Garden\TwigTrait` that supports rendering twig templates.
- Adds a couple of utility methods to our twig instance (translation and url building).

### DIFF
--- a/library/Vanilla/Contracts/Web/LegacyViewHandlerInterface.php
+++ b/library/Vanilla/Contracts/Web/LegacyViewHandlerInterface.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @author Adam Charron <adam.c@vanillaforums.com>
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace Vanilla\Contracts\Web;
+
+/**
+ * A web asset interface.
+ */
+interface LegacyViewHandlerInterface {
+    /**
+     * Render the given view.
+     *
+     * @param string $path The path to the view's file.
+     * @param \Gdn_Controller $controller The controller that is rendering the view.
+     */
+    public function render($path, $controller);
+}
--- a/library/Vanilla/Contracts/Web/LegacyViewHandlerInterface.php
+++ b/library/Vanilla/Contracts/Web/LegacyViewHandlerInterface.php
@@ -8,7 +8,7 @@
 namespace Vanilla\Contracts\Web;
 
 /**
- * A web asset interface.
+ * An interface for a view handler for Gdn_Controller.
  */
 interface LegacyViewHandlerInterface {
     /**
--- a/library/Vanilla/Web/LegacyTwigViewHandler.php
+++ b/library/Vanilla/Web/LegacyTwigViewHandler.php
@@ -9,7 +9,7 @@ namespace Vanilla\Web;
 use Vanilla\Contracts\Web\LegacyViewHandlerInterface;
 
 /**
- * Gdn_Controller view handler for smarty.
+ * Gdn_Controller view handler for twig.
  */
 class LegacyTwigViewHandler implements LegacyViewHandlerInterface {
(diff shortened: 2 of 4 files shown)


library/Vanilla/Contracts/Web/LegacyViewHandlerInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ * An interface for a view handler for Gdn_Controller.
+ */
+interface LegacyViewHandlerInterface {
+    /**
+     * Render the given view.
+     *
+     * @param string $path The path to the view's file.
+     * @param \Gdn_Controller $controller The controller that is rendering the view.
+     */
+    public function render($path, $controller);
```
alex-mtl: I am not sure what is this interface you need for.
But in my opinion `render()` should not get `controller` as an argument but `data`.

library/Vanilla/Web/LegacyTwigViewHandler.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * @inheritdoc
+     *
+     * @param string $path
+     * @param \Gdn_Controller $controller
+     */
+    public function render($path, $controller) {
+        $path = str_replace(PATH_ROOT, '', $path);
+        echo self::twigInit()->render($path, $controller->Data);
```
alex-mtl: And here as I mentioned before if you would pass `$controller->Data` as an incoming argument,
that would have better structure imho

library/core/class.smarty.php
```diff
@@ -12,7 +12,7 @@
 /**
  * Vanilla implementation of Smarty templating engine.
  */
-class Gdn_Smarty {
+class Gdn_Smarty implements \Vanilla\Contracts\Web\LegacyViewHandlerInterface {
```
alex-mtl: Now I see why. 
You took `render()` from here with `controller` argument

But I would say it seem that analogue of `twig->render`  is `smarty->display`.

library/Vanilla/Contracts/Web/LegacyViewHandlerInterface.php
initvector: `Gdn_Controller` relies on its view handlers having a specific signature for their `render` methods [[1]](https://github.com/vanilla/vanilla/blob/d54d29e307000cbcce3b8a232f72df9dc4f2828b/library/core/class.controller.php#L763) [[2]](https://github.com/vanilla/vanilla/blob/d54d29e307000cbcce3b8a232f72df9dc4f2828b/library/core/class.controller.php#L2036). The interface is to ensure the legacy view handlers obey the same contract.

Only passing in data would definitely be the way to go, if we were writing a new system, but the Twig view handler has to be backwards-compatible to work with `Gdn_Controller`. I think we're stuck doing it the "legacy" way.

library/Vanilla/Web/LegacyTwigViewHandler.php
```diff
@@ -0,0 +1,27 @@
...
+use Vanilla\Contracts\Web\LegacyViewHandlerInterface;
+
+/**
+ * Gdn_Controller view handler for twig.
+ */
+class LegacyTwigViewHandler implements LegacyViewHandlerInterface {
+    use \Garden\TwigTrait;
+
+    /**
+     * @inheritdoc
```
initvector: Coding standards check doesn't like the first line of a method's doc block being a tag.

```suggestion
     * Render the given view.
```

library/core/class.smarty.php
charrondev: Except the controller does not call display, it call render.

PR #8190, opened 2019-01-09
Add ability for all addons to create settings controllers & add imageuploadpreview to Gdn_Form::simple()

charrondev:
Blocking https://github.com/vanilla/knowledge/issues/485

The old dispatcher (which currently does all dispatching for the dashboard) for the dashboard, has the ability to dispatch settings controllers for applications if:

- The request path is `{addonKey}/settings/*`
- The controller name is `{AddonKeySettingsController}`

After discussing with @tburry I've extended this support to all addons, not just the legacy applications.

**ImageUploadPreview**

Add's support for the existing `Gdn_Form::imageUploadPreview()` in `Gdn_Form::simple()`.


alex-mtl:
I approved this. But just a question.
This change is for to scan not only "applications", but '"plugins" as well?
As usually, I think about performance and side effects.
findController() - is our router?  
```
/**
     * Find the controller object corresponding to a request path.
     *
     * @param array $parts The path parts in lowercase.
     * @return array Returns an array in the form `[$controllerName, $parts]` where `$parts` is the remaining path parts.
     * If a controller cannot be found then an array in the form of `['', $parts]` is returned.
     */
    private function findController(array $parts) {
```
I have some concerns that this extension of scope (much more plugin folders to scan) can bring some conflicts.

linc:
This created issue https://github.com/vanilla/support/issues/156

PR #8191, opened 2019-01-09
Add support for converting Garden Schema exceptions to Gdn_Validation

initvector:
This update adds basic support for converting a Garden Schema exception object into its `Gdn_Validation` equivalent. This allows API controllers to be used within legacy controllers and models, where `Gdn_Validation` is still heavily used and expected to report relevant messaging. A Garden Schema validation exception can be caught by these legacy classes, converted to a `Gdn_Validation` object and then handled with existing validation capabilities (e.g. displaying relevant errors on a form).

### DIFF
--- a/library/Vanilla/Utility/ModelUtils.php
+++ b/library/Vanilla/Utility/ModelUtils.php
@@ -16,6 +16,29 @@ use Gdn_Validation;
  * Class ModelUtils.
  */
 class ModelUtils {
+
+    /**
+     * Convert a Garden Schema validation exception into a Gdn_Validation instance.
+     *
+     * @param ValidationException $exception
+     */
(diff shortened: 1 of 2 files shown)


library/Vanilla/Utility/ModelUtils.php
```diff
@@ -16,6 +16,29 @@
...
 class ModelUtils {
+
+    /**
+     * Convert a Garden Schema validation exception into a Gdn_Validation instance.
+     *
+     * @param ValidationException $exception
+     */
+    public static function validationExceptionToValidationResult(ValidationException $exception): Gdn_Validation {
+        $result = new Gdn_Validation();
+        $validation = $exception->getValidation();
```
charrondev: This looks to be unused.

PR #8200, opened 2019-01-11
API v2 Roles: Fix permissions updates overwriting all saved permissions for a category

cchabilall83:
**Issue:**

When the saving custom category permissions by role via the api endpoint, it overwrites the custom permissions set on the category.

**Solution:**

Send a flag whether to overwrite all the category permissions in Category::Save.  Unit test to make sure a call to /roles/{id)/permissions doesn't overwrite other roles permissions.

**To Test:**

1)  Create category with custom permissions
2) Create 2 roles and give them different permissions.
3) send a call to /api/v2/roles/35/permissions to modify a roles permissions.  
4) the other roles permissions should have disappeared.

closes vanilla/support#111

### DIFF
--- a/tests/APIv2/RolesTest.php
+++ b/tests/APIv2/RolesTest.php
@@ -187,6 +187,53 @@ class RolesTest extends AbstractResourceTest {
         $this->assertFalse($this->hasPermission('site.manage', 'global', $permissions));
         $this->assertFalse($this->hasPermission('comments.add', 'category', $permissions, 1));
     }
+    /**
+     * Test updating permissions with PATCH /roles/:id/permissions
+     */
+    public function testPatchPermissionOverWrite() {
+        $role = $this->getPermissionsRole([[
+            'type' => 'category',
+            'id' =>  1,
+            'permissions' => [
+                'discussions.view' => true,
+                'discussions.add' => true,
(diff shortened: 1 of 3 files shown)


tests/APIv2/RolesTest.php
```diff
@@ -187,6 +187,53 @@ public function testPatchPermissionEndpoint() {
...
+                    ]
+                ]
+            ]
+        );
+
+        $permissions1 = $this->getPermissions($role['roleID']);
+        $permissions2 = $this->getPermissions($role2['roleID']);
+
+        $this->assertTrue($this->hasPermission('discussions.add', 'category', $permissions1, 1));
+        $this->assertFalse($this->hasPermission('comments.add', 'category', $permissions1, 1));
```
alex-mtl: Why do we still expect 1 for `comments.add` ?
We just patched it few lines before to false?

cchabilall83: assertFalse verfies the condition is false from my understanding.

https://phpunit.readthedocs.io/en/7.4/assertions.html#assertfalse

alex-mtl: :-) My bad sorry! :-)

Issue #139, opened 2009-10-08
Cannot install PDO::__construct(): [2002] Invalid argument (trying to connect via unix://)


qubodup:
I don't see what the problem is here. code bug? -- [screen](http://www.box.net/shared/m97hbgommd) --

Could it be some uname(dont be root)/pwlength requirement that is not explained?

PR #8243, opened 2019-01-21
Pinterest: Fix Canadian url embeds when using advanced editor

dannimajdalani:
related [#139](https://github.com/vanilla/support/issues/139)

### Details
Pinterest does not auto-embed for Canadian users. They are re-directed to pinterest.ca and the code is set up only for .com.

This PR fixes Pinterest auto-embeds for Canadian URL using Advanced Editor.


charrondev:
@nervoustwit I wouldn't worry to much about it at the moment. There's a lot of movement going on in `Gdn_Format` right now, and this whole legacy embed system will likely be pulled out soon in favour of the new embed system.

I'd be more concerned with this issue being present in the rich editor variant of this embed rather the rest of legacy embeds which surely suffer from numerous other deficiencies.

Issue #3112, opened 2015-09-20
Delete Draft Button on the quick reply form


linc:
What if we just let it auto-save and removed the Save Draft button?

(credit to Tim who filed this internally a long time ago)

charrondev:
Completed in the 2.8 release.

PR #8247, opened 2019-01-22
Boilerplate: hide draft button

igraziatto:
Reference issue: https://github.com/vanilla/vanilla/issues/3112


themes/theme-boilerplate/src/scss/components/_forms.scss
```diff
@@ -265,6 +265,10 @@ div.token-input-dropdown {
     }
 }
 
+.DraftButton {
```
slafleche: Shouldn't this only be applied when the rich editor is on?

igraziatto: I checked with @charrondev and it should be hidden all times.

PR #3572, opened 2016-03-02
Add NBBC parser to core

initvector:
This update adds a BBCode parser to core, based on the [NBBC plug-in](https://github.com/vanilla/addons/tree/master/plugins/NBBC) and utilizing [vanilla/nbbc](https://github.com/vanilla/nbbc), which has been added as a Composer dependency.

PR #8250, opened 2019-01-22
Delete BBCode fallback parser

charrondev:
This bespoke fallback BBCode parser was vanilla's original BBCode parsing before https://github.com/vanilla/vanilla/pull/3572. Previously that parser using nbbc was it's own plugin.

Since that PR we have had a proper BBCode parser registered in the factory/container in all installations so this fallback path was not used.

[Sometimes we override it in the container](https://github.com/search?q=org%3Avanilla+BBCodeFormatter&type=Code) but it is never removed in first party code.

This PR removes the bespoke formatter and only allows the formatter to be pulled from the container.

_Note:_ I'd recommend viewing the diff with whitespace disabled.

### DIFF
--- a/library/core/class.format.php
+++ b/library/core/class.format.php
@@ -258,97 +258,29 @@ class Gdn_Format {
     public static function bbCode($mixed) {
         if (!is_string($mixed)) {
             return self::to($mixed, 'BBCode');
-        } else {
-            // See if there is a custom BBCode formatter.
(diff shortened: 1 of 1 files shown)


thunderer:
Hi @charrondev, @initvector - would you be interested in evaluating a new BBCode processing library? I would like to hear your thoughts about https://github.com/thunderer/Shortcode. I'll be happy to answer all your questions and feature requests. Thanks!

Issue #7880, opened 2018-09-29
Change addons in core and defaults for new installs


linc:
Ops ticket for before deploy https://github.com/vanilla/ops-tickets/issues/3

Addon swaps:

* [x] Move Button Bar from core to addons.
  - Adding to `vanilla/addons` https://github.com/vanilla/addons/pull/682
  - Removing from `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8293
* [x] Move OpenID from core to addons.
  - Adding to `vanilla/addons` https://github.com/vanilla/addons/pull/683
  - Removing from `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8295
* [x] Delete Google+.
  - Removing from `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8284
* [x] Delete Debugger.
  - Removing from `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8283
* [x] Add debugbar to addons (currently closed source).
  - Adding to `vanilla/addons` https://github.com/vanilla/addons/pull/673
  - Removing from `vanilla/internal` https://github.com/vanilla/internal/pull/1798
* [x] Add Akismet to core from addons.
  - Adding to `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8277
  - Removing from `vanilla/addons` https://github.com/vanilla/addons/pull/679
* [x] Add Hero Image to core (currently closed source).
  - Adding to `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8172
  - Removing from `vanilla/internal` https://github.com/vanilla/internal/pull/1796
* [x] Add Pockets to core.
  - Adding to `vanilla/vanilla` https://github.com/vanilla/vanilla/pull/8173
  - Removing from `vanilla/addons` https://github.com/vanilla/addons/pull/672
* [x] Add Keystone theme to core (new) - https://github.com/vanilla/vanilla/pull/8179

Settings changes:

* [x] Default to Rich Editor being enabled on new installs.  https://github.com/vanilla/vanilla/pull/8302
* [x] Default to Keystone theme being enabled on new installs. https://github.com/vanilla/vanilla/pull/8302
* [ ] Add open source stub content about enabling Akismet & how to use Pockets.
* [ ] Add open source stub content about the new import feature flag.
* [ ] Rewrite Addonstest::testConflictingAddons test #8304

charrondev:
>  Remove Debugger from core.

[...]

PR #8277, opened 2019-01-28
Transfer Akismet plugin to Vanilla from addons

cchabilall83:
Issue: #7880
Ops Ticket: vanilla/ops-tickets#3

Hold this PR until this is coordinated with Ops. Please review and approve it before that time though.

This brings the Akismet plugin and its git history along from the addons repo using git filter-branch and git merge --allow-unrelated-histories. The PR for deletion from internal is here


linc:
The CLAs are fine. All were employees, yes.

PR #8280, opened 2019-01-29
Add components for KB's Help Centre Home Page

slafleche:
Goes with https://github.com/vanilla/knowledge/pull/594

New components for KB's Help Centre home page.


charrondev:
A few style notes:

- The column width is really small narrow here. It should be widened (pending confirmation from @ElisaAlmeida). The mobile width is also keeping this, even when we've dropped down to 1 column (should be 100% width).
- The links are missing focus/hover states.

PR #8282, opened 2019-01-29
Add frontend component for prev/next pagination

initvector:
This update introduces the `SimplePager` frontend component, as well as `SimplePagerModel` for parsing data to be used with the component.

### Overview
1. Add `SimplePager` component. This component will display 1-2 buttons, depending on the available pages.
1. Add `SimplePagerModel`. This model currently contains a single method, which allows parsing page parameters out of a link HTTP header for "previous" and "next" pages.

### DIFF
--- a/library/src/scripts/simplePager/SimplePager.tsx
+++ b/library/src/scripts/simplePager/SimplePager.tsx
@@ -29,7 +29,6 @@ export default class SimplePager extends React.Component<IProps> {
                 {prev && (
                     <LinkAsButton
                         className={classNames(["simplePager-button", "simplePager-prev", { isSingle }])}
-                        key="simplePagerPrev"
                         to={this.makeUrl(prev)}
                     >
                         {t("Previous")}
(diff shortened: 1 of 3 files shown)


library/src/scripts/simplePager/SimplePager.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    className={classNames(["simplePager-button", "simplePager-next", { isSingle }])}
+                    key="simplePagerNext"
+                    to={this.makeUrl(next)}
+                >
+                    {t("Next")}
+                </LinkAsButton>,
+            );
+        }
+
+        return <div className="simplePager">{buttons}</div>;
```
slafleche: This works, but you could have less lines of code if you wrote it like this:

```
{next && (
    <LinkAsButton className={classNames(["simplePager-button", "simplePager-next", { isSingle }])} key="simplePagerNext" to={this.makeUrl(next)} >
        {t("Next")}
    </LinkAsButton>
)}
```

library/src/scripts/simplePager/SimplePager.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+        const buttons = [] as JSX.Element[];
+        const isSingle = (prev && !next) || (!prev && next);
+
+        return (
+            <div className="simplePager">
+                {prev && (
+                    <LinkAsButton
+                        className={classNames(["simplePager-button", "simplePager-prev", { isSingle }])}
+                        key="simplePagerPrev"
```
slafleche: When you're not in a loop, you don't need a key.

PR #8320, opened 2019-02-06
Convert vanilla header to use TypeStyle

slafleche:
Converted the Vanilla Header component to use typestyle. Should look/behave the same as before.


applications/dashboard/Gruntfile.js
```diff
@@ -124,7 +124,7 @@ module.exports = function (grunt) {
             src: [
               'bootstrap/LICENSE'
               , 'bootstrap/scss/*.scss'
-              , 'bootstrap/scss/mixins/*.scss'
+              , 'bootstrap/scss/styleUtilities/*.scss'
```
charrondev: Did you intend to touch this? My gut tells me it's find/replace that reached too far.

library/src/scripts/components/headers/vanillaHeaderStyles.ts
```diff
@@ -0,0 +1,555 @@
...
+import { color, percent, px } from "csx";
+import { globalVariables } from "@library/styles/globalStyleVars";
+import { flexHelper, getColorDependantOnLightness } from "@library/styles/styleHelpers";
+import { layoutStyles } from "@library/styles/layoutStyles";
+import { style } from "typestyle";
+import { formElementsVariables } from "@library/components/forms/formElementStyles";
+import { userPhotoVariables } from "@library/styles/userPhotoStyles";
+import { vanillaMenuVariables } from "@library/styles/vanillaMenu";
+
+// const sampleThemeOverwrite = {
```
charrondev: Commented code here

library/src/scss/components/_vanillaHeader.scss
```diff
@@ -4,399 +4,396 @@
...
-
-    @include mediaQuery-panelLayout_oneColumn {
-        height: $vanillaHeader_mobile_height;
-    }
-}
-
-.vanillaHeaderHome {
-    min-height: $vanillaHeader_mobile_height * 2;
-}
+//$vanillaHeader-count_size: 18px;
```
charrondev: Are we going to delete this now?

library/src/scripts/components/headers/vanillaHeaderStyles.ts
```diff
@@ -0,0 +1,555 @@
...
+                            borderRadius: vars.button.borderRadius,
+                        },
+                    },
+                },
+            },
+        },
+        mediaQueries.oneColumn({ height: px(vars.sizing.mobile.height) }),
+    );
+
+    const meBox = style({
```
charrondev: Could we break this into a separate partial in the `components/mebox` directory? I feel like this partial is starting to get a little large.

slafleche: These are "vanillaHeader-meBox" styles, not just meBox styles

PR #8351, opened 2019-02-08
Add unified breadcrumb solution for developers

charrondev:
## Backend

- Implements new `BreadcrumbModel` and supporting classes.
- Implements `ForumBreadcrumbProvider` to supply breadcrumbs for forum categories, discussions, and comments.
- Implement passing of JSON-LD breadcrumbs from the breadcrumb model to the `default-master.twig` template.
- Updates the frontend `Breadcrumbs` component with the ability to display at its usual size, even when no breadcrumbs are provided.


applications/vanilla/settings/class.hooks.php
```diff
@@ -10,13 +10,23 @@
...
  * Vanilla's event handlers.
  */
 class VanillaHooks implements Gdn_IPlugin {
 
+    /**
+     * Handle the container init event to register things with the container.
+     *
+     * @param Container $dic
+     */
+    public function container_init(Container $dic) {
```
alex-mtl: I don't think this is the best place to addProvider.

Why not in bootstrap.php?

charrondev: Because the class is from an addon (This is addon declaring it's contribution to the breadcrumb manager). If the forum application was disabled the `ForumBreadcrumbProvider` class would not be available. That's why the the registration lives in the addon initialization.

PR #8352, opened 2019-02-08
Splash component and variables for buttons

slafleche:
This PR is for the "Splash" component, featured on the KB's home page.

There were a few rabbit holes along the way. I did a lot of work to move over the button styles I needed, as well as a new generic "search component" and the backLink styles.


slafleche:
Found a few bugs. I'm fixing them now.

library/src/scripts/components/Search.tsx
```diff
@@ -0,0 +1,85 @@
...
+    cancelContentClassName?: string;
+    theme?: object;
+}
+
+interface IState {
+    query: string;
+}
+
+/**
+ * Implements Compact Search component for header
```
charrondev: Could we update this docblock?

charrondev: I'm also thinking this class name could be more specific. Maybe `IndependentSearch` or `SplashSearch` and moved to the splash directory.

charrondev:
1 other thing I'm noticing is that the `cursor` value seems to be `pointer` over the entire search input. It should be `text` while hovering over the text input and `pointer` while hovering over the button or results.

PR #8372, opened 2019-02-11
Feature/error pages

slafleche:
These updates are for the KB's error pages.

Changes:
- Fixed tabindex for "LinkAsButton" component. If it mimics a button, it would get 0 as a tab index
- Added icon for KB error pages
- Tweaked font size for small titles
- added "inheritHeightClass" helper function
- Fixed height of editor

### DIFF
--- a/library/src/scripts/components/LinkAsButton.tsx
+++ b/library/src/scripts/components/LinkAsButton.tsx
@@ -18,6 +18,7 @@ interface IProps extends IOptionalComponentID, LinkProps {
     title?: string;
     ariaLabel?: string;
     baseClass?: ButtonBaseClass;
+    tabIndex?: number;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


library/src/scripts/components/LinkAsButton.tsx
```diff
@@ -29,14 +30,14 @@ export default class LinkAsButton extends React.Component<IProps> {
...
-        const { baseClass, className, title, ariaLabel, to, children, ...restProps } = this.props;
+        const { baseClass, className, title, ariaLabel, to, children, tabIndex, ...restProps } = this.props;
         const componentClasses = classNames(baseClass, className);
         return (
             <SmartLink
                 className={componentClasses}
                 title={title}
                 aria-label={ariaLabel || title}
-                tabIndex={-1}
+                tabIndex={tabIndex ? tabIndex : 1}
```
alex-mtl: To not have it hardcoded back and fourth, probably it is better to use some kind of object property or constant?

PR #8373, opened 2019-02-11
Fix breadcrumbs prop

slafleche:
I had compilation problems on my localhost because there were missing "forceDisplay" props on "Breadcrumbs" components. I made it optional, as it will default to false.


slafleche:
Already fixed

PR #8380, opened 2019-02-13
Session::checkRankedPermission - Add Garden.Curation.Manage to permissions ranked list

cchabilall83:
This allows for Garden.Curation.Manage to be used when checkedRankPermission is called.

required for Q&A plugin


linc:
Someone from R&D should sign off on this. I could've sworn Curation was already a ranked permission.

cchabilall83:
> Someone from R&D should sign off on this. I could've sworn Curation was already a ranked permission so now I wanna be sure I didn't inadvertently reverse some decision I missed.

looking at the history it appears like it was never there.

initvector:
> Someone from R&D should sign off on this. I could've sworn Curation was already a ranked permission so now I wanna be sure I didn't inadvertently reverse some decision I missed.

I could not find evidence of this. The file history didn't seem to include it. Going back to the original "spec" from '15, only these original four permissions were mentioned. I think this is okay to add.

PR #8382, opened 2019-02-14
Migrate to static OpenAPI definitions

tburry:
Rather than inspect controllers, OpenAPI definitions must now be explicit YAML files. This change adds a tool to combine these files and serve them for the purposes of documentation.


tburry:
This PR is ready to go now @initvector.

initvector:
The `GLOB_BRACE` constant is not available on all systems. Per PHP's [manual page on `glob`](http://php.net/glob):

> Note: The GLOB_BRACE flag is not available on some non GNU systems, like Solaris.

I see the behavior exhibited while running under Alpine Linux, due to the distro's use of musl libc over glibc. As a fix, I've added a simple fallback if `GLOB_BRACE` is not available.

PR #8392, opened 2019-02-14
Add toggle-able labels in SiteNav components

slafleche:
I had done most of https://github.com/vanilla/knowledge/issues/617, but I was missing this last piece to make the guide categories toggable instead of links in the side panel nav.

Other updates
- Converted button loader to use CSS in JS styles to have the colour changeable.
- Renamed buttonVariables const for clariy
- Added links variables in globalVariables
- Fixed max-width of search in header


library/src/scripts/components/forms/select/SearchBar.tsx
```diff
@@ -119,7 +121,7 @@ export default class SearchBar extends React.Component<IProps, IState> {
                 menuIsOpen={this.isMenuVisible}
                 classNamePrefix={this.prefix}
                 className={classNames(this.prefix, className)}
-                placeholder={this.props.placeholder}
+                placeholder={this.props.placeholder || ""}
```
initvector: It looks like you're using `||` to implement a default value for this property. Would it be better to set `placeholder` to an empty string in `defaultProps`?

library/src/scss/components/_siteNav.scss
```diff
@@ -146,8 +145,11 @@ $siteNav-node_padding               : $utility-baseUnit - $siteNav-node_borderWi
 
 
 .siteNavNode-buttonOffset {
+    position: relative;
     display: flex;
     justify-content: flex-end;
     width: $siteNav-nodeToggle_width;
     margin-left: -#{$siteNav-nodeToggle_width};
+    top: #{(37 / 2)}px;
```
alex-mtl: Seem like hardcoded value (37). Could we base it on some global variable value? do we have default `line-height` or `siteNav-button-height` or `siteNav-height` ?

slafleche: I'll fix it when I convert the styles to JS.

alex-mtl: Why not now? Never mind

Issue #175, opened 2009-11-28
Search results need to be improved


linc:
Trying to find a specific discussion is nearly impossible with the current search. I've been repeatedly unable to refer new members to past discussions because I can't find them, and they report not finding the discussions either when they searched before posting.

Once a discussion cycles off the first page, it might as well be gone forever as it stands. :-/

PR #8398, opened 2019-02-18
Reported Posts: fix reported activity posts displaying as html

dannimajdalani:
closes [#175](https://github.com/vanilla/support/issues/175)

This bug was noticed after a recent change in [#7724](https://github.com/vanilla/vanilla/pull/7724)
**To Replicate** 

- Find a profile with a comment in their Activity
- Using an admin/mod account, flag that comment for the report. 
- The author of the comment will be sent to 'Reported Posts' with the expanded html as the title. 

image

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -1804,7 +1804,7 @@ if (!function_exists('getRecord')) {
                         throw permissionException();
                     }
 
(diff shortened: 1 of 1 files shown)


cchabilall83:
This looks good, it doesn't look it affects other areas this is called.

Issue #8422, opened 2019-02-19
Make page scroll to embedded comments when logging in.


nervoustwit:
Vanilla has an embed comments feature for blogs and articles. It allows forum owners to use the commenting feature of their forum on an article or blog.

If a user arrives on the article page where comments are embedded and they are not logged into the forum, their is a link to log the user in. Unfortunately, when the user logs in, he/she is brought back to the top of the article page. Articles can be long. We should create a way so that when the log in process is completed and the user is redirected back to the article page that they scroll down to the comment section.

PR #8423, opened 2019-02-19
Make page scroll down to embedded comments when user logs into embedded comments.

nervoustwit:
This PR would close https://github.com/vanilla/vanilla/issues/8422

Pass the DOM element id of the embedded comments (#vanilla-comments) to the sign in URL on embedded comments. This hash will be passed as part of the redirect URL through the log in workflow so that when the authenticated user will come back to the page where the embedded comments are they will be brought to the commenting form.

### DIFF
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -615,7 +615,7 @@ if (!function_exists('writeEmbedCommentForm')) :
 
                 // If we aren't ajaxing this call then we need to target the url of the parent frame.
                 $returnUrl = $controller->data('ForeignSource.vanilla_url', Gdn::request()->pathAndQuery());
-                $returnUrl = trim($returnUrl,'/').'#vanilla-comments';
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -615,6 +615,7 @@ function writeEmbedCommentForm() {
 
                 // If we aren't ajaxing this call then we need to target the url of the parent frame.
                 $returnUrl = $controller->data('ForeignSource.vanilla_url', Gdn::request()->pathAndQuery());
+                $returnUrl = trim($returnUrl,'/').'#vanilla-comments';
```
charrondev: I don't think our linter is enforcing this at the moment but we have a bit more white space in our expressions now.
```suggestion
                $returnUrl = trim($returnUrl, '/') . '#vanilla-comments';
```

linc:
My only concern here would be how this could affect backwards compatibility or cause a site to start behaving differently. I can't think of anything since this is an established anchor point and we're adding it to a fairly narrow context here.

PR #8424, opened 2019-02-19
wip/queue integration

egarciajulia:
Work in progress PR, the purpose is to get some feedback. **Not intended to be merged**.
Relates to: https://github.com/vanilla/garden-message-queue/pull/2

## bootstrap.php
+ added `\Garden\MessageQueue\JobQueue` bootstrap. Called before `/bootstrap.late.php` to give an opportunity to re-handle this later. 
+ Fire event `BeforeShutdown` in a register_shutdown_function after all the bootstrap process. That would make the callback the last function called (before shutdown). Firing the event at that moment would allow processing others register_shutdown_function, for example, the one handling cookies.
+ Be aware in case a Job is scheduled, garden-message-queue default's implementation would                 make a session_write_close() and a fastcgi_finish_request();

## job/SendMailJob.php
+ Naive implementation of a Job that sends mail. Gets a VanillaMailer instance and just send() it

## library/core/class.email.php
+ small tweak to use the SendMailJob

## library/core/functions.compatibility.php
+ Define function "setvalr" only if it doesn't exists yet

## job/EchoJob.php
+ Echo Job. Basic Job Implementation

### DIFF
--- a/library/core/functions.compatibility.php
+++ b/library/core/functions.compatibility.php
@@ -452,51 +452,55 @@ if (!function_exists('valr')) {
     }
 }
 
-/**
- * Set a key to a value in a collection.
- *
- * Works with single keys or "dot" notation. If $key is an array, a simple
- * shallow array_merge is performed.
- *
- * @param string $key The key or property name of the value.
- * @param array &$collection The array or object to search.
- * @param mixed $value The value to set.
- * @return mixed Newly set value or if array merge.
- */
-function setvalr($key, &$collection, $value = null) {
-    if (is_array($key)) {
-        $collection = array_merge($collection, $key);
-        return null;
-    }
-
-    if (strpos($key, '.')) {
-        $path = explode('.', $key);
-
-        $selection = &$collection;
-        $mx = count($path) - 1;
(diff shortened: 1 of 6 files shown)


R-J:
It would be great to see a queue implementation!

May I suggest to make it compatible to the [queue interop standard](https://github.com/queue-interop/queue-interop)? Vanilla is moving more and more towards interoperatibility anyway and designing new components right from the start in that spirit would be a good habit

library/core/functions.compatibility.php
```diff
@@ -452,51 +452,55 @@ function valr($key, $collection, $default = false) {
...
-    }
-
-    if (strpos($key, '.')) {
-        $path = explode('.', $key);
-
-        $selection = &$collection;
-        $mx = count($path) - 1;
-        for ($i = 0; $i <= $mx; ++$i) {
-            $subSelector = $path[$i];
+if (!function_exists('setvalr')) {
```
alex-mtl: I don't see how this `setvalr` function is related to `queue` implementation. 
Do I miss something?

egarciajulia:
PR closed. Not needed anymore.

Issue #5555, opened 2017-05-09
Adding "fulltext" key (almost) always fails


R-J:
In class Gdn_MySQLStructure the method `_modify()` fails to handle fulltext keys if c('Database.ForceStorageEngine') isn't set exactly as "MyISAM".

The reason is, that `$this->_TableStorageEngine` is only set if the config value above is set. So even if a table is already MyISAM, that property isn't set which leads to a failure when the method `_supportsFulltext()` is called.
This method return `false` if the class property `_TableStorageEngine` isn't "MyISAM".

So the only way to add a fulltext key is to temporarily set the config 'Database.ForceStorageEngine' to "MyISAM", which isn't intuitive when your tables already are MyISAM.

Since version 5.6 of MySQL, InnoDB also supports fulltext search. Since the only condition under which fulltext key can by added is that the storage engine is forced to be MyISAM, there is no way (using the structure class) to add fulltext keys to InnoDB tables.
That's why `_supportsFulltext()` isn't suited at all for InnoDB.

I would suggest to
a) using an '$engine' parameter with the method `_supportsFulltext($engine)`  and do the comparison with the property only if the parameter isn't set:
~~~
    protected function _supportsFulltext($engine) {
        if (!$engine) {
            $engine = $this->_TableStorageEngine;
        }
        return strcasecmp($engine, 'myisam') == 0;
    }
~~~
This would make the method more versatile.
Setting the property `_TableStorageEngine` would be another option, but it would require changing conditional code blocks in _modify, while adding a parameter only would require changing the way the method is called from _modify.

b) extending this check to allow fulltext keys for other engines. Either by doing a MySQL version check or by a using a config key like `'Database.FulltextSupportingEngines' = array('myisam', 'innodb');` (although I know you don't like arrays in your config any more). Maybe `'Database.AlwaysAllowFulltext' = true`?
~~~
    protected function _supportsFulltext($engine) {
        if (c('Database.AlwaysAllowFulltext') == true) {
            return true;
        }
        if (!$engine) {
            $engine = $this->_TableStorageEngine;
        }
        return strcasecmp($engine, 'myisam') == 0;
    }
~~~

I saw another issue which has something to do with indexes in this class. Maybe this problem can be considered when [the other problem](https://github.com/vanilla/vanilla/issues/5441) is being looked at

Issue #6029, opened 2017-10-11
Default all table to InnoDB if MySQL version > 5.6


DaazKu:
*Note: I'm hijacking this issue mainly to get the PR I reworked reviewed.*

InnoDB support fulltext in MySQL 5.6+ so we should create all the table using InnoDB instead of using MyISAM if the MySQL version allows it.

The whole "supportFullText" feature also needs a revamp. More details about that there: https://github.com/vanilla/vanilla/issues/5555

R-J:
This should be done with 2.8 since requirements now are MySQL 5.7
https://docs.vanillaforums.com/developer/installation/self-hosting/#requirements

initvector:
Closed by #9470

PR #8440, opened 2019-02-21
Get rid of forcing tables to MyISAM when FULLTEXT key is used

R-J:
Vanilla always checked if a table had a FULLTEXT key in one row and forced that table to MyISAM. The minimum required MySQL version for Vanilla already supports fulltext keys for InnoDB tables, therefore that logic isn't needed any more.

The Gdn_MySqlStructure class had a helper function "_supportsFulltext. Under the assumption that this will always return `true`, I've deleted all code lines that became unnessecary.

This will close 
https://github.com/vanilla/vanilla/issues/5555
https://github.com/vanilla/vanilla/issues/6029


tburry:
Closing in favor of #9470.

PR #8442, opened 2019-02-21
Add some additional typestyle variables and clean up "any" type usage

charrondev:
Blocking https://github.com/vanilla/knowledge/pull/709
Closes https://github.com/vanilla/knowledge/issues/676

This PR:

- Fixes `subComponentStyles()`. Previously this had an implicit return type of `any` which when unioned with other types would cause them to be any. As a result many of the core variable objects had a type of `any`. I alterred the return type to be `object` and now the union types work properly. As a result I had to fix a few broken types. These broken types masked by `any` were the cause of https://github.com/vanilla/knowledge/issues/676
- Added variables for `meta` text and `baseUnit`.

### DIFF
--- a/library/src/scripts/styles/buttonStyles.ts
+++ b/library/src/scripts/styles/buttonStyles.ts
@@ -42,34 +42,34 @@ export function buttonStyles(theme?: object) {
 type Unit = string | number;
 
 export interface IButtonType {
-    fg: string;
-    bg: string;
-    spinner: string;
+    fg: ColorHelper;
+    bg: ColorHelper;
+    spinner: ColorHelper;
     border: {
-        color: string;
+        color: ColorHelper;
         width: Unit;
         style: string;
         radius: Unit;
     };
     hover: {
-        color: string;
-        backgroundColor: string;
-        borderColor: string;
+        fg: ColorHelper;
+        bg: ColorHelper;
+        borderColor: ColorHelper;
     };
     focus: {
-        color: string;
-        backgroundColor: string;
-        borderColor: string;
+        fg: ColorHelper;
+        bg: ColorHelper;
+        borderColor: ColorHelper;
     };
     active: {
-        color: string;
-        backgroundColor: string;
-        borderColor: string;
+        fg: ColorHelper;
+        bg: ColorHelper;
+        borderColor: ColorHelper;
     };
     focusAccessible: {
-        color: string;
-        backgroundColor: string;
-        borderColor: string;
+        fg: ColorHelper;
+        bg: ColorHelper;
+        borderColor: ColorHelper;
     };
 }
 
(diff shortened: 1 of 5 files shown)


library/src/scripts/styles/buttonStyles.ts
```diff
@@ -39,34 +39,36 @@ export function buttonStyles(theme?: object) {
...
-        width: string;
+        width: Unit;
         style: string;
-        radius: string;
+        radius: Unit;
     };
     hover: {
-        fg: string;
-        bg: string;
+        color: string;
```
slafleche: Why change this? fg can bd an icon color, or a border, I like it better the way it is.

library/src/scripts/styles/buttonStyles.ts
```diff
@@ -39,34 +39,36 @@ export function buttonStyles(theme?: object) {
...
+
 export interface IButtonType {
     fg: string;
     bg: string;
-    spinner: ColorHelper;
+    spinner: string;
     border: {
         color: string;
-        width: string;
+        width: Unit;
```
slafleche: We should use the proper types from TypeStyle (`WidthProperty`) instead.

library/src/scripts/styles/buttonStyles.ts
```diff
@@ -39,34 +39,36 @@ export function buttonStyles(theme?: object) {
     return { padding, sizing, border };
 }
 
+type Unit = string | number;
+
 export interface IButtonType {
     fg: string;
     bg: string;
-    spinner: ColorHelper;
+    spinner: string;
```
slafleche: Why are you getting rid of ColorHelper? It's better, we can modify it.

library/src/scripts/styles/buttonStyles.ts
```diff
@@ -78,7 +80,7 @@ export function buttonVariables(theme?: object) {
     const standard: IButtonType = {
         fg: globalVars.mainColors.fg.toString(),
         bg: globalVars.mainColors.bg.toString(),
-        spinner: globalVars.mainColors.primary,
+        spinner: globalVars.mainColors.primary.toString(),
```
slafleche: I had issues not doing a toString... How come you don't?

library/src/scripts/styles/globalStyleVars.ts
```diff
@@ -10,6 +10,8 @@ export const globalVariables = (theme?: object) => {
     const colorPrimary = color("#0291db");
     const themeVars = componentThemeVariables(theme, "globalVariables");
 
+    const baseUnit = 6;
```
slafleche: This doesn't really apply anymore, we don't really use it to scale things and I find it's less easy to read than a real value. I was getting rid of it.

charrondev:
@slafleche I've addressed all of your feedback.

PR #8449, opened 2019-02-22
First batch of converting components to use CSS in JS

slafleche:
This is the first batch of updates for converting our styles from SASS to JS.


charrondev:
Some visual regressions I noticed here:

## Splash

- The independent search box does not have the same border colour all around. The border here should be white.

|Actual|Mockups|
|---|---|
|image|image|

It should be white all the way around.

## Mobile Header

|Actual|Mockups|
|---|---|
|image|image

- Gutter padding is too large.
- Padding between user icon and search seems to be off.
- **Back button disappears on hover**

|Actual|Mockups|
|---|---|
|image|image|

- Spacing
- Icon 



## Desktop MeBox

|Actual|Mockups|
|---|---|
|image|image|

- Button colour is incorrect.
- Spacing to the search icon is incorrect.

|Actual|Mockups|
|---|---|
|image|image


- Placeholder text doesn't appear
- Incorrect color & border radiuses



- A bunch of extra padding on the top and bottom of the dropdown.
- This also seems to affect the article dropdown

## Nav Manager

|Actual|Mockups|
|---|---|
|image|image|

- Too much horizontal padding.

Issue #8379, opened 2019-02-13
Normalize Reaction model


alex-mtl:
Move `ownerType, RecordType and reactionType` into dedicated model reactionTypeModel

Relates to: #8294

PR #8454, opened 2019-02-25
Normalize reaction Model

alex-mtl:
Move ownerType, reactionType and RecordType to separate table.

# WARNING
We need to drop GDN_reaction table and reinitialize it after since few columns where deleted and we can't delete them through utility/update. 
 
Relates to: #8379


charrondev:
Did you test this?

I'm getting errors like the following when requesting an article:

`Column 'ownerType' in where clause is ambiguous`

PR #8457, opened 2019-02-26
Implement form error component

charrondev:
image

This PR lays the groundwork for more robust error handling.

- Adds a `<FormError />` component and accompanying styles.
- Modifies `IApiError` type definition to be more accurate. It's actual an AxiosError which _potentially_ has a Server error nested inside. Some cases where this assumption was not true was if the network time out or a some edge node (caching server, etc) throws a 500, or we have a fatal error early enough in the process that we don't return a full API error.
- Modifies `getGlobalErrorMessage` and `getFieldErrors`  to account for the new error. I update the usages in the authentication form to match the new signature.


library/src/scripts/components/forms/formElementStyles.ts
```diff
@@ -58,6 +62,19 @@ export function formElementsVariables(theme?: object) {
         ...themeVars.subComponentStyles("colors"),
     };
 
+    const errors = {
+        bg: color("#FFF3D4"),
```
slafleche: Use `feedbackColors` from globalStyleVars

charrondev: Ok I've done so but needed to update them. Luckily they've been made consistent already in all mockups so I've used those.

PR #8464, opened 2019-02-28
Backport: Make page scroll down to embedded comments when user logs into embedded comments.

nervoustwit:
Backport of #8423

> Pass the DOM element id of the embedded comments (#vanilla-comments) to the sign in URL on embedded comments. This hash will be passed as part of the redirect URL through the log in workflow so that when the authenticated user will come back to the page where the embedded comments are they will be brought to the commenting form.


nervoustwit:
We will release this in the next release, no backport.

Issue #8465, opened 2019-02-28
Add consistent navigation to the keystone header


charrondev:
Mobile & Desktop should both have the following links:

- Community.
- Help (with KB) if enabled.

ElisaAlmeida:
Adam,

please find below the link with the mockup:
Desktop:
https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5c7ff205ede03b34df4a5476
Mobile:
https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5c7ff205081f474d0097dfe3

charrondev:
Ah oops, looks like were sharing the same task. This is my sprinted task for implementation of the feature.

PR #8475, opened 2019-03-02
Fix Advanced Editor flyouts on mobile with keystone

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8465

This whole legacy flyout situation is probably the biggest definition of spaghetti code that I've ever seen. I'm very glad we don't use it anymore.

Heres what I ended up doing:

- Improve the selectors of the keystone style flyout handling to close more things. Advanced Editor and core fight over settings handlers a lot, so I'm sorry if it's kind of ugly.
- A kludge for the WYSIWYG editor was covering too many selectors and causing flyouts to close and open again immediately. I narrowed the scope of that event handler.

## Testing

I've manually tested various permutations of the following:

- **Themes:** Keystone, Deflector, Basic Mobile
- **Editors:** Markdown, BBCode, WYSIWYG
- Desktop & Mobile devices.

### DIFF
--- a/js/flyouts.js
+++ b/js/flyouts.js
@@ -29,11 +29,18 @@
         $(document).delegate(".Hijack, .js-hijack", "click", handleHijackClick);
         $(document).delegate(".ButtonGroup > .Handle", "click", handleButtonHandleClick);
         $(document).delegate(".ToggleFlyout", "click", handleToggleFlyoutClick);
-        $(document).delegate(".ToggleFlyout a", "mouseup", handleToggleFlyoutMouseUp);
-        $(document).delegate(".mobileFlyoutOverlay", "click", closeAllFlyouts);
-        $(document).delegate(".Flyout", "click", function(e) {
+        $(document).delegate(".ToggleFlyout a, .Dropdown a", "mouseup", handleToggleFlyoutMouseUp);
+        $(document).delegate(".mobileFlyoutOverlay", "click", function (e) {
+            e.preventDefault();
+            e.stopPropagation();
+            closeAllFlyouts();
+        });
+        $(document).delegate(".Flyout, .Dropdown", "click", function (e) {
             e.stopPropagation();
         });
+        $(document).on("click", function (e) {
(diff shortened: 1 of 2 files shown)


michaeltyson:
This fixes the Advanced Editor but seems to break at least the "Edit Profile" flyout on a profile page. The menu appears on tap, but disappears when trying to select an item (like Ban - guess it's spammers' lucky day!)

charrondev:
@michaeltyson I think I've got it fixed. I was a bit overzealous with the event handler which affected this page.

michaeltyson:
Yep! That's got it - cheers, @charrondev

michaeltyson:
Ah! It's also broken the top nav menu on mobile. Tapping the menu items now shows nothing (see it at forum.audiob.us)

michaeltyson:
Ah - it's broken the flout from the mobile nav bar. It can't be dismissed any more (see it at forum.audiob.us)

charrondev:
Got it again and double checked all of the interactions. I really wish we could just use our `<DropDown />` component but that would be a hell of a re-factoring.

js/flyouts.js
```diff
@@ -263,6 +272,7 @@
      * Close all of the flyouts unless we are clicking on a button inside of a flyout.
      */
     function handleToggleFlyoutMouseUp() {
+        console.log("flyout mouseup");
```
michaeltyson: Was that meant to stay in?

charrondev: Good catch. We don't have our linter setup on these files either, which adds to the pain.

It's a bit tough, being used to work in unit tested, CIed coded and coming into these.

michaeltyson:
Yep, that seems to have fixed it! Not sure if it's related, but on the discussion pages, the flyout menu doesn't appear at all (https://open.vanillaforums.com/discussion/37087/profile-menu-on-keystone-on-mobile-hidden-on-discussion-pages#latest). This hasn't changed with this PR, though.

PR #8480, opened 2019-03-04
AddonUpload - Address "Could not parse addon info array." error being thrown when addon.json exists

cchabilall83:
When that uploading an addon an _"Could not parse addon info array."_  is thrown if the addon.json isn't set to the first entry in the entries array.  The $info variable we are passing to `UpdateModel::checkAddon` is `false` if none of the conditions in the IF block are met.   `UpdateModel::checkAddon` will then throw the error because it is expecting an array.  This checks if $info is `false` and is a valid array, if it is we iterate through the foreach again.

To reproduce:
Pre-requisite : Community Repo
1 ) Enable Addons Directory Application
2) navigate to /addon and try to upload the zip file below

Closes vanilla/community#145
Related to https://open.vanillaforums.com/discussion/36924/could-not-parse-addon-info-array-posting-update-to-addon#latest

### DIFF
--- a/applications/dashboard/models/class.updatemodel.php
+++ b/applications/dashboard/models/class.updatemodel.php
@@ -196,11 +196,11 @@ class UpdateModel extends Gdn_Model {
                         $info = self::parseInfoArray($entry['Path']);
                     }
 
-                    if ($info === false || !is_array($info)) {
-                        continue;
+                    $result = self::checkAddon($info, $entry);
+                    if (!empty($result)) {
+                      continue;
                     }
 
-                    $result = self::checkAddon($info, $entry);
                     $addon = self::buildAddon($info);
                 }
             }
--- a/applications/dashboard/models/class.updatemodel.php
+++ b/applications/dashboard/models/class.updatemodel.php
@@ -148,19 +148,6 @@ class UpdateModel extends Gdn_Model {
                 $deleteEntries = true;
             }
 
-            // Set addon.json to the first entry if it exists and isn't already.
-            // This is a patch to avoid parseArrayInfo errors because getInfoZip is returning the entries
-            // in the right order.
-            $names = array_column($entries, 'Name');
-            $addonJSONExists = in_array('/addon.json', $names);
-
-            if ($entries[0]['Name'] !== '/addon.json' && $addonJSONExists) {
-                $tempArray = $entries[0];
-                $arrayKey = array_search('/addon.json', $names);
-                $entries[0] = $entries[$arrayKey];
-                $entries[$arrayKey] = $tempArray;
-            }
-
             foreach ($entries as $entry) {
                 if ($entry['Name'] == '/environment.php') {
                     // This could be the core vanilla package.
@@ -209,10 +196,11 @@ class UpdateModel extends Gdn_Model {
                         $info = self::parseInfoArray($entry['Path']);
                     }
 
-                    $result = self::checkAddon($info, $entry);
-                    if (!empty($result)) {
-                        break;
+                    if ($info === false || !is_array($info)) {
+                        continue;
                     }
+
+                    $result = self::checkAddon($info, $entry);
                     $addon = self::buildAddon($info);
                 }
             }
--- a/applications/dashboard/models/class.updatemodel.php
+++ b/applications/dashboard/models/class.updatemodel.php
@@ -198,7 +198,7 @@ class UpdateModel extends Gdn_Model {
 
                     $result = self::checkAddon($info, $entry);
                     if (!empty($result)) {
-                      continue;
+                        continue;
                     }
 
                     $addon = self::buildAddon($info);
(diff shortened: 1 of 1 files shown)


cchabilall83:
[UserAgent.zip](https://github.com/vanilla/vanilla/files/2927613/UserAgent.zip)

dannimajdalani:
@nervoustwit I am getting Fatal Error in PHP.gdn_ErrorHandler();  The file is added to GDN_Addon.
Are you getting this error ? i'll will debug it next.
The error occurred on or near: /srv/vanilla-repositories/vanillaspawn/library/database/class.database.php

```
Variables in local scope:
[sql] 'select c.AddonVersionID as `AddonVersionID`, c.CoreVersionID as `CoreVersionID`, COUNT(c.ConfidenceID) as `TotalVotes`, SUM(c.Weight) as `TotalWeight`, AVG(c.Weight) as `AverageWeight`, av.Version as `CoreVersion`
from `GDN_Confidence` `c`
join `GDN_AddonVersion` `av` on c.CoreVersionID = av.AddonVersionID
where `c`.`AddonVersionID` = :cAddonVersionID
 and `c`.`CoreVersionID` is null
group by `c`.`AddonVersionID`'
[inputParameters] array (
  ':cAddonVersionID' => 3,
)
[options] array (
  'Type' => 'select',
  'Slave' => NULL,
  'ReturnType' => 'DataSet',
)
[returnType] 'DataSet'
[tries] 2
[try] 0
[pDO] array (
)
[pDOStatement] false
[state] '42000'
[code] 1055
[message] '\'vanillafactory_h1.c.CoreVersionID\' isn\'t in GROUP BY'
```

charrondev:
@dannimajdalani I have the fix here but you could do the same.

https://github.com/vanilla/community/pull/147/files#diff-7006a6fed011f21027082df9a1079831L112

It's a strict mode error. Just add it to GroupBy.

linc:
> This error seems to be because of the number of lines in the addon.json file.

Can you explain how that's the case?

Did you identify what the logical error is that makes it such that having addon.json not-first causes a problem?

cchabilall83:
>Can you explain how that's the case?
>Did you identify what the logical error is that makes it such that having addon.json not-first causes a >problem?

When using the addon.json attached to the PR,  UpdateModel::getInfoZip creates an array $entries 

`$entries =>[
  0 => [ 
              ['Name'] => '/class.useragent.plugin.php',
              ['Path'] => '...',
              [''] => ,
            ],
  1 =>    ['Name'] => '/addon.json',
              ['Path'] => '...',
              [''] => ,
            ],
]`

if index 0 is not addon.json we will end up here

https://github.com/vanilla/vanilla/blob/be968a0c3980140b61f3c559e096393368b09abc/applications/dashboard/models/class.updatemodel.php#L196

where $entry is $entries[0], UpdateModel::parseInfoArray tries to parse the Path and returns null which in turn causes 
self::checkAddon to throw the error.

When '/addon.json' is in the first index $entries[0].  We arrive here 

https://github.com/vanilla/vanilla/blob/be968a0c3980140b61f3c559e096393368b09abc/applications/dashboard/models/class.updatemodel.php#L193

where the json file is correctly parsed.  

If $entries[0]['Name'] is not '/addon.json' the the logic block below fails.  $entries in this case is built based off of UpdateModel::getInfoZip.  If I alter the addon.json and add a blank line, $entries is in the correct order and the addon will be built.

This is a patch meant to reorganize the $entries array so that addon.json entries is set to first index.  I wanted to avoid re-structuring the method, to keep backwards compatibly.

linc:
> where `$entry` is $`entries[0]`, `UpdateModel::parseInfoArray` tries to parse the `Path` and returns `null` which in turn causes `self::checkAddon` to throw the error.

Wouldn't it make more sense to add a check so it doesn't error? That would let the loop simply continue until it successfully found what it was looking for, which seems like why we're doing a foreach loop and not just looking at the first entry.

linc:
If I were to guess, I'd say `checkAddon` didn't used to throw an error when it got an invalid file and allowed the loop to continue. That, or this never worked correctly and we've just been lucky that it was always first. Of those options, the former seems more likely.

linc:
In this last revision, I wasn't clear what the `break` was originally for nor why you removed it.

So, I went and looked at a blame of the `checkAddon` method and saw it's never been modified since its creation, implying there's no new bug. Then I looked at its execution and I believe the docblock is in error - it can only return an array, full stop. And in fact, that array is only ever error messages. So then, why can't our call to it handle that obvious and intended return condition?

I believe that's what the `if` statement you delete in that last commit was intended for. But there is one critical typo: That `break` was intended to be a `continue`. If you just changed that one word and left everything else as-is, this bug would disappear I believe.

cchabilall83:
@linc changing the break for a continue has resolved the issue.

linc:
FYI I've enabled Squash Merging for future situations where we want to simplify the commit history in a PR where it could otherwise be misleading.

PR #8481, opened 2019-03-04
Add native JS event for iCheck change event

initvector:
Vanilla's dashboard makes use of [iCheck](https://github.com/fronteed/icheck), a jQuery plugin for customizing checkboxes and radio buttons. Part of what the plugin does includes hijacking the native JavaScript events triggered by the controls when they're clicked or changed and re-triggering them as jQuery events. Those jQuery events cannot be easily handled by native JavaScript. Since Vanilla trying to move away from using jQuery, this isn't an ideal situation. As such, we're implementing a re-firing of these jQuery events as native JavaScript events.

### DIFF
--- a/applications/dashboard/js/dashboard.js
+++ b/applications/dashboard/js/dashboard.js
@@ -1,169 +1,170 @@
-(function($, window, document, undefined) {
-    "use strict";
+;(function ($, window, document, undefined) {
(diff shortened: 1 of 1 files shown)


slafleche:
Even better would be to replace them with the label technique. No JS required.

PR #8486, opened 2019-03-05
Fixed height of button on splash component search

slafleche:
Fix for height of search button in Splash of KB home page.

### DIFF
--- a/library/src/scripts/styles/searchBarStyles.ts
+++ b/library/src/scripts/styles/searchBarStyles.ts
@@ -30,7 +30,7 @@ export function searchBarVariables(theme?: object) {
     };
 
     const sizing = {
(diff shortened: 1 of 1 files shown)


library/src/scripts/styles/searchBarStyles.ts
```diff
@@ -30,7 +30,7 @@ export function searchBarVariables(theme?: object) {
     };
 
     const sizing = {
-        height: formElementVars.sizing.height,
+        height: 40,
```
alex-mtl: I hope that is temporary solution, I mean hardcoded `40`

charrondev: It's not a temporary solution. This is the declaration of the constant. `40` is the size specified in the mockups.

PR #8492, opened 2019-03-06
Fix invalid rich posts being created while uploads are in-progress

### LINKED ISSUE #8419: Rich Editor posting invalid json
Currently there are some posts are being saved with invalid json in the body.  This is causing requests to fail and throw 500 errors.  I'm in the process of gathering more data and reproduction setups.  

details can be seen in Vanilla/Support#212.

We should do the following:

1) Prevent invalid json from being posted
2) Fortify the filter method we are using currently.

More info to come.

cchabilall83:
When the using the rich editor there is an issue when trying to post a discussion or comment while a file upload isn't complete.  This arises with slower connections and high traffic sites.  Trying to post a discussion\comment when an upload isn't complete will cause and throw an error and you won't be able to reach that tread after.  This is because the JSON in the body isn't saved correctly and fails when it's being render.

The address this a small change was made to the filterer to resort the operations that are being built for for the parser.  A Filterer unit test class was created ensure it is working.  The Parser was fortified so that if some invalid JSON was posted, that error will be caught and logged, instead of throwing the stoping the render.  This is was done with the assistance of @charrondev 

**Steps to reproduce**

1) Create new discussion 
2) In Chrome\FireFox throttle your network connection
I put 150kb dl, 150kb ul
3) Try to upload a file
4) While the file is still uploading, try to post the discussion

Closes #8419


dannimajdalani:
Tested locally and on CL10012

### locally
Screen Shot 2019-03-11 at 9 01 57 AM



### On the test cluster
 no error, but the attachment was stripped from the comment when i simulated a slow connection (10Kb/s)

cchabilall83:
This is normal we are throwing an error and logging it if there are issues uploading files on a slow connections.  If you turn debug off, the error will be caught and logged.

library/Vanilla/Formatting/Formats/RichFormat.php
```diff
@@ -61,12 +61,12 @@ public function __construct(Quill\Parser $parser, Quill\Renderer $renderer, Quil
     public function renderHTML(string $content): string {
         try {
             $operations = Quill\Parser::jsonToOperations($content);
-        } catch (FormattingException $e) {
+            $blotGroups = $this->parser->parse($operations);
+            return $this->renderer->render($blotGroups);
+        } catch (\Throwable $e) {
```
initvector: Is there no way to narrow this down?

charrondev: It is intentional here and it can't be narrowed down because it literally is meant to catch _every single fatal error_ and convert them to a warning. There is __no__ circumstance where we want some bug in post formatting to take down a page.

initvector: I appreciate the intention of avoiding bad formatting taking down a page, but I think there are better ways to accomplish it than hiding every single error and exception encountered during the formatting step. This is covering up points of failure that might be resolved with defensive coding practices or at least more clearly indicated with specific exceptions.

PR #8497, opened 2019-03-06
Add logo and mobileLogo to theme API resource responses

initvector:
The themes API resource has been updated to include the site's desktop and mobile logos, if present.

### Overview
1. Add `logo` and `mobileLogo` asset to themes API resource responses. These assets are only included if the site has configured them.
1. Add `ImageAsset` class to represent an image asset of a resource.
1. Add tests for verifying the proper desktop and mobile logos are returned by the API when configured on a site.
1. Update OpenAPI spec for the themes resource to include the new assets.
1. Fix location of `ThemesApiController`. This controller was previously under Vanilla. Now, it lives under Dashboard.

### Notes
The original spec had "logo" and "mobileLogo" as top-level fields in the response of a theme. However, they were flagged as TBD, which I take to mean they were only there as placeholders for future discussion. After further internal conversation, it has been decided the logos will be presented as assets. As such, the previous top-level keys have been removed.

### DIFF
--- a/tests/APIv2/ThemesTest.php
+++ b/tests/APIv2/ThemesTest.php
@@ -6,6 +6,8 @@
 
 namespace VanillaTests\APIv2;
 
+use Gdn_Configuration;
+use Gdn_Upload;
 use Vanilla\Addon;
 use Vanilla\AddonManager;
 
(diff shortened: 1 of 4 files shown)


tests/APIv2/ThemesTest.php
```diff
@@ -82,4 +84,32 @@ public function testGetByName() {
...
     }
+
+    /**
+     * Test getting a theme's logo.
+     *
+     * @depends testGetByName
+     */
+    public function testLogo() {
+        $logo = "logo.png";
+        self::container()->get(Gdn_Configuration::class)->set("Garden.Logo", $logo);
```
charrondev: I don't know if you're aware but we do have a `MockConfig` object.

See usage https://github.com/vanilla/vanilla/blob/d79e5924f7d0ccc7df8f7e27ad65d44ad892c947/tests/Library/Vanilla/Formatting/GdnFormatLinksTest.php#L35

initvector: This would be great for tests with minimal config requirements. Unfortunately, our API integration tests require a lot of additional keys to get the site up and running and in a state where it can respond to API requests. Using `MockConfig` for API/integration tests is probably something we could look into doing in the future, but would require some more groundwork that's outside the scope of this PR.

PR #8502, opened 2019-03-08
Fix application folder name under Windows 

### LINKED ISSUE #7263: Cannot enable Applications on Windows/XAMPP
Windows/XAMPP

After resolving issue of Vanilla plugin not being enabled (https://github.com/vanilla/vanilla/issues/7262) cannot enable Vanilla applications.

In Dashboard, the slider moves across and the setting is written to conf.php

// EnabledApplications
$Configuration['EnabledApplications']['Vanilla'] = '/applications/vanilla';
$Configuration['EnabledApplications']['Conversations'] = '/applications/conversations';

However, the application (e.g. conversations) doesn't work, returning a 'Page not found' message if the envelope icon is clicked.

Returning to the Dashboard, the slider is reset to off, seemingly ignoring the conf.php setting:

austins:
Fixes #7263.

### Bug Description

Vanilla currently returns an incorrect folder name for applications when Vanilla is run on Windows. This causes applications to not be enabled because Vanilla depends on a properly set value for elements in the EnabledApplications array in config.php.

**Actual values on Windows:**
```
$Configuration['EnabledApplications']['Conversations'] = '/applications/conversations';
$Configuration['EnabledApplications']['Vanilla'] = '/applications/vanilla';
```

**Expected values:**
```
$Configuration['EnabledApplications']['Conversations'] = 'conversations';
$Configuration['EnabledApplications']['Vanilla'] = 'vanilla';
```

The Conversations and Vanilla applications are just two examples to show how the EnabledApplications array looks like on Windows.

On Linux environments, the values are good as expected.

You can reproduce this on a Windows environment by (1) a new installation of Vanilla or (2) by enabling any application via the dashboard.

**Environments used for reproducing this issue:**
1. Windows Server 2019, IIS 10, PHP 7.2.7 (VC15 x64), MariaDB 10.3.13
2. Windows 10 with WAMP/XAMPP/Laragon software stack. My Laragon setup has: Apache 2.4.37, PHP 7.3.2, MariaDB 10.3.13

### Fix Description

Note: Windows supports both backslashes and forward slashes as a path separator.

$addon->getSubdir() returns a string with forward slashes (e.g. `/applications/conversations`) in both Windows and Linux.  DS is `\` on Windows and `/` on Linux; therefore, `explode(DS, $addon->getSubdir())` won't work as intended. I suggest changing DS to `'/'`.

### DIFF
--- a/library/core/class.applicationmanager.php
+++ b/library/core/class.applicationmanager.php
@@ -101,7 +101,7 @@ class Gdn_ApplicationManager {
      */
     private function calcOldInfoArray(Addon $addon) {
         $info = Gdn_pluginManager::calcOldInfoArray($addon);
-        $directories = explode(DS, $addon->getSubdir());
+        $directories = explode('/', $addon->getSubdir());
         $info['Folder'] = $directories[count($directories) - 1];
 
         return $info;


charrondev:
@tburry What do you think of us just updating that constant across the application?

tburry:
Hmm. I’m not against trying to update DS but that could have a big
regression impact. Truthfully we want to just stop using DS in favor of “/“
so this change is good in and of itself imo.

Issue #253, opened 2010-02-22
Undefined function imagecreatefromjpeg()


JoshuaChi:
When you try to upload a new photo in user profile page, the error was outputed.
Fatal error: Call to undefined function imagecreatefromjpeg() in Garden\library\core\class.uploadimage.php on line 95

PR #8509, opened 2019-03-08
Add index to GDN_Comment.Score

dannimajdalani:
closes [#253](https://github.com/vanilla/support/issues/253)

Details in the referenced ticket.

Score column had no index causing slow queries post deploy.

### DIFF
--- a/applications/vanilla/settings/structure.php
+++ b/applications/vanilla/settings/structure.php
@@ -212,7 +212,7 @@ $Construct
     ->column('InsertIPAddress', 'ipaddress', true)
     ->column('UpdateIPAddress', 'ipaddress', true)
     ->column('Flag', 'tinyint', 0)
-    ->column('Score', 'float', null)
+    ->column('Score', 'float', null, ['index'])
     ->column('Attributes', 'text', true)
     //->column('Source', 'varchar(20)', true)
     ->set($Explicit, $Drop);


R-J:
An index in Score doesn't make sense in **application Vanilla** since it isn't used there. That index should be created when Reactions (I guess) is enabled. 


[Although it can be tempting to create an indexes for every possible column used in a query, unnecessary indexes waste space and waste time for MySQL to determine which indexes to use.] (https://dev.mysql.com/doc/refman/5.5/en/optimization-indexes.html)

nervoustwit:
I understand that too many indexes is the same as not enough. The comments table is far from having too many. Score is searched on in more than just Reactions. I think we could index it regardless of if a Forum uses Reactions.

R-J:
So you basically say: _I know that this will hurt performance, but here at this place we haven't yet wasted resources and therefore we'll do it anyway._ 

I know that this [most probably will not be noticed](https://redmondmag.com/articles/2014/07/16/too-many-sql-server-indexes.aspx) by anyone but the same goes for most of the performance tweaks. Nevertheless nobody would argue that improving performance is superfluous.

In the end I'm just after it because I like a clean separation of applications and plugins and thus keep everything focused and efficient. But that's something I'm a little bit obsessed with...
Having Score indexed can be an advantage if that column is used later on by another plugin and at that point someone doesn't think of adding an index to it.
But until then every OS forum will have an additional, useless index.

PR #8510, opened 2019-03-08
Implement theming header, footer, styles, and fix panel scrolling issues

charrondev:
Required companion PR https://github.com/vanilla/knowledge/pull/771

Closes https://github.com/vanilla/knowledge/issues/529
Closes https://github.com/vanilla/knowledge/issues/680

_I recommend testing using the `chip-test` theme in this PR https://github.com/vanilla/themes/pull/265. I provided a header/footer/styles and some variables in it._

- `Add themeHeader` & `themeFooter` each with the contents of the theme styles inside of a `<noscript />` tag.
- When the `<ThemeProvider />` mounts it will:
  - Take the contents of the header & footer
  - Create a new `<header>` or `<footer>` tag to replace the `<noscript>` tag.
  - Create a new shadow root inside of the them to scope their contents.
  - Place the HTML contents from the nocript inside.
- Update the header to use position sticky. This allows it sit relatively under the client header.
- Update the fixed panels on the side of the `PanelLayout` to use position sticky. The complicated and buggy (see https://github.com/vanilla/knowledge/issues/680) bespoke sticky implementation was removed.
- Adds a webcomponent polyfill for ShadowDOM.

I've also fixed some cross browser bugs I found while testing

- Fix SiteNav styles in Safari.
- Fix a javascript error preventing the knowledge base from loading in IE11.
- Fix url generation in `<SmartLink />` in IE11 by using the `Url` class. (polyfilled by the webcomponent poylfill).


resources/views/default-master.twig
```diff
@@ -42,9 +42,11 @@
     <body class="{{ cssClasses|join(' ') }}">
         <div id="page" class="page">
             {%- if header -%}
-                <div id="themeHeader">{{ header|raw }}</div>
+                <noscript id="themeHeader">
+                    {{ header|raw }}
+                </noscript>
             {%- endif -%}
-            <div id="vanillaHeader"></div>
+            <header id="vanillaHeader"></header>
```
slafleche: Don't put "header" or "footer" tags for the client. As covineant as it may seem, many clients want to drop in their headers/footers without having to touch the HTML/CSS. It's likely to cause duplicate header/footer tags.

slafleche: We could, however, find an easy way to include it, or detect if it exists in the client's HTML.

charrondev: This our own header. I've removed the header inside that was created by react, because the extra level of nesting was causing `position: sticky` to fail and `display: contents` is not widely supported yet.

charrondev: End result for example
image

PR #8512, opened 2019-03-11
DiscussionModel - fire modifyQuery Event

cchabilall83:
This fires the modifyQuery event to allow for addons to hook in and modify queries in the DiscussionModel::GetWhere and DiscussionModel::GetCount methods.

Require for vanilla/internal#1834


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -710,6 +710,8 @@ public function getWhere($where = false, $orderFields = '', $orderDirection = ''
 
         $sql = $this->SQL;
 
+        $this->fireEvent('modifyQuery');
```
initvector: Why couldn't you use the "BeforeGet" event?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1550,6 +1552,7 @@ public function getCount($wheres = [], $unused = null) {
         }
 
         $wheres = $this->combineWheres($this->getWheres(), $wheres);
+        $this->fireEvent('modifyQuery');
```
initvector: Why couldn't you use the "BeforeGetCount" event?

applications/vanilla/models/class.discussionmodel.php
initvector: If you need to introduce a new event, naming it "modifyQuery" is probably going to be too generic. There's isn't any context about what type of query you're attempting to modify (e.g. delete, update, insert). It would also be beneficial to know where it is being fired (e.g. some way to distinguish between the event being fired in `getWhere` versus `getCount`).

cchabilall83:
I will use BeforeGet and BeforeGetCount, closing this PR

PR #8514, opened 2019-03-11
JS Styles - body, button, dropdown, frame, loader, mebox, meta

slafleche:
More conversions of styles from CSS to JS for KB. This PR mostly touches:

- body
- buttons
- dropdowns
- frames
- loader
- mebox
- meta


charrondev:
I'm noticing a few issues here:

- MeBox notification and messages dropdowns seem a bit messed up. Opening makes the header break or disappear, layouts are misaligned, and the hover effects are missing.

image

- There are some issues with the default flyout styles now:
  - No dropshadow
  - Hover colours are incorrect.
  - Sometimes no background colour.

image

image

- The header search has a couple of issues
  - Missing background
  - misalignment of the X

image

PR #8519, opened 2019-03-12
Theme Vars - Modals, buttons and category management

slafleche:
Here are the converted SASS for the KB's modals, buttons and category management in JS.


charrondev:
A few things I've noticed:

- Modals & MeBox dropdowns are missing their `minHeight` now. Aside from being off-spec this also causes them to loaders inside of them to get cut off.

charrondev:
This PR also seems to have busted up the category management styles a bit.

|Before|After|
|---|---|
|image|image

charrondev:
Sorry there's one more.

image

The editor title doesn't seem to implement :focus-visible properly. It always displays a focus style.

Issue #8466, opened 2019-02-28
Convert user content & embeds to typestyle


charrondev:
Convert all user content & embeds currently written in SASS to typestyle so they work with the upcoming theme changes.

PR #8520, opened 2019-03-13
Implement dynamically themed user content styles

charrondev:
Requires companion PR https://github.com/vanilla/knowledge/pull/779

Closes https://github.com/vanilla/vanilla/issues/8466

- Converts the core user content styles to use typestyles.
- Paragraphs, lists, inline styles, & code blocks are completely converted.
- At the end I started running out of time, so only the bare minimum of colour conversions were done for embeds, blockquotes & spoilers.
- Numerous style fixes were implemented for embed loading and error states.
- Fixed a bug where code syntax highlighting was not applying properly.
- User content is mostly usable in storybook now. I've also setup some core styles to wrap all stories. Try `yarn storybook`.
- Some sass partials & normalize have been moved into core. A kind of shared base partial has been created `_base.scss`. It's basically our "reset".


library/src/scripts/user-content/userContentStyles.ts
```diff
@@ -0,0 +1,282 @@
...
+    });
+
+    return { fonts, list, blocks, code, codeInline, codeBlock, embeds };
+});
+
+/**
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+export const userContentStyles = useThemeCache(() => {
```
slafleche: shoudl be suffixed "Classes". "Styles" is for the file name

PR #8526, opened 2019-03-13
Add logo support to dynamic theming

initvector:
Frontend components were recently updated to utilize the theming API. However, the recent introduction of the `logo` and `mobileLogo` fields have not yet been implemented. This update remedies that.

### Overview
1. Add `logo` and `mobileLogo` to pre-loaded theme data in `ThemedPage`. The `variables` key was previously the only thing sent down. Now, the structure has shifted to account for more assets. Theme variables have been moved into a new `assets` key, along with the logos.
1. Add `logo` and `mobileLogo` to the local store. To mimic the API resource's structure, the top-level `theme` key in the store now has an `assets` key. This key houses the new `logo` and `mobileLogo` values. The previous `theme.variables` key has also been moved under `assets`.
1. Add `ThemeLogo` component for getting the theme's desktop and mobile logos. If the specific theme logo is not found, the default Vanilla SVG will be used.
1. Add `ThemeLogo` usage to `VanillaHeader` and `VanillaMobileHomeHeader`.

Closes vanilla/knowledge#733


library/src/scripts/theming/ThemeProvider.tsx
```diff
@@ -62,9 +66,29 @@ class BaseThemeProvider extends React.Component<IProps> {
...
+    let logo;
+
+    if (logoType === LogoType.DESKTOP) {
+        logo = assets.logo || null;
+    } else if (logoType === LogoType.MOBILE) {
+        logo = assets.mobileLogo || null;
+    }
+
+    if (!logo) {
+        return undefined;
```
charrondev: For consistencies sake can we return null out of this? I'm not the biggest fan of mixing null/undefined together when it's not necessary.

library/src/scripts/components/headers/VanillaHeader.tsx
```diff
@@ -94,6 +95,16 @@ export class VanillaHeader extends React.Component<IProps, IState> {
         const containerElement = this.props.container || document.getElementById("vanillaHeader")!;
         containerElement.classList.value = outerCssClasses;
 
+        const themeLogo = getLogo(LogoType.DESKTOP);
+        let logoData = dummyLogoData;
```
charrondev: Can we rename this delete the `dummyData` file and just put this inline as the default?

library/src/scripts/components/headers/VanillaHeader.tsx
```diff
@@ -94,6 +95,16 @@ export class VanillaHeader extends React.Component<IProps, IState> {
         const containerElement = this.props.container || document.getElementById("vanillaHeader")!;
         containerElement.classList.value = outerCssClasses;
 
+        const themeLogo = getLogo(LogoType.DESKTOP);
```
charrondev: This breaking the encapsulation we get from redux, which we shouldn't do unless we have to. Right now the logo wouldn't update for example if it changed in the store.

It should probably come in through some `mapStateToProps`.

library/src/scripts/components/headers/pieces/VanillaMobileHomeHeader.tsx
```diff
@@ -46,14 +48,24 @@ export class VanillaMobileHomeHeader extends React.Component<IProps> {
         const headerClasses = vanillaHeaderClasses();
         const classes = vanillaHeaderHomeClasses();
 
+        const themeLogo = getLogo(LogoType.MOBILE);
+        let logoData = dummyLogoData;
```
charrondev: I see this is actually duplicated. It might be best to wrap it up a redux connected `<ThemeLogo />` component and share it. It can use redux's connect function to map in the data.

Issue #7180, opened 2018-05-07
Revert #7179


tburry:
Revert #7179 once it has found everything it needs to. Don't do this until July 2018.

PR #8547, opened 2019-03-16
Remove suppression of ArgumentCountError exceptions from Gdn_Pluggable

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7180


charrondev:
None as far as I can tell.

image

Issue #8260, opened 2019-01-23
Add ability to set canonical URL for discussions via the API


initvector:
Vanilla automatically generates canonical URLs for discussions in its `DiscussionController` and `DiscussionsController`. The ability to override this value over the API is  needed.

### Requirements

1. A user should be able to override the canonical URL for a discussion using the discussions resource in API v2. This can only be done once. If the canonical is attempted to be set again, an error is thrown.
1. A message should be displayed at the top of the discussions page if a custom canonical URL has been detected (e.g. "This discussion has a more recent version here."). This message will link to the custom canonical URL. Actually visualization of the UI to be done [here](https://github.com/vanilla/vanilla/issues/8261).
1. Canonical URL tags generated by the page should use the custom canonical URL, if available.

alex-mtl:
Closed by #8556

PR #8556, opened 2019-03-19
Add ability to set the canonical URL for a discussion over the API

alex-mtl:
New DiscussionsAPI endpoints:

```
PUT /discussions/{id}/canonical 

DELETE /discussions/{id}/canonical
```
Update openapi schema.
Update discussion controller  to update meta tag `canonical` with custom `canonicalUrl` when present

Relates to: #8260

### DIFF
--- a/applications/vanilla/controllers/api/DiscussionsApiController.php
+++ b/applications/vanilla/controllers/api/DiscussionsApiController.php
@@ -674,7 +674,7 @@ class DiscussionsApiController extends AbstractApiController {
      * @param int $id The ID of the discussion.
      * @param array $body The request body.
      *       Ex: ["canonicalUrl" => "https://mydomain.com/some+path/"]
-     * @throws NotFoundException if unable to find the discussion.
+     * @throws NotFoundException If unable to find the discussion.
      * @return array
      */
     public function put_canonical($id, array $body) {
@@ -708,7 +708,7 @@ class DiscussionsApiController extends AbstractApiController {
      * Remove canonical url for a discussion.
      *
      * @param int $id The ID of the discussion.
-     * @throws NotFoundException if unable to find the discussion.
+     * @throws NotFoundException If unable to find the discussion.
      * @return array
      */
     public function delete_canonical($id) {
@@ -726,7 +726,7 @@ class DiscussionsApiController extends AbstractApiController {
             $this->discussionModel->categoryPermission('Vanilla.Discussions.Edit', $categoryID);
         }
         $attributes = $row['Attributes'];
-        if(!empty($attributes['CanonicalUrl'] ?? '')) {
+        if (!empty($attributes['CanonicalUrl'] ?? '')) {
             unset($attributes['CanonicalUrl']);
             $this->discussionModel->setProperty($id, 'Attributes', dbencode($attributes));
         }
--- a/applications/vanilla/controllers/api/DiscussionsApiController.php
+++ b/applications/vanilla/controllers/api/DiscussionsApiController.php
@@ -695,7 +695,7 @@ class DiscussionsApiController extends AbstractApiController {
             $this->discussionModel->categoryPermission('Vanilla.Discussions.Edit', $categoryID);
         }
 
-        $attributes = $row['Attributes'];
+        $attributes = $row['Attributes'] ?? [];
         $attributes['CanonicalUrl'] = $body['canonicalUrl'];
         $this->discussionModel->setProperty($id, 'Attributes', dbencode($attributes));
 
--- a/applications/vanilla/controllers/api/DiscussionsApiController.php
+++ b/applications/vanilla/controllers/api/DiscussionsApiController.php
@@ -30,6 +30,9 @@ class DiscussionsApiController extends AbstractApiController {
     private $discussionPostSchema;
 
     /** @var Schema */
+    private $discussionPutCanonicalSchema;
+
+    /** @var Schema */
     private $idParamSchema;
 
     /** @var UserModel */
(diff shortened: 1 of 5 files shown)


applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -645,4 +667,68 @@ public function put_bookmark($id, array $body) {
...
+            throw new \Garden\Web\Exception\ClientException('Canonical url already set for this discussion.', 409);
+        };
+        $categoryID = $row['CategoryID'];
+        if ($row['InsertUserID'] !== $this->getSession()->UserID) {
+            $this->discussionModel->categoryPermission('Vanilla.Discussions.Edit', $categoryID);
+        }
+
+        $attributes = $row['Attributes'];
+        $attributes['CanonicalUrl'] = $body['canonicalUrl'];
+        $this->discussionModel->setProperty($id, 'Attributes', dbencode($attributes));
```
charrondev: Pretty sure this will blow away anything existing in that attributes column? @initvector do we have a standard way of handling this?

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -645,4 +667,68 @@ public function put_bookmark($id, array $body) {
...
+        if (!empty($row['Attributes']['CanonicalUrl'] ?? '')) {
+            throw new \Garden\Web\Exception\ClientException('Canonical url already set for this discussion.', 409);
+        };
+        $categoryID = $row['CategoryID'];
+        if ($row['InsertUserID'] !== $this->getSession()->UserID) {
+            $this->discussionModel->categoryPermission('Vanilla.Discussions.Edit', $categoryID);
+        }
+
+        $attributes = $row['Attributes'];
+        $attributes['CanonicalUrl'] = $body['canonicalUrl'];
```
charrondev: @initvector Do we have some in-built attributes handling in the models? I thought they were supposed to automatically serialize/unserialize these values.

initvector: `DiscussionModel::getID` will handle unserializing the column on the way out. The column should be serialized with `DiscussionModel::save` on the way back in. However, `DiscussionModel::setProperty` doesn't perform any special field formatting and would require this manual re-serializing of the field contents. As an alternative, you might try something like:

```php
$this->discussionModel->save([
    "DiscussionID" => $id,
    "Attributes" => $attributes,
]);
```

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -645,4 +667,68 @@ public function put_bookmark($id, array $body) {
...
+        $row = $this->discussionByID($id);
+        if (!empty($row['Attributes']['CanonicalUrl'] ?? '')) {
+            throw new \Garden\Web\Exception\ClientException('Canonical url already set for this discussion.', 409);
+        };
+        $categoryID = $row['CategoryID'];
+        if ($row['InsertUserID'] !== $this->getSession()->UserID) {
+            $this->discussionModel->categoryPermission('Vanilla.Discussions.Edit', $categoryID);
+        }
+
+        $attributes = $row['Attributes'];
```
initvector: If the row's Attributes column was `null` in the database, I believe it'll be `null` here too. The following assignment seems to rely on PHP converting the `null` to an array when referenced it as an array. I think it would be better if we could implement some null-coalescing in this assignment to default it to an empty array.

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -645,4 +667,68 @@ public function put_bookmark($id, array $body) {
...
+    /**
+     * Set canonical url for a discussion.
+     *
+     * @param int $id The ID of the discussion.
+     * @param array $body The request body.
+     *       Ex: ["canonicalUrl" => "https://mydomain.com/some+path/"]
+     * @throws NotFoundException If unable to find the discussion.
+     * @return array
+     */
+    public function put_canonical($id, array $body) {
```
initvector: We're referring to this as a "canonical URL" just about everywhere else, aside from the endpoint. What do you think about renaming this from "/discussions/{id}/canonical" to "/discussions/{id}/canonical-url"?

applications/vanilla/controllers/api/DiscussionsApiController.php
alex-mtl: I've updated the code, but it was working fine with nulls.

applications/vanilla/controllers/api/DiscussionsApiController.php
alex-mtl: As you wish. Updated

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -645,4 +667,68 @@ public function put_bookmark($id, array $body) {
...
+     * Set canonical url for a discussion.
+     *
+     * @param int $id The ID of the discussion.
+     * @param array $body The request body.
+     *       Ex: ["canonicalUrl" => "https://mydomain.com/some+path/"]
+     * @throws NotFoundException If unable to find the discussion.
+     * @return array
+     */
+    public function put_canonicalUrl($id, array $body) {
+        $this->permission('Garden.SignIn.Allow');
```
initvector: We should really get some stricter permissions on this endpoint.

alex-mtl: That is kind the same way we have for discussions `post()`
https://github.com/vanilla/vanilla/blob/29ab4e729ac0b14fb19998a4f3ce923cd5591d7d/applications/vanilla/controllers/api/DiscussionsApiController.php#L596

initvector: You're absolutely right. I skimmed over the permission checks you have, further down. Sorry!

PR #8566, opened 2019-03-25
[ProfileExtender] Translate field label with locale definition

vintprox:
Just guessing it's necessary to translate field labels everywhere including profile section.

I just wrapped label with `t` function.


R-J:
Thanks for the contribution! You have used the handy `t()` function but all those helper functions will be deprecated and should be avoided. Please see [this issue](https://github.com/vanilla/vanilla/issues/7696) for more information.

The `t()` function is defined like that:

~~~
function t($code, $default = false) {
    return Gdn::translate($code, $default);
}
~~~

Therefore you can simply use `Gdn::translate()` instead of `t()`.

vintprox:
Ahem, I replaced it, just in case you didn't get a notification or thing

PR #8573, opened 2019-03-25
Add informational message to discussion if it has a user-provided canonical URL

### LINKED ISSUE #8261: Implement discussion canonicalization notification UI
Implement UI for the notification that a discussion has been moved to a knowledge article.

This UI should be implemented as some HTML before the rendered output of the discussion using info message style. For example see the rich editor error HTML.

The text content of this message should be 

```
This discussion has a more [recent version](https://link-to-article)
```

Where the markdown there is an actual link. This should __not__ be part of the discussions contents.

initvector:
The ability to set a discussion's canonical URL was recently added. There we no UI indicators that a discussion could potentially be pointing to a new resource. Those UI indicators are added here.

### Overview
1. Add `Vanilla\Message` class. This is a basic class that represents a site message. It is built to follow the established convention of message classes and structure (see [class.messagecontroller.php](https://github.com/vanilla/vanilla/blob/master/applications/dashboard/controllers/class.messagecontroller.php)).
1. Add backward-compatible display of messages when viewing a discussion. The BeforeDiscussionDisplay event is hooked into and messages are output at that point.
1. Add ability to set messages in `DiscussionController`. Messages may be added any time before rendering. When a discussion is rendered, the messages will be included before it.
1. Add a message to a discussion if a user-provided canonical URL is detected.

### Testing

To see the message on a discussion, you'll need to set its canonical URL. You can do this in a couple ways:
1. Set a discussion's canonical URL using the API (PUT /api/v2/discussions/{id}/canonical-url).
1. Edit a discussion in the database. Set its Attributes column to `{"CanonicalUrl":"/path/to/resource"}`.

Closes #8261

### DIFF
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -172,6 +172,14 @@ class DiscussionController extends VanillaController {
         if (empty($canonicalUrl)) {
             $this->canonicalUrl(discussionUrl($this->Discussion, pageNumber($this->Offset, $Limit, 0, false)));
         } else {
+            $canonicalMessage = new Message(
+                sprintf(
+                    \Gdn::locale()->translate('This discussion has a more <a href="%1$s">recent version</a>.'),
+                    htmlspecialchars(\Gdn::request()->url($canonicalUrl))
+                ),
+                Message::TYPE_INFORMATION
+            );
+            $this->addMessage($canonicalMessage);
             $this->canonicalUrl($canonicalUrl);
         }
         $this->checkPageRange($this->Offset, $ActualResponses);
--- a/applications/vanilla/controllers/class.discussioncontroller.php
+++ b/applications/vanilla/controllers/class.discussioncontroller.php
@@ -173,9 +173,10 @@ class DiscussionController extends VanillaController {
             $this->canonicalUrl(discussionUrl($this->Discussion, pageNumber($this->Offset, $Limit, 0, false)));
         } else {
             $canonicalMessage = new Message(
-                sprintf(
-                    \Gdn::translate('This discussion has a more <a href="%1$s">recent version</a>.'),
-                    htmlspecialchars(\Gdn::request()->url($canonicalUrl))
+                str_replace(
+                    ["<0>", "</0>"],
+                    ['<a href="' . htmlspecialchars(\Gdn::request()->url($canonicalUrl)) . '">', "</a>"],
+                    \Gdn::translate('This discussion has a more <0>recent version</0>.')
                 ),
                 Message::TYPE_WARNING
             );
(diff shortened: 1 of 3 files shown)


initvector:
Need some styling assistance from @slafleche but otherwise this should be good to go.

slafleche:
@initvector How can I test this?

initvector:
> If we're using the old message system, we should be using Message::TYPE_WARNING instead of Message::TYPE_INFORMATION. It's pretty close and Todd said it was fine.

It is done!

applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -158,6 +172,14 @@ public function index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
         if (empty($canonicalUrl)) {
             $this->canonicalUrl(discussionUrl($this->Discussion, pageNumber($this->Offset, $Limit, 0, false)));
         } else {
+            $canonicalMessage = new Message(
+                sprintf(
+                    \Gdn::locale()->translate('This discussion has a more <a href="%1$s">recent version</a>.'),
```
charrondev: Any reason for the longer form here?

I think `\Gdn::translate()` is the standard.

charrondev: Do we have a PR moving adding this to the locales repo?

applications/vanilla/settings/class.hooks.php
```diff
@@ -1110,6 +1110,23 @@ public function pluginController_tagsearch_create() {
...
     }
 
+    /**
+     * Hook in before a discussion is rendered and display any messages.
+     *
+     * @param mixed DiscussionController $sender
+     * @param array array $args
+     */
+    public function discussionController_beforeDiscussionDisplay_handler($sender, array $args) {
+        if (!is_object($sender) || !method_exists($sender, "getMessages")) {
```
charrondev: Any reason we're not just doing an `instanceof` check here?

library/Vanilla/Message.php
```diff
@@ -0,0 +1,76 @@
...
+    /** @var string */
+    private $type;
+
+    /**
+     * Create a new message.
+     *
+     * @param string $body
+     * @param string $type
+     */
+    public function __construct(string $body, string $type) {
```
charrondev: Any way we could reference these as `$htmlBody` or `$unsafeHtmlBody`? At least something indicating that the contents here should be safely processed before being inserted.

I could picture a mistake being made in the future where something one of these is created with some user content reflecting inside of it.

applications/vanilla/controllers/class.discussioncontroller.php
initvector: We do now! See vanilla/locales#118

initvector: My original thought process:
1. I want to avoid using `Gdn` functions where possible.
1. `Gdn::locale` is a `Gdn` function, **but** it could be swapped out with a `Locale` instance if we are ever able to safely use DI with these old controllers.

I'm fine with swapping it out to `Gdn::translate`.

applications/vanilla/settings/class.hooks.php
initvector: Being overly defensive. A `instanceof` check will probably be sufficient.

library/Vanilla/Message.php
initvector: I'm not keen on prefixing a required parameter with the word "unsafe", but I agree with making it more apparent that it will be treated as HTML. I think `$htmlBody` sounds good.

PR #8574, opened 2019-03-26
Fix auto marking categories as read

R-J:
Categories that has never been marked as read weren't automatically marked as read when all discussions in such a discussion have been read.

This PR solves issue https://github.com/vanilla/vanilla/issues/6856

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -716,7 +716,7 @@ class CommentModel extends Gdn_Model {
 
         } else {
             // Make sure the discussion isn't archived.
-            $archiveDate = c('Vanilla.Archive.Date', false);
+            $archiveDate = Gdn::config('Vanilla.Archive.Date', false);
             if (!$archiveDate || (Gdn_Format::toTimestamp($discussion->DateLastComment) > Gdn_Format::toTimestamp($archiveDate))) {
                 $newComments = true;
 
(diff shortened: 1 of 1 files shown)


charrondev:
I tested and validated this locally. I also rebased it on master for you. Sorry for letter it linger so ling @R_J.

applications/vanilla/models/class.commentmodel.php
```diff
@@ -741,27 +741,25 @@ public function setWatch($discussion, $limit, $offset, $totalComments) {
...
         if (!$category) {
             return;
         }
-        $dateMarkedRead = val('DateMarkedRead', $category);
-        if (!$dateMarkedRead) {
-            return;
+        $wheres = ['CategoryID' => $categoryID];
+        $dateMarkedRead = $category->DateMarkedRead;
+        if ($dateMarkedRead) {
+            $wheres['DateLastComment>'] = $dateMarkedRead;
```
alex-mtl: I have some concerns about these changes. It is not 100% equal to previous approach.
Before if we had no `$dateMarkedRead` we returned early and we did not get discussions from `$discussionModel`.
Now we don't do early return and even more we are getting from the model with weaker filtering.
I think that can change user experience.

R-J: I understand those concerns, but exiting at this point was the main reason why the functionality wasn't working as intended. This code blocks purpose is to determine if the category can be marked as read. "Categories that has never been marked as read weren't automatically marked as read when all discussions in such a discussion have been read."
This fix requires that `$dateMarkedRead` is handled in another way.

The UX was broken, therefore a change was needed.

Issue #8575, opened 2019-03-26
Form->color() cannot be created with a value set


R-J:
The default value of the form element always is "#ffffff": `$colorAttributes['value'] = '#ffffff';`. It needs to accept a value parameter in order to be useful for e.g. settings.

It should be easy to implement since the color picker already accepts an $options parameter where "value" can be used in.

PR #8580, opened 2019-03-26
Allow starting values for color picker form control

R-J:
This fixes: https://github.com/vanilla/vanilla/issues/8575

The color picker always started with a default value. With this change the starting value can be passed to the form control like that: ` $form->color('FieldName', ['Value' => '#0099aa'])`

### DIFF
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -381,7 +381,8 @@ class Gdn_Form extends Gdn_Pluggable {
 
         // Default starting color for color input. Color inputs require one, Chrome will throw a warning if one
         // doesn't exist. The javascript will override this.
-        $colorAttributes['value'] = '#ffffff';
+        $colorAttributes['value'] = $options['value'] ?? '#ffffff';
+        $valueAttributes['value'] = $colorAttributes['value'];
 
         $cssClass = 'js-color-picker color-picker input-group';
         $dataAttribute = $allowEmpty ? 'data-allow-empty="true"' : 'data-allow-empty="false"';
--- a/library/core/class.form.php
+++ b/library/core/class.form.php
@@ -373,25 +373,25 @@ class Gdn_Form extends Gdn_Pluggable {
     public function color($fieldName, $options = []) {
         Gdn::controller()->addJsFile('colorpicker.js');
 
-        $valueAttributes['class'] = 'js-color-picker-value color-picker-value Hidden';
-        $textAttributes['class'] = 'js-color-picker-text color-picker-text';
-        $colorAttributes['class'] = 'js-color-picker-color color-picker-color';
-
-        // Default starting color for color input. Color inputs require one, Chrome will throw a warning if one
-        // doesn't exist. The javascript will override this.
-        $colorAttributes['value'] = $options['Value'] ?? '#ffffff';
-        $valueAttributes['value'] = $colorAttributes['value'];
-
         $cssClass = 'js-color-picker color-picker input-group';
+        $allowEmpty = $options['AllowEmpty'] ?? '';
+        $dataAttribute = $allowEmpty ? ' data-allow-empty="true"' : ' data-allow-empty="false"';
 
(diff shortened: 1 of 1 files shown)


library/core/class.form.php
```diff
@@ -381,10 +379,17 @@ public function color($fieldName, $options = []) {
 
         // Default starting color for color input. Color inputs require one, Chrome will throw a warning if one
         // doesn't exist. The javascript will override this.
-        $colorAttributes['value'] = '#ffffff';
+        $colorAttributes['value'] = $options['Value'] ?? '#ffffff';
+        $valueAttributes['value'] = $colorAttributes['value'];
```
initvector: I believe this is gonna overwrite the form data value. For example, if you implemented this for the "Mobile Address Bar Color" field of the "Branding" page in the dashboard, it'll only ever show the starting/default color. The value from the `Garden.MobileAddressBarColor` config will be overwritten when the form is rendered.

R-J: Astonishing, you are right. But take a look at the markup generated from the current Form->color

~~~
<div id="Garden-dot-MobileAddressBarColor" class="js-color-picker color-picker input-group" data-allow-empty="true">
<input ... id="Form_Garden-dot-MobileAddressBarColor" ... value="#00ff00" ...>
<input ... id="Form_Garden-dot-MobileAddressBarColor-text" value="" ...>
<span class="js-color-picker-preview color-picker-preview" style="background-color: rgb(0, 255, 0);"></span>
<input ... id="Form_Garden-dot-MobileAddressBarColor-color" ... value="#ffffff"...>
</div>
~~~

I've shortened it a bit. The html markup shows that default value of #fff from the Form class in the input#color, whereas the custom color #0f0 is in the first input and the span.
I wouldn't expect to see the default color in the form element and although it doesn't have consequences I think it should be avoided to have different information in one and the same form element.

As far as I can see the process right now is this:
1. the color element is created with the default value #fff only in the input#color
2. the configuration module sets the value from the config to the "main" form input#Form_Garden-dot-MobileAddressBarColor which has the class .js-color-picker-value
3. colorpicker.js picks the value from input.js-color-picker-value and sets it as the text and for the preview color

It definetly is a problem that the js overrides the default color value. But if I see it right that color value is a dummy value, the code shown above proves this. I'll try another approach...
I'll spend some more time on it to understand what is going on there...

R-J:
Change is tested now with a) the usage in the ConfigurationModule (Brandings settings) and b) a simple view like that:
~~~
<?php

echo $this->Form->open(), $this->Form->errors();
echo $this->Form->color('hooray', ['Value' => '#00f']);
echo $this->Form->close();
~~~

Both show the expected markup, though the color control created through the ConfigurationModule still shows the dummy color (in an not relevant input). I will take a look at the Form->simple() method to check if this can be avoided when the color control is in the list of directly named controls

library/core/class.form.php
```diff
@@ -365,28 +365,33 @@ public function linkButton($code, $destination = '', $attributes = []) {
...
+            // Default dummy starting color for color input. Color inputs require one, Chrome
+            // will throw a warning if one doesn't exist. The javascript will ignore this.
+            $colorAttributes['value'] = '#ffffff';
+        }
 
         $cssClass = 'js-color-picker color-picker input-group';
-        $dataAttribute = $allowEmpty ? 'data-allow-empty="true"' : 'data-allow-empty="false"';
 
-        return '<div id="'.$this->escapeFieldName($fieldName).'" class="'.$cssClass.'" '.$dataAttribute.'>'
+        $allowEmpty = $options['AllowEmpty'] ?? '';
```
alex-mtl: Since we only use this variable as a boolean, why to set it by default to empty string?
```suggestion
        $allowEmpty = $options['AllowEmpty'] ?? false;
```

R-J: You are completely right, "false" is correct here

Issue #249, opened 2010-02-15
V2 installation error Smarty Error


cardassian-tailor:
this is the normal error i get
cache/Smarty/compile' does not exist, or is not a directory.
this error occurred on or near library/vendors/Smarty-2.6.25/libs/Smarty.class.php
Fatal Error in PHP.ErrorHandler();
library/vendors/Smarty-2.6.25/libs/Smarty.class.php

1272:         if ($display && !$this->caching && count($this->_plugins['outputfilter']) == 0) {
1273:             if ($this->_is_compiled($resource_name, $_smarty_compile_path)
1274:                     || $this->_compile_resource($resource_name, $_smarty_compile_path))
1275:             {
1276:                 include($_smarty_compile_path);
1277:             }
1278:         } else {
1279:             ob_start();
1280:             if ($this->_is_compiled($resource_name, $_smarty_compile_path)

Backtrace

/library/core/functions.error.php:132] PHP::include();
/library/vendors/Smarty-2.6.25/libs/Smarty.class.php:1276] PHP::ErrorHandler();
/library/vendors/Smarty-2.6.25/libs/Smarty.class.php:1276] Smarty::fetch();
/library/vendors/Smarty-2.6.25/libs/Smarty.class.php:1126] Smarty->fetch();
/library/core/class.smarty.php:54] Smarty->display();
/library/core/class.controller.php:985] Gdn_Smarty->Render();
/library/core/class.controller.php:775] Gdn_Controller->RenderMaster();/library/core/class.pluggable.php:192] Gdn_Controller->xRender();
/library/core/class.pluggable.php:192] Gdn_Pluggable->__call();
/applications/garden/controllers/home.php:42] HomeController->Render();
/library/core/class.dispatcher.php:230] HomeController->FileNotFound();
/index.php:41] Gdn_Dispatcher->Dispatch();

Additional information for support personnel:

```
* Application: Garden
* Application Version: 1.0
* PHP Version: 5.2.5
```
- Request Uri: /v3/index.php/garden/gardensetup
- Controller: PHP
- Method: ErrorHandler

post on vanilla forums before making issue
http://vanillaforums.org/discussion/10676/installation-error-help/

PR #8607, opened 2019-03-29
Fix: imgur embeds not rendering correctly

dannimajdalani:
closes [#249](https://github.com/vanilla/support/issues/249)

### Details

Using rich-editor, some imgur image embeds with image extensions were failing.
On imgur site, links looks something like https://i.imgur.com/1okCy50. this link embed is currently handled. 
What users are doing is 
Screen Shot 2019-03-28 at 2 17 59 PM 

this gives https://i.imgur.com/1okCy50.jpg which we don't handle in our embeds because of the image extension at the end

Our [ImgurEmbed.php](https://github.com/vanilla/vanilla/blob/master/library/Vanilla/Formatting/Embeds/ImgurEmbed.php) does not handle this type of link.

Based on a feedback from RnD, we should strip those extensions.

### To Test

- Using rich-editor, try to post these 2 images
- https://i.imgur.com/RuFna0O.jpg (should fail)
- https://imgur.com/gallery/mzto8oD (should work)
image

### DIFF
--- a/library/Vanilla/Formatting/Embeds/ImgurEmbed.php
+++ b/library/Vanilla/Formatting/Embeds/ImgurEmbed.php
@@ -28,7 +28,7 @@ class ImgurEmbed extends Embed {
      */
     public function matchUrl(string $url) {
         $data = [];
-
+        $url = str_replace(['.jpg', '.jpeg', '.png'], '', $url);
         if ($this->isNetworkEnabled()) {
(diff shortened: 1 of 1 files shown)


library/Vanilla/Formatting/Embeds/ImgurEmbed.php
```diff
@@ -28,7 +28,7 @@ public function __construct() {
      */
     public function matchUrl(string $url) {
         $data = [];
-
+        $url = str_replace(['.jpg', '.jpeg', '.png'], '', $url);
```
charrondev: This seems to work, but could we get an inline comment on this explaining the situation. Either that or maybe list the types of URL structures supported in the docblock above?

Issue #8496, opened 2019-03-06
Add breadcrumb to results on search endpoint of categories API resource


initvector:
GET requests to [/api/v2/categories/search](https://github.com/vanilla/vanilla/blob/0aedb5a837b4515f7d55543722d2576c85625fa6/applications/vanilla/controllers/api/CategoriesApiController.php#L191) should include the ability to expand the item's breadcrumbs. This could be an update to the endpoint's expand parameter.

charrondev:
See `Vanilla\Navigation\Breadcrumbs`

tburry:
I would like to lessen the breadcrumbs that we display to include just the following:

- **Guides**: Just the knowledge base link.
- **Help Center**: Knowledge base > category > category > etc.

PR #8608, opened 2019-03-29
Add  expanding breadcrumbs to categories API search

alex-mtl:
Extend  categories/search api when "expand"
to add breadcrumbs data

Closes: #8496

### DIFF
--- a/applications/vanilla/controllers/api/CategoriesApiController.php
+++ b/applications/vanilla/controllers/api/CategoriesApiController.php
@@ -211,10 +211,7 @@ class CategoriesApiController extends AbstractApiController {
                 'minimum' => 1,
                 'maximum' => 200
             ],
-            'expand:b?' => [
-                'default' => false,
-                'description' => 'Expand with the parent record.'
-            ]
+            'expand?' => ApiUtils::getExpandDefinition(['parent', 'breadcrumbs'])
         ])->setDescription('Search categories.');
         $out = $this->schema([':a' => $this->schemaWithParent($query['expand'])], 'out');
 
@@ -223,10 +220,10 @@ class CategoriesApiController extends AbstractApiController {
         list($offset, $limit) = offsetLimit("p{$query['page']}", $query['limit']);
         $rows = $this->categoryModel->searchByName(
             $query['query'],
-            $query['expand'],
+            $this->isExpandField('parent',$query['expand']),
             $limit,
             $offset,
-            ($query['expand'] ?? true) ? ['breadcrumbs'] : []
+            $this->isExpandField('breadcrumbs',$query['expand']) ? ['breadcrumbs'] : []
         );
 
         foreach ($rows as &$row) {
@@ -243,7 +240,7 @@ class CategoriesApiController extends AbstractApiController {
     /**
      * Get an ID-only category record schema.
      *
-     * @param string $type The type of schema.
+     * @param string $type The type of schema.ApiUtils::getExpandDefinition
      * @return Schema Returns a schema object.
      */
     public function idParamSchema($type = 'in') {
--- a/applications/vanilla/controllers/api/CategoriesApiController.php
+++ b/applications/vanilla/controllers/api/CategoriesApiController.php
@@ -5,11 +5,15 @@
  */
 
 use Garden\Schema\Schema;
+use Vanilla\Utility\InstanceValidatorSchema;
 use Garden\Web\Data;
 use Garden\Web\Exception\ClientException;
 use Garden\Web\Exception\NotFoundException;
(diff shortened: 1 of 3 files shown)


applications/vanilla/controllers/api/CategoriesApiController.php
```diff
@@ -219,7 +225,8 @@ public function get_search(array $query) {
             $query['query'],
             $query['expand'],
             $limit,
-            $offset
+            $offset,
+            ($query['expand'] ?? true) ? ['breadcrumbs'] : []
```
initvector: The description on the expand parameter say it's to "expand with the parent record", but that isn't accurate any more. I recommend altering the expand parameter definition to be built with the `\Vanilla\ApiUtils::getExpandDefinition` method. This will allow you to use specific fields (e.g. "parent" and "breadcrumbs") and should be backwards-compatible with the boolean expand this endpoint currently has. From there, you can use the controller's `isExpandField` method for explicitly expanding certain items.

applications/vanilla/models/class.categorymodel.php
```diff
@@ -93,6 +99,10 @@ public function __construct() {
         parent::__construct('Category');
         $this->collection = $this->createCollection();
         $this->eventManager = Gdn::getContainer()->get(EventManager::class);
+
+        $container = Gdn::getContainer();
+        $this->breadcrumbModel = $container->get(BreadcrumbModel::class);
+        $this->breadcrumbModel->addProvider($container->get(ForumBreadcrumbProvider::class));
```
initvector: I think Vanilla has [a hook that's already adding `ForumBreadcrumbProvider` to instances of  `BreadcrumbModel`](https://github.com/vanilla/vanilla/blob/ccbcaceb7343d3ce9986d9ed5127fe9eeaf7a327/applications/vanilla/settings/class.hooks.php#L26) that are instantiated by the container. Are you sure you need to add the provider it here too?

alex-mtl: Yes, for some reason `container` returns empty list of providers for BreadcrumbModel here. I think that is some legacy incompatibility. As you can see we do not receive container in constructor as a param. And when I try to add anything there - it is not passed. 
I don't want to do "yak shaving" here. This is just a "patch". If you have an idea how I can get correctly instantiated BreadcrumbModel with all providers I would appreciate any help.

applications/vanilla/controllers/api/CategoriesApiController.php
alex-mtl: Transformation is done. Please check updated.

applications/vanilla/controllers/api/CategoriesApiController.php
```diff
@@ -236,7 +239,7 @@ public function get_search(array $query) {
     /**
      * Get an ID-only category record schema.
      *
-     * @param string $type The type of schema.
+     * @param string $type The type of schema.ApiUtils::getExpandDefinition
```
initvector:
```suggestion
     * @param string $type The type of schema.
```

applications/vanilla/models/class.categorymodel.php
alex-mtl: @initvector I had discussion with @charrondev , and we've found the issue. 
I moved initializing of BreadcrumbModel out of constructor, it seem better now. 
Please check updated.

applications/vanilla/controllers/api/CategoriesApiController.php
```diff
@@ -217,9 +219,10 @@ public function get_search(array $query) {
         list($offset, $limit) = offsetLimit("p{$query['page']}", $query['limit']);
         $rows = $this->categoryModel->searchByName(
             $query['query'],
-            $query['expand'],
+            $this->isExpandField('parent',$query['expand']),
```
initvector:
```suggestion
            $this->isExpandField('parent', $query['expand']),
```

applications/vanilla/controllers/api/CategoriesApiController.php
```diff
@@ -217,9 +219,10 @@ public function get_search(array $query) {
...
             $limit,
-            $offset
+            $offset,
+            $this->isExpandField('breadcrumbs',$query['expand']) ? ['breadcrumbs'] : []
```
initvector:
```suggestion
            $this->isExpandField('breadcrumbs', $query['expand']) ? ['breadcrumbs'] : []
```

initvector:
Just need to update the expand parameter definition in the endpoint's OpenAPI doc and I think this'll be good to go!

charrondev:
Semi-blind approval

charrondev:
Semi-blind approval

Issue #8558, opened 2019-03-20
Add deployment key header to API responses


initvector:
A deployment key, somehow derived from `Garden.Deployed` (see `assetVersion`), should be sent down as a header along with API responses. This header can then be used for client-side scripts to determine if a new code deployment/version change has happened since the page was loaded.

PR #8612, opened 2019-03-29
Implement Vanilla Deployment Key (Vdk) header for all http responses

alex-mtl:
Implement Vanilla Deployment Key (Vdk) header for all http responses

Closes: #8558


alex-mtl:
@initvector I doubt that implementing middleware will have the same coverage like if I do this in Data directly. Data object has been created by many different ways and cover many exceptions even after middleware got applied. We can discuss it. I am not against to try middleware approach. But imho Data class is a very good place to inject that.

charrondev:
If the header is not done in middleware, maybe we should mark the `DeploymentCacheBuster` as shared in the container? Otherwise we'll be making a lot more of them than we would really need (A request may dispatch multiple API endpoints creating Data objects internally).

initvector:
> I doubt that implementing middleware will have the same coverage like if I do this in Data directly.

That's probably a good thing. I think we only want the header to be on responses sent to a client, not every time the class might be instantiated and used, internally.

alex-mtl:
> > I doubt that implementing middleware will have the same coverage like if I do this in Data directly.
> 
> That's probably a good thing. I think we only want the header to be on responses sent to a client, not every time the class might be instantiated and used, internally.

A lot of our 400, 404 and etc are returned by those Exception scenarios.  On top of that, why this task came on the table -> because we want to track version of resources but not only (and mostly not for) html pages generated.

initvector:
> A lot of our 400, 404 and etc are returned by those Exception scenarios.

Error responses are still run through middleware. Putting the deployment key in a middleware will not stop it from being sent on 400s or 404s. Making the deployment key a middleware will ensure it gets sent to every API response, while still allowing us to avoid including it where not needed.

charrondev:
@alex-mtl This is specifically only needed for the API & page responses. Currently every single item there get's run through the middleware. I can't really think of a more straightforward case for middleware than an HTTP header though.

library/Vanilla/Web/DeploymentHeaderMiddleware.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Invoke the deployment key on a request.
+     *
+     * @param RequestInterface $request The incoming request.
+     * @param callable $next The next middleware.
+     * @return mixed Returns the response of the inner middleware.
+     */
+    public function __invoke(RequestInterface $request, callable $next) {
+        $response = Data::box($next($request));
+        $cacheBuster = \Gdn::getContainer()->get(DeploymentCacheBuster::class);
```
initvector: You should be able to inject this dependency, instead of manually grabbing it from the container.

library/Vanilla/Web/DeploymentHeaderMiddleware.php
```diff
@@ -0,0 +1,35 @@
...
+        $response->setHeader(
+            self::VANILLA_DEPLOYMENT_KEY,
+            APPLICATION_VERSION.'-'.$cacheBuster->value()
```
initvector: This is mostly an observation, but if you don't have a value in `Garden.Deploy`, the header value becomes `{version}-{version}` (e.g. "2.8.1+2019.005-2.8.1+2019.005"). For the purposes of this header, that's probably fine. But if we can easily avoid the unnecessary duplication, that would be swell.

PR #8613, opened 2019-03-30
Fixed quote mention to user is also required in WYSIWYG

vintprox:
As soon as `userAnchor` function respects `Text` option, it will always return what we expect in WYSIWYG mode. Well, what if I wanted it to return some "Display Name" by default? I just copy `userAnchor` code and paste slightly modified version in my addon. And there is no better solution than using mention-style username atm, so this practice should always be there in Quotes plugin.

### DIFF
--- a/plugins/Quotes/class.quotes.plugin.php
+++ b/plugins/Quotes/class.quotes.plugin.php
@@ -516,7 +516,7 @@ BQ;
 
                     break;
                 case 'Wysiwyg':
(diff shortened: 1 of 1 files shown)


charrondev:
I see you've jumped straight into a pull request. While contributions are appreciated you seem to have skipped over all of the pull request criteria that was pre-filled when you opened your pull request.

* [ ] Title is formatted in present imperative (ex: "Fix X by doing Y").

charrondev:
Closing a due to lack of communication.

PR #8614, opened 2019-03-30
[Flagging] Fix for translated link to content

vintprox:
I just noticed context part was left original. Idea here is to give proper "Link to content" by providing common pattern under 'FlagLinkFormat' key with three arguments: *translated* context ("Discussion"/"Comment"), content ID and author's username, which was left out of pattern before.

### DIFF
--- a/plugins/Flagging/views/flag.php
+++ b/plugins/Flagging/views/flag.php
@@ -16,8 +16,7 @@ echo $this->Form->errors();
                 <?php echo t('FlagForReview', "You are about to flag this for moderator review. If you're sure you want to do this,
          please enter a brief reason below, then press 'Flag this!'."); ?>
             </div>
-            <?php echo t('FlagLinkContent', 'Link to content:').' '.anchor(t('FlagLinkFormat', "{$UcContext} #{$ElementID}"), $URL); ?> &ndash;
-            <?php echo htmlspecialchars($this->data('Plugin.Flagging.Data.ElementAuthor')); ?>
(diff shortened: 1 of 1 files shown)


charrondev:
I'm repeating myself here but please make sure you follow the pull request template.

* [x] Title is formatted in present imperative (ex: "Fix X by doing Y").

vintprox:
I have zero idea how this change should be mentioned in release log relatively to other items, but "link to content" part directly refers to what we see in popup as "Link to content: _whatever_". Also I don't know how I should tag plugins which were affected by pull request (`[Flagging]`).

After all, red-tape stuff can be modified by users with write access to `vanilla/vanilla`, I allow and 
encourage it.

charrondev:
The title's fine actually. What red tape is there on this? I'm not asking you to tag anything. Really it's:

- File an issue.
- Link to that issue.
- Explain how the change fixes the issue.

On the pull request itself, I understand it's quite short, but it's really hard to make sense of. You've essentially merged 2 already quite complicated lines into 1 super long 230 character line. Some intermediary variables along the way would go a long way.

vintprox:
I'm sorry Adam that I was kind of inconvenient at the sense. Issues are really a good way to take many variables into consideration before writing another chunk of code. 👍 Who knows! What if next week anything bad will be noticed about this whole view of `Flagging`? ⛓

charrondev:
Closed due to staleness and unaddressed feedback.

Issue #6768, opened 2018-03-06
Rich Editor - Handle copy/paste of nested lists & integration tests


charrondev:
- Gather numerous examples.
- Setup an integration test harness for pasting them in and examining the outputted HTML/JSON.

charrondev:
This is no longer going to be necessary for MVP as our current fallback to the advanced editor works pretty well.

PR #8615, opened 2019-04-01
Implement copy/pasting of lists into Rich Editor

charrondev:
Closes https://github.com/vanilla/vanilla/issues/6768

- Add/improve some utilities for rich editor testing.
- Adds integration tests for pasting simple text content into rich editor.
- Add integration tests for pasting nested lists into rich editor.
- Add support for pasting nested lists into rich editor.

## Testing

When testing keep the following in mind:

- An item must be fully "selected around" in order to be pasted properly.
- Selecting partial lists may not bring all of the list formatting as it originally existed.

The simplest way I've found to test this is to make a post in an older format & use the `Convert to Article` menu option. This will move it to the knowledge base and "paste" it into the rich editor.


charrondev:
@slafleche Both of those are this issue which I already have in my sprint. I consider it outside of the scope of this PR.

(The issue comes we try to split the UL/OL in have. There is some buggy behaviour there)

PR #8617, opened 2019-04-01
Add React components for displaying global UI messages

slafleche:
Implements components "MessageAndRefresh" and "MessageAndRefreshFixed".

To test, I suggest adding "MessageAndRefresh" in the ArticleLayout's "middleBottom" section. and "MessageAndRefreshFixed" in the article page.

Note that in the mockup, it seems like it follows the width of the middle column, but Todd said it should have its own width. He also asked for it to overlap slightly with the header and stay put, even if the header animates away.

Closes: https://github.com/vanilla/knowledge/issues/791


library/src/scripts/messages/messageStyles.ts
```diff
@@ -0,0 +1,168 @@
+import { styleFactory, useThemeCache, variableFactory } from "@library/styles/styleUtils";
+import { globalVariables } from "@library/styles/globalStyleVars";
+import { unit, userSelect, colorOut, paddings, borders, fonts, allButtonStates } from "@library/styles/styleHelpers";
```
charrondev: You're imports are above your file doc block here.

PR #8625, opened 2019-04-02
Implement script content security policy for new pages

alex-mtl:
Relates to: #vanilla/knowledge/issues/132


charrondev:
This looks great! I think we need some additional default providers.

## Embed CSP provider

It will need to become more flexible in the future, but the following domains should be allowed by default. They should not be allowed though if `Garden.Format.DisableUrlEmbeds` is set to `true`.

- `embed-cdn.gettyimages.com`
- `s.imgur.com`
- `platform.instagram.com`
- `platform.twitter.com`

## Followup issue

I think we'll need a followup issue for documenting how to change these values or to search out some additional defaults. Maybe a followup for putting these in a dashboard page somewhere?

PR #8626, opened 2019-04-02
Added styles and classes for Rich Editor Errors

slafleche:
These are components and classes/styles used for errors in the Rich Editor (modern). Note that these are not hooked up yet.


library/src/scripts/forms/formElementStyles.ts
```diff
@@ -99,14 +99,32 @@ export const formElementsVariables = useThemeCache(() => {
     };
 });
 
+export const AccessibleErrorClasses = useThemeCache(() => {
```
charrondev: Was the capitalization here intentional? All the other ones are camel cased.

PR #8653, opened 2019-04-05
Add tick API v2 resource for collecting page view stats

alex-mtl:
- Add new TickApiController to register basic pageviews
- Add openapi spec yml file for tick endpoint

Closes: https://github.com/vanilla/knowledge/issues/83

### DIFF
--- a/applications/dashboard/openapi/tick.yml
+++ b/applications/dashboard/openapi/tick.yml
@@ -4,12 +4,6 @@ paths:
   /tick:
     post:
       summary: Register a page view for statistical and analytical purpose.
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/TickRequest'
-        required: false
       responses:
         '201':
           description: Success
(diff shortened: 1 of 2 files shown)


alex-mtl:
@tburry Could you please confirm that this minimal implementation is good to go and we don't need to dive deeper and refactor `Gdn_Statistics` class?

applications/dashboard/openapi/tick.yml
```diff
@@ -0,0 +1,32 @@
...
+        '201':
+          description: Success
+      tags:
+      - Statistics
+components:
+  schemas:
+    TickRequest:
+      description: Both body request parameters are optional
+      properties:
+        Path:
```
tburry: These should adhere to our naming conventions and be lower camel case.

applications/dashboard/openapi/tick.yml
```diff
@@ -0,0 +1,32 @@
...
+          description: Page path
+          minLength: 1
+          type: string
+          example: /discussions/123
+        ResolvedPath:
```
tburry: Can you leave this one out of the docs? It is kind of outdated and I'd rather new code doesn't depend on it.

applications/dashboard/openapi/tick.yml
alex-mtl: I was trying to keep backward compatibility with old endpoint.

PR #8535, opened 2019-03-14
Fix regression disabling of categories

tburry:
I bungled another #8448 which resulted in enabled nested/flat categories in the post discussion dropdown. I'm assigning two senior reviewers to bail me out here.

Closes vanilla/support#259

PR #8654, opened 2019-04-05
Backport: Fix regression disabling of categories (release/2019.004)

dannimajdalani:
backporting #8535


charrondev:
I think I’ve got the build fixed. I backported the babel version changes from master. Likely some downstream babel package wasn’t pinning its children version tight enough.

If we need to backport to another branch,  we may have to do the same thing.

Issue #8638, opened 2019-04-04
Google Sign-In Plugin


cchabilall83:
**Google Sign-In Plugin**

Add the ability for end-users to login to Vanilla using their Google account.  This will replace the current Google+ which has been deprecated by Google.

> Google Sign-In is a secure authentication system that reduces the burden of login for your users, by enabling them to sign in with their Google Account—the same account they already use with Gmail, Play, and other Google services.

**User Story**

As a user I would like to allow members to login to the community using their Google Account.

**Acceptance Criteria**

- Create basic plugin with plugin class, settings view, js folder, design folder for CSS.
- Add a toggle to '/social/manage' for Google Sign-In (Similar to Google+).
   - Add settings menu to configure 
- Add Plugin to extend Vanilla's Oauth2 class
  - Create a config form for dashboard:
    - Create a view (twig template).
    - Create form fields/validation: Client ID and Client Secret
    - Write user instructions.
  - Create button to link to Google Sign In
  - Make call to Google's OpenID discovery to get the OAuth2 endpoints.
- Create client facing docs (with links to Google's docs).
- Create technical docs (README).
  - Add testing steps.
- Delete Google Plus plugin.
- Create graphics (create issue).
- Update existing Google+ UnitTest.

**Additional context**

**Mockups**

**Related Issues**

vrjrobitaille:
Epic seems to be completed.

PR #8656, opened 2019-04-05
Create a Google sign in feature.

nervoustwit:
Recently Google discontinued their Google + feature which we offered as a method for authenticating users based on their GMail accounts. This PR will replace that plugin with a plugin that connects with a Google OpenID Connect OAuth 2.0 app to authenticate users.

For more information about the Google Sign In OpenID Connect workflow go to [developers.google.com/identity/protocols/OpenIDConnect](https://developers.google.com/identity/protocols/OpenIDConnect).

This PR references #8638 

**Testing**
In order to test this plugin you will have to 
  - Create an OAuth2 web app on https://console.developers.google.com.
  - Turn on the Google Sign In plugin.
  - Plug in your Client ID and Client secret from Google.
  - Sign out, and sign back in again using the Google button.


plugins/googlesignin/class.googlesignin.plugin.php
```diff
@@ -0,0 +1,124 @@
...
+ * Expose the functionality of the core class Gdn_OAuth2 to create SSO workflows.
+ */
+
+class GoogleSignInPlugin extends Gdn_OAuth2 {
+    CONST AUTHORIZEURL = 'https://accounts.google.com/o/oauth2/v2/auth';
+    CONST TOKENURL = 'https://oauth2.googleapis.com/token';
+    CONST PROFILEURL = 'https://openidconnect.googleapis.com/v1/userinfo';
+    CONST PROFILENAME = 'name';
+    CONST ACCEPTEDSCOPE = 'email openid profile';
+    /**
```
alex-mtl: Code sniffer check failed because of  `const` keyword, should be lowercase.
And i would recommend to use regular php naming convention for const names with underscore as a word separator:
```
const AUTHORIZE_URL = 'https://accounts.google.com/o/oauth2/v2/auth';
    const TOKEN_URL = 'https://oauth2.googleapis.com/token';
    const PROFILE_URL = 'https://openidconnect.googleapis.com/v1/userinfo';
    const PROFILE_NAME = 'name';
    const ACCEPTED_SCOPE = 'email openid profile';
```

plugins/googlesignin/class.googlesignin.plugin.php
```diff
@@ -0,0 +1,124 @@
...
+        } else {
+
+            $form->setFormValue('AuthenticationKey', $this->getProviderKey());
+
+            $sender->Form->validateRule('AssociationKey', 'ValidateRequired', 'You must provide a unique AccountID.');
+            $sender->Form->validateRule('AssociationSecret', 'ValidateRequired', 'You must provide a Secret');
+
+            // To satisfy the AuthenticationProviderModel, create a BaseUrl.
+            $baseUrlParts = parse_url($form->getValue('AuthorizeUrl'));
+            $baseUrl = (val('scheme', $baseUrlParts) && val('host', $baseUrlParts)) ? val('scheme', $baseUrlParts).'://'.val('host', $baseUrlParts) : null;
```
alex-mtl: Long line warning. I would recommend to set up dedicated function for this and avoid using global `val()` function.
```
// $baseUrlParts = parse_url($form->getValue('AuthorizeUrl'));
// $baseUrl = (val('scheme', $baseUrlParts) && val('host', $baseUrlParts)) ? val('scheme', $baseUrlParts).'://'.val('host', $baseUrlParts) : null;
$baseUrl = $this->getBaseUrl($form->getValue('AuthorizeUrl'));
```

And new method:

```
    /**
     * Extract base url from url
     *
     * @param string $url
     * @return boolean|string
     */
    public function getBaseUrl(string $url) {
        $baseUrlParts = parse_url($url);
        if (($baseUrlParts['scheme'] ?? false) && ($baseUrlParts['host'] ?? false)) {
            $baseUrl = $baseUrlParts['scheme'].'://'.$baseUrlParts['host'];
        } else {
            $baseUrl = false;
        }
        return $baseUrl;
    }
```

plugins/googlesignin/class.googlesignin.plugin.php
```diff
@@ -0,0 +1,124 @@
...
+            if ($baseUrl) {
+                $form->setFormValue('BaseUrl', $baseUrl);
+                $form->setFormValue('SignInUrl', $baseUrl); // kludge for default provider
+            }
+            if ($form->save()) {
+                $sender->informMessage(t('Saved'));
```
alex-mtl: Do we translate interface elements and messages for dashboard generally?
If **yes**, all message strings in this plugin need to be wrapped with `t()` function.
Or if **no**, then we can remove `t()` call here.

plugins/googlesignin/class.googlesignin.plugin.php
```diff
@@ -0,0 +1,124 @@
...
+
+        if (!$form->authenticatedPostBack()) {
+            $provider = $this->provider();
+            $form->setData($provider);
+        } else {
+
+            $form->setFormValue('AuthenticationKey', $this->getProviderKey());
+
+            $sender->Form->validateRule('AssociationKey', 'ValidateRequired', 'You must provide a unique AccountID.');
+            $sender->Form->validateRule('AssociationSecret', 'ValidateRequired', 'You must provide a Secret');
```
alex-mtl: Validation messages break the layout with not styled messages and escaped html tags:
Screen Shot 2019-04-06 at 1 20 46 PM

If you change twig  line to `raw` format that fix the issue:
```
{{ form.errors() | raw }}
```

plugins/googlesignin/views/settings.twig
```diff
@@ -0,0 +1,11 @@
...
+<div class="padded alert alert-warning">
+    {{ Instructions|raw }}
+</div>
+ {{ form.open()|raw }}
+ {{ form.errors() }}
+ {{ form.simple(formData)|raw }}
+<div class="modal-footer js-modal-fixed">
+{{ form.button('Save')|raw }}
+</div>
+ {{ form.close()|raw }}
```
alex-mtl: Too much usage of `raw` format. That makes twig less useful. 
One of the purpose of using templating library (such as twig) is to avoid XSS vulnerability, when templating engine is capable to auto escape special html characters and reduce vulnerability level.

I think that is ok to go for this PR. But we need to think about converting or extending base classes.
For example in this particular case we need `form->errors()` to return plain array with messages without any html.
And then decorate those messages with html twig template.

PR #8658, opened 2019-04-08
UI consistency (KB)

slafleche:
This PR includes many design fixes, mostly hover/focus styles. 

See mockups: https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5ca4e0edb031df05fb919092

You can test the header (excluding MeBox), buttons - next/previous and was this helpful (example article page), input text focus styles (search page main box and advanced search), button toggle, modal (delete article)

More changes to come


charrondev:
A few style issues that seem to have been introduced hear or still remain. If you aim to tackle these in your next set of fixes that's fine, but I'm still going to list them.

## Dropdowns & Selects

**This is only one here that is absolutely a regression and should be fixed before merging**

The various `<Select />` & `<Token />` inputs not seem to have extra padding in the top/bottom of their dropdowns.

image

This also affects the MeBox notifications & user dropdowns as well as the search autocomplete dropdown.

## Buttons

- Buttons "lose" their border in their active state. This gives them the appearance of shrinking when pressed.
- Buttons have a black border when focused.

image

## Location Picker Buttons

The contents of the location picker buttons aren't visible when hovered.

image

## New Category dropdown

The hover state for the buttons here incorrectly has a border.

image

I think these just have the wrong button type, because other modals (like `<ModalConfirm />` are correct.) 

image

## Nav Manager

I'm not sure if these just weren't fully specified in the mockups, but the hover behaviour for these buttons feels incorrect.

image

Additionally the rename box seems to have an ugly outline no matter what.

image

PR #8659, opened 2019-04-08
Fix imgur embeds

dannimajdalani:
closes https://github.com/vanilla/support/issues/249

### Details

Using rich-editor, some imgur image embeds with image extensions were failing.
On imgur site, links looks something like https://i.imgur.com/1okCy50. this link embed is currently handled. 
What users are doing is 
Screen Shot 2019-03-28 at 2 17 59 PM 

this gives https://i.imgur.com/1okCy50.jpg which we don't handle in our embeds because of the image extension at the end

Our [ImgurEmbed.php](https://github.com/vanilla/vanilla/blob/master/library/Vanilla/Formatting/Embeds/ImgurEmbed.php) does not handle this type of link.

Based on a feedback from RnD, we should strip those extensions.

### To Test

- Using rich-editor, try to post these 2 images
- https://i.imgur.com/RuFna0O.jpg (should fail)
- https://imgur.com/gallery/mzto8oD (should work)
image

### DIFF
--- a/library/Vanilla/Formatting/Embeds/ImgurEmbed.php
+++ b/library/Vanilla/Formatting/Embeds/ImgurEmbed.php
@@ -14,7 +14,7 @@ use Exception;
 class ImgurEmbed extends Embed {
 
     /** @inheritdoc */
-    protected $domains = ['imgur.com'];
+    protected $domains = ['imgur.com', 'i.imgur.com'];
 
     /**
      * ImgurEmbed constructor.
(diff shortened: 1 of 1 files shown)


cchabilall83:
This works and I will approve. 

 But I think Adam had suggested that we parse the url host first and match against allowed imgur hosts.  Then continue matching the query string.

```
$url = "http://imgur.com";
$domains = ["imgur.com", "i.imgur.com"];
$url = parse_url($url); 
in_array($url, $domains);

continue...
```

PR #8660, opened 2019-04-08
Extend site navigation styles to implement icon style for links

alex-mtl:
Add new style for inner icon element of link

Relates to: https://github.com/vanilla/knowledge/issues/726


library/src/scripts/navigation/siteNavAdminLinksStyles.ts
```diff
@@ -46,5 +46,11 @@ export const siteNavAdminLinksClasses = useThemeCache(() => {
         ...debug.name("link"),
     });
 
-    return { root, item, divider, link };
+    const linkIcon = style({
+        marginLeft: px(-6),
```
slafleche: Remove negative margin and get rid of this one instead: https://github.com/vanilla/vanilla/blob/master/library/src/scripts/navigation/siteNavAdminLinksStyles.ts#L45

slafleche:
Don't forget the state colours mentioned in https://github.com/vanilla/knowledge/issues/726

PR #8661, opened 2019-04-08
Add favicon and theme-color to twig master template

initvector:
The default Twig template does not support either the site's configured favicon or mobile address bar color (theme-color). These settings can be found under the Branding area of the dashboard.

`Vanilla\Web\Page` has been updated to set these values and hand them down to the default Twig template. The default Twig template has been updated to use them. `Vanilla\Web\ThemedPage` has been updated to account for the new dependencies in `Vanilla\Web\Page`.

### DIFF
--- a/library/Vanilla/Web/Page.php
+++ b/library/Vanilla/Web/Page.php
@@ -7,6 +7,7 @@
 
 namespace Vanilla\Web;
 
+use Gdn_Upload;
 use Garden\CustomExceptionHandler;
 use Garden\Web\Data;
 use Garden\Web\Exception\ServerException;
@@ -17,10 +18,10 @@ use Vanilla\Navigation\Breadcrumb;
 use Vanilla\Navigation\BreadcrumbModel;
 use Vanilla\Web\Asset\WebpackAssetProvider;
 use Vanilla\Web\ContentSecurityPolicy\ContentSecurityPolicyModel;
-use Vanilla\Web\ContentSecurityPolicy\DefaultContentSecurityPolicyProvider;
 use Vanilla\Web\JsInterpop\PhpAsJsVariable;
 use Vanilla\Web\JsInterpop\ReduxAction;
 use Vanilla\Web\JsInterpop\ReduxErrorAction;
+use Vanilla\Contracts\ConfigurationInterface;
 
 /**
  * Class representing a single page in the application.
@@ -31,6 +32,12 @@ abstract class Page implements InjectableInterface, CustomExceptionHandler {
 
     /** @var string */
     private $canonicalUrl;
+
+    /** @var string */
+    private $favIcon;
+
+    /** @var string */
+    private $mobileAddressBarColor;
 
     /** @var string */
     private $seoTitle;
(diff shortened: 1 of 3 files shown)


library/Vanilla/Web/Page.php
```diff
@@ -110,21 +117,32 @@ abstract class Page implements InjectableInterface, CustomExceptionHandler {
...
         $this->cspModel = $cspModel;
+        $this->config = $config;
+
+        if ($favIcon = $this->config->get("Garden.FavIcon")) {
+            $this->setFavIcon(Gdn_Upload::url($favIcon));
+        }
+
+        if ($mobileAddressBarColor = $this->config->get("Garden.MobileAddressBarColor")) {
+            $this->setMobileAddressBarColor($mobileAddressBarColor);
+        }
```
alex-mtl: This is very generic core class. I don't like the idea to pollute with new method and properties every time when we need to add some `<meta>` tag. We have `addMetaTag()` method which is enough to achieve the goal here and to avoid necessity to create 2 new methods.

And if we go with `addMetaTag()` we don't need to touch twig. It will reflect automatically.
What do you think?

initvector: Seems like a sane update to me!

library/Vanilla/Web/Page.php
```diff
@@ -32,6 +33,12 @@ abstract class Page implements InjectableInterface, CustomExceptionHandler {
     /** @var string */
     private $canonicalUrl;
 
+    /** @var string */
+    private $favIcon;
+
+    /** @var string */
+    private $mobileAddressBarColor;
```
alex-mtl: I think we don't need this property anymore?

library/Vanilla/Web/Page.php
```diff
@@ -211,6 +227,26 @@ protected function addReduxAction(ReduxAction $action): self {
         return $this;
     }
 
+    /**
+     * Get the page's "favorite icon".
+     *
+     * @return string|null
+     */
+    protected function getFavIcon() {
```
alex-mtl:
```suggestion
    protected function getFavIcon(): ?string {
```

alex-mtl: Since php 7.1 we can declare nullable returns :-)

library/Vanilla/Web/Page.php
```diff
@@ -110,21 +114,32 @@ abstract class Page implements InjectableInterface, CustomExceptionHandler {
...
      */
     public function setDependencies(
         SiteMeta $siteMeta,
         \Gdn_Request $request,
         \Gdn_Session $session,
         WebpackAssetProvider $assetProvider,
         BreadcrumbModel $breadcrumbModel,
-        ContentSecurityPolicyModel $cspModel
+        ContentSecurityPolicyModel $cspModel,
+        ConfigurationInterface $config
```
charrondev: Could we put that favicon in SiteMeta? I've been trying to isolate `Page` from using the configuration directly.

initvector: Works for me. I'm in favor of avoiding adding another dependency to this class.

Favicon and the "mobile address bar color" both need to see the config, so they've both been moved into `SiteMeta`.

Issue #8307, opened 2019-02-05
Add count data to <MeBox /> component


charrondev:
The new `<MeBox />` component is not displaying any counts.

Items in the user dropdown menu and their counts need to come from some API endpoint (will need to be created).

## Remove

- Particated
- Dynamic addons.

PR #8662, opened 2019-04-08
Implement user dropdown counts in MeBox

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8307

Required companion Pr https://github.com/vanilla/knowledge/pull/859.

- Implements real counts in the MeBox user menu using the `/api/v2/users/me-counts` endpoint.
- Removes links that come from addons (partipated & badges). These may be added back in the future when they are wired up dynamically.
- Deletes hard coded count data.
- Updates the `/api/v2/users/me-counts` API to return numbers instead of strings for counts.
- Refactors `userModel` & `UserActions` to use FSA style actions.
- Refactors `UserDropDown` to use `UserDropDownContents` internally.
- Fixes a scrolling issue in the mobile panels of the MeBox.

## The counts

The actual counts are loaded over the API when the dropdown is opened. The counts will be refreshed if the user loads opens the dropdown more than 60 seconds after the initial loading.


charrondev:
@slafleche That will be separate issue. It was not in the scope of my sprint. It can be counted as design feedback for next sprint.

slafleche:
Ok, then link me the issue and I'll approve it.

PR #8663, opened 2019-04-08
Add support for custom fonts to dynamic theming system

initvector:
The dynamic theme system didn't have support for customizing the page's font usage. Now, it has some basic support.

### Overview
1. Add `loadThemeFonts` frontend function. This function loads a theme's fonts (read from the store) using [Web Font Loader](https://github.com/typekit/webfontloader).
1. Add support for configuring a theme's body font styles with a new global theme variable: `fonts.families.body`.
1. Add default theme font: Open Sans. If a theme does not include its own font assets, Open Sans will be available. Open Sans is also the default font family for a page.
1. Add `fontFamilyWithDefaults` frontend function. This function allows generating a valid CSS `font-family` value that includes Vanilla's fallbacks.
1. Add font asset to `IThemeAssets` interface.
1. Update Vanilla's fallback font families. New font fallbacks mirror those found on GitHub.
1. Remove hardcoded loading Open Sans with Google's web font loader from the default master Twig template.
1. Remove default for `ContentSecurityPolicy.ScriptSrc.AllowedDomains` config. This previous default included "https://ajax.googleapis.com", which should no longer be necessary after the removal of the hardcoded web font loader from the default master Twig template.

### Testing
1. Get the [Dark Theme Test](https://github.com/vanilla/themes/tree/master/theme-variables-dark) theme.
1. Find a nice font on [Google Fonts](https://fonts.google.com). I recommend [Libre Barcode 39](https://fonts.google.com/specimen/Libre+Barcode+39). Select it to get its CSS URL (e.g. https://fonts.googleapis.com/css?family=Libre+Barcode+39).
1. Modify the theme's fonts.json asset. You can replace the current font config or just add a new font using the current value as an example. Use the CSS URL from the previous step.
1. Edit the theme's variables asset. Edit the `global.fonts.families.body` key, or create it if it doesn't exist. Set this to an array that includes the name of the font you chose (e.g. `["Libre Barcode 39"]`).
1. Load up a page that uses the dynamic theming system. You should see the new font being used on the page.

Closes vanilla/knowledge#796


library/src/scripts/theming/loadThemeFonts.ts
```diff
@@ -0,0 +1,38 @@
...
+                    families: fonts.map(font => font.name),
+                    urls: fonts.map(font => font.url),
+                },
+            };
+
+            if (webFontConfig.custom && webFontConfig.custom.urls && webFontConfig.custom.urls.length > 0) {
+                WebFont.load(webFontConfig);
+            } else {
+                // Default font loading
+                WebFont.load(defaultFontConfig);
```
slafleche: We never get here, because fonts is undefined...

conf/config-defaults.php
```diff
@@ -40,8 +40,6 @@
 
 $Configuration['HotReload']['IP'] = '127.0.0.1';
 
-$Configuration['ContentSecurityPolicy']['ScriptSrc']['AllowedDomains'] = ['https://ajax.googleapis.com/'];
-
```
alex-mtl: Can we leave it here with empty array `= [];` please?

PR #8667, opened 2019-04-09
Remove Google+ plugin

dannimajdalani:
related #8638
This removes the Google+ plugin.
We should first verify that GooglePlus service is down before merging this PR.


dannimajdalani:
@nervoustwit can you approve this?

dannimajdalani:
We have more than 55+ clients with Google+ plugin enabled.

dannimajdalani:
We have decided to set this plugin as hidden for the time being instead of deleting it, since we still have multiple clients using it.

tburry:
I think we should probably close this PR at this time. My reasons for this suggestion is the following:

1. We have a lot of sites with it enabled and need to go through a different, longer process to get them disabled.
2. We have just marked the addon hidden.
3. I don't want to keep seeing the reminder of this PR forever.
4. The longer we wait, the more difficult to merge this PR is.
5. It would be easy to re-do this PR in the future.

Give this comment a thumbs up or thumbs down. Otherwise, I'm starting the clock.

PR #8670, opened 2019-04-10
UI Updates (mostly states) to match latest mockups

slafleche:
This is has the rest of the style updates to consistency (first pass) We will do more next sprint through QA.

Closes: https://github.com/vanilla/knowledge/issues/826


charrondev:
A few buttons still have some issues:

## Search tabs

image

These are very tall.

## Standard buttons

These still lose their border when in the `:active` state. As a result they give the appearance of "shrinking" when pressed.

charrondev:
Additionally:



## Category Modal

There seems to be a regression in the button spacing here.

image

Issue #8675, opened 2019-04-10
Parameter $attributes of MediaItemModule->addButton() should have a default value


R-J:
That method would be used like that:

~~~
$item = new MediaItemModule();
$item->addButton(
    $buttonText,
    $buttonUrl,
    $attributes
);
~~~

All parameters are required, but if you use an empty array for attributes, `$attributes = []`, the method checks for an array element "class" and if that is not set, it defaults to "btn btn-secondary":

~~~
        if (!isset($attr['class'])) {
            $attr['class'] = 'btn btn-secondary';
        }
~~~

Although adding a default value for the attributes parameter wouldn't make the check for the class unnecessary, it would become more convenient to use.

PR #8676, opened 2019-04-10
Add default value for MediaItemModule->addButton() $attributes parameter

R-J:
This PR refers to [this issue](https://github.com/vanilla/vanilla/issues/8675). It changes the MediaItemModules addButton method to the following definition:
`public function addButton($text, $url, $attributes = ['class' => 'btn btn-secondary'])`
During process flow an empty $attributes parameter would have been replaced by exactly that array. 

Therefore passing it as default
1. makes this behaviour more obvious and
2. makes using this method more convenient


tburry:
Closes #8675.

Issue #8687, opened 2019-04-11
Improve ability to debug dynamic theme styles


charrondev:
User story from @slafleche 

It can be difficult to know what input you are giving to typestyle when our mixins our used. Adding the ability to log what is passed into the a `styleFactory()` statement would be useful.

Issue #8686, opened 2019-04-11
JS `log` utils are not working properly


charrondev:
**Describe the bug**
Nothing using the `log()` statement is being outputted.

**To Reproduce**
1. Set `$Configuration['Debug'] = true`
2. Use `log()` in the frontend
3. Notice nothing is logged.

**Expected behavior**
The item is logged out only while debug is enabled.

PR #8688, opened 2019-04-11
Add the ability to debug styleFactory calls

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8686
Closes https://github.com/vanilla/vanilla/issues/8687

This PR adds the ability for someone using `styleFactory` to debug their input into typestyle.

To do so just pass the imported `DEBUG_STYLES` symbol as the first param to the `style()` call.

I've used a symbol instead of a boolean or string so that it has an explicit name and it won't be confused with the first string parameter of the output name.

## Also

- I tightened up the type returned from `style()`. It's always a string. Then I fixed some broken places that were trying to use this string as a mixin.
- Fixed an issue where the debug value on the frontend wasn't being set properly in KB with hot reload enabled.

### DIFF
--- a/library/src/scripts/headers/vanillaHeaderStyles.ts
+++ b/library/src/scripts/headers/vanillaHeaderStyles.ts
@@ -382,7 +382,7 @@ export const vanillaHeaderClasses = useThemeCache(() => {
         },
     });
 
-    const tabButtonActive = style("tabButtonActive", {
+    const tabButtonActive = {
         color: globalVars.mainColors.primary.toString(),
         $nest: {
             ".vanillaHeader-tabButtonContent": {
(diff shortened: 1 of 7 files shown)


library/src/scripts/headers/vanillaHeaderStyles.ts
```diff
@@ -382,7 +382,7 @@ export const vanillaHeaderClasses = useThemeCache(() => {
         },
     });
 
-    const tabButtonActive = style("tabButtonActive", {
+    const tabButtonActive = {
```
slafleche: Why did you edit this line? You're creating an object here, not a style.

Issue #8690, opened 2019-04-11
Rich Editor: disabling it leaves "Rich" being the default format


R-J:
When the rich-editor is enabled it sets the format automatically to "Rich". When it is disabled it deletes the InputFormatter line from the config. But the formatter in the config-default is also set to Rich so the admin has to manually create the keys in the config and assign safe values to them.

a) config-defaults values shouldn't depend on a special plugin.
b) instead of deleting the keys in onDisable() it would be more comfortable for admins if the keys persist and some safe other format, a format that doesn't require a special plugin, is set.

R-J:
I just noticed that the rich editor most probably needs those two settings in the config-defaults because it **gets** not enabled on installation, but installation process is **simulated** by simply adding plugins initial config values to the config-defaults...

charrondev:
This one might need a little bit of discussions @R-J. I don't have time today, but I'll try to write down my thoughts next week, but the long and short of it is that writing to the config on enable and disable is going to have some edge cases no matter how you do it. I'm going to try to get some user stories together and see if we can decide which one is the best (which may end up being what's in your PR).

R-J:
Okay, but I think I have not make clear right now what happens when Rich Editor is disabled: 
![grafik](https://user-images.githubusercontent.com/3996187/56062868-1284f580-5d6e-11e9-8839-a74e4dc9c750.png)

Without an additional admin action no one can create new posts after the Rich Editor is disabled.

charrondev:
Closed by https://github.com/vanilla/vanilla/pull/8692

PR #8692, opened 2019-04-11
Set Markdown as input formatter when Rich Editor is disabled

R-J:
As described [here](https://github.com/vanilla/vanilla/issues/8690), when the Rich Editor is disabled it removes the input formatter lines from the config which causes Vanilla to use the values from config-defaults, but "Rich" is the format in config-defaults. Therefore the remains "Rich" which causes errors and is not the intention of those lines in the onDisable() method.

When Rich Editor is disabled the format must be set to some format which doesn't require a special editor. I've chosen Markdown but that was quite arbitrary. I feel allowing HTML "surprisingly" might be a bad decision, Text would be  to barebone and Markdown feels somewhat more modern than BBCode.

### DIFF
--- a/plugins/rich-editor/RichEditorPlugin.php
+++ b/plugins/rich-editor/RichEditorPlugin.php
@@ -32,8 +32,8 @@ class RichEditorPlugin extends Gdn_Plugin {
     }
 
     public function onDisable() {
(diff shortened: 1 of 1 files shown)


charrondev:
Markdown used to be the default editor so it's definitely the right choice here.

PR #8694, opened 2019-04-12
Add ability for addons to extend theme variables

initvector:
The variables asset of records returned from the themes API v2 resource can now be expanded on by addons.

### Overview
1. Add the `Vanilla\Theme\VariablesProviderInterface` interface. This defines the contract for addons who wish to register a theme-variables provider.
1. Add support for variable providers to `ThemeModel`. Currently, providers can only be registered and retrieved. No processing is applied through the model, since the variables asset is generated by `ThemesApiController`.
1. Add support for variable providers to `ThemesApiController`. Normalizing a theme now includes adjusting its variables asset, based on any variable providers configured in the controller's `ThemeModel`. It would be much better if this lived in the model, but it's been added to `ThemesApiController` since that's where all the asset generation is currently happening.

Relates to vanilla/knowledge#803

### DIFF
--- a/applications/dashboard/controllers/api/ThemesApiController.php
+++ b/applications/dashboard/controllers/api/ThemesApiController.php
@@ -260,6 +260,22 @@ class ThemesApiController extends AbstractApiController {
             $res["assets"][$assetKey] = $this->generateAsset($assetKey, $asset, $theme);
         }
 
+        // Allow addons to add their own variable overrides. Should be moved into the model when the asset generation is refactored.
+        $additionalVariables = [];
+        foreach ($this->themeModel->getVariableProviders() as $variableProvider) {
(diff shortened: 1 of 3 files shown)


charrondev:
Should this type of thing really be a modification of the API output? Couldn't an addon say, register a `Processor` with the `ThemeModel`?

initvector:
> I remain unconvinced that we should add additional dynamically dispatched events anywhere.

I don't think adding everything into the container is necessarily a better alternative. Adding a provider and configuring it via new rules in the container is going to require a hook. It will also mean that every time Vanilla's bootstrap is executed, that rule will be added. Vanilla's container is already pretty heavy and I am hesitant to continue packing it full of new rules for every new thing in the name of avoiding events.

The event hook lookup has its own performance considerations, but it will only be necessary when this specific API call is made.

PR #8701, opened 2019-04-15
Styles for mobile dropdown with drafts/revisions

slafleche:
Styles for mobile drafts/revisions dropdown on mobile.

Mockup: https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5c915482c2d2d2052d257d49

For: https://github.com/vanilla/knowledge/issues/838


charrondev:
I see a few styling issues here:


## Back button

image

The back button here isn't aligned properly. Additionally on the editor page the same back button overlaps with the saved draft text.

## Mockups

|Mockup|Actual|
|---|---|
|image|image|

- In the mockup we appear to have some offset from the bottom that is not present here.
- In the mockup the top corners do not have border radiuses.

library/src/scripts/headers/pieces/MobileDropDown.tsx
```diff
@@ -73,51 +73,53 @@ export default class MobileDropDown extends React.Component<IProps, IState> {
...
-                                        >
-                                            {chevronUp(classes.closeModalIcon)}
-                                        </Button>
-                                    </FrameFooter>
-                                </Frame>
-                            </Panel>
+                                    </PanelWidgetHorizontalPadding>
+                                </Container>
+                            </header>
+                            <FrameBody className={this.props.frameBodyClassName}>{children}</FrameBody>
```
charrondev: Since this has overflow: "auto" on it, it needs a wrapper of `<TouchScrollable />`. I'd recommend giving this a test in browser stack.

charrondev: image

Example from my real iOS device. Nothing here is scrollable.

slafleche:
@charrondev I checked with Elisa, dropdowns on mobile are not centered, they are from the top and they grow up to 100% height. 

Could not reproduce the cancel button being off. Is it still broken on your end?

PR #8725, opened 2019-04-23
Add custom background image support to Splash component

initvector:
`splashStyles` now returns a `customBackground` member, which is a function used to generate a background-image style. This is used by the `Splash` component to render a custom background.

### Overview
1. Add `customBackground` member to the result of `splashStyles`. This is a function that generates a background-image style.
1. Add `backgroundImage` prop to `Splash` component. If provided, this value will be used as the background image for the rendered element.
1. Fix styling inconsistencies in `splashStyles`. Thanks [Prettier](https://prettier.io).

Relates to vanilla/knowledge#936

### DIFF
--- a/library/src/scripts/splash/Splash.tsx
+++ b/library/src/scripts/splash/Splash.tsx
@@ -35,13 +35,7 @@ export class Splash extends React.Component<IProps> {
 
         return (
             <div className={classNames(className, classes.root)}>
-                <div
-                    className={classNames(
-                        classes.outerBackground,
-                        this.props.outerBackgroundImage &&
-                            classes.customOuterBackground(this.props.outerBackgroundImage),
-                    )}
-                />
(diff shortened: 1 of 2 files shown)


library/src/scripts/splash/Splash.tsx
```diff
@@ -34,7 +35,12 @@ export class Splash extends React.Component<IProps> {
 
         return (
             <div className={classNames(className, classes.root)}>
-                <div className={classes.outerBackground} />
+                <div
+                    className={classNames(
+                        classes.outerBackground,
+                        this.props.backgroundImage && classes.customBackground(this.props.backgroundImage),
```
slafleche: Can you please make it `outerBackgroundImage` and `customOuterBackground`. For now, we only have the outer, but the component is already set up for supporting the inner background too.

PR #8726, opened 2019-04-23
Fix KB's home header, count styles, border radiuses

slafleche:
Various small styling bug fixes. most notably fixing the special header styles for the KB's home page, fixing border radiuses and the count styles in the meBox header (red background)


slafleche:
@charrondev should be fixed!

PR #8729, opened 2019-04-25
Fixed text input styles for iOS devices

slafleche:
Fixes issue on iOS devices where some input text fields are being cut off.

I split the "textInputSizing" into 2 functions, one that calculates its own height based on the font size and padding and the other that figures out the padding based on the target height.

### DIFF
--- a/plugins/rich-editor/src/scripts/editor/richEditorFormClasses.ts
+++ b/plugins/rich-editor/src/scripts/editor/richEditorFormClasses.ts
@@ -6,15 +6,15 @@
 import {
     paddings,
     placeholderStyles,
-    textInputSizing,
     colorOut,
     unit,
     absolutePosition,
     pointerEvents,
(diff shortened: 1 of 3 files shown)


plugins/rich-editor/src/scripts/editor/richEditorFormClasses.ts
```diff
@@ -6,15 +6,15 @@
...
     colorOut,
     unit,
     absolutePosition,
     pointerEvents,
     margins,
     negative,
+    textInputSizingFromSpacing,
 } from "@library/styles/styleHelpers";
-import { styleFactory, useThemeCache } from "@library/styles/styleUtils";
+import {styleFactory, useThemeCache} from "@library/styles/styleUtils";
```
charrondev: What happened to the whitespace in this file?

PR #8732, opened 2019-04-25
Fix ProxyRequest so that it can handle content-less 204 responses.

nervoustwit:
Presently in the ProxyRequest class we expect successful responses to have a response body. Many APIs will return no content and a 204 response code when PATCHing or PUTing. This is being interpreted as a failure. 

This PR will check if the response code is not 2XX before concluding there was a negative response.

[See this document for more information about 204 responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)

### DIFF
--- a/library/core/class.proxyrequest.php
+++ b/library/core/class.proxyrequest.php
@@ -140,10 +140,12 @@ class ProxyRequest {
         $startTime = microtime(true);
         $response = curl_exec($handler);
         $this->ResponseTime = microtime(true) - $startTime;
-        $this->ResponseStatus = curl_getinfo($handler, CURLINFO_HTTP_CODE);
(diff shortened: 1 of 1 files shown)


library/core/class.proxyrequest.php
```diff
@@ -140,12 +140,10 @@ protected function curlReceive(&$handler) {
         $startTime = microtime(true);
         $response = curl_exec($handler);
         $this->ResponseTime = microtime(true) - $startTime;
-
-        if ($response == false) {
```
alex-mtl: I suspect that the only change we need here is strict comparison operator:
```
    if ($response === false) {
```

alex-mtl:
Can we cover this with unit test?

PR #8735, opened 2019-04-26
Add support to make PATCH request through ProxyRequest.

nervoustwit:
Presently we do not treat PATCH requests differently from GET requests which any integral modern API will not support. This PR will package cURL request properly when the PATCH verb is invoked.

This PR closes: https://github.com/vanilla/vanilla/issues/8749

### DIFF
--- a/library/core/class.proxyrequest.php
+++ b/library/core/class.proxyrequest.php
@@ -337,6 +337,7 @@ class ProxyRequest {
         $requestMethod = strtoupper($requestMethod);
         switch ($requestMethod) {
             case 'PUT':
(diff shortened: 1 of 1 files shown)


alex-mtl:
Can we cover this case with unit test somehow?

nervoustwit:
> Can we cover this case with unit test somehow?

I can make an issue to create a test, but for now I really need this deployed in the next release.

charrondev:
I'll add an approval once I see that issue @nervoustwit. Please add the `Domain: PHP Tests` and `Type: Technical Debt` tags.

Issue #8576, opened 2019-03-26
Move robots.txt support into core


initvector:
Support for robots.txt currently lives in the [Sitemaps addon](https://github.com/vanilla/addons/tree/master/plugins/Sitemaps). The file is important to give direction to bots/crawlers of any site, indicating what should/shouldn't be cached, providing the location of sitemaps, etc. You can [read more here](http://www.robotstxt.org).

This feature should be pulled into core to allow consistent control of a site's robots.txt. A site should be able to dictate its crawl-able content directives without relying on an addon.

The [Vanilla\Web\Robots](https://github.com/vanilla/vanilla/blob/e27058a064fcabdcee4ea7b2e257b003a3d46714/library/Vanilla/Web/Robots.php) class was recently added to core. It should be used as part of the core robots.txt support.

## Implementation

- Move the utility controller extension from the Robots plugin into the utility controller directly.
- Move the internal routing from `robots.txt` -> The new robots location.
- Add the ability for a config setting to update implement the functionality of https://github.com/vanilla/addons/blob/master/plugins/InvisibilityCloak/class.invisibilitycloak.plugin.php
- Update the invisibility cloak addon to virtually set this config setting.

## Validation

- Ensure Invisibility cloak addon works properly with this.
- Ensure the sitemaps plugin still works.
- Ensure knowledge base sitemaps still works.

PR #8744, opened 2019-04-29
Add support for robots.txt

initvector:
Vanilla now has core support for serving robots.txt.

### Overview
1. Add `robots` endpoint to `UtilityController`. This endpoint essentially replaces the endpoint of the same name from the [Sitemaps addon](https://github.com/vanilla/addons/tree/master/plugins/Sitemaps). A config-based list of robots.txt rules will be loaded and rendered. The lookup of configs is backwards-compatible with Sitemaps. A default list of rules will be added if no config is found. Addons can hook in to modify the output using the `robots_init` event. This is an existing, backwards-compatible event.
1. Add route to internally redirect `/robots.txt` to `/utility/robots` as part of Dashboard's structure routine.
1. Add `Robots.Invisible` config to toggle the usage of "invisible" robots.txt content which is intended to tell crawlers the site should not be indexed.
1. Add migration of robots config rules from Sitemaps. If no robots rules already exist in the `Robots.Rules` config, the Sitemaps robots config (`Sitemap.Robots.Rules`) will be migrated.

Relates to #8576


applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -38,8 +68,16 @@ class UtilityController extends DashboardController {
...
+     * @param EventManager $eventManager
+     * @param SessionInterface $session
      */
-    public function __construct() {
+    public function __construct(ConfigurationInterface $configuration, EventManager $eventManager, SessionInterface $session) {
+        $this->configuration = $configuration;
+        $this->eventManager = $eventManager;
+        $this->session = $session;
+
         parent::__construct();
```
alex-mtl: Is there any specific reason why we need to extend UtilityController instead of create new one?

charrondev: I noticed the same thing, but it was part of the acceptance criteria that we had initially discussed.

applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -8,14 +8,44 @@
...
+    /** Default robots.txt contents. */
+    private const ROBOTS_DEFAULT = <<<ROBOTS_DEFAULT
+User-agent: *
+Disallow: /entry/
+Disallow: /messages/
+Disallow: /profile/comments/
+Disallow: /profile/discussions/
+Disallow: /search/
+ROBOTS_DEFAULT;
+
```
alex-mtl: If we move this to core, shouldn't it be better for these hardcoded `const`s to live in `Vanilla\Web\Robots` class?

applications/dashboard/controllers/class.utilitycontroller.php
alex-mtl: It looks like "bad practice in action" to me.

charrondev: Only that we've planned it poorly again. I think the purpose was to push this quickly to satisfy the ops team.

If we were to really make a small project out of it, I would have advocated for making this directly using `Vanilla\Web\Page`.

initvector: > Any reason to advocate that decision?

I believe the original reasoning was partially that this migration of functionality from Sitemaps to core should not include breaking changes. If we put it in its own controller (i.e. move it to a different URL), then the /utility/robots URL from Sitemaps would be broken.

applications/dashboard/controllers/class.utilitycontroller.php
initvector: My thoughts on this are:

1. They can't be private constants if they're on `Vanilla\Web\Robots`. Also, I'm _pretty sure_ they would only ever be used here. If I'm wrong and we do need to move them to `Robots`, having them as private constants means we can replace their usage without fear of breaking references in the future.
1. The contents of the default rules are application-specific URLs. I don't think they should live in general-purpose library classes. That said, the "invisible" rules could go in `Vanilla\Web\Robots`.

applications/dashboard/controllers/class.utilitycontroller.php
alex-mtl: But isn't it possible to have any controller/method bind to the route here:
```
Gdn::router()->setRoute("robots.txt", "/utility/robots", "Internal");
```
?

initvector: It's definitely possible to route robots.txt to any other URL. But that doesn't change the fact that /utility/robots was a functioning URL in Sitemaps that will no longer be available.

alex-mtl:
```
Gdn::router()->setRoute("robots.txt", "/robots/general", "Internal");
Gdn::router()->setRoute("utility/robots", "/robots/general", "Internal");
```
kind?

Issue #8714, opened 2019-04-17
Make sign-in pages use AJAX for form submissions


initvector:
Vanilla's entry pages are one of the only areas of the application that require a guest user to have a cookie. This is due to transient key checks that are performed in some requests. Alternative, cookie-less methods of these checks should be implemented on at least the registration and sign-in pages.

The proposed implementation is forgoing the cookie for guests, hijacking their entry form submissions with JavaScript and instead submitting them as AJAX requests, which should include the X-Requested-With header. Vanilla supports this as an alternative to its transient-key-based CSRF protection (introduced in #7068). The method is based on [an OWASP article](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#use-of-custom-request-headers), which proposes the use of custom headers, along with host and referrer validation.

charrondev:
Investigate what's going to `.js-form` (currently in `global.js`). I'm bumping up the point total because I'm unsure of what the exactly uses this.

alex-mtl:
Please work same branch [feature/8578-x-request-with](url) as: https://github.com/vanilla/vanilla/pull/8763
feature/8578-x-request-with

Issue #8578, opened 2019-03-26
Add X-Requested-With support as alternative to TK check on sign-in and registration pages


initvector:
Vanilla's entry pages are one of the only areas of the application that require a guest user to have a cookie. This is due to transient key checks that are performed in some requests. Alternative, cookie-less methods of these checks should be implemented on at least the registration and sign-in pages.

The proposed implementation is forgoing the cookie for guests, hijacking their entry form submissions with JavaScript and instead submitting them as AJAX requests, which should include the X-Requested-With header.

Verify guest users are not unnecessarily issued TK cookies.

PR #8763, opened 2019-05-01
Activate X-Request-With header validation instead of Transient key for entry controller

alex-mtl:
Remove unnecessary calls of

Gdn::session()->ensureTransientKey();

since we have a check down there, which can be confirmed with X-Request-With header

$this->Request->isAuthenticatedPostBack()

Closes: #8578 
Closes: #8714

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -71,6 +71,8 @@ class EntryController extends Gdn_Controller {
         parent::initialize();
         Gdn_Theme::section('Entry');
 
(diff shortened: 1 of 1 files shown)


alex-mtl:
@nervoustwit Could I ask you please to check how safe are our changes?
Could you please test [feature/8578-x-request-with](feature/8578-x-request-with) branch with some custom /entry/signin  endpoints?

nervoustwit:
I tested this using the native Vanilla log in, as well as the OAuth2 plugin and in a local instance of Arenanet and it all worked.

Issue #6798, opened 2018-03-14
Changing cache folder to path outside forum root will fail


R-J:
A [user reported](https://open.vanillaforums.com/discussion/36058/tmp-being-filled-with-files-called-atomic) that his /tmp folder is flooded with Vanilla files and showed an extract of a cached theme file.
He hasn't responded by now but I assume he uses /tmp as his cache folder.

When looking at the sources I found one hardcoded "/cache" appearance ([PR already created(https://github.com/vanilla/vanilla/pull/6797)). At first I thought that this might be the problem, but this method seems not to be used anywhere.

The problem could be in AddonManager->setCacheDir which already states "If this doesn't include **PATH_ROOT** then it will be prepended."
Obviously this will "fail" for anyone trying to use `/tmp`. This might be intended behaviour, but I would recommend to change either this or the passage in the constants.php file:

~~~
// If you want to change where these paths are located on your server, edit these constants.
if (!defined('PATH_CACHE')) {
    define('PATH_CACHE', PATH_ROOT.'/cache');
}
~~~

charrondev:
Interesting related issue from infra. https://github.com/vanilla/dev-inter-ops

x00:
`setCacheDir` clearly does not allow for the cache folder to be outside PATH_ROOT or it will append path root to it.  This will cause insentiences. 
```
    public function setCacheDir($cacheDir) {
        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0) {
            $cacheDir = PATH_ROOT.$cacheDir;
        }
        $this->cacheDir = $cacheDir;
        return $this;
    }
```

x00:
`setCacheDir` clearly does not allow for the cache folder to be outside PATH_ROOT or it will append path root to it.  This will cause issue, on scalable setup, when you have to give location outside of the forum directory.  Also becuase there is inconsistency you can't just symlink an hope for the best. 
```
    public function setCacheDir($cacheDir) {
        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0) {
            $cacheDir = PATH_ROOT.$cacheDir;
        }
        $this->cacheDir = $cacheDir;
        return $this;
    }
```

x00:
If you are using boostap.before.php to route config/cache on a milt-tenant setup you can't use symlinks, especially if you want to keep data separate, which cna be a security requirement.

x00:
[...]

PR #8764, opened 2019-05-02
setCacheDir allow for the cache folder to be outside PATH_ROOT

x00:
https://github.com/vanilla/vanilla/issues/6798#issuecomment-485390630

### DIFF
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -1216,9 +1216,6 @@ class AddonManager implements Contracts\AddonProviderInterface {
      * @return AddonManager Returns `$this` for fluent calls.
      */
     public function setCacheDir($cacheDir) {
-        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0) {
-            $cacheDir = PATH_ROOT.$cacheDir;
-        }
         $this->cacheDir = $cacheDir;
         return $this;
     }


library/Vanilla/AddonManager.php
```diff
@@ -1216,9 +1216,6 @@ public function getCacheDir() {
      * @return AddonManager Returns `$this` for fluent calls.
      */
     public function setCacheDir($cacheDir) {
-        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0) {
```
alex-mtl: Wouldn't it be better to extend this function with new parameter for better backward compatibility:
```
    public function setCacheDir($cacheDir, bool $relative = true) {
        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0 && $relative) {
            $cacheDir = PATH_ROOT.$cacheDir;
        }
        $this->cacheDir = $cacheDir;
        return $this;
    }
```

x00: What I removed wasn't backward compatible in the first place

If you read
https://github.com/vanilla/vanilla/issues/6798#issuecomment-485390630

Adam says he can't think of any reason why it would be a problem. 

It defeats the whole ability to set cache root, which used to be possible. Just like you can set the uploads root. 

So we have a situation where is some of the time cache root is correct and in other case it is needlessly appending the path root. So you are in some situation it is no possible to win. 

not the first time I encounter a situation like this one time a method had two parameter reversed break backward com. 

You can't enforce backward compatibility of a change that is already breaking backward compatibility, especially for no good reason. 

The is whole architecture is fairly new, and those line where included based on a wrong assumption.

x00: It basically breaks many applications of multi-tenancy setups or scalable setus, where symlinks are not viable.

charrondev:
@x00 I'm afraid you just missed the boat on breaking changes. 3.0 just went out and I don't want a 4.0 right after. If you want this merged it will have to be in a backwards compatible manner.

library/Vanilla/AddonManager.php
alex-mtl: Sorry, I can't `$relative = false` since it will break backward compatibility.

x00: well it obviously would but @charrondev implied it was better.  

```
We have more stringent requirements at this point about backwards compatibility, specifically because of backwards compatibility issues in the past. If you want to make this change, it will have to be in a backwards compatible manner.

There are few of options for this.

Make a new method with the defaults that make sense, and deprecate the old one.
Update the existing method with a less ideal default.
```

You introduced this breaking change not me.

x00:
@charrondev @alex-mtl 
2.8 doesn't have `$relative = false`

    https://github.com/vanilla/vanilla/blob/Vanilla_2.8.3/library/Vanilla/AddonManager.php#L1218

The release candidate doesn't have `$relative = false`

https://github.com/vanilla/vanilla/blob/Vanilla_3.0-rc1/library/Vanilla/AddonManager.php#L1218

You introduced this breaking change.

library/Vanilla/AddonManager.php
x00: `$relative` is not even in the release candidate

alex-mtl: You probably did not get my message correctly.
That was an example of code which:
-  would respect backward compatibility (the way this method  was called before and the way was functioning before remains absolutely the same).
- and would bring new functionality and new behavior you need for some cases same time
- and that is acceptable to me to approve such PR

x00: My code is not calling the methods. The framework is through the constructor, which is ultimately a part of boostraping. 

1. I could do this entirely within the method. if the if I detect if the directory depth is deeper than the framework, which would work in some cases but not others. 
2. Could introduce another config value to force absolute paths, which is probably the more likely candidate. 

It is also a mixed bug, it is not as if there is a consistent result with absolute paths as it is stands. 

I'm supporting your breaking code, which I no idea why you implemented nor what the application. All I know is it wasn't done with developers in mind.

The developer program is almost dead after 10 years. Almost nobody supports addons over versions, it is just too much hassle. The drop off rate is stark. You are not going to get contributors if focus exclusively on your clients, and stuff like branding the dashboard. I'm left wondering why bother? There are very few of us who have stuck it out for this long.

alex-mtl: > Almost nobody supports addons over versions,

Which addons are we talking about?

 Just create an issue in related repository please.

x00: I'm not talking about addon developed by the core team, I;m pointing out the incentives for developing this framework are diminished. There is a drop of in interest in people developing the framework, and supporting their addons across version of the framework. This has ab impact on contributors.

x00: I will go for the config override. It wouldn't be the first an probably won't be the last, because of bugs of this nature.

linc:
Hey @x00, can you help me understand the timeline here? My understanding is that this breaking change was made prior to the Vanilla 2.3 release. We rebuilt our addon caching system from scratch here: https://github.com/vanilla/vanilla/pull/3693

This was a huge undertaking and you're correct that that breaking change was an oversight in the process, but it feels like a bit of water under the bridge after 3 years and 4 (going on 5) major releases. The only big change I see since then was making the caching optional here: https://github.com/vanilla/vanilla/pull/5496

We do care that we made a breaking change in 2016, but it feels pretty bad to respond to that by saying "let's break it again" after this much time has elapsed. Further, throwing sand in our eyes by insulting the popularity of the platform is an egregious strawman argument. You have two experienced developers spending their time in good faith trying to negotiate a solution to your issue and you're claiming they don't care. I didn't make it a job requirement for them to spend their afternoon arguing on our public repo, so questioning their motives is both unfair and counterproductive.

It feels like you're saying this breakage is new in either 3.0 or 2.8 (or maybe even 2.5?), but that doesn't appear to be the case. Am I wrong in my above assessment? If you're waiting years to upgrade your Vanilla install, this feels like a risk that goes with that territory. If there's something else I'm missing, we can talk more about how to mitigate that change for the 3.0 release.

x00:
I'm not exactly sure when it happened, however I certainly didn't use 2.3 with this client I haven't known them that long. I was setting the constant value and it worked. I don't remember having to patch this prior 2.6, I may have used the relative method but that is not possible now. This is bit more complicated than it appears, it might have been how this method was used prior rather than the method itself. 

I will do a PR with a config based override, which is backward compatible becuase it won't change the existing behaviour. I will still have to patch this for some time till it makes it to a release. 

I'm not questioning their motives personally. I'm saying that if vanilla as a company wants contributors it has to support a viable developer program, and that has declined in recent years and this is frustration coming out after many years.

If I degraded them in any way I'm sorry, that wasn't my intention.

linc:
> This is bit more complicated than it appears, it might have been how this method was used prior rather than the method itself.

I agree. It definitely requires a deeper dive to unwind this, and we're a bit short on time given the circumstances.

If you get us a config override PR today, we can squeeze that in the 3.0 release so you can forgo further hacks.

> I'm saying that if vanilla as a company wants contributors it has to support a viable developer program, and that has declined in recent years and this is frustration coming out after many years.

I hope Alex and Adam working with you here, as well as other developers assisting in open source releases, is evidence that we have been encouraging interaction with the community as our company grows to a size that we aren't as thinly spread in our responsibilities.

x00:
Sorry Only just saw this bank holiday. Appreciated. I will put do a config override immediately, after closing this.

x00:
https://github.com/vanilla/vanilla/pull/8772

PR #8766, opened 2019-05-02
Fix inconsistencies in Panel Layout and convert its styles to JS

slafleche:
The panel layout still had SCSS styles. This PR converts those styles and fixes styling issues.


plugins/rich-editor/src/scripts/editor/richEditorFormClasses.ts
```diff
@@ -23,7 +23,7 @@ import { richEditorVariables } from "@rich-editor/editor/richEditorVariables";
 
 export const richEditorFormClasses = useThemeCache((legacyMode: boolean = false) => {
     const globalVars = globalVariables();
-    const headerVars = titleBarVariables();
+    const titleBarVars = titleBarVariables();
```
charrondev: I'm pretty sure this file is just deleted at this point. It's been moved into KB as `editorFormStyles`, but probably got caught up in those merge conflicts a couple weeks ago. Could we just delete this file?

slafleche:
@charrondev @initvector I ran into some issues with the media queries not cascading as they normally do. In the Typestyle documentation they recommend making them "mutally exclusive". I did that, but I also made some new ones that are not. Poke me if you need more explanation. It only matters if you're using multiple media queries for the same class.

PR #8767, opened 2019-05-02
Embedded Comments - Inherit dashboard post format type

### LINKED ISSUE #8748: Embedded comments do not respect formatting config or dashboard settings
**Describe the bug**
When using embedded comments, the site's format settings are not respected.

**To Reproduce**
1. Set up embedded comments.
2. Set your text format as `Rich`.
3. Go to a page with embedded comments.
4. Notice a plain text area.

Expected behavior:
The format that I set in my dashboard should be used.

or

1. Set up embedded comments.
2. Set your text format as `Markdown`.
3. Go to a page with embedded comments.
4. Put markdown in the text area.
5. Make a post.
6. Notice the markdown did not get converted into HTML.

Expected behavior:

Markdown is rendered when the text format is set configured to be markdown.

**Additional context**
There's been some back and forth about this such as in https://github.com/vanilla/support/issues/272.

@tburry To me locking these into an older format is worse than before, because now, on top of not using the editor, if my forum format is set to markdown it will not be respected.

cchabilall83:
Per vanilla/vanilla#8738 we mounted an editor for embedded content and set the format type for those comments to TextEx.  The format of TextEx causes a mismatch with that of the form and would cause content to render incorrectly.  Quotes rendering was especially off (see below).  

This allows for the comment form to inherit the format type set in the dashboard so that there would be no conflict in rendering.  So changes had to be made in embed view and the rich editor plugin to ensure the data being accessed was an object instead of an array.  This was done to make sure rich content would display correctly on the page where the embedded comments were displayed.

Example of Embedded Comment with a quote
```
> @user1 said:
> Quoted Text form the body of the comment
> 

This looks weird ^
```

## To test:

Ensure you have embedded comments functioning

1) Test posting and embedded comment
2) Test quoting the the posted comment
3) Repeat steps with all different formats available

Close #8748

### DIFF
--- a/applications/vanilla/views/discussion/embed.php
+++ b/applications/vanilla/views/discussion/embed.php
@@ -9,6 +9,7 @@ if (!function_exists('WriteComment'))
     include($this->fetchViewLocation('helper_functions', 'discussion'));
 
 ?>
+
 <div class="Embed">
     <?php
     echo '<span class="BeforeCommentHeading">';
@@ -38,6 +39,7 @@ if (!function_exists('WriteComment'))
         if ($this->Pager->lastPage()) {
             $LastCommentID = $this->addDefinition('LastCommentID');
             if (!$LastCommentID || $this->Data['Discussion']->LastCommentID > $LastCommentID)
+                $this->Data['Discussion'] = (is_array($this->Data['Discussion'])) ? (object)$this->Data['Discussion'] : $this->Data['Discussion'];
                 $this->addDefinition('LastCommentID', (int)$this->Data['Discussion']->LastCommentID);
             $this->addDefinition('Vanilla_Comments_AutoRefresh', Gdn::config('Vanilla.Comments.AutoRefresh', 0));
         }
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -66,6 +66,8 @@ if (!function_exists('writeComment')) :
         // Whether to order the name & photo with the latter first.
         static $userPhotoFirst = null;
 
+        $comment = (is_array($comment)) ? (object)$comment: $comment;
+
         if ($userPhotoFirst === null) {
             $userPhotoFirst = c('Vanilla.Comment.UserPhotoFirst', true);
         }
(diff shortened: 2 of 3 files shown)


applications/vanilla/views/discussion/embed.php
```diff
@@ -38,6 +39,7 @@
         if ($this->Pager->lastPage()) {
             $LastCommentID = $this->addDefinition('LastCommentID');
             if (!$LastCommentID || $this->Data['Discussion']->LastCommentID > $LastCommentID)
+                $this->Data['Discussion'] = (is_array($this->Data['Discussion'])) ? (object)$this->Data['Discussion'] : $this->Data['Discussion'];
```
charrondev: Do you remember why this was added to the PR? The same question applies for the other 2 new seemingly unused variables.

applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -66,6 +66,8 @@ function writeComment($comment, $sender, $session, $currentOffset) {
         // Whether to order the name & photo with the latter first.
         static $userPhotoFirst = null;
 
+        $comment = (is_array($comment)) ? (object)$comment: $comment;
+
```
cchabilall83: https://github.com/vanilla/vanilla/blob/20a36a47f1e4103b6bc1b95af20d6a654c9581b3/applications/vanilla/views/discussion/helper_functions.php#L75

applications/vanilla/views/discussion/embed.php
cchabilall83: https://github.com/vanilla/vanilla/blob/20a36a47f1e4103b6bc1b95af20d6a654c9581b3/applications/vanilla/views/discussion/embed.php#L43

plugins/rich-editor/RichEditorPlugin.php
```diff
@@ -119,10 +119,13 @@ public function base_afterFlag_handler($sender, $args) {
     protected function addQuoteButton($sender, $args) {
         // There are some case were Discussion is not set as an event argument so we use the sender data instead.
         $discussion = $sender->data('Discussion');
+        $discussion = (is_array($discussion)) ? (object)$discussion : $discussion;
+
```
cchabilall83: https://github.com/vanilla/vanilla/blob/20a36a47f1e4103b6bc1b95af20d6a654c9581b3/plugins/rich-editor/RichEditorPlugin.php#L133

PR #8768, opened 2019-05-02
Add intermediate caching to webpack builds

charrondev:
With the build times creaping up again, I figured adding a little disk caching here would help. I've used this plugin before. It stores intermediate cache files tied with cache keys generated based on the current environment.

This significantly speeds up builds.

## Stats

Here's some stats from my localhost

**Production Builds**

Build Type|Build Number|Duration|Difference to baseline
---|---|---|---
Production|Initial (No Caching)|74s|N/A
Production|Followup (No Caching)|71s|N/A
Production|Initial (w/Caching)|89s|**+20%**
Production|Followup (w/Caching)|48s|**-33%**
Development|Initial (No Caching)|35s|N/A
Development|Followup (No Caching)|33s|N/A
Development|Hot Rebuild (No caching)|11s|N/A
Development|Initial (w/ Caching)|36s|**+3%**
Development|Followup (w/ Caching)|10s|**-70%**
Development|Hot Rebuild (w/ caching)|0.6s/1.5s|**-80/-90%**

## Production builds take a little bit of a hit the first time, but as long as a significant portion of the code stays the same that should only ever be experienced once on any given installation or after clearing of the cache directory.

This is mostly because the minification/uglification takes ~45seconds minimum. Even if 98% of everything before is cached on disk, it will still have to redo the magnification from scratch and it's quite intensive.

This doesn't apply to the dev builds which is why there is a such a massive speedup present there.

## Purging the cache

I talked briefly with @DaazKu. I was told that a 2gb limit was fine. A build cache is about ~100-300mb depending on if there's already a cache file. After the amount to 2GB it will automatically start purging the older cache files.


DaazKu:
Please make **absolutely sure** that the 2G will start purging itself out
thanks

charrondev:
@alex-mtl It logs out the currently used cache space with every build, so it should become pretty quickly apparent if it is now.

image

Some tests on my local indicate that it purges properly.

PR #8772, opened 2019-05-07
A config override for relative/absolute path in setCacheDir

x00:
I used `Cache.` config prefix becuase it already exists. I didn't use `Cache.Filecache.` becuase this is a lower level thing than filecache, and involves the object/class/addon caching regardless of type of cache.

Default is relative for backwards compatibility. 

Setting false will allow you to set an absolute path to the cache directory, as a constant in a similar way to the upload directory.  Useful for multi-tenancy arrangements.

### DIFF
--- a/library/Vanilla/AddonManager.php
+++ b/library/Vanilla/AddonManager.php
@@ -1216,7 +1216,7 @@ class AddonManager implements Contracts\AddonProviderInterface {
      * @return AddonManager Returns `$this` for fluent calls.
      */
     public function setCacheDir($cacheDir) {
-        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0) {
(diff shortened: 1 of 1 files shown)


library/Vanilla/AddonManager.php
```diff
@@ -1216,7 +1216,7 @@ public function getCacheDir() {
      * @return AddonManager Returns `$this` for fluent calls.
      */
     public function setCacheDir($cacheDir) {
-        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0) {
+        if ($cacheDir !== null && strpos($cacheDir, PATH_ROOT) !== 0 && Gdn::config('Cache.DirRelative', true)) {
```
alex-mtl: Where `Cache.Relative` is declared and defined?

charrondev:
I'm targeting `3.0` with a backport for this. It is not required for our infra.

PR #8778, opened 2019-05-13
KB - Fixed visual issues with search page

slafleche:
Visual fixes for categories page:
1. Next/Prev button double in size if only 1 button is there
2. Reduced spacing between title and meta data
3. Refactored simple pager component to use newer style standard
4. Modified result interface to accept no excerpt, since we do not want any for the categories page
5. Removed excerpt from the categories page

For: https://github.com/vanilla/knowledge/issues/903


charrondev:
A few notes:

## Top padding
The spacing between the title and the list is accurate to the mockups, but also feels out of space because we haven't implemented the pagination counts or the sorting options yet.

image

Could we trim this down a bit until we implement that UI?

## Search pagination

I don't see the updated/fixed pagination styles on the search page. I think we need to use the same approach there as well.

image

slafleche:
@charrondev ready for re-review.

PR #8783, opened 2019-05-14
Fix editor focus issues

charrondev:
Apparently we even had unit tests for this behaviour that I'll need to modify.

Fixes https://github.com/vanilla/knowledge/issues/628

Previously a different behaviour had been intentionally implemented. I've changed it to comply with the requirements.


charrondev:
Closing because more discussion is necessary. See the notes in https://github.com/vanilla/knowledge/issues/628.

PR #8809, opened 2019-05-15
KB: Location picker and embed flyout design tweaks

slafleche:
Design tweaks for embed flyout:
- Now wider on desktop
- Fixed vertical positioning to match the emoji picker's position
- Updated placeholder text to be "https" instead of "http"

Design tweaks for location picker: 
- Fixed styles for back chevron "<"
- Fixed paddings so the selected state goes all the way to the edge

For: https://github.com/vanilla/knowledge/issues/914

### DIFF
--- a/plugins/rich-editor/src/scripts/flyouts/EmbedFlyout.tsx
+++ b/plugins/rich-editor/src/scripts/flyouts/EmbedFlyout.tsx
@@ -4,7 +4,7 @@
  * @license GPL-2.0-only
  */
 
-import DropDown from "@library/flyouts/DropDown";
+import DropDown, { FlyoutSizes } from "@library/flyouts/DropDown";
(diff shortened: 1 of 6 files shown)


plugins/rich-editor/src/scripts/flyouts/EmbedFlyout.tsx
```diff
@@ -96,7 +99,7 @@ export class EmbedFlyout extends React.PureComponent<IProps, IState> {
                                     className={classNames("InputBox", classesInsertMedia.insert, {
                                         inputText: !this.props.legacyMode,
                                     })}
-                                    placeholder={t("http://")}
+                                    placeholder={t("https://")}
```
alex-mtl: what do we expect when translate **http** or **https**?

PR #8815, opened 2019-05-16
Fix some OpenAPI docs

tburry:
- Fix typo in pagination and move to a component.
- Remove HTML tags. OpenAPI expects pure Markdown.
- Create components for `dateInserted` and `dateUpdated` parameters.
- Remove some duplicate `x-addon` properties.

### DIFF
--- a/applications/dashboard/openapi/search.yml
+++ b/applications/dashboard/openapi/search.yml
@@ -5,7 +5,14 @@ paths:
     get:
       parameters:
       - description: |
-          <p>Restrict the search to the specified main type(s) of records.</p>
+          Filter the records using the supplied terms.
+        in: query
+        name: query
+        schema:
+          type: string
+        x-search-filter: true
+      - description: |
+          Restrict the search to the specified main type(s) of records.
         in: query
         name: recordTypes
         schema:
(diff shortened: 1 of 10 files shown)


applications/dashboard/openapi/search.yml
```diff
@@ -32,59 +39,52 @@ paths:
...
+          Allow search in archived categories.
         in: query
         name: includeArchivedCategories
         schema:
           default: false
           type: boolean
       - description: |
-          <p>Filter the records using the supplied terms.</p>
-        in: query
-        name: query
```
alex-mtl: Is it intentionally removed `query` param?

applications/dashboard/openapi/search.yml
```diff
@@ -32,59 +39,52 @@ paths:
...
-        schema:
-          type: string
-        x-search-filter: true
-      - description: |
-          <p>Filter the records by matching part of their name.</p>
```
alex-mtl: I believe that only last line supposed to be replaced. All the rest get to this PR by accident I guess.
Because param with: `name: query` is one of our key param for  advanced search api:
https://github.com/vanilla/internal/blob/4671a163429f705d18de89b72c881abc1fed3c48/plugins/AdvancedSearch/controllers/api/SearchApiController.php#L153

PR #8839, opened 2019-05-21
Line Height Fix and Splitting "frame" styles into multiple files.

slafleche:
The `lineHeightAdjustment` we were using was incorrect. I fixed it.

I also split up frameStyles into multiple files to make easier to find.

This PR is to work towards closing https://github.com/vanilla/knowledge/issues/998, but I split up this part since the diff was getting too big.

I also snuck in some alignment and state style issues for the search icon and the me box (both in the toolBar)

Closes: https://github.com/vanilla/knowledge/issues/995
Closes: https://github.com/vanilla/knowledge/issues/994


library/src/scripts/routing/links/backLinkStyles.ts
```diff
@@ -55,11 +55,12 @@ const backLinkClasses = useThemeCache(() => {
     );
 
     const link = style("link", {
-        display: "inline-flex",
-        alignItems: "center",
-        justifyContent: "flex-start",
+        // display: "inline-flex",
```
charrondev: Looks like a few lines were left commented out here.

library/src/scripts/routing/links/backLinkStyles.ts
```diff
@@ -68,6 +69,12 @@ const backLinkClasses = useThemeCache(() => {
         },
     });
 
+    const forHeading = lineHeight => {
```
charrondev: Is the usage of this in an upcoming PR? I was looking for the usage at this point in time and was unable to find it.

The name seems a little vague though. I'm wondering if we should name the function to generate style classes a little differently though. Maybe `headingLinkMixin()` or something like that?

library/src/scripts/styles/typographyStyles.ts
```diff
@@ -23,18 +24,25 @@ export const typographyClasses = useThemeCache(() => {
         {
             fontSize: unit(globalVars.fonts.size.title),
             lineHeight: globalVars.lineHeights.condensed,
-            $nest: lineHeightAdjustment(globalVars.lineHeights.condensed, true),
-        },
+            transform: `translateX(${em(globalVars.fonts.alignment.headings.horizontal)})`,
+            $nest: lineHeightAdjustment() as NestedCSSSelectors,
```
charrondev: I have the same note here about the `NestedCSSSelectors` typing. This would allow us to remove the imports and remove both `as` declarations here.

PR #8840, opened 2019-05-22
Replace node-sass with sass

initvector:
This is done to address the [failing Travis CI cron builds](https://travis-ci.com/vanilla/vanilla/jobs/201993934).

See https://github.com/vanilla/vanilla/pull/8840#issuecomment-495147600 for additional context.


charrondev:
I've kind of hijacked this a little bit, but for ~the last year `libsass` (The C library that `node-sass` wraps)  has __not__ been the primary sass implementation.

[The team created a dart implementation](https://sass-lang.com/dart-sass) which is available in a plain javascript implementation here in the `sass` package. This package is almost entirely API compatible with `node-sass`.

I was recently experiencing some personal difficulty with long `node-sass` compile times. That in addition to the long deployment times it causes, convinced me to replace our implementations with just `sass`.

I did this for

- The new build process.
- The legacy dashboard build process.
- The legacy advanced editor build process.

**Notes**

- The dashboard and editor required upgrading some grunt plugins as well.
- We had 1 mixin that was not compatible with newer implementations of sass. I fixed it.

alex-mtl:
@charrondev But isn't it node-sass  4.12.0  require node 12?
Why did you set node dependencies to 10?

charrondev:
@alex-mtl `node-sass` was removed. Instead `sass` is now being used which is not a native module.

In any case both modules have a minimum node version required of ~Node 8, but `node-sass` tends to break with every node release because it's a native module.

image

We don't actually get much benefit of using the native module anymore though. Switching from `node-sass` to `sass` gives us a slight slowdown (The sass goes from ~0.1s to 0.3s), and gives us numerous benefits:

- **Not** a native module, so it doesn't break with every node release.
- **Not** a native module, so it machines installation don't need C++ compiler toolchain + to compile from source (Saves 60-120 seconds during installation of `node_modules`).
- Is the primary maintained official Sass distribution.

alex-mtl:
Ok. I see. But you need to change PR description then.
And this PR seem pretty dangerous to me. Since we switch the library for our builds.
Probably we should warn services, support and ops with some label?

charrondev:
I don't believe any notification is needed to services, support or ops around this. It is not a breaking change, the builds still run successful, and build requirements have actually been loosened here.

PR #8846, opened 2019-05-23
Refactor emoji extender to improve code standards and remove useless files

charrondev:
## Release Notes

_Releases notes for the open source release. These are not necessary for our cloud deployment strategy._

Some changes renaming has occurred within the EmojiExtender plugin and additional upgrade steps are required on certain systems to prevent having duplicate versions of the plugin.

1. Delete `EmojiExtender` from your plugins directory. The plugin directory now matches its addon key `emojiextender`.

## Changes

This plugin actually has no JS or CSS at all to run. However, the plugin author seems to have used an automated tool to bring in a bunch of alpha junk dependencies. They don't even install/build on modern systems so I've removed all of it wholesale. The important thing to note here _the gulpfile + lintfile + yofile + dependencies didn't actually do anything._

- Deleted useless files.
- Update our diff-standards command to work on added, modified, or copied files. Previously it could try to run CodeSniffer on a deleted file which would cause it to fail.
- Updated the plugin file to pass our linter.
- Update the casing of the plugin directory to match it's `addon.json` key (fixes a warning).


plugins/emojiextender/EmojiExtenderPlugin.php
```diff
@@ -152,43 +158,59 @@ public function emoji_init_handler($sender, $args) {
...
+            $hasManifestIcon = isset($manifest['icon']);
+            $icon = $hasManifestIcon
+                ? img(
+                    $emojiSet['basePath'] . '/' . $manifest['icon'],
+                    [
+                        'alt' => $manifest['name'],
+                        'class' => 'label-selector-image'
+                    ]
+                )
+                : '';
```
alex-mtl: Can we use `if` statements for such constructions. It seem a bit **big** for ternary operator.

plugins/emojiextender/EmojiExtenderPlugin.php
```diff
@@ -152,43 +158,59 @@ public function emoji_init_handler($sender, $args) {
...
+                '</div>' .
+                '</div>' .
+                '<div emojiset-body>' .
+                '<div><b>' . htmlspecialchars($manifest['name']) . '</b></div>' .
+                (empty($manifest['author'])
+                    ? ''
+                    : '<div class="emojiset-author info">' . sprintf(t('by %s'), $manifest['author']) . '</div>') .
+                (empty($manifest['description'])
+                    ? ''
+                    : '<p class="emojiset-description">' . Gdn_Format::wysiwyg($manifest['description']) . '</p>') .
```
alex-mtl: This fragment is a good candidate to be extracted to some small twig template or React component or at least dedicated method.

charrondev: Sure. I was attempting to be more minimal and just update the formatting (pretty sure it only works with Advanced Editor anyways, but I'll give it a go.)

charrondev: I've converted it to a twig view.

plugins/emojiextender/EmojiExtenderPlugin.php
charrondev: This got restructured with the twig view anyways.

PR #8852, opened 2019-05-24
Add styling and wireup for editor placeholder

charrondev:
Currently rich editor doesn't have a way to enable a placeholder.

This PR adds

- Support for providing quill options with the `options` parameter on rich editor.
- Styling support for this placeholder. This might seem a little odd, but a `::before` element is the only easy way to get some relatively positioned content into the editor that's not actually part of the content editable. The `.ql-blank` class is how quill applies this in its own styles (which we don't use). https://github.com/quilljs/quill/issues/1831#issuecomment-351938711

### DIFF
--- a/library/src/scripts/content/userContentStyles.ts
+++ b/library/src/scripts/content/userContentStyles.ts
@@ -274,6 +274,10 @@ export const userContentClasses = useThemeCache(() => {
         lineHeight: globalVars.lineHeights.base,
         fontSize: vars.fonts.size,
         $nest: {
+            // A placeholder might be put in a ::before element. Make sure we match the line-height adjustment.
+            "&::before": {
+                marginTop: lineHeightAdjustment()["&::before"]!.marginTop,
+            },
             ...headings,
             ...lists,
             ...paragraphSpacing,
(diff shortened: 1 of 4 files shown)


slafleche:
This isn't accessible though... The placeholder should be using the placeholder attribute...

I found this solution, which may not work in old browsers, but not the worst:

```
[contenteditable=true]:empty:before {
  content: attr(placeholder);
  display: block; /* For Firefox */
}
```

charrondev:
@slafleche It should be fixed now.

slafleche:
That's fine, but it needs to be there for the screen reader

slafleche:
I'm getting errors:

```
2:36:59 p.m. - File change detected. Starting incremental compilation...
plugins/knowledge/src/scripts/modules/editor/EditorForm.tsx(265,30): error TS2322: Type '{ placeholder: string; }' is not assignable to type 'IntrinsicAttributes & IProps'.
  Property 'placeholder' does not exist on type 'IntrinsicAttributes & IProps'.
plugins/knowledge/src/scripts/modules/editor/EditorPageModel.test.ts(10,27): error TS2305: Module '"../../../../../../../../../../../../Users/stephanelafleche/Documents/vagrant-vanilla/git_repos/vanilla/plugins/knowledge/src/scripts/modules/editor/EditorPageModel"' has no exported member 'IInjectableEditorProps'.
plugins/knowledge/src/scripts/modules/editor/EditorPageModel.test.ts(27,36): error TS2339: Property 'getInjectableProps' does not exist on type 'typeof EditorPageModel'.
plugins/knowledge/src/scripts/modules/editor/EditorPageModel.test.ts(111,36): error TS2339: Property 'getInjectableProps' does not exist on type 'typeof EditorPageModel'.
2:37:03 p.m. - Found 4 errors. Watching for file changes.
```

slafleche:
Could not check, I get an error.

PR #8860, opened 2019-05-27
Add TK information to csrf failure event

cchabilall83:
This adds the users and sites transient key to the the db logger when there is a csrf failure event.


nervoustwit:
This will be closed when https://github.com/vanilla/vanilla/pull/8860 will be merged and deployed.

Issue #8862, opened 2019-05-28
Signin forms do not handle errors


charrondev:
**Describe the bug**
The signin forms submitted over ajax do not properly handle errors.

**To Reproduce**
1. Go to `/entry/signin`
2. Submit the form with missing fields or anything else to trigger an error.
3. Observe a JS error and the spinner going forever.

**Expected behavior**

The form errors are displayed.

nervoustwit:
This is only a bug on forms that are not appearing in in-page popups. If I set my settings to Garden.SignIn.Popup it works fine.

tburry:
I just had this happen on /entry/connect too.

charrondev:
It's all entry connect pages. https://github.com/vanilla/vanilla/pull/8863 has the fix for all entry pages.

PR #8863, opened 2019-05-28
Fix error handling in entry AJAX forms

charrondev:
Fixes https://github.com/vanilla/vanilla/issues/8862

Starting with https://github.com/vanilla/vanilla/pull/8763 our entry pages are submitting through AJAX. Unfortunately there were some bugs in error handling.

This PR.

- Fixes the ajax submission logic.
- Adjust all entry forms to have consistent data structure and explictly put `AjaxForm` on their containers.
- Allow displaying of form errors returned from the ajax submission.

**Note on signin.php**

The `signin.php` view was particularly complicated because it contained 2 forms, and returning back an error rendered the full view. Therefor every error would cause both the forms to be rendered and returned back, while our old ajax form submit functionality expects only 1 form to be rendered as part of the view.

We already had a separate view and routing for the "Forgot?" password view though, so I've removed the kludgy and inaccessible javascript that toggled the 2 views and now the 2 forms are just on different pages.

This PR will need to be followed up with updating any view overrides in custom themes.


nervoustwit:
My first tests it works great. The only place I found that it doesn't work is on /entry/password

charrondev:
@nervoustwit Thanks I probably made a mistake there.

charrondev:
@nervoustwit Are you sure? I can't seem to reproduce this issue on the password page.

nervoustwit:
> @nervoustwit Are you sure? I can't seem to reproduce this issue on the password page.

OK my bad. I just looked again (cleared cache, etc.)  and it works.

PR #8864, opened 2019-05-28
Update exception class name

dannimajdalani:
This PR will rename an exception class that was created in [#8859](https://github.com/vanilla/vanilla/pull/8859)

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -1129,8 +1129,8 @@ class EntryController extends Gdn_Controller {
                                 'Reason' => 'Password',
                             ]);
                         }
-                    } catch (Gdn_CoreException $ex) {
-                        $errorMessage = htmlspecialchars(strip_tags($ex->getMessage()));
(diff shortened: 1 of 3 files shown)


applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -1129,8 +1129,8 @@ public function signIn($method = false, $arg1 = false) {
                                 'Reason' => 'Password',
                             ]);
                         }
-                    } catch (Gdn_CoreException $ex) {
-                        $errorMessage = htmlspecialchars(strip_tags($ex->getMessage()));
+                    } catch (Gdn_SanitizedUserException $ex) {
+                        $errorMessage = htmlspecialchars($ex->getMessage());
                         $errorMessage .= sprintf(' Click <a href=%s>here</a> to reset your password', url('/entry/passwordrequest'));
```
charrondev: You should leave the link back in it's original spot now. The idea of the `SanitizedUserException` would be that the message is able to be used directly and we can leave the message where it originally came from.

dannimajdalani: even without htmlspecialchars?

PR #8881, opened 2019-05-31
Twitter addon fixes

tburry:
- Twitter’s API changed `callback_url` to `oauth_callback` resulting in broken profile connects.
- Remove the ability to profile connect on another user’s profile. Twitter now whitelists their callback URLs so we can only redirect back to the generic profile URL which belongs to the current user. This also comes with an additional check that the profile that initiated the connect matches the current session user.
- Fix typo in the `profileConnectUrl()` function name.


tburry:
Closes https://github.com/vanilla/vanilla-patches/issues/555.

Issue #8838, opened 2019-05-21
Backports and bugfixes needed - 3.0


charrondev:
Before shipping out the final `release/3.0` we'll need a few backports and fixes.

## Needs backport

- [ ] https://github.com/vanilla/vanilla/issues/8680
- [x] https://github.com/vanilla/vanilla/issues/8748
- [ ] https://github.com/vanilla/vanilla/pull/8784
- [ ] https://github.com/vanilla/vanilla/pull/8816
- [ ] https://github.com/vanilla/vanilla/issues/8683
- [ ] Any medium + security fixes completed since the release was forked.

charrondev:
Rebranched with https://github.com/vanilla/vanilla/pull/8887

PR #8888, opened 2019-05-31
Fix issues with the phing build process

charrondev:
Needs backport to release/3.0 (so we can actually get the release out).

Related https://github.com/vanilla/vanilla/issues/8838

- Fixes some `addon.json` versions.
- Prevents build from running twice with the release script.
- Updates the environment version.
- Updates build to cleanup the `build/temp` directory. Previously only a message about it was emitted.
- Prevent some knowledge dist files from being distributed (Nothing important was there, only the rich-editor entry, but it was questionable).
- **Ensure FS permissions are preserved while zipping with phing.** Phing sets all permissions to 777 with it's ZIP command. That's terrible so I've switched to shelling out to `zip` which preserves the permissions.


build/package-exclude.txt
```diff
@@ -23,5 +23,6 @@ applications/dashboard/styleguide/**
 applications/dashboard/template/**
 applications/dashboard/scss/**
 applications/dashboard/js/src/**
+dist/knowledge/**
```
alex-mtl: Shouldn't this build to be knowledge agnostic?

charrondev: We had a discussion in person. Theses are files from the forum targeting the KB search page (the forum category filter input). We may revisit in the future.

PR #8891, opened 2019-06-03
Add the ability to use HTTP basic auth for token requests in Gdn_OAuth2 

tburry:
OAuth2 provides for the client ID and secret to be passed via HTTP basic authentication and some servers may only support this option. Here we are adding the option into the core class.

I tested this by using Google's OAuth settings in the OAuth2 addon.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -55,11 +55,6 @@ class Gdn_OAuth2 extends Gdn_Plugin {
     protected $settingsView;
 
     /**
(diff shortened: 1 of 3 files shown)


tburry:
This PR got borked. No idea.

initvector:
> This PR got borked. No idea.

Merge into master broke some tests. Fix is forthcoming.

tburry:
I'll redo the PR in a bit.

PR #8905, opened 2019-06-04
Tweak the locale settings help text

tburry:
While reviewing some locale strings I noticed two very similar strings. On top of that neither of them were used in the app, but rather a third string that we don’t currently translate.

This PR changes the string to one in our locales repo with a translation.

### DIFF
--- a/applications/dashboard/views/settings/plugins.php
+++ b/applications/dashboard/views/settings/plugins.php
@@ -18,7 +18,7 @@ if ($addonType === 'applications') {
     $title = '';
     $helpTitle = sprintf(t('About %s'), t('Locales'));
     $pathHelp = sprintf(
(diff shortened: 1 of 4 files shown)


applications/dashboard/views/settings/plugins.php
```diff
@@ -18,7 +18,7 @@
     $title = '';
     $helpTitle = sprintf(t('About %s'), t('Locales'));
     $pathHelp = sprintf(
-        t('LocaleHelp', 'Locales allow you to support other languages on your site. Once a locale has been added to your %s folder, you can enable or disable it here.'),
+        t('Locales allow you to support other languages on your site.'),
```
alex-mtl: If we remove `Once a locale has been added to your %s folder` 
what is the reason to use `sprintf()` function?

tburry: Good point. Please stand by for a commit.

PR #8912, opened 2019-06-06
Update garden-http to version 2

charrondev:
Our Garden HTTP version was out of date. I've brought it up to the latest version.

There was a slight change in the `HttpRequest` class provided by garden http that required minor updates to some strengthen tighten up some types.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "symfony/yaml": "^3.2",
         "tburry/pquery": "~1.1",
         "vanilla/garden-container": "^1.3.4",
-        "vanilla/garden-http": "~1.0",
+        "vanilla/garden-http": "~2.0",
         "vanilla/garden-schema": "~1.0",
         "vanilla/garden-password": "~1.0",
         "vanilla/htmlawed": "~2.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "276632fa2d34259857e6e4314b6f044e",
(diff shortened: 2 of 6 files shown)


tburry:
Do you think it's okay into include updates to other dependencies in PRs like this? I think there are pros and cons so am uncertain. In this case twig came in with a bug fix.

charrondev:
If we didn't want to allow something like twig to update itself on any random PR, we should be restricting the specified version more, like we do in our frontend.

That's my take at least. If we didn't blindly trust a `0.0.x` update of twig we should not have the version specified as `^2.5`.

initvector:
> Do you think it's okay into include updates to other dependencies in PRs like this? I think there are pros and cons so am uncertain. In this case twig came in with a bug fix.

It isn't ideal, but I'm okay with it in this case.

If the intent here is to update garden-http, then we should be updating garden-http, not all dependencies with a relevant update. It increases the potential for other library compatibility issues without a reasonable expectation for such issues. If I see a PR to update garden-http get merged, and a couple days later we notice an issue in Twig, that garden-http PR probably isn't going to immediately come to mind.

charrondev:
That's the whole purpose of dependency ranges. If we aren't comfortable with it, once again we should be specified tighter version ranges on our dependencies. I've done that on the our frontend for the most part, but am considering loosening it slightly.

initvector:
I'm not arguing against the use of version ranges. I'm saying it isn't ideal to update unrelated dependencies when you do not have to. There is not a reasonable expectation for anyone viewing this PR title to assume Twig was updated. If we are updating garden-http, then update garden-http. If we want to update all dependencies, then we open a PR clearly stating all dependencies are being updated.

I think everyone is comfortable using version ranges. My concern is in obscuring dependency updates.

PR #8929, opened 2019-06-11
Fix the APIv0 build

tburry:
The APIv0 tests were failing because of a mismatch from the upgrade to garden-http 2.0 (see line 245). This PR also includes some other inspection fixes that helped troubleshoot the issue.

The composer.json was only upgraded to include ext-pdo which should not have made a material change to the config.

### DIFF
--- a/tests/APIv0/APIv0.php
+++ b/tests/APIv0/APIv0.php
@@ -120,7 +120,7 @@ class APIv0 extends HttpClient {
     /**
      * Get a connection to the database.
      *
-     * @param bool $db Whther or not to add the db name to the DSN.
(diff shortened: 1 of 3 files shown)


tests/APIv0/APIv0.php
```diff
@@ -120,6 +120,7 @@ public function getConfigPath() {
     /**
      * Get a connection to the database.
      *
+     * @param bool $db Whther or not to add the db name to the DSN.
```
alex-mtl: Typo
```suggestion
     * @param bool $db Whether or not to add the db name to the DSN.
```

PR #8931, opened 2019-06-11
Implement HSTS 

alex-mtl:
Implement Http Strict Transport Security (HSTS) headers to increase security for sites when https.
(https://hstspreload.org/)

- [x] HSTS Model  - generates header string according to site configuration
- [x] HSTS Middleware - applies HSTS header for "new" controllers
- [x] Update class.controller.php to support HSTS for basic core controllers

Relates to: https://github.com/vanilla/vanilla-patches/issues/554

### Note
Once merged and deployed we need to enable HSTS for our major sites and clients one by one with some test period to check consequences.


applications/dashboard/controllers/class.settingscontroller.php
```diff
@@ -8,6 +8,7 @@
  * @since 2.0
  */
 use Vanilla\Addon;
+use Vanilla\Web\HttpStrictTransportSecurity\HttpStrictTransportSecurityModel as HstsModel;
```
charrondev: What was your reasoning for not just calling this `HstsModel`? Bit of mouthful eh?

alex-mtl: I prefer to have descriptive name to understand what is the thing. Not sure what is our standards and policy to use abbreviation for class names. 
To me there are not that many places where I need to call this class.
I've set up local short alias for this just for settingscontroller, because it was difficult to read the code with long name. In all other cases that was ok to keep full name for this class.

charrondev: Sounds good.

Issue #7459, opened 2018-07-13
Move remaining items using Gdn_Factory to use the Container


charrondev:
Interestingly enough `Gdn_Factory` is still used for quite a lot of things. I'm seeing it in:

- Configuration
- Dashboard hooks
- SearchController
- AssetModel
- SettingsController
- VanillaHooks

We should also deprecate `Gdn::factory()`. Currently `Gdn::getFactory()` is deprecated, but if we want everything to use the container instead, both should be deprecated.

PR #8970, opened 2019-06-17
Implement CSP 'frame-ancestors'  header for Gdn_Controller

alex-mtl:
## Release Notes

_These are mostly important for the OSS release._

### Upgrade Notes

- Delete `container.html` from the root of the vanilla installation.
- Run `/utility/update`.

## PR

Output `container.html` page through GDN_Controller to apply all standard checks and headers

apply CSP 'frame-ancestors' header for `Gdn_Controller`

Closes: https://github.com/vanilla/vanilla-patches/issues/551
**Note** Require to run utility/update to add new route for container.html (no DB change)


library/core/class.controller.php
```diff
@@ -275,6 +277,10 @@ public function __construct() {
         $hsts = Gdn::factory('HstsModel');
         $this->_Headers[HttpStrictTransportSecurityModel::HSTS_HEADER] = $hsts->getHsts();
 
+        $cspModel = Gdn::factory(ContentSecurityPolicyModel::class);
```
charrondev: I'm trying to get rid of `Gdn::factory()` (it just uses the container internally at this point). https://github.com/vanilla/vanilla/issues/7459

The standard way for this is `Gdn::getContainer()->get()` instead.

PR #8974, opened 2019-06-18
StoryBook configuration and utility components

slafleche:
Updated Storybook config with out break points, and accessibility plugin.

Added a heading and separator component for StoryBook use only.

Snuck in a minor color fix.


charrondev:
Note: This PR contains a unit test fix for an inconsistent test (async behaviour that we can't easily track). We bumped up the timeout to a full second which should make things work better.

PR #8975, opened 2019-06-18
Don’t allow the passing of the RankID directly to connect endpoints.

nervoustwit:
This PR will remove the RankID from being filtered out of the data being inserted into the User table and instead, filter it out in the entry endpoint. This way SSO plugins can allow trusted sources to pass the RankID. This is the same way we handle Roles.


nervoustwit:
related https://github.com/vanilla/vanilla/pull/8975

PR #9005, opened 2019-07-04
Oauth2: OIDC compliance - Added prompt param authorize uri

cchabilall83:
Epic Vanilla/Support#464

closes Vanilla/Support#461
closes Vanilla/Support#467


library/core/class.oauth2.php
```diff
@@ -466,6 +470,10 @@ protected function realAuthorizeUri(array $state = []): string {
         $state['token'] = $this->ssoUtils->getStateToken();
         $get['state'] = $this->encodeState($state);
 
+        if (array_key_exists('Prompt', $provider) && isset($provider['Prompt'])) {
+            $get['Prompt'] = $provider['Prompt'];
```
nervoustwit: By passing the `prompt` as `Prompt` it didn't work with my Authentication provider. It is case sensitive. It needs to be `prompt`.

library/core/class.oauth2.php
```diff
@@ -400,7 +401,9 @@ public function settingsEndpoint($sender, $args) {
...
             'TokenUrl' => ['LabelCode' => 'Token Url', 'Description' => 'Endpoint to retrieve the authorization token for a user.'],
             'ProfileUrl' => ['LabelCode' => 'Profile Url', 'Description' => 'Endpoint to retrieve a user\'s profile.'],
-            'BearerToken' => ['LabelCode' => 'Authorization Code in Header', 'Description' => 'When requesting the profile, pass the access token in the HTTP header. i.e Authorization: Bearer [accesstoken]', 'Control' => 'checkbox']
+            'BearerToken' => ['LabelCode' => 'Authorization Code in Header', 'Description' => 'When requesting the profile, pass the access token in the HTTP header. i.e Authorization: Bearer [accesstoken]', 'Control' => 'checkbox'],
+            'Prompt' => ['LabelCode' => 'Prompt', 'Description' => 'Prompt Parameter to append to Authorize Url', 'Control' => 'DropDown', 'Items' => [ 'none' => 'none', 'consent' => 'consent', 'login' => 'login', 'consent and login' =>  'consent and login']],
+            'Response Type' => ['LabelCode' => 'Code Token', 'Description' => 'The response type send in the authorize uri (note: certain values can change the work flow ie. id_token)'],
```
nervoustwit: Here I think the wording could be more descriptive.

The LabelCode should be "Response Type". The description could be something like:

 > `response_type` to be sent in the Authorization Request.

Also, can we add 'code token' to the defaults in the structure() function when you turn on the plugin.

library/core/class.oauth2.php
```diff
@@ -453,9 +456,10 @@ protected function realAuthorizeUri(array $state = []): string {
         $uri = val('AuthorizeUrl', $provider);
 
         $redirect_uri = '/entry/'.$this->getProviderKey();
+        $reponse_type = array_key_exists('Response Type', $provider) ? $provider['Response Type'] : 'code';
```
nervoustwit: This should be `$provider['Response_Type']`

PR #9007, opened 2019-07-05
Build dashboard

dannimajdalani:
Someone did a PR and forgot to run the build.


alex-mtl:
What has been changed? Could you point to the change or to the PR that broke the style?
```
applications/dashboard/js/dashboard.js
```
Is not just a build issue. Isn't it?

PR #9010, opened 2019-07-08
Allow users to view their own email address on their profile pages

tburry:
Users should be able to see their own email address when they are viewing their own profile.


bleistivt:
Why? This will most likely just confuse people into thinking their email is public.

Issue #8450, opened 2019-02-22
Rich Editor: Add a spoiler button title


tburry:
Currently if I hover over the spoiler toggle button I just get the title from the SVG which looks a bit odd to me. Can we add a `title` attribute to the `<button>` that reads: "Toggle Spoiler".

PR #9038, opened 2019-07-11
Added title attribute to spoiler button

slafleche:
As per https://github.com/vanilla/vanilla/issues/8450, add title on spoiler toggle button.

### DIFF
--- a/library/src/scripts/content/UserContent.story.tsx
+++ b/library/src/scripts/content/UserContent.story.tsx
@@ -46,7 +46,7 @@ storiesOf("CoreComponents/UserContent", module)
            }
        }
        </pre><p><br></p><h2>H2 Here. Spoiler next</h2><div class="spoiler"><div contenteditable="false" class="spoiler-buttonContainer">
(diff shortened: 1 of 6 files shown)


library/src/scripts/content/UserContent.story.tsx
```diff
@@ -46,7 +46,7 @@ storiesOf("CoreComponents/UserContent", module)
            }
        }
        </pre><p><br></p><h2>H2 Here. Spoiler next</h2><div class="spoiler"><div contenteditable="false" class="spoiler-buttonContainer">
-       <button class="iconButton button-spoiler js-toggleSpoiler">
+       <button title="Toggle Spoiler" class="iconButton button-spoiler js-toggleSpoiler">
```
initvector: Should this also be wrapped in the translation function?

slafleche: I checked with Adam, it was just a quick dump of rendered HTML. It will eventually be replaced by "real" rendered content.

Issue #9039, opened 2019-07-11
Rich Editor toolbars missing border radii


charrondev:
**Describe the bug**
Border radii from the rich editor toolbars are missing after the last border utility refactoring.

**Screenshots**

![image](https://user-images.githubusercontent.com/1770056/61082315-36041e80-a3f7-11e9-90a8-1eb8ce5edbb3.png)

PR #9040, opened 2019-07-11
Fix border radius for go back to home and for editor flyouts

slafleche:
Fixes a button issue (on the KB not found page) and border radii issues (same page and in the editor.

Removed the "type" from the back to home button, since it doesn't need one and was adding 2 conflicting classes.

Removed if statement bypassing all border styles if none were given. They should use defaults (which they now do).

Added fallback to global radius if none were found.

Closes: https://github.com/vanilla/vanilla/issues/9039


initvector:
> Fix/border radii

PR #9045, opened 2019-07-15
Refactored the way we use "type" for buttons

slafleche:
We were using "types" on buttons both for their HTML effects (like in forms) and also for classes. We had several class conflicts, so I decided to refactor the types. They no longer affect the classes, "baseClass" is the proper way to do that now.

"button" is the assumed type, but "submit" can also be specified when applicable.

### DIFF
--- a/library/src/scripts/forms/Button.tsx
+++ b/library/src/scripts/forms/Button.tsx
@@ -12,7 +12,6 @@ import classNames from "classnames";
 interface IProps extends IOptionalComponentID {
     children: React.ReactNode;
     className?: string;
-    type?: string;
     disabled?: boolean;
     prefix?: string;
     legacyMode?: boolean;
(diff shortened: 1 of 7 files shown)


library/src/scripts/forms/Button.tsx
```diff
@@ -61,7 +61,7 @@ export const getDynamicClassFromButtonType = (type: ButtonTypes | undefined) =>
             case ButtonTypes.CUSTOM:
                 return classes.custom;
             default:
-                return classes.standard;
```
alex-mtl: Shouldn't it return at least class `button`? or whatever?

slafleche: I'm going to leave it as is. This function is to explicitly set a standard class. If you don't want a standard one, you're on your own.

PR #9051, opened 2019-07-16
Fix embed styles and various rich editor bugs

charrondev:
This PR includes various styling and focusing issues in the editor. Many of these were just regressions from the refactoring and don't have particular issues.

- Attachments work again. They were broken before.
- Twitter embeds no longer flash while loading in the editor.
- The editor no longer appears focused (even when it's not) while an embed link loader is present.
- Embed link focus styles have been corrected.
- Fixes an issue where the the `<DocumentTitle />` would constantly re-render in the `<EditorForm />`.
- Fixes an issue where `<FlyoutToggle />` would constantly call `forceRenderStyles()` outside of any lifecycle method. I've placed it into a `useEffect()` hook, which prevents the constant recalling of this method in Rich Editor.
- Fixes scraped image embeds, which were not functioning properly

## Format Compat

In order to address some of the edit issues, a new service `FormatCompatService` has been introduced. This converts old rich embeds to new ones on edit.

It includes:

- The service itself.
- Unit tests and fixtures for the service.
- A trait for easy usage.
- Integration on the `PostController` edit methods.
- Integration on the `/api/v2/discussions`, `/api/v2/comments`, and `/api/v2/drafts`.
- A new default test case in `AbstractResourceTest` so that all API controllers with edit endpoints will be tested. Exclusions have been added for the `MessagesTest` because it does not have an edit endpoint (can't edit messages).

Knowledge base API integrations https://github.com/vanilla/knowledge/pull/1088
ReactionsTest exclusion https://github.com/vanilla/internal/pull/1925


library/src/scripts/embeddedContent/FileEmbed.tsx
```diff
@@ -7,6 +7,7 @@ import React from "react";
 import { IBaseEmbedProps } from "@library/embeddedContent/embedService";
 import Attachment from "@library/content/attachments/Attachment";
 import { mimeTypeToAttachmentType } from "@library/content/attachments/attachmentUtils";
+import { EmbedContent } from "@library/embeddedContent/EmbedContent";
```
initvector: I think we can ditch this, yeah?
```suggestion
```

charrondev:
@initvector This should be ready to go.

> Editing a Rich comment and canceling the edit will render the post with a broken post menu. Clicking the "gear" does nothing.

This needs a separate issue. It even effects Advanced Editor, so it's an old bug and completely unrelated. It's not a firefox specific bug either.

Issue #9026, opened 2019-07-10
The rich editor has odd behavior when hitting hard returns in lists


tburry:
**Describe the bug**

When I am in a list and hit return somewhere in the middle of the line then rather than creating a new list item the editor instead creates two list items: one blank line then the correct line with the item.

**To Reproduce**

1. Create a list with multiple items.
2. In the middle of the list, place the cursor in the middle of the line.
3. Hit return.
4. After you notice the bug, hit backspace.

**Expected behavior**

1. Hitting return should create a new item and then place the remaining part of the line in it.
2. Hitting backspace should remove the list item and restore the list to the starting state. I'm also okay with this removing the list formatting if that is our behavior, but it shouldn't remove the list formatting from a blank item above too.

**Screenshots**

![odd-list-behaviour](https://user-images.githubusercontent.com/97432/60988666-0af3cf00-a312-11e9-938d-113ff0c67089.gif)

Issue #8684, opened 2019-04-11
Rich Editor: Inserting new item at start of list is buggy


linc:
Create a simple list:

<img width="685" alt="Screen Shot 2019-04-11 at 11 05 25 AM" src="https://user-images.githubusercontent.com/117672/55968373-deb0af80-5c49-11e9-92f4-593fca74b99f.png">

Place cursor just before "One" (after the bullet) and press Return:

<img width="643" alt="Screen Shot 2019-04-11 at 11 05 16 AM" src="https://user-images.githubusercontent.com/117672/55968410-eff9bc00-5c49-11e9-85b8-8e4d91978cc2.png">

A new line was created at an unexpected place.

Add text (cursor hasn't moved) and press Return:

<img width="666" alt="Screen Shot 2019-04-11 at 11 05 40 AM" src="https://user-images.githubusercontent.com/117672/55968449-00119b80-5c4a-11e9-95c4-0b46c2c0ab7e.png">

Now we have an extra line.

PR #9059, opened 2019-07-16
Carry over and make a new line in the bullet list for the mid and end of the line cases.

planitnov:
Fixes https://github.com/vanilla/vanilla/issues/9026
Fixes https://github.com/vanilla/vanilla/issues/8684

- Fixes the bug named in the issue.
- Adds multiple test cases for the given behaviour.


alex-mtl:
I think that `composer.phar` shouldn't be here.

PR #9065, opened 2019-07-18
Move dom utils into @vanilla/dom-utils & @vanilla/react-utils

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8822
Closes https://github.com/vanilla/vanilla/issues/8823

This PR continues the refactoring of vanilla's `@library` module until smaller packages.

I've moved all of the stuff from `@library/dom/*` into `@vanilla/dom-utils` and `@vanilla/react-utils`.

A few notes

**@vanilla/dom-utils**

Only the pure dom stuff is left here. Anything react related went to react-utils. This includes the hooks for some of the classes. For example, `FocusWatcher` lives in `@vanilla/dom-utils` and `useFocusWatcher` lives in `@vanilla/react-utils`.

The single `@library/dom/domUtils.ts` was really large and was broken up into a lot of different pieces.

A few dependencies were able to be taken from the top level `package.json` and moved into this package. They are

- `focus-visible` - Polyfill for `:focus-visible` css selector.
- smoothscroll-polyfill - Polyfill for `window.scrollTo(, {behaviour: "smooth"})`.
- twemoji - Polyfill for emojis.

**@vanilla/react-utils**

This is some of our custom utility hooks, are react-dom mounting wrapper, and the `<Hoverable />` component.

- The `resize-observer-polyfill` was moved from the top level package.json for the `use-measure` hook.
- `react`, `react-dom`, and `typestyle` have been added as peer-dependencies of this package.

## Updating imports

All imports in this repo have been updated.

Import updates for KB are in https://github.com/vanilla/knowledge/pull/1087


library/src/scripts/flyouts/FlyoutToggle.tsx
```diff
@@ -4,18 +4,17 @@
...
+import Button from "@library/forms/Button";
+import { ButtonTypes } from "@library/forms/buttonStyles";
 import Modal from "@library/modal/Modal";
 import ModalSizes from "@library/modal/ModalSizes";
+import { t } from "@library/utility/appUtils";
+import { uniqueIDFromPrefix } from "@library/utility/idUtils";
+import classNames from "classnames";
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { forceRenderStyles } from "typestyle";
+import { useFocusWatcher, useEscapeListener } from "@vanilla/react-utils";
```
alex-mtl: Is there any way to commit only real changes but not reordered lines?
In this case 2 lines where deleted:
```
import FocusWatcher, { useFocusWatcher } from "@library/dom/FocusWatcher";
import EscapeListener, { useEscapeListener } from "@library/dom/EscapeListener";
```
and replaced with one new line:
```
import { useFocusWatcher, useEscapeListener } from "@vanilla/react-utils";
```
That would help to review and track repo changes.

packages/vanilla-dom-utils/src/events.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @param scopeSelector - And element to scope the event listener to.
+ *
+ * @returns The hash of the event. Save this to use removeDelegatedEvent().
+ */
+export function delegateEvent(
+    eventName: string,
+    filterSelector: string,
+    callback: (event: Event, triggeringElement: HTMLElement) => boolean | void,
+    scopeSelector?: string | HTMLElement,
+): string | undefined {
```
alex-mtl: In which cases this function can return `undefined`?

library/src/scripts/flyouts/FlyoutToggle.tsx
charrondev: This type of thing is mostly the cause of using the typescript refactoring tool "Cleanup Imports". It rewrites all the import declarations, sorted, and with no unused values.

Personally I don't think the imports are of much concern for a review. If the are incorrect the build will fail.

packages/vanilla-dom-utils/src/events.ts
charrondev: I cannot. Good catch.

library/src/scripts/flyouts/FlyoutToggle.tsx
alex-mtl: >  If the are incorrect the build will fail.
Completely disagree to this argument. :-)
Do we need a review at all if the build does not fail?

packages/vanilla-dom-utils/src/events.ts
charrondev: I think that it could at some point in the past but was changed a few months ago.

library/src/scripts/flyouts/FlyoutToggle.tsx
charrondev: It's not garbage. I trust the import generation of the typescript language. I also trust that broken imports break the build. If the file does not exist, it will fail loudly and plainly.

PR #9069, opened 2019-07-18
Implement js  source map files protection

alex-mtl:
Implements source maps delivery through backend app controller.

Closes: vanilla/knowledge#648


conf/config-defaults.php
```diff
@@ -87,6 +87,8 @@
 $Configuration['Garden']['Security']['Hsts']['Preload'] = false;
 $Configuration['Garden']['Security']['Hsts']['MaxAge'] = 604800;
 
+$Configuration['Garden']['Security']['SourceMaps']['Enabled'] = false;
```
charrondev:
```suggestion
```

I think having nothing in config-defaults would be a better default. That way, unless explicitly configured it will map to the debug value.

With this setup, the fallback value specified in the in the static controller above will never be reached.

build/scripts/configs/makeProdConfig.ts
```diff
@@ -96,6 +96,12 @@ export async function makeProdConfig(entryModel: EntryModel, section: string) {
         ],
     };
 
+    baseConfig.plugins!.push(new SourceMapDevToolPlugin({
+        namespace: `vanilla-${section}`,
+        filename: `sourcemaps/`.concat(Math.random().toString(36).slice(-5), '/', '[chunkhash]'),
+        publicPath: `/sourcemaps/${section}/`,
```
charrondev:
```suggestion
        publicPath: `/sourcemaps/${section}/`, // PHP-FPM will serve these files with some permission checks in StaticContentController.
```

conf/config-defaults.php
```diff
@@ -87,6 +87,7 @@
 $Configuration['Garden']['Security']['Hsts']['Preload'] = false;
 $Configuration['Garden']['Security']['Hsts']['MaxAge'] = 604800;
 
+
```
alex-mtl:
```suggestion
```

applications/dashboard/controllers/api/SourcemapsApiController.php
```diff
@@ -0,0 +1,40 @@
...
+class SourcemapsApiController extends AbstractApiController {
+
+    /**
+     * Get source map.
+     *
+     * @param string $path Relative path to source map file
+     * @return Data
+     */
+    public function get(string $path): Data {
+        $sourceMapsEnabled = Gdn::config()->get(
```
initvector: Why are we using globals here instead of injecting the config object?

applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -44,7 +44,7 @@ public function __get($name) {
                 return $this->data('Comments');
                 break;
         }
-        throw new Exception("DiscussionController->$name not found.", 400);
+        return null;
```
initvector: Is this supposed to be part of the PR?

applications/dashboard/controllers/api/SourcemapsApiController.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class SourcemapsApiController extends AbstractApiController {
+
+    /**
+     * Get source map.
+     *
+     * @param string $path Relative path to source map file
+     * @return Data
+     */
+    public function get(string $path): Data {
```
initvector: If we're sure we really don't want any kind of permission check on here, we should at least have an empty permission check (e.g. `$this->permission()`).

applications/dashboard/controllers/api/SourcemapsApiController.php
```diff
@@ -0,0 +1,40 @@
...
+        $sourceMapsEnabled = Gdn::config()->get(
+            'Garden.Security.SourceMaps.Enabled',
+            Gdn::config()->get('Debug')
+        );
+
+        $result = new Data('');
+        if ($sourceMapsEnabled) {
+            $fullPath = PATH_ROOT.DS.'dist'.DS.$path;
```
initvector: Are you certain this won't potentially expose a directory-traversal issue?

applications/dashboard/controllers/api/SourcemapsApiController.php
alex-mtl: I'll update.

applications/vanilla/controllers/class.discussioncontroller.php
alex-mtl: it is unrelated, but intentional.

applications/dashboard/controllers/api/SourcemapsApiController.php
alex-mtl: Does empty permission check does nothing?
I hate it :-(
But ok, I'll do.

applications/dashboard/controllers/api/SourcemapsApiController.php
alex-mtl: Update is coming, good catch.

applications/vanilla/controllers/class.discussioncontroller.php
charrondev: This was previously a `Gdn_Controller`, not an API controller, and checking for the existence of a properly on `DiscussionController` would cause an exception to be thrown. Pretty ridiculous.

applications/dashboard/controllers/api/SourcemapsApiController.php
charrondev: @initvector There should not be a permission check there. Even if we're in private communities, these are static assets. They should be served regardless of the user's permission level if the configuration is setup for it.

initvector: I understand opening up the permissions, but API endpoints are supposed to always have a permission check, even if it's empty. That convention is as old as API v2. I'm open to discussing if that should change, though.

charrondev: It should not apply to static asset controllers. That would mean `ThemesApiController` get endpoints, `LocalesApiController`, and `SourcemapsApiController`.

Do banned users not get served static assets served through an API? @tburry Thoughts?

charrondev: If we really want to keep it for consistency we could pass some constant. `ALWAYS_ALLOW` that would bypass any checks.

PR #9072, opened 2019-07-19
Fix message link visibility by changing styling

alexbrohman:
Changes colors of links and adds underlines to make it more obvious when an anchor tag is present messages.


alexbrohman:
New Coral appearance: 
image

New Dusk appearance: 

image

PR #9075, opened 2019-07-19
Fix slow pruning of access token tables with a high number of rows

dannimajdalani:
closes https://github.com/vanilla/support/issues/624
closes https://github.com/vanilla/support/issues/630
### Details

GDN_DateExpires does not have an index, causing queries to generate a token to run slow.


initvector:
This is necessary due to the pruning queries the model runs, [based on the `DateExpires` field](https://github.com/vanilla/vanilla/blob/5206e3e41e8b1a35a0aeb798f04ad95d06f74a68/library/core/class.accesstokenmodel.php#L30).

PR #9078, opened 2019-07-22
Fix longform Vimeo URLs not embedding

charrondev:
This regression was introduced with the refactoring. Noticed while doing some testing for https://github.com/vanilla/support/issues/631

I've added the URL type as well as the test for ti.

### DIFF
--- a/library/Vanilla/EmbeddedContent/Factories/VimeoEmbedFactory.php
+++ b/library/Vanilla/EmbeddedContent/Factories/VimeoEmbedFactory.php
@@ -43,7 +43,7 @@ class VimeoEmbedFactory extends AbstractEmbedFactory {
      * @inheritdoc
      */
     protected function getSupportedPathRegex(string $domain): string {
-        return "`^(/[a-z,[0-9]+){1,2}$`";
+        return "`^/\d+(/[a-z0-9]+)?$`";
     }
 
     /**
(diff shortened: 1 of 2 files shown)


library/Vanilla/EmbeddedContent/Factories/VimeoEmbedFactory.php
```diff
@@ -43,7 +43,7 @@ protected function getSupportedDomains(): array {
      * @inheritdoc
      */
     protected function getSupportedPathRegex(string $domain): string {
-        return "`^/?\d+(\?|$)`";
+        return "`^(/[a-z,[0-9]+){1,2}$`";
```
initvector: I _believe_ it has to be an all-numeric first segment and then an alphanumeric second segment. This would mean we shouldn't be allowing "," or "[" for either segment.

How about the following?

```suggestion
        return "`/\d+(/[a-z0-9]+)?$`";
```

alex-mtl: don't we need to escape slashes? and I would keep `^` start line check
`^(\/\d+)(\/[a-z0-9]+)?$`

initvector: > don't we need to escape slashes?

We're using the back tick (`) as the delimiter here, so escaping the slashes is not necessary.

Issue #9082, opened 2019-07-23
KB - Alt Text Dropdown


slafleche:
A few problems we have with the dropdown:

* Page jumps up when you click on image 
* Opens/closes once correctly then flickers
* Would be nice to have a confirmation popup if you close without saving
* useDevice() doesn't work properly. Always returning desktop (can't test the popup style on mobile)

PR #9083, opened 2019-07-23
Alt Text UI for images

slafleche:
This PR adds a menu to images in edit mode to add alt text.

Note that this PR is only for visual UI and not the functional aspect. There are also a few lingering issues that I put in an issue: https://github.com/vanilla/vanilla/issues/9082

To test, edit an existing post with the rich editor (or make a new one) with an image. Click on the image to select it. A menu will appear and you can open it by clicking on the icon. Note that the input is disabled for now.


slafleche:
Closes in favour of merge into feature branch.

PR #9085, opened 2019-07-23
Add hooks to allow modifying the form data during connection.

nervoustwit:
This PR is to add hooks to the EntryController so that we can modify the the form data before saving users either when registering, signing in or connecting over SSO.


nervoustwit:
At present we don't need this. This was originally intended to be used by the Terms Manager that needs to be completely over-hauled at some point.

PR #9111, opened 2019-07-24
Fix file themes not receiving variables in API responses

initvector:
If a theme on the filesystem does not have a variables asset, the themes API will not include variables contributed by addons. The asset will simply be returned as not existing.

This update essentially introduces a persistent variables asset for all themes. If a theme does not actually have a variables asset, it will receive an empty array, which addons can modify. If they do have a variables asset, it will be merged with whatever addons provide.


initvector:
Closing this in favor of #9150

PR #9114, opened 2019-07-24
Improve mention sorting by adding better test cases

charrondev:
As shown in the demo today https://github.com/vanilla/vanilla/pull/9067 didn't 100% accomplish what it set out to do.

This PR builds on it by:

- Improving the test case so that more edge cases where the `old` names would come first in a normal sort.
- Improve the test case by changing the casing of the "exact" match to be of a different case.


alex-mtl:
Your last commit to remove unnecessary `only()` seem broke the build tests

charrondev:
Good catch. I fixed a bit of a logic error in one of the sorting functions.

PR #9110, opened 2019-07-24
Add editor stylesheets from the main page in the iframe

irina-vanillaforums:
@tburry @nervoustwit @dannimajdalani @cchabilall83 

https://app.zenhub.com/workspaces/services-59bffff3fb3fbe0b3572c070/issues/vanillaforums/ea/168
Include other stylesheets from the main page in the iframe for the advanced editor.

PR #9115, opened 2019-07-25
Add current theme url to site meta

slafleche:
This PR adds to the gdn object the current theme's url path. This can be used to import assets from the theme with JS.

### DIFF
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -287,6 +287,10 @@ class Gdn_Controller extends Gdn_Pluggable {
 
         parent::__construct();
         $this->ControllerName = strtolower($this->ClassName);
+
+        // Add current theme path to GDN object
+        $themeInfo = Gdn::themeManager()->getThemeInfo(Gdn::themeManager()->currentTheme());
+        $this->addDefinition('currentThemePath', $themeInfo["Dir"]);
     }
 
     /**
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -290,8 +290,9 @@ class Gdn_Controller extends Gdn_Pluggable {
         $this->ControllerName = strtolower($this->ClassName);
 
         $currentTheme = Gdn::getContainer()->get(\Vanilla\AddonManager::class)->getTheme();
-        $this->addDefinition('currentThemePath', $currentTheme->getSubdir());
-
+        if ($currentTheme) {
+            $this->addDefinition('currentThemePath', $currentTheme->getSubdir());
+        }
     }
 
     /**
--- a/library/core/class.controller.php
+++ b/library/core/class.controller.php
@@ -12,7 +12,6 @@
  * @abstract
  */
 
-use Vanilla\Utility\ContainerUtils as ContainerUtilsAlias;
 use \Vanilla\Web\Asset\LegacyAssetModel;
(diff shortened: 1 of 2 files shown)


tburry:
We originally used editor.css in the editor and that was removed. Did you find the cause for this? Creating another css file without knowing what happened to the first bit of functionality is something we should try and avoid.

charrondev:
That was my primary concern with https://github.com/vanilla/vanilla/pull/9110 as well.

The line is commented out, but why was it removed in the first place?

slafleche:
So, the code seems to have been commented since the first commit in 2015:

https://github.com/vanilla/vanilla/commit/51207e75f5a9c1fe8cdbd028548d1214055179a3#diff-4cdf1a850be5a6b9c69af45f60dbf08b

There IS and editor CSS file, but it lives in the plugin. We don't want to overwrite that, because we'll need to rewrite all the styles. I think a separate sheet is the best option here @charrondev

library/core/class.controller.php
```diff
@@ -287,6 +287,10 @@ public function __construct() {
 
         parent::__construct();
         $this->ControllerName = strtolower($this->ClassName);
+
+        // Add current theme path to GDN object
+        $themeInfo = Gdn::themeManager()->getThemeInfo(Gdn::themeManager()->currentTheme());
```
charrondev: The theme manager is deprecated.

The easiest way to get this now is likely with:

```php
/** Addon $currentTheme */
$currentTheme = Gdn::getContainer()->get(ContainerUtils::currentTheme());
```

library/core/class.controller.php
```diff
@@ -12,6 +12,7 @@
  * @abstract
  */
 
+use Vanilla\Utility\ContainerUtils as ContainerUtilsAlias;
```
initvector: Unused, so I think we can dump it.

```suggestion
```

plugins/editor/js/editor.js
```diff
@@ -47,6 +39,9 @@
             stylesheetsInclude.push(editorWysiwygCSS + '?v=' + editorVersion);
         }
 
+        // If you want to include a custom stylesheet inside the iFrame of the wysiwyg
+        stylesheetsInclude.push(gdn.meta.currentThemePath + '/design/editorWysiwyg.css?v=' + editorVersion);
```
initvector:
```suggestion
        var currentThemePath = gdn.definition("currentThemePath", null);
        if (currentThemePath) {
            stylesheetsInclude.push(currentThemePath + "/design/editorWysiwyg.css?v=" + editorVersion);
        }
```

Even still, isn't this just gonna throw a 404 in the background and cause a pointless request to the server if the theme doesn't have an editorWysiwyg.css?

initvector:
It looks like Advanced Editor [already mostly supported this](https://github.com/vanilla/vanilla/blob/c246ed34c1f20883d4066fe20d45c98aca20c55c/plugins/editor/class.editor.plugin.php#L545) by adding the right file to your theme.

slafleche:
I was afraid overwriting it would remove important styles, but it looks like all the styles in the wysiwyg file also exist in editor.css... so they could of just left the file blank, but they didn't.

I'm going to leave the theme path url because I think it's pretty handy for theming.

PR #9116, opened 2019-07-25
Add re-orienting of images uploaded with media API

dannimajdalani:
Creating a new image via the media API v2 endpoint does not reorient an image based on its EXIF orientation data. This is addressed here by pre-processing new images with `ImageResizer::resize`, which performs several operations on images, including orientation correction.

### Testing
1. Upload an image with orientation EXIF data through Rich Editor. It should be re-oriented.
1. Enable image dimension limits in the dashboard (under "Posting").
1. Upload the same image again, ensuring its dimensions exceed the limits you set in the previous step. It should be resized and reoriented.

If you need an example image, please let me know.

Closes vanilla/support#409


charrondev:
With the recent team re-org I'm taking myself off of this one and assigning Ryan.

slafleche:
This only seems to work for one orientation. I've got examples here for all possible transforms and it only worked for 1 of them. https://github.com/recurser/exif-orientation-examples
 
@initvector

PR #9119, opened 2019-07-25
Reduce JS bundle size through optimized imports

alexei-paj:
reduced the size of the minified bundle by ~30%

- `highlight.js` dynamically imported (goes into it's own chunk and only used when needed).
- `moment.js` locales have been filtered down to the most common ones. There's not a great solution at the moment to dynamically import these. If some other date time format ends up being needed we can add it to the whitelist.
- Added a lint rule mandating single lodash imports.
- Removed the full lodash import in `<EditorContent />`.

Relates to: https://github.com/vanilla/knowledge/issues/695

### DIFF
--- a/build/scripts/configs/makeBaseConfig.ts
+++ b/build/scripts/configs/makeBaseConfig.ts
@@ -124,6 +124,7 @@ ${chalk.green(aliases)}`;
         },
         performance: { hints: false },
         plugins: [
+            new webpack.ContextReplacementPlugin(/moment[/\\]locale$/, /en|es|fr|pt|ch/),
             new webpack.DefinePlugin({
                 __BUILD__SECTION__: JSON.stringify(section),
             }),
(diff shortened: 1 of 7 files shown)


build/scripts/configs/makeBaseConfig.ts
```diff
@@ -124,6 +124,7 @@ ${chalk.green(aliases)}`;
         },
         performance: { hints: false },
         plugins: [
+            new webpack.ContextReplacementPlugin(/moment[/\\]locale$/, /en|es|fr|pt|ch/),
```
alex-mtl: what we are trying to get here? Where this list of locales came from?
I don't see any list in original issue.

plugins/rich-editor/src/scripts/editor/EditorContent.tsx
```diff
@@ -49,6 +49,29 @@ export default function EditorContent(props: IProps) {
     return <div className="richEditor-textWrap" ref={quillMountRef} />;
 }
 
+let hljs: any;
```
alex-mtl: I doubt about variable name. Usually we keep long explicit version of names like: `hightLightJs`

PR #9123, opened 2019-07-26
Adds a hover state on icons in editor for the kb and forum

alexei-paj:
Adds hover state for the icons in the editor.


alex-mtl:
Please always point original issue in PR description section with `Relates to: <link>` or `Closes: <link>`. In this case:

Closes: https://github.com/vanilla/knowledge/issues/1060

PR #9124, opened 2019-07-26
Fix pasting URL in Rich Editor's embed menu

slafleche:
This PR fixes an issue with the rich editor's embed menu. When pasting a valid URL, it did not validate it properly without a change to the text.

I also took the liberty of setting focus inside the menu, since it's a single action dropdown, we can set the focus there.

Closes: https://github.com/vanilla/support/issues/627


plugins/rich-editor/src/scripts/flyouts/EmbedFlyout.tsx
```diff
@@ -143,4 +147,9 @@ export default function EmbedFlyout(props: IProps) {
             </DropDown>
         </>
     );
+
+    function handleVisibilityChange() {
```
charrondev: Can we move this up above the return statement? I know it get's hoisted, but it just feels really weird to have function definitions after a return statement.

PR #9126, opened 2019-07-28
Added explicit top and left positioning to absolute positioned hero image

slafleche:
The hero image is absolutely positioned in Keystone, but was missing explicit top and left values. IE 11's default is weird, which is why it looked broken on the client's site.

For: https://github.com/vanilla/support/issues/629

### DIFF
--- a/themes/keystone/src/scss/components/_herobanner.scss
+++ b/themes/keystone/src/scss/components/_herobanner.scss
@@ -118,6 +118,8 @@
 
 .Herobanner-bgImage {
     position: absolute;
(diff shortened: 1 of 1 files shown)


slafleche:
@igraziatto do I need to bump up a version number? What's the process for getting this out to clients?

igraziatto:
> @igraziatto do I need to bump up a version number? What's the process for getting this out to clients?

This fix is specific to Keystone, so no. If you ever do a fix to the boilerplate you need to bump the version and publish on npm.

Issue #9049, opened 2019-07-15
Embedding a KB article gives a 422 on the scrape endpoint


tburry:
**Describe the bug**

I tried embedding a link to a KB article and got a 422 error.

**To Reproduce**

1. Start a new discussion.
2. Paste this link: https://success.vanillaforums.com/kb/articles/80-2019-005
3. Hit enter to trigger auto embed.

**Expected behavior**

The article should embed properly just like a website.

**Screenshots**

![embed-error](https://user-images.githubusercontent.com/97432/61232145-9fd43f00-a6fb-11e9-96cb-3a6a39c62ae8.gif)

Error response:

```json
{
    "message": "Validation Failed",
    "status": 422,
    "errors": [
        {
            "field": "body",
            "code": "missingField",
            "status": 422,
            "message": "body is required."
        }
    ]
}
```

**Additional context**

I'm not sure if this is widespread or just particular URLs.

charrondev:
I'm pretty sure this was fixed in the embed refactor. Just to double check before actually sprinting the issue.

charrondev:
Confirmed to not be fixed after the embed refactor.

cchabilall83:
The issue here is the because some previously published kb articles don't have excerpts.  The description tag of an article page is built with the excerpt and if that's not set, then an empty string is put in it's place.  Link embeds are built using the page description meta tag.  If the content of that is an empty string, the link embed will fail.  

https://github.com/vanilla/knowledge/blob/e6651148a9018d52338802b954a329e771d9a323/plugins/knowledge/Controllers/Pages/ArticlePage.php#L40

alex-mtl:
But we still need to fix it on scraper side I think. So if description is  not present or content of the page is absolutely "empty" we still need to generate embed frame I guess. @charrondev ?

PR #9147, opened 2019-07-30
Allow body parameter to be optional for link embeds

cchabilall83:
Currently the link embed schema requires the body to parameter.  A link embed's body is built with the content attribute of the page, if that's an empty string it will cause the media scrape to fail and throw a 422 error code.  This relaxes the schema and makes the body parameter optional.

closes vanilla/vanilla#9049


cchabilall83:
> Please update the frontend interface as well.
> 
> https://github.com/vanilla/vanilla/blob/bb771e6480d131c895dfedfeb901d49c8b8bf147/library/src/scripts/embeddedContent/LinkEmbed.tsx#L18

done

Issue #8982, opened 2019-06-19
Create Alt Text UI for images in Rich Editor


slafleche:
Create Dropdown component that gets added in edit mode on images to edit alt text. The dropdown will be in the margin on desktop (to the left) and over top the image on mobile. Note that it will be absolute positioned and should fit in the spacing.

Mockup Desktop:
https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5c92a20540d00405f5b54dea
Mockups Mobile:
https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5c93a78a4e90c024d1dac713
https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5c93a78ad12e87057b33637c

Issue #8981, opened 2019-06-19
Update quill image embed blot to support alt text


charrondev:
The server side will respect the `name` parameter saved in the embed data as the alt text.

The frontend component needs to be updated to do the following:

1. Present action button when focused.
2. Display the react component overlay.
3. Reflect changes from that form into the embed data.

PR #9156, opened 2019-08-02
Implement alt-text input for rich editor

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8981
Closes https://github.com/vanilla/vanilla/issues/8982

Required KB PR: https://github.com/vanilla/knowledge/pull/1094/files

- Implements alt-text UI in the `<ImageEmbed />`
- Implements way for rendered embeds to pass data back to their embed blot. All embeds are now passed `syncBackEmbedValue` and the `quill` instance.
- Implements an `EditorEventWall` that prevents events from propagated into the editor. This allows stuff like our flyout and input classes to be rendered inside of the editor without issue.
- Makes dropdown styles more consistent. We had some type-checking being omitted, but for consistency there are now 2 styles of flyout. `FlyoutType.LIST` and `FlyoutType.FRAME`. LIST types are smaller in size, contain some padding, and are meant to have a list of items. FRAME has no padding is meant for use to use the `<Frame />` component as it's content.

## Image styling

After some discussion with @tburry The current layout with different sized images was agreed upon, but is not fully defined in mockupss


charrondev:
> Enter should save

Done.

> "Insert" is an odd choice of word. Why not "Save"?

I changed it. I'll ask Elisa to update the mockup as well. It's probably just a holdover from the insert embed flyout.

> If I modify the text and don't save, I expect it to go back to the original text, ideally with a confirmation modal in between.

Fixed.

> I thought the margins for the images were supposed to be adjusted so the icon could be int he margin without bumping the content down

The margin wasn't enough, so this margin has been added. With margin collapsing it's generally not noticeable. When demoing to Todd he said to just add the margin here rather than re-adjusting all the margins in the application. In addition on the forum we have no guarantee there is enough margin so we have to place it ourselves. Once again because of margin collapsing this doesn't have too much of an impact. In any case I did go over this with @tburry.

PR #9161, opened 2019-08-07
Add support for images in webp format to AdvancedEditor

paveldedik:
AdvancedEditor doesn't insert images with `webp` extension. The format should be supported.

https://developers.google.com/speed/webp/

### DIFF
--- a/plugins/editor/js/editor.js
+++ b/plugins/editor/js/editor.js
@@ -1244,7 +1244,7 @@
                 // Match image URLs
                 var imageUrl = e.target.value;
                 // Doesn't have to be perfect, because URL can still not exist.
-                var imageUrlRegex = /(?:\/\/)(?:.+)(\/)(.*\.(?:jpe?g|gif|png))(?:\?([^#]*))?(?:#(.*))?/i;
+                var imageUrlRegex = /(?:\/\/)(?:.+)(\/)(.*\.(?:jpe?g|gif|png|webp))(?:\?([^#]*))?(?:#(.*))?/i;
                 if (imageUrlRegex.test(imageUrl)) {
 
                     editorInstance.focus();


charrondev:
Unfortunately there is some additional work that would be required to take this as a core change in Vanilla. Not all browsers support `.webp` with the most notable browser without support being Safari and iOS Safari, which makes up 10-30% of browser usage varying by region. https://caniuse.com/#feat=webp

For us to add this into core, it would require converting `.webp` images into a compatible format and displaying that those to particular browsers.

There have been some team discussions about creating a single endpoint to serve images through that would generate thumbnails/resize images/convert to different formats for use in a  `` but there is a lot a planning and implementation details to work out for that.


## In the meantime

In the meantime updating the regex to instead use the allowed extensions set in the site configuration would allow you to accomplish the same thing. You should find them available through javascript as `window.gdn.meta.allowedImageExtensions`.

paveldedik:
@charrondev Thank you for the fast response. I don't understand what `window.gdn.meta.allowedImageExtensions` is for, in my case, it contains also `tiff` and `bmp`, but I cannot insert images using the Attach Image button. And even if I update it with `webp`, it doesn't help - still not possible to insert `webp` images.

The problem is that I cannot insert images using the Attach Image button, I want to just add the URL of the image to the comment but it obviously doesn't pass some regex validation and inserts nothing (the popup window with the image URL field doesn't even disappear). If I manually write `.jpg` extension to the URL, it inserts the image URL correctly.

charrondev:
@paveldedik So Vanilla has a configuration setting `$Configuration['Garden']['Upload']['AllowedFileExtensions']`. By default it contains numerous file extensions. In order to allow an extension to be uploaded it needs to be allowed from both the backend and the frontend.

The way this is done in rich editor is by the frontend looking at some javascript variable set in the page, which is made up of these values.

Right now advanced editor is not following these configuration settings. I'm suggesting you update your PR to do that, and set your configuration setting to allow `webp` if that is acceptable for your forum.

**Overview**

- We don't want to allow `.webp` by default until we've done additional work converting images for backwards compatibility.
- We have an existing configuration setting `Garden.Upload.AllowedFileExtensions` that allows any site to configure what file extensions they want to support.
- Rich Editor respects this configuration setting. Advanced editor does not.
- The `window.gdn.meta.allowedImageExtensions` global variable will give you access to these configured extensions.
- A robust fix would involve using the values set in that variable to construct the regex.

paveldedik:
@charrondev OK, thank you very much for the explanation, I understand very clearly now. I don't think it has enough priority for us to properly fix it, so I'm not that motivated at this point (and since it works in Rich Editor...). But if I do, I'll open a new pull request. I'm gonna close this one.

Issue #9168, opened 2019-08-08
Dashboard's "Visit Site" icon too big


slafleche:
The dashboard's "visit site" icon is too big on the latest master.

How it was:

<img width="318" alt="Screen Shot 2019-08-08 at 13 35 04" src="https://user-images.githubusercontent.com/19651440/62724610-6385bd00-b9e1-11e9-95cb-3dcb6e998318.png">

How it is:

<img width="269" alt="Screen Shot 2019-08-08 at 13 35 10" src="https://user-images.githubusercontent.com/19651440/62724621-67b1da80-b9e1-11e9-9148-ac353b9dbddb.png">

slafleche:
Looks like multiple icons have been affected.

PR #9169, opened 2019-08-08
Fix icons sizes in dashboard and updated classes on React Icons

slafleche:
Looks like the dashboard styles weren't compiled properly in a previous PR. The icons in the dashboard were too big.

I've removed the sizing for ".icon" which is used by a common styleshsheet for both in the forum, the KB and the dashboard. It shouldn't only be applied to the KB.

Most of the classes for icons are now converted to CSS in JS.

the rest of the diff is from a partial that was removed last sprint, but wasn't compiled. It shouldn't affect anything. InputBlocks are currently only used on the KB and they use CSS in JS styles.

Closes: https://github.com/vanilla/vanilla/issues/9168


charrondev:
Looked fine aside from the busted import. Fix that it and it should be good.

PR #9172, opened 2019-08-09
Refactor additional (mostly time) functions out of Gdn_Format to add test coverage

charrondev:
## Deprecations

**Soft Deprecations**

These methods had too many usages to realistically refactor at the same time as this, so I've only marked them as deprecated in the doc block.

- `Gdn_Format::date()`
- `Gdn_Format::dateFull()`
- `Gdn_Format::defaultFormat()`
- `Gdn_Format::fuzzyTime()`
- `Gdn_Format::seconds()`
- `Gdn_Format::toDate()`
- `Gdn_Format::toTimestamp()`
- `Gdn_Format::toTimezone()`

**Real deprecations**

- `Gdn_Format::clean()`
- `formatDateCustom`
- `Gdn_Format::timespan()`

Using these again in the future will fail tests that use them.

## New classes

The new core code is primarily made up of

- `Vanilla\Formatting\DateConfig`
- `Vanilla\Formatting\DateTimeFormatter`
- `Vanilla\Formatting\TimeUnit`

Most of it is in `DateTimeFormatter`. Tests have been adding bringing `DateTimeFormatter` up to 97% coverage.

`Gdn_Format::fuzzyTime()` and `Gdn_Format::seconds()` (now `DateTimeFormatter::formatRelativeTime()` and `DateTimeFormatter::formatSeconds()`) had some logic errors around when we should return certain string values. Various inconsistencies were resolved while adding the tests.

Additionally some test utilities were modified.

## Test harness

- Updates `MockLocale` to extend `MockConfig` so it can have values set/initialized.
- Updates `MinimalContainerTestCase` to configure the container in `setUp` instead of `setUpBeforeClass`. This prevents container pollution issues when a test needs to run in its own process.

## Test coverage

This improves the test coverage of the `Vanilla\Formatting` namespace

image

`DateTimeFormatter` and `DateConfig` are doing pretty well too. I did not have the time to get this to 100%, but this is certainly and improvement over the 0% it was at before this PR.

image


library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * Convert a datetime to a timestamp.
+     *
+     * @param string $dateTime The Mysql-formatted datetime to convert to a timestamp. Should be in one
+     * of the following formats: YYYY-MM-DD or YYYY-MM-DD HH:MM:SS.
+     * @param mixed $fallback The value to return if the value couldn't be properly converted.
+     * @return int A timestamp or now if it couldn't be parsed properly.
+     */
+    public static function dateTimeToTimeStamp($dateTime = '', $fallback = null) {
```
alex-mtl:
```suggestion
    public static function dateTimeToTimeStamp(string $dateTime = '', $fallback = null): int {
```

library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -0,0 +1,410 @@
...
+     * @see DateTimeFormatter::formatSeconds()
+     *
+     * @param string $datetime The time to convert.
+     * @param int|null $from What time to be relative to.
+     * @return int
+     */
+    public static function dateTimeToSecondsAgo($datetime, $from = null): int {
+        if ($from === null) {
+            $from = time();
+        }
```
alex-mtl: $from = $from ?? time()

library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -0,0 +1,406 @@
...
+    private $dateConfig;
+
+    /**
+     * @param DateConfig $dateConfig
+     */
+    public function __construct(DateConfig $dateConfig) {
+        $this->dateConfig = $dateConfig;
+    }
+
+    const NULL_TIMESTAMP_DEFALT_VALUE = '-';
```
alex-mtl: Could we keep `const`s declaration on the top before class methods starts and before properties  please?

library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -0,0 +1,406 @@
...
+     */
+    public function formatDate($timestamp = '', bool $isHtml = false, string $format = ''): string {
+        // Was a mysqldatetime passed?
+        if ($timestamp !== null && !is_numeric($timestamp)) {
+            $timestamp = self::dateTimeToTimeStamp($timestamp);
+        }
+
+        if ($timestamp === null) {
+            return self::t('Null Da' .
+                'te', self::NULL_TIMESTAMP_DEFALT_VALUE);
```
alex-mtl: what `Null Da te` string means?

library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -0,0 +1,406 @@
...
+        }
+
+        $gmTimestamp = $timestamp;
+        $timestamp = $this->adjustTimeStampForUser($timestamp);
+
+        if ($format == '') {
```
alex-mtl: I prefer either `=== ''` or `empty(...)`

library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -0,0 +1,406 @@
...
+        $result = strftime($format, $timestamp);
+
+        if ($isHtml) {
+            $fullFormat = $this->dateConfig->getDefaultDateTimeFormat();
+            $fullFormat = $this->normalizeFormatForTimeStamp($fullFormat, $timestamp);
+            $result = wrap(
+                $result,
+                'time',
+                [
+                    'title' => strftime($fullFormat, $timestamp),
```
alex-mtl: Are we taking into account that `$fullFormat` is now partially rendered string but not just format pattern? (since `normalizeFormatForTimeStamp()` call)

charrondev: I believe it was. This is essentially a 1:1 translation of what was there before. It looks like it just preformats certain portions. 

Since `strftime()` is so flexible I don't see any issues there. If you can think of a test case I'd be glad to add it though.

library/Vanilla/Formatting/DateTimeFormatter.php
charrondev: It was the existing translation string being used. 

See https://github.com/vanilla/vanilla/pull/9172/files#diff-b4fbc949423c122721c207f9d731c9d3L427

The line break looks it was some formatting mistake so I've fixed that.

library/Vanilla/Formatting/DateTimeFormatter.php
charrondev: Fixed.

library/Vanilla/Formatting/DateTimeFormatter.php
charrondev: Done

Issue #9143, opened 2019-07-30
Rich Editor - Emojis Fail To Render On Windows Computers


tburry:
Many of the emojis for the Rich Editor fail to render on Windows computers (tested with Chrome and IE11, Windows 10 PC):

![image](https://user-images.githubusercontent.com/19805402/60117332-e27fa880-9747-11e9-968d-9af545952995.png)

They render just fine on MacOS.  

May be related to an issue with: https://github.com/vanilla/vanilla/pull/6644

Closes https://github.com/vanilla/support/issues/535

PR #9173, opened 2019-08-09
Improve Emoji support for Windows

slafleche:
I did some tests and moving the "sans-serif" to the end of the font fallbacks gives much better results for emojis on windows.

I also took the opportunity to fix some IE 11 bugs in the emoji picker. Many misaligned button icons and the title was off centered vertically.

I renamed richEditorClasses.ts to richEditorStyles.ts, to match our standards.

Closes: https://github.com/vanilla/vanilla/issues/9143


slafleche:
Good catch, here it is: https://github.com/vanilla/knowledge/pull/1139

PR #9183, opened 2019-08-14
Add route instance caching to RouteHandler

charrondev:
Fixes https://github.com/vanilla/knowledge/issues/1103
Required companion https://github.com/vanilla/knowledge/pull/1146

The issue was caused by the component re-mounting quickly when we updated the previous history entry.

If you look here we replace the current add URL with an editor URL, then push a new entry for the article page.

https://github.com/vanilla/knowledge/blob/94e9b49b8a2624543c20e6842cd64f31869abb61/plugins/knowledge/src/scripts/modules/editor/EditorPageActions.ts#L379-L391

Here's a timeline. Bolded items were happening, but should not have been happening.

1. Publish
2. Replace `/kb/articles/add?...` with `/kb/articles/:id/editor` so that navigating backwards works.
3. __`react-router` notices a different routes component (even if the child `<EditorPage` is the same)__.
4. __`<EditorPage />` un-mounts and reset the page data.__ 
5. __`<EditorPage />` re-mounts. Finds no data, and triggers an AJAX request for the edit article data.__
6. Replace `/kb/articles/:id/editor` with `/kb/articles/:id`.
7. `<EditorPage />` un-mounts and resets the form.
8. `<ArticlePage />` mounts.
9. __Sometime later the AJAX request returns and pollutes the form with data.__

## The fix

- Update React router to `5.0` which now supports having an array of paths for a single route.
- Update `RouteHandler` to support an array of paths.
- https://github.com/vanilla/knowledge/pull/1146 ensures there is only 1 shared editor route, by passing an array of paths and removing `EditorAddRoute`.


alex-mtl:
History back is not working and article EDIT page is broken

charrondev:
Fixed with the latest changes https://github.com/vanilla/knowledge/pull/1146#pullrequestreview-275086729

PR #9184, opened 2019-08-14
Improve user content list styles

charrondev:
After discussing with @ElisaAlmeida I've updated the bullet list styles to be more inline with our mockups.

I also reduced the left margin from 3em to 2em to align with the mockups.

|Before|After|
|---|---|
|image|image|


library/src/scss/userContent/_lists.scss
```diff
@@ -47,7 +47,14 @@ $nestedList-margin: 0 0 0 1em;
 .userContent {
     ul {
         li {
-            list-style: disc !important;
+            list-style: none !important;
+            position: "relative";
+
+            &::before {
+                content: `"•"`,
```
slafleche: Make sure to specify the font to something safe, like "arial" so it doesn't mess up if a bad font is selected.

applications/dashboard/design/style.css
```diff
@@ -6214,15 +6198,9 @@ img.jcrop-preview {
 }
 
 .flex {
-  display: -webkit-flex;
-  display: -ms-flexbox;
   display: flex;
-  -webkit-justify-content: space-between;
```
slafleche: Looks like we lost a lot of auto prefixing? Did our settings change?

library/src/scss/userContent/_lists.scss
charrondev: Done

applications/dashboard/design/style.css
charrondev: Whoops. Fixed.

PR #9186, opened 2019-08-14
Fix notifications rendering with null format

charrondev:
Closes https://github.com/vanilla/support/issues/703
Cleanup from https://github.com/vanilla/vanilla/pull/8980

Prior to https://github.com/vanilla/vanilla/pull/8980 `Gdn_Format::plainText()` and `Gdn_Format::excerpt()` redirected their format through `Gdn_Format::to()` which has explicit null handling.

During the refactoring this was removed causing "No formatter is installed for format" messages if a `null` format was was passed https://github.com/vanilla/support/issues/703

I've added handling of null to these 2 methods, and updated the 1 call-site that was unknowingly passing in null. This both fixes the proper cause of https://github.com/vanilla/support/issues/703 and addresses other potential backwards compatibility issues if there is some other call site doing this.

### DIFF
--- a/applications/dashboard/controllers/class.notificationscontroller.php
+++ b/applications/dashboard/controllers/class.notificationscontroller.php
@@ -8,6 +8,8 @@
  * @since 2.0
  */
 
+use Vanilla\Formatting\Formats\HtmlFormat;
+
 /**
  * Handle /notifications endpoint.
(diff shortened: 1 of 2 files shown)


charrondev:
Sorry guys!

PR #9193, opened 2019-08-15
Add title type to configurationModule & splash image filter

charrondev:
This PR implements 2 to things to unblock https://github.com/vanilla/knowledge/pull/1149

## `ConfigurationModule` section title support

This PR adds the option to the `ConfigurationModule` to allow section titles in between. This allows a long form to be more easily broken up into different pieces.

## Implement `splashVariables` filter variables.

Two new variables have been added to `splashVariables`:

- `outerBackground.useFilter` - Boolean value to enable a scrim overlay on top of the splash. Defaults to off.
- `outerBackground.brightnessFilterPercentage` - Percentage number of brightness to adjust the image by if `useFilter` is enabled. Defaults to 70% with a bright contrast FG color and `130%` with a dark contast foreground color. (Eg. Darkens the image if we overlay light text, and brightens the image if we overlay dark text).

## Implement titleBar border variables

After discussing with Elisa some of the header variables a variable was needed to allow a header border/shadow. I've added the variable `titleBar.border.type` which can be `none|border|shadow`. A dropdown for selecting the style is provided in https://github.com/vanilla/knowledge/pull/1149


library/src/scripts/splash/splashStyles.ts
```diff
@@ -60,12 +60,15 @@ export const splashVariables = useThemeCache(() => {
...
 
-    const outerBackground: IBackground = makeThemeVars("outerBackground", {
+    const isContrastLight = colors.contrast instanceof ColorHelper && colors.contrast.lightness() >= 0.5;
+    const outerBackground = makeThemeVars("outerBackground", {
         color: colors.primary,
         backgroundPosition: "50% 50%",
         backgroundSize: "cover",
         image: assetUrl("/resources/design/fallbackSplashBackground.svg"),
         fallbackImage: assetUrl("/resources/design/fallbackSplashBackground.svg"),
+        useFilter: false,
```
slafleche: I wouldn't use the word "filder" here because it makes me think of CSS filters: https://developer.mozilla.org/en-US/docs/Web/CSS/filter

I think "overlay" or "colorOverlay" is easier to understand.

charrondev: Implemented. The variables are now `useOverlay` and `overlayColor`. I've also made the overlay its own element instead of a filter.

applications/dashboard/design/admin.css
```diff
@@ -6706,6 +6706,12 @@ label,
   border: 0.0625rem solid #ccc;
 }
 
+.form-sectionTitle {
```
slafleche: I don't think this is correct. There's "Subheading" and "Subheading Block" available in the styleguide:

`/vanilla/applications/dashboard/styleguide/section-1.html`

I think it would be proper to have the "Global Site Colors" as a block and the "Header Variables" as a block.

slafleche:
This wasn't part of the sprint. This should of been discussed with myself and @ElisaAlmeida  before rushing it to PR. We should of discussed success criteria as a team for this.

applications/dashboard/design/admin.css
charrondev: Because this is a `ConfigurationModule` I can't make the sections themselves blocks, but I can remove my custom CSS and re-use `subheading`

charrondev:
## IE11

I don't see any padding issue with the overlay here:

image

## Edge 16

I also don't see any issues here

image

charrondev:
After a bit more digging I found the some CSS classes are being applied to the titlebar at all in IE. I've opened an issue here https://github.com/vanilla/vanilla/issues/9213

PR #9201, opened 2019-08-19
Set the error reporting to include E_USER_WARNING

tburry:
I want to be able to trigger warnings for some things that we plan to change in the future, but are uncertain of the production implications.

This change will send E_WARNING and E_USER_WARNING errors to the log rather than discarding them.

**Note: I added E_WARNING to the list of warnings we report because it seems like the right thing to do, but that may end up being too noisy. I feel like I remember some really dumb things escalating to E_WARNING in PHP, but I just can't remember what.**

### DIFF
--- a/index.php
+++ b/index.php
@@ -9,7 +9,7 @@
  */
 
 // Report and track all errors.
-error_reporting(E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR);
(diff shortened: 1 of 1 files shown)


charrondev:
If we're doing this we need to take an aggressive stance to cleaning up deprecated function calls in our code base. Currently we have enough that we added a separate `\Vanilla\Utility\Deprecate::log` that does does write the deprecations to the log (and the error log).

tburry:
Are the two necessarily related?

I want a warning so that we can model runtime behavior in production. This is something we can't easily do another way.

Deprecated function use should be able to be modeled with static analysis. Furthermore, the use of a deprecated function doesn't necessarily impact our product. It's usually just a function we no longer want to use. My guess the fact that we haven't made a huge effort to remove deprecated functions in our core repos means it's not actually as high a priority to the team.

PR #9204, opened 2019-08-20
Fix plain text emails rendering with HTML tags

initvector:
Vanilla's plain text emails have been rendering with HTML tags since the recent formatting service improvements were introduced. `EmailTemplate` has been updated to generate HTML that is then converted to plain text using the new formatting service before. This ensures the message is rendering consistently without HTML, as expected.

### Recommended Testing

1. Visit the email settings page of the Vanilla dashboard. Ensure HTML email is **off**.
1. Send a test email.
1. Verify the email you receive is a plain text message with no visible HTML tags. The formatting of the message should also separate all relevant portions onto their own newlines (e.g. the title and the message should be separated by at least one new line).

Closes vanilla/support#719

### DIFF
--- a/applications/dashboard/library/class.emailtemplate.php
+++ b/applications/dashboard/library/class.emailtemplate.php
@@ -4,6 +4,7 @@
  * @license GPL-2.0-only
  */
 
+use Vanilla\Formatting\Formats\WysiwygFormat;
 use Vanilla\Web\Asset\LegacyAssetModel;
 
 /**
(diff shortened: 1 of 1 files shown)


dannimajdalani:
Tested on 10012

**Before the fix**

Screen Shot 2019-08-20 at 2 47 15 PM

**After the fix**

Screen Shot 2019-08-20 at 2 48 00 PM

Issue #7702, opened 2018-08-30
Recursive class lookups from the root of an addon do not work


charrondev:
Due to `''` not being an actual path, recursive lookups were not happen so classes that are not in the explicitly defined folders are not working. This means a class place in say a top level `Utils` folder will not be loaded. I'm mainly looking for acceptance criteria of a fix for this.

### Remove the root lookup

If we decide we only want to classes to be in specific subfolders than we should removed the first `''` entry because it doesn't match anything and gives the _impression_ that you can have classes in the root.

### Allow lookups from the root

If we want to allow classes to be looked up from the root the primary reason would likely be to allow cleaner namespacing:

Eg:

```
Vanilla\Addons\SomeAddon\Utils -> /vanilla/plugins/some-addon/Utils

// instead of

Vanilla\Addons\SomeAddon\Utils -> /vanilla/plugins/some-addon/library/Utils
```

If we do this though we may need to add exclusions for these lookups. We don't want to be looking in folders such as `node_modules`, `bower_components`, or `vendors`.

### Start all namespaces in a single directory

The biggest reason why a library folder feels a little bit weird is that your structure ends up like the following:

```
/vanilla/plugins/some-addon/library/Utils/MyUtils.php -> Vanilla\Addons\SomeAddon\Utils\MyUtils
/vanilla/plugins/some-addon/controllers/MyController.php -> Vanilla\Addons\SomeAddon\Controllers\MyController
/vanilla/plugins/some-addon/models/MyModel.php -> Vanilla\Addons\SomeAddon\Models\MyModel
```

The primary issue here is that we have things of the same namespace with different gaps based on the pre-whitelisted folder. Instead something like:

```
/vanilla/plugins/some-addon/library/Utils/MyUtils.php -> Vanilla\Addons\SomeAddon\Utils\MyUtils
/vanilla/plugins/some-addon/library/Controllers/MyController.php -> Vanilla\Addons\SomeAddon\Controllers\MyController
/vanilla/plugins/some-addon/library/Models/MyModel.php -> Vanilla\Addons\SomeAddon\Models\MyModel
```

Would map more consistently.

tburry:
We need to make classes in the root work. I'm reluctant to search every subfolder, but perhaps we can search all subfolders that:

1. Match our current whitelist or:
2. Begin with a capital letter to match our namespace convention.

There are unit tests around all of this so adding new functionality should be fairly straightforward.

charrondev:
[...]

PR #9209, opened 2019-08-20
Allow custom namespaces from the root of an addon

charrondev:
Closes https://github.com/vanilla/vanilla/issues/7702

- Implements namespaced PHP classes from the root of an addon if they are in an Uppercase subdirectory from the root.
- Refactors `AddonManagerTest::testClassDirectoryRecursion()` to use a data provider.
- Adds 2 cases to the new data provider for arbitrary classes to be loaded.


library/Vanilla/Addon.php
```diff
@@ -531,13 +530,17 @@ private function scanDirPhp($dir) {
...
 
         // Get all of the php files from subdirectories.
         foreach ($this->glob("$dir/*", true) as $subdir) {
+            // Don't recursively scan the root of an addon unless it is uppercase.
+            // This means it may be part of a namespace.
+            $trimmedDir = ltrim($subdir, '/\\');
+            $isUppercaseDirName = strlen($trimmedDir) > 0 && ctype_upper($trimmedDir[0]);
+            if ($isDirRoot && !$isUppercaseDirName) {
+                continue;
+            }
```
tburry: I'm okay approving your change if you don't want to do this, but it seems like the logic you are going with would be better put in `scanClassPaths()` to make it more complete. Like it can return all standard lowercase directories plus uppercase directories which would make it more complete and logical IMO.

charrondev: I agree actually. That would probably be a better solution. I'll get around to today

PR #9217, opened 2019-08-22
Add authorization token support to /utility/update

tburry:
Sometimes it is difficult to run utility/update while signed in. However, running updates without a session is also difficult because a lot of addons need to hydrate data which includes `insertUserID` fields.

This change introduces the concept of an update token that an admin can use to run an update as system if they cannot be signed in during an update. It also removes flood control checks because the update token or `Garden.Settings.Mange` is required for updates.

Note that this change puts the functionality behind the `updateTokens` feature flag. In order to test it you'll need to set that feature in your config.


tburry:
This is a fairly significant amount of work to review so it probably needs some QA in addition to review.

My intention is that this PR is being handed off and will be checked and any necessary fixes will be committed by one of you two.

initvector:
This is ready for review per @tburry so I'm taking off the WIP label.

applications/dashboard/models/class.updatemodel.php
```diff
@@ -733,4 +755,40 @@ public function runStructure($captureOnly = false) {
...
+    }
+
+    /**
+     * Save an update status to the config to aid support.
+     *
+     * @param string $status One of the `STATUS_*` constants.
+     * @param string|null $message A message to go along with the status
+     */
+    private function saveStatus(string $status, string $message = null) {
+        $now = new \DateTimeImmutable('now', new \DateTimeZone('America/Toronto'));
```
initvector: > America/Toronto

🤔

applications/dashboard/views/utility/update-legacy.php
```diff
@@ -0,0 +1,28 @@
+<?php if (!defined('APPLICATION')) exit(); ?>
```
initvector: This file is a good candidate for a Twigging.

applications/dashboard/views/utility/update.php
```diff
@@ -1,28 +1,39 @@
-<?php if (!defined('APPLICATION')) exit(); ?>
```
initvector: Should probably Twig this one too.

index.php
```diff
@@ -9,7 +9,7 @@
  */
 
 // Report and track all errors.
-error_reporting(E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_WARNING | E_USER_WARNING | E_RECOVERABLE_ERROR);
```
initvector: I believe this is unintentional.

library/core/class.session.php
```diff
@@ -690,7 +690,7 @@ public function validateTransientKey($foreignKey, $validateUser = true) {
         if (!isset($return)) {
             // Checking the postback here is a kludge, but is absolutely necessary until we can test the ValidatePostBack more.
             $return = ($forceValid && Gdn::request()->isPostBack()) ||
-                (hash_equals($this->_TransientKey, $foreignKey) && !empty($this->_TransientKey));
+                (is_string($foreignKey) && hash_equals((string)$this->_TransientKey, $foreignKey) && !empty($this->_TransientKey));
```
initvector: I'm assuming some of this was done to deal with a notice or warning.

```suggestion
                (!empty($this->_TransientKey) && is_string($foreignKey) && hash_equals((string)$this->_TransientKey, $foreignKey));
```

applications/dashboard/controllers/class.utilitycontroller.php
```diff
@@ -653,4 +641,61 @@ public function showTouchIcon() {
...
+            preg_match('`Bearer\s+(.+)`i', $this->Request->getHeader('Authorization'), $m)) {
+            $token = $m[1];
+        } elseif ($this->Request->post('updateToken', '')) {
+            $token = $this->Request->post('updateToken');
+        }
+
+        $isTokenUpdate = false;
+        if (!empty($token)) {
+            $knownString = (string)Gdn::config()->get('Garden.UpdateToken', '');
+            if (!empty($knownString) && hash_equals($knownString, $token)) {
```
initvector: Nice job thwarting the timing attack!

initvector:
I've gone in and made some fixes to the code. It looks good to me, now. I've done some basic local testing and it works as expected. I'll followup by testing on infrastructure and make my final approval if all goes well over there.

initvector:
### Recommended Testing
1. Make sure `Feature.updateTokens.Enabled` is set to `false` in your site config.
1. Change the size of a field in your database, like increasing the length of `Comment.Format` from 20 to 25.
1. Attempt to run a utility-update as a guest.
1. Review the size of `Comment.Format`. It should be resized back to 20.
1. Set `Feature.updateTokens.Enabled` to `true` in your site config.
1. Resize a database field, as before.
1. Attempt to run a utility-update as a guest without providing any token when prompted. This action should fail.
1. Review the size of `Comment.Format`. It should still be at 25.
1. Get the update token from your site config (`Garden.UpdateToken`). If you do not have one in your config, create one. It can be any random alphanumeric string, 40 characters long.
1. Attempt to run a utility-update as a guest, using the update token obtained from the previous step.
1. Review the size of `Comment.Format`. It should be resized back to 20.
1. Resize the database field for a third time.
1. Sign into the site as a user with `Garden.Settings.Manage` permission.
1. Attempt to run a utility-update as the signed-in user, without providing a token.
1. Review the size of `Comment.Format`. It should be resized back to 20.

PR #9229, opened 2019-08-28
Add navigation to storybook

slafleche:
Added the navigation component to storybook


slafleche:
@charrondev Moved to another file. I leave the cleanup to the end if I have time, but I've got more important tasks to finish this sprint.

PR #9242, opened 2019-09-03
Add class to prevent self scraping and apply to the embed service & page scraper

charrondev:
Prepares for https://github.com/vanilla/internal/pull/1956
Fixes https://github.com/vanilla/dev-inter-ops/issues/23

**Breaking the infinite loop**

I've left plenty of inline comments here I think, but the _TL;DR_ is if the embed service was used in a get request (or the page scraper), and someone pointed the scraper to that page, then the site would take itself out pretty quickly in an infinite loop.

I've created a utility to check the request type and throw an exception and applied that to the page scraper and embed service.

Note this breaks the reporting plugin which was recently updating to use the `EmbedService` in a get request (on accident).

I've created an update to the reporting plugin as well to fix this.

**Utility**

I've also added a utility for normalizing user fragments (from outside of the API controller). This is useful for when an embed must be constructed manually to avoid this loop.

### DIFF
--- a/library/Vanilla/Web/RequestValidator.php
+++ b/library/Vanilla/Web/RequestValidator.php
@@ -31,14 +31,15 @@ class RequestValidator {
      * Ensure the request is not of a particular type.
      *
      * @param string $requestMethod The method of the request, Eg. POST, PATCH, GET. See Gdn_Request constants.
-     * @param string|null $message The message to print in the exception.
+     * @param string $extraMessage Additional message content to add to the exception.
      *
      * @throws ClientException If the request method doesn't match.
      *
      */
-    public function blockRequestType(string $requestMethod, ?string $message = null) {
-        if ($message === null) {
-            $message = "Request method $requestMethod is not allowed.";
+    public function blockRequestType(string $requestMethod, string $extraMessage = "") {
+        $message = "Request method $requestMethod is not allowed.";
+        if ($extraMessage !== "") {
+            $message .= '\n' . $extraMessage;
         }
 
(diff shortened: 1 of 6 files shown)


library/Vanilla/Web/RequestValidator.php
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param string $requestMethod The method of the request, Eg. POST, PATCH, GET. See Gdn_Request constants.
+     * @param string|null $message The message to print in the exception.
+     *
+     * @throws ClientException If the request method doesn't match.
+     *
+     */
+    public function blockRequestType(string $requestMethod, ?string $message = null) {
+        if ($message === null) {
+            $message = "Request method $requestMethod is not allowed.";
```
alex-mtl: Why not just set this as default value?

library/Vanilla/Web/RequestValidator.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Ensure the request is not of a particular type.
+     *
+     * @param string $requestMethod The method of the request, Eg. POST, PATCH, GET. See Gdn_Request constants.
+     * @param string|null $message The message to print in the exception.
+     *
+     * @throws ClientException If the request method doesn't match.
+     *
+     */
+    public function blockRequestType(string $requestMethod, ?string $message = null) {
```
alex-mtl:
```suggestion
    public function blockRequestType(string $requestMethod, string $message = '') {
```

Issue #9241, opened 2019-09-01
Missing HTML encoding in MeBox Inbox


R-J:
This issue has been observed on the current open.vanillaforums.com installation. I've received a message with an ampersand in it and this is how it looks like:
![encoding](https://user-images.githubusercontent.com/3996187/64079257-c4a16780-cce5-11e9-8c3c-0c56365674c0.jpg)

To recreate simply write a message with a "&" in it to yourself and look in the MeBox drop down

charrondev:
PR out now: https://github.com/vanilla/vanilla/pull/9243

If you notice any more of these please report them and we'll fix them quickly. Some recent changes in the formatting pipeline for security purposes might mean we are double escaping things in certain places. They are easy to fix when noticed, but need to be individually verified so we don't accidentally introduce XSS.

I wish more of these views were using twig or react which are automatically escaped. I may do a couple PRs to start converting these views.

PR #9243, opened 2019-09-03
Fix double HTML encoding in messages

charrondev:
Fixes https://github.com/vanilla/vanilla/issues/9241

We were double encoding in some of these views.

### DIFF
--- a/applications/conversations/views/messages/conversations.php
+++ b/applications/conversations/views/messages/conversations.php
@@ -32,7 +32,7 @@ foreach ($this->data('Conversations') as $Conversation) {
     $CssClass .= ' '.($Conversation->CountNewMessages <= 0 ? 'Read' : 'Unread');
 
     $JumpToItem = $Conversation->CountMessages - $Conversation->CountNewMessages;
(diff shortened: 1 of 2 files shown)


applications/conversations/views/messages/conversations.php
```diff
@@ -32,7 +32,7 @@
     $CssClass .= ' '.($Conversation->CountNewMessages <= 0 ? 'Read' : 'Unread');
 
     $JumpToItem = $Conversation->CountMessages - $Conversation->CountNewMessages;
-    $Message = (sliceString(Gdn_Format::plainText($Conversation->LastBody, $Conversation->LastFormat), 100));
+    $Message = (sliceString(Gdn::formatService()->renderExcerpt($Conversation->LastBody, $Conversation->LastFormat), 100));
```
charrondev: First `htmlspecialchars` was part of `Gdn_Format::plainText`.

Second one is here on the output line: https://github.com/vanilla/vanilla/blob/97472338dc7ca97e044eb2fea7c068eb318fc74c/applications/conversations/views/messages/conversations.php#L69

`renderPlainText()` from the format service does not apply this html escaping.

PR #9244, opened 2019-09-04
Update Vanillicon email hashes to truncated SHA1

initvector:
The Vanillicon addon uses MD5 hashes of user emails to generate avatars with the [Vanillicon service](http://vanillicon.com). This has been updated to use the SHA1 algorithm for email hashing. In addition, email hashes will be truncated from forty characters, down to thirty-two (the length of the current MD5 hashes). These two steps, together, should provide  increased protection of user information. SHA1 is a more secure algorithm and truncating its length by twenty percent ensures it's dramatically more difficult to reverse.

Closes vanilla/vanilla-patches#281


tburry:
Can you give the CSMs a heads up on this one? It should also be noted in the release notes which may be sufficient for notifying affected customers.

This might be a good thing to tell the OSS community too.

plugins/vanillicon/class.vanillicon.plugin.php
```diff
@@ -114,7 +117,7 @@ function userPhotoDefaultUrl($user, $options = []) {
         if (!$email) {
             $email = val('UserID', $user, 100);
         }
-        $hash = md5($email);
+        $hash = substr(sha1($email), 0, VanilliconPlugin::TRUNCATED_HASH_LENGTH);
```
tburry: Can you make this change in the Gravatar plugin too?

initvector: This is all set. I also hoped to update the version used to Vanillicon v2, but Gravatar doesn't play nicely with SVGs.

PR #9246, opened 2019-09-04
Prepare core for new subcommunity chooser

charrondev:
Implemented for https://github.com/vanilla/multisite/pull/241  and https://github.com/vanilla/multisite/issues/226

## New

- Add stub to keystone and boilerplate for the subcommunity chooser.
  Example:
```html
<div data-react="subcommunity-chooser"></div>
```
- Allow using the a small dropdown without requiring the `LIST` style.
- Allow adding extra items to `TitleBar` and `TitleBarNav` from custom plugins.
- Allow getting a ref to a `<DropDownItemButton />`.
- Make `name` optional in various dropdown items (you can just pass a ReactNode child instead).
- Add a new `<FrameHeaderMinimal />` variation.

## Fixes

- Fixes a dropdown focus stealing losing bug (adding back `event.stopImmediatePropagation` and juggled a couple other things).


slafleche:
I was unable to add that component elsewhere in Keystone's default master.

charrondev:
> I was unable to add that component elsewhere in Keystone's default master.

It's there if you drop it on the body. We don't exactly have automatic contrast detection, so that would be up to whoever integrates it to handle.

image

PR #9247, opened 2019-09-04
Add icons to the categoriesModule and hide them by default

irina-vanillaforums:
Many clients requested categories tiles on their homepage. The CategoriesModule doesn't have the icons so we used to hardcode those icons.
See examples here: 
1. https://app.zeplin.io/project/5cd9e211e615ba34728f6856/screen/5cd9e228f8edf03befd161a2
https://drive.google.com/drive/u/0/folders/11Mp4aT7lpBYyaqNwBXe1Q6Ve6rbPzpo7
https://smartsheet-stage.vanillastaging.com/
2. https://drive.google.com/drive/u/0/folders/1kIr54KSi8jmVthMgG2xEYYvS-YQaRVdS

I added them directly in the CategoriesModule and hide them by default so they can be used only if needed.


applications/vanilla/views/modules/categories.php
```diff
@@ -44,6 +44,12 @@
                     $CountText = '';
                 }
 
+                if( $Category->PhotoUrl !== "" ){
+                    echo '<div class="Item-Icon PhotoWrap PhotoWrap-Category">';
+                    echo '<img src="' .$Category->PhotoUrl. '" class="CategoryPhoto" alt="' .$Category->Name. '">';
```
charrondev: This would be an XSS. Escape the name or use twig which will do it automatically.

irina-vanillaforums: It's a big scoop to convert the file to twig

PR #9250, opened 2019-09-04
Fix emoji suggestions not displaying

initvector:
Emoji suggestions are currently broken. Typing a colon should trigger the emoji suggestion menu in a default install of Vanilla. This is not happening due to [recent revisions to Vanilla's formatting](https://github.com/vanilla/vanilla/pull/8980). Vanilla started instantiating the `Emoji` class earlier for DI. The `Emoji` class had code in it to add its config to the controller's definitions whenever it was instantiated. Since we started instantiating it earlier, there is no controller available at the time.

Having the controller definitions added at the time the class is instantiated is weird. Adding the config to the controller has been moved to `base_render_before` in the Dashboard hooks file. This should be in a place where it's always available and only added when needed.

### Recommended Testing
1. Navigate to the new discussion form.
1. Enter a single colon (:).
1. Observe the emoji suggestions menu is displayed.
1. Type characters to update the suggestions.
1. Click a suggestion to insert it in the content and dismiss the menu.

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -28,19 +28,36 @@ class DashboardHooks extends Gdn_Plugin {
     /** @var \Vanilla\AddonManager */
     private $addonManager;
 
+    /** @var Emoji */
+    private $emoji;
+
     /**
      * Constructor for DI.
      *
      * @param \Vanilla\AddonManager $addonManager
      * @param Contracts\ConfigurationInterface $config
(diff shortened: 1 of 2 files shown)


dannimajdalani:
Drop-down is working, however, emoji gets posted in the middle of the screen.
Screen Shot 2019-09-04 at 4 17 52 PM

initvector:
Unable to get a second review in time and I want to get this backported. Gonna flex and merge without a second approval.

PR #9251, opened 2019-09-04
Added more dropdown examples to storybook

slafleche:
Added some more dropdown examples to storybook. Unfortunately, I couldn't get the actual components to work because they do multiple API calls to get their data and don't accept props only. This is the best I can do with the time I have.


library/src/scripts/embeddedContent/StoryExampleDropDownMessages.tsx
```diff
@@ -0,0 +1,366 @@
...
+    open: boolean;
+}
+
+const messagesData = [
+    {
+        authors: [
+            {
+                userID: 25,
+                name: "Joe With Spaces",
+                photoUrl: "https://dev.vanilla.localhost/applications/dashboard/design/images/defaulticon.png",
```
initvector: Is there any way to make these URLs relative to Storybook and not reliant on Vanilla also running alongside it?

initvector: Most other images in Storybook seem to be linking to an external CDN, which is probably better than making them relative. My primary concern here is the additional requirement of a web server running on the same host as Storybook to serve up these requests to `dev.vanilla.localhost`. If we need to use relative URLs as an interim solution or something, I'm cool with that.

Issue #9245, opened 2019-09-04
Fix FOUC introduced with new style system


charrondev:
**Describe the bug**
There is a FOUC when loading the new style system starting with https://github.com/vanilla/vanilla/pull/9240.

**To Reproduce**
1. Make a discussion with a spoiler.
2. Open that page in a new tab.
3. Observe the flash of unstyled content.

**Expected behavior**
Nothing should be shown until all styles are applied.

A simple white overlay until everything is done would suffice.

PR #9257, opened 2019-09-05
Fixed FOUC on userContent, forum side

slafleche:
Since the styles for the userContent are in JS and the scripts are deferred, we had some FOUC on the forum side. In the legacy styles, I've added "visibility: hidden" for userContent and set back the visibility in the JS styles. This will not affect the KB, as it doesn't have the legacy styles.

Also fixed position of spoiler icon, it was a off centered in the rich editor menu.

Fixed loader zIndex for forum. It was going under the comments in some themes.

Closes: https://github.com/vanilla/vanilla/issues/9245


charrondev:
A few notes:

I don’t think `visibility: hidden` on just the `userContent` will be sufficient for a few reasons.

- We still will have document reflow appearing when the style come in. Items outside the user content may jump up, when say the spoiler content styles get applied.
- We will likely have more and more stuff like this coming in the future (especially with the theme system your working on this sprint being applied to the forum), so a user content specific fix would just leave us aback he

I think we’ll need to do the whole document body rather than just a subsection

slafleche:
How's this @charrondev ?

PR #9266, opened 2019-09-06
Remove E_WARNING from error reporting

tburry:
TOO LOUD!!!

### DIFF
--- a/index.php
+++ b/index.php
@@ -9,7 +9,7 @@
  */
 
 // Report and track all errors.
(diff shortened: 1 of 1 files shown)


charrondev:
We should probably get this in a backport to OSS when we do the backports. I'm sure they're getting swamped too.

charrondev:
Worst throwing out this though:

https://wiki.php.net/rfc/engine_warnings

I have a feeling at least some current warnings are going to become exceptions in PHP 8. That's a ways away, but we should really be on top of this.

initvector:
> We should probably get this in a backport to OSS when we do the backports. I'm sure they're getting swamped too.

I'm actually thinking of forking 3.2.1 when we do the next infrastructure release, next week. There are a few bug fixes that need to go in there, in addition to this.

PR #9268, opened 2019-09-07
Add the PrivateCommunityMiddleware

tburry:
This will allow the APIv2 to sensibly be called with private communities enabled.

APIv2 endpoints will fail all permission checks when there is a private community and no user is signed in. Endpoints can opt out of the ban using an opt-out permission.

This is a fairly large PR so I'll try and lay out what it includes.

1. I added a bypass of the default private community block for `/api/v2`.
2. I added the `PrivateCommunityMiddleware` to handle private communities in the API.
3. I added some unit tests for various private community functionality. To help test I added bypasses to some logical endpoints: `/users/register` and `/users/request-password`. These endpoints were technically bugged before so this also counts as a fix.
4. The unit tests uncovered some idempotence issues in `UsersTest` which I fixed. This lead to adding some nice convenience methods in the `AbstractAPIv2Test` to run tests in private community mode and with other scenarios.
5. I added `assertNoSession()` to the base class so you'll notice it being removed from some subclasses.
6. I added a basic private community test for discussions to ensure that private communities throw exceptions on endpoints without a bypass.


shaunamcclemens:
psssttt @initvector sounds like this is still waiting for your review, but this is also blocking 
https://github.com/vanilla/internal/pull/1960

Would you happen to have an ETA for this that I can share with the customer?

armstrongkatie:
@tburry @initvector Client is wondering what the eta on this is. Any chance we could get this done sooner rather than later as it's a service they are paying for?

tburry:
This has been assigned to this sprint and @initvector is going to see the PR through.

Issue #8842, opened 2019-05-22
Create core interfaces for subcommunities


charrondev:
**Describe the Feature**

Over time we've had numerous issues with extending extending sub-communities in the past. We should create some core interface for segmenting communities into multiple pieces and break apart and be using that interface wherever possible.

**User Story**

As a developer I do **not** want to ever interact with the sub-community model directly. I want to interact with some core interface that works whether or not the plugin is enabled or not.

**Acceptance Criteria**

This interface should be able to provide

- All the different site sections.
- Each section should specify a URL slug a locale and an ID.*Mockups**
- A core implementation with a default locale and empty slug should be specified.

PR #9272, opened 2019-09-09
Add SiteSection interfaces and default implementation

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8842

- Adds core interfaces for site section.
- Adds default implementation.
- Adds tests for default implementation.


tests/Library/Vanilla/Site/SingleSiteSectionProvider.php
```diff
@@ -0,0 +1,76 @@
...
+use Vanilla\Site\DefaultSiteSection;
+use Vanilla\Site\SingleSiteSectionProvider;
+use VanillaTests\MinimalContainerTestCase;
+
+/**
+ * Tests for SingleSiteSectionProvider.
+ */
+class SingleSiteSectionProviderTests extends MinimalContainerTestCase {
+
+    const LOCALE_KEY = 'AB-df';
```
alex-mtl: What does `AB-df` mean? I expect either `en_US` or `en-us`.
We should try to keep test data to contain pretty explicit and valid values.
If you want it to be different from default one we can use `fr` or `es` variants

charrondev: I had chosen this as it's a value that will never be set on any actual system ever. Would you prefer something like `fake-locale`?

charrondev: Adjusted to be `en_US`

PR #9276, opened 2019-09-10
Create tests to ensure that arrayTranslate() function works.

nervoustwit:
Add a test to make sure that arrays translate.

### DIFF
--- a/tests/Library/Core/ArrayTranslateTest.php
+++ b/tests/Library/Core/ArrayTranslateTest.php
@@ -39,7 +39,8 @@ class ArrayTranslateTest extends TestCase {
                 ['badKey' => 'a.b'],
                 ['badKey' => 'fixedKey'],
                 1,
-                ['fixedKey' => 'a.b']],
+                ['fixedKey' => 'a.b']
+            ],
             'longAssocAdd' => [
                 ['badKey' => 'a.b', 'anotherIndex' => 'b.c'],
                 ['badKey' => 'fixedKey'],
--- a/tests/Library/Core/ArrayTranslateTest.php
+++ b/tests/Library/Core/ArrayTranslateTest.php
@@ -2,17 +2,16 @@
 /**
  * @author Patrick Kelly <patrick.k@vanillaforums.com>
  * @copyright 2009-2019 Vanilla Forums Inc.
- * @license Proprietary
+ * @license GPL-2.0-only
  */
 
 namespace VanillaTests\Library\Core;
-
-use VanillaTests\SharedBootstrapTestCase;
+use PHPUnit\Framework\TestCase;
 
 /**
  * Tests for arrayTranslate().
  */
-class ArrayTranslateTest extends SharedBootstrapTestCase {
+class ArrayTranslateTest extends TestCase {
 
     /**
      * Test  {@link arrayTranslate()} against several scenarios.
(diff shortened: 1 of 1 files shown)


tests/Library/Core/ArrayTranslateTest.php
```diff
@@ -0,0 +1,46 @@
+<?php
+/**
+ * @author Patrick Kelly <patrick.k@vanillaforums.com>
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license Proprietary
```
charrondev: We've got the wrong license here.

```suggestion
 * @license GPL-2.0-only
```

tests/Library/Core/ArrayTranslateTest.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace VanillaTests\Library\Core;
+
+use VanillaTests\SharedBootstrapTestCase;
+
+/**
+ * Tests for arrayTranslate().
+ */
+class ArrayTranslateTest extends SharedBootstrapTestCase {
```
charrondev: Since this is a library test w/ no container usage, or addon startup, I'd recommend using the PHPunit base test case.

`SharedBootstrapTestCase` brings along a lot of baggage with it (starting session, db connection, sql driver, addon autoloader, etc).

You should be able to just use `PHPUnit\Framework\TestCase`. for cases where you still need the container, our `MinimalContainerTestCase` would be appropriate too. I'll be overviewing this Friday during my lunch & learn in more detail.

tests/Library/Core/ArrayTranslateTest.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Provide test data for {@link arrayTranslate()}.
+     *
+     * @return array Returns an array of test data.
+     */
+    public function provideTranslateableArrays() {
+        $r = [
+            'simpleAssocAdd' => [['badKey' => 'a.b'], ['badKey' => 'fixedKey'], [0 => 1], ['fixedKey' => 'a.b']],
```
charrondev: If possible, could you break up the the parameters into separate lines? I find it can make parsing these test cases a lot easier to read.

Some of these lines are particularly long.

tests/Library/Core/ArrayTranslateTest.php
```diff
@@ -0,0 +1,64 @@
...
+     * Provide test data for {@link arrayTranslate()}.
+     *
+     * @return array Returns an array of test data.
+     */
+    public function provideTranslateableArrays() {
+        $r = [
+            'simpleAssocAdd' => [
+                ['badKey' => 'a.b'],
+                ['badKey' => 'fixedKey'],
+                [0 => 1],
```
tburry: You can just make these booleans. The data provider doesn't have to have arrays for every argument. It just passes whatever you specify to the test using `call_func_array`.

tests/Library/Core/ArrayTranslateTest.php
```diff
@@ -0,0 +1,64 @@
...
+                ['fixedKey' => 'a.b']],
```
tburry: You have a lint error on your build. You can call `composer lint-diff` on your local machine to see what you need to fix.

I suspect the closing bracket here should be on the next line, but you may have other lint errors.

tests/Library/Core/ArrayTranslateTest.php
tburry: Go to your PHPStorm preferences and search for "File and Code Templates". You can edit the default header on a per project basis to include the correct header. I always miss the license too so I found the preference very useful.

tests/Library/Core/ArrayTranslateTest.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @return array Returns an array of test data.
+     */
+    public function provideTranslateableArrays() {
+        $r = [
+            'simpleAssocAdd' => [
+                ['badKey' => 'a.b'],
+                ['badKey' => 'fixedKey'],
+                1,
+                ['fixedKey' => 'a.b']],
```
initvector:
```suggestion
                ['fixedKey' => 'a.b']
            ],
```

PR #9278, opened 2019-09-11
Improvement on FOUC fix, do not hide content from no JS users

slafleche:
We added a hidden style on the body for for forum to prevent some FOUC. This PR is an improvement where we don't apply that style when the user has no JS enabled.


slafleche:
@charrondev slipped in the fix for the code blocks. Let me know if it checks out for you.

PR #9281, opened 2019-09-11
Improve twig utilities

charrondev:
Blocking https://github.com/vanilla/vanilla/pull/9282 which contains some view conversions.

## Twig setup changes

I've improved our twig setup in a few different ways.

- Unify traits.
  We had 2 traits for twig rendering. They have been joined together into `\Vanilla\Web\TwigRenderTrait`. Nothing was using `\Garden\TwigTrait`.
- More robust environment setup.
  - Caching is now used by the twig renderer. Caching is in `/cache/twig` (or whatever environmentally specified filesystem cache directory).
  - `strict_variables` is now enabled in debug mode.
- Function loading has been moved into `\Vanilla\Web\TwigEnhancer`
  - Contains numerous additional utilities.
  - Fetching with caching for: permissions, translations, config.
  - Some very common utility functions.
  - Some app interaction features
    - firing events and capturing echoed output (before returning to the twig renderer)
    - Rendering out a helpAsset to its theme section in the dashboard.

## `Gdn_Form` proxy object

`\Vanilla\Web\TwigFormWrapper` is a wrapper around `Gdn_Form` that proxies calls over to it. A certain whitelist of methods will have their string HTML result wrapped in a `\Twig\Markup`. This prevent twig from double encoding the output of these methods.

## Updates to the legacy renderer.

The legacy renderer has been updated to pass in certain common data.

- `form` (automatically wrapped in the proxy) if it exists.
- `pluggable` The controller instance if it's a pluggable. This is needed for compatibility with the old events so it may be passed to `firePluggableEchoEvent`. 
- `discussion` & `category` if they exist - These are pretty commonly used in a lot of old controller views.

## Add twig namespaces for all enabled applications.

Every addon has gained some twig namespace for their views directory. An addons namespace is a their addon key and it points to their view directory.

As an example: `@dashboard/someView.twig` will try to lookup in `/applications/dashboard/views` and `@rich-editor/my-view.twig` will lookup in `/plugins/rich-editor/my-view.twig`.

For IDE support the symfony extension can be added and configured for PHPStorm/IntelliJ

image


charrondev:
@alex-mtl It has no usage directly. It is only used view `TwigRenderTrait` (because the base twig trait doesn't even expose a translation function).

PR #9299, opened 2019-09-13
Add subcommunity chooser's to keystone & theme-boilerplate

charrondev:
Closes https://github.com/vanilla/multisite/issues/242
Desktop Design https://zpl.io/aRXoRre
Mobile Design https://zpl.io/2jW9Op6

## Chooser update

The chooser required a few updates to go with this PR. Those can be found here: https://github.com/vanilla/multisite/pull/260

## DropDown fixes

Some fixes for the dropdown on the forum were necessary:

- Dropdown can now automatically chooses it's direction based on the quadrant of the screen it is added to if the prop `openDirection` is set to `DropDownOpenDirection.AUTO`.
- FlyoutToggle will now call `closeAllFlyouts` if it is set from a legacy script. This has been defined for some time here https://github.com/vanilla/vanilla/blob/master/js/flyouts.js#L164

## New Button Types

2 new button types have been defined

- `ButtonTypes.TITLEBAR_LINK` - Applies a borderless style with a colors taken from the titlebar.
- `ButtonTypes.RESET` - Provides a reset that removes all browser specific button styles. This is an alternative to `ButtonTypes.CUSTOM`.

## Smarty tag

A smarty tag `{community_chooser}` is now defined. It takes 3 parameters:

- buttonType: Any defined `buttonType`
- buttonClass: An arbitrary CSS class to apply to the chooser button.
- fullWidth: Whether the button should stretch to the full width of it's container (also makes the chevron icon appear in a slightly different position to match the mockups).

## Keystone/boilerplate integration

- Apply a chooser to the header next to the mebox on desktop.
- Apply a chooser as the last item in the mobile navigation.
- Apply a chooser in the center of the footer.

|Desktop|Mobile|Footer|Footer Mobile|
|---|---|---|---|
|image|image|image|image|

## Keystone variables

Due to this components' reliance on the new theme system, it looked a little out of place in some keystone theme options.

To combat this I've:

- Added support for theme options to `FsThemeProvider`.
- Defined 6 variables assets for keystone. 1 for each option.

## Other style fixes

While working with the variables for some of the darker theme options I had to make a couple of other fixes.

- Add a button reset to some rich editor buttons (in addition to their normal styles). Without it the buttons were appearing with white text on a white background.
- Removed the rich editor color definition in `richEditorLegacy.scss`. This was made redundant when user content styles were migrated in the new theming system. This was causing the text color to not properly respect theming variables (and could make for some poor color contrast).


library/SmartyPlugins/function.community_chooser.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param array $params The options for the component if it exists.
+ * - buttonType: 'primary', 'standard', 'titleBarLink', 'transparent', 'transluscent'
+ * - fullWidth: Whether or not the chooser button shoudl take the full width of it's container.
+ *
+ * @return string Some HTML.
+ */
+function smarty_function_community_chooser(array $params): string {
+    $props = json_encode($params);
+    return "<span data-react='subcommunity-chooser' data-props='$props'></span>";
```
alex-mtl: Couldn't we remove this html out of here to react component somehow?

charrondev: We can't. This is what the react component actually mounts on top of (The `data-react` is determines what registered component is mounted).

The `data-props` are what get passed to the component. You can see the registration in the frontend here. https://github.com/vanilla/multisite/blob/master/plugins/subcommunities/src/scripts/entries/forum.tsx#L9

If that isn't registered (eg. subcommunities not enabled) then this is just an empty span tag.

charrondev: I would make it a twig view, but the utility I would need to render it in a global function is sitting in my other PR.

library/Vanilla/Models/FsThemeProvider.php
```diff
@@ -54,6 +57,7 @@ public function __construct(
         $this->addonManager = $addonManager;
         $this->request = $request;
         $this->config = $config;
+        $this->themeOptionValue = $this->config->get('Garden.ThemeOptions.Styles.Value', null);
```
alex-mtl: Could we set some default value instead of `null`? Like: `''` or `[]`.
That would help to determine if property was set to empty or it was never set and because of that it is `null`.

library/Vanilla/Models/FsThemeProvider.php
```diff
@@ -238,14 +258,15 @@ public function getAssetData($themeKey, string $assetKey): string {
...
+     *
+     * @param string $themeID
      *
-     * @param string $id
      * @return mixed
      * @throws NotFoundException Throws an exception if asset not found.
      */
-    private function getAssets(string $id): array {
-        $theme = $this->getThemeByName($id);
+    private function getAllDefinedAssets(string $themeID): array {
```
alex-mtl: Why do we need to rename function? I don't see any profit.

library/Vanilla/Models/FsThemeProvider.php
charrondev: In the end it doesn't really matter, because there is a value in `config-defaults`, but I can do that.

library/Vanilla/Models/FsThemeProvider.php
charrondev: Done

library/Vanilla/Models/FsThemeProvider.php
charrondev: Done

library/SmartyPlugins/function.community_chooser.php
alex-mtl: To me it is not even about twig but about react. But up to you. I just feel this is dirty :-)

Issue #9311, opened 2019-09-17
formatTableName not properly formatting when table name is prefixed with database


patches:
## Issue

The issue was first encountered when performing a merge using the "Forum Merge" plugin. [An issue](https://github.com/vanilla/addons/issues/685) was opened in the addons repository a while ago, but the issue is not in the plugin. 

In this case, the error occurs when trying to get the structure of a table from a different database. We need to query the structure of the tables we are about to merge for the plugin to work properly.

```php
$oldColumns = Gdn::structure()->get($oldDatabase.'.'.$oldPrefix.$table)->columns();
```

Here is the error we get.

<img width="1069" alt="Screen Shot 2019-09-17 at 11 35 55 AM" src="https://user-images.githubusercontent.com/5429361/65056709-5c868e80-d93f-11e9-8f73-11c07e39496b.png">

As you can see in the screenshot, there is an additional set of backticks escaping the database and table names. The backticks are added in the `formatTableName` function of the "Gdn_MySQLDriver" class.

## Expected behavior

Instead of having backticks wrapping the whole table reference, we want each parts separately escaped, as they currently are. Below is the expected output.

```sql
show colums from `vf_dev`.`GDN_User`
```

## How to reproduce

To reproduce this error, you need to have two Vanilla databases.

1. Go in the dashboard and enable the "Forum Merge" plugin.
2. In the dashboard, go in the merge section.
3. Enter the database name and tables prefix (should be GDN_).
4. Press merge

PR #9312, opened 2019-09-17
Prevent formatTableName from adding redundant set of backticks by adding validation

patches:
Here is more details on the problematic. https://github.com/vanilla/vanilla/issues/9311

## Description

The issue was that an additional set of backticks was wrapping table name when the database was added as a prefix. I think that the database name can only be found in "Forum Merge" queries, hence the reason why we never heard of this issue before.

Instead of instantly wrapping the table name, I added a validation that will prevent this addition if the table name is already escaped with backticks.

```php
$table = str_replace('.', '`.`', $table);

if (!preg_match('#^`.+`$#', $table, $matches)) {
    $table = "`$table`";
}
```

The regex is pretty straight forward, maybe there are better solutions to get over this problem.

### DIFF
--- a/library/database/class.mysqldriver.php
+++ b/library/database/class.mysqldriver.php
@@ -206,7 +206,11 @@ class Gdn_MySQLDriver extends Gdn_SQLDriver {
                 $databaseTable = '`'.str_replace('.', '`.`', $matches[1]).'`';
                 $table = str_replace($matches[1], $databaseTable, $table);
             } else {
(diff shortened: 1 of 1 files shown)


tburry:
@patches, is this change still necessary? The PR is getting a bit stale.

patches:
Yes, thanks for the reminder. I will put this in the backlog and try to take care of it somewhere this week.

patches:
I will re-open or open a new PR with the requested changes when the time comes. This PR has been stale for a while now.

PR #9314, opened 2019-09-18
Refactor dashboardSymbol and heading to use twig templates

charrondev:
This PR refactors 2 functions from `functions.render.php` to use a twig template, and soft deprecates them.

- Converts `dashboardSymbol` and `heading` to have a twig template and twig macro.
- Adds some unit tests for `dashboardSymbol`. Unfortunately to properly load the the dashboard alias, `SiteTestTrait` had to be used to start the addon manager in the test.
- Makes the twig cache configurable through the container. It now defaults to off, but is configured in the out primary site `bootstrap` to use `/conf/twig`. The caching was interfering with unit tests, so it is now disabled in all tests suites.
- Fix an infinite loop in `TwigStaticRenderer`.
- Put back the `editcategory` view that was accidently deleted in https://github.com/vanilla/vanilla/pull/9281. It will come in a followup pull request.

**Expected results**

- Dashboard looks the same.


library/core/functions.render.php
```diff
@@ -46,22 +47,32 @@ function alternate($odd = '', $even = 'Alt', $attributeName = 'class') {
...
-            $attr['class'] = $class.' ';
-        } else {
-            $attr['class'] = isset($attr['class']) ? $attr['class'].' ' : '';
+        if (isset($attr['class'])) {
+            unset($attr['class']);
         }
 
-        $baseCssClass = 'icon icon-svg-'.$name;
-        $attr['class'] .= $baseCssClass;
+        $attrs = attribute($attr);
```
alex-mtl: What is this `$attrs` variable is for?

alex-mtl:
```suggestion
```

PR #9319, opened 2019-09-18
Fix bugs from demo

charrondev:
- Fix translation of KB header nav items.
- Fix variable merging in ThemeVariablesTrait (resulting in bad titlebar color for KB).
- Remove footer community chooser from keystone.
- Add button reset to dropdown actions. This causes invisible text on dark themes.


applications/dashboard/design/admin.css
```diff
@@ -7911,15 +7911,15 @@ label[id^=iCheck] {
...
+  max-height: calc(92vh - 108px);
   overflow-y: auto;
   padding: 0 18px;
 }
 .modal-confirm .modal-body {
   padding: 18px;
 }
 .modal-full-height .modal-body {
-  height: calc(90vh - (2 * 48px));
+  height: calc(92vh - 108px);
```
alex-mtl: Just for the record. 
`2*48 = 96      !=     108`
Is it intentional?

charrondev: Yup. It's weird. Not sure why, but I don't want to spend more time in our old modal.

We use a new one with better calculations for everything going forward.

PR #9325, opened 2019-09-20
Fix frontend URL generation when site is served in a subdirectory

charrondev:
Goes with https://github.com/vanilla/knowledge/pull/1202

Fixes https://github.com/vanilla/internal/issues/1970
Fixes https://github.com/vanilla/support/issues/781
Fixes https://github.com/vanilla/support/issues/551
Fixes https://github.com/vanilla/multisite/issues/219

- Format full URLs in `makeProfileUrl()`.
- Use `<SmartLink />` in `RouteHandler`.
- Have `RouteHandler` wrap it's URLs in `formatUrl()` -> This change required a fix in KB https://github.com/vanilla/knowledge/pull/1202
- Fix a react warning in SiteNav caused by having an unneeded fragment wrapper.
- Fix generation of the OpenApi URL when site is served from a subdirectory (test added).
- Fix our PHP request test fixture having backwards logic about setting a site root.

## Testing

@tburry has a WIP PR that lets you serve your site out of `vanilla.localhost/dev` instead of `dev.vanilla.localhost` here https://github.com/vanilla/vanilla-docker/pull/60
It's pretty useful for testing this and is what I used.

### DIFF
--- a/library/Vanilla/OpenAPIBuilder.php
+++ b/library/Vanilla/OpenAPIBuilder.php
@@ -105,7 +105,7 @@ class OpenAPIBuilder {
         // Fix the server URL.
         $openApi['servers'] = [
             [
-                'url' => $this->request->urlDomain(true) . $this->request->getRoot() . '/api/v2',
+                'url' => $this->request->urlDomain(true) . $this->request->getAssetRoot() . '/api/v2',
             ]
         ];
 
--- a/library/Vanilla/OpenAPIBuilder.php
+++ b/library/Vanilla/OpenAPIBuilder.php
@@ -7,6 +7,7 @@
 
 namespace Vanilla;
 
+use Garden\Web\RequestInterface;
 use Symfony\Component\Yaml\Yaml;
 
 /**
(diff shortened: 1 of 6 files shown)


library/Vanilla/OpenAPIBuilder.php
```diff
@@ -81,9 +87,31 @@ public function getFullOpenAPI(): array {
         }
 
         $result = require $this->cachePath;
+
+        // Reapply URL even after pulling from cache.
+        // A site may be accessed from multiple URLs and share the same cache.
+        $result = $this->applyCorrectApiBaseUrl($result);
```
alex-mtl: I think naming convention should be to avoid `Correct` or any other unrelated adjectives to be in method name.
Such kind of names bring the idea that there is `applyIncoorectApiBaseUrl` or `applyWrongApiBaseUrl` somewhere else. Or I could be wrong and we have theme elsewhere?
```suggestion
        $result = $this->applyApiBaseUrl($result);
```

tests/fixtures/src/Request.php
```diff
@@ -176,7 +176,7 @@ public function getRoot() {
      */
     public function setRoot($root) {
         $root = trim($root, '/');
-        $root = $root ? '' : "/$root";
+        $root = $root ? "/$root" : '';
```
alex-mtl: This change looks very suspicious to me. Visually new version is correct.
But how this code has been working for a long time?

charrondev: Because it's a test fixture, and not the real request.

alex-mtl: :-)))))) sorry, missed the path of the file

library/Vanilla/OpenAPIBuilder.php
charrondev: We don't necessarily have a method for them, but the default was certainly incorrect.

I've renamed it to `applyRequestBasedApiBasePath()`

tests/fixtures/src/Request.php
charrondev: It does point to a glaring lack of coverage around around code that uses  the root for anything significant though.

PR #9326, opened 2019-09-22
Fix Attributes JSON encoding for special data types

tburry:
The `Attributes` class is meant to properly JSON encode object-like data, but was not handling date/time objects properly, nor IP addresses.

This fix grabs functionality from the `Data` class into a trait that can be used by both classes. I cut/pasted the methods into the traits rather than using a refactoring tool.


tburry:
Closes https://github.com/vanilla/addons/issues/742. I think.

PR #9329, opened 2019-09-24
Extract post notifications into dedicated methods

### LINKED ISSUE #9285: Refactor post notifications
Refactor notifications processing for comments and discussions so that their code is isolated and easy to extract.

initvector:
Vanilla's notification system is beginning to undergo an overhaul. Instead of doing this in one giant leap, we're taking small incremental steps. This particular step moves notification logic outside post save methods and into dedicated methods on the model. This is done so that when the new notification system is introduced, changes can confidently be made without affecting the rest of the record's save routine. For the sake of simplifying future updates, the process behind generating notifications has been normalized between the comment and discussion models. Previously, each had their own way of handling notifications.

No functional changes have been introduced. Notifications should function the same as they currently do.

Closes #9285

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -536,6 +536,150 @@ class CommentModel extends Gdn_Model {
     }
 
     /**
+     * Notify users of a new comment.
+     *
+     * @param array $comment
+     * @param array $discussion
+     */
+    private function notifyNewComment(?array $comment, ?array $discussion) {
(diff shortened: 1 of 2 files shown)


dannimajdalani:
### Tested on 10012

Commented on a bookmarked discussion
@mentioned a user

In both cases pop-up and email notifications were received.
Screen Shot 2019-09-24 at 12 09 55 PM
Screen Shot 2019-09-24 at 12 09 40 PM

PR #9333, opened 2019-09-24
Add deferment of post notifications

### LINKED ISSUE #9285: Refactor post notifications

### LINKED ISSUE #9286: Defer notifications processing to an in-process job
Using the [message queue](https://github.com/vanilla/vanilla/pull/9046), defer notifications processing to an in-process job. Initial implementation should place this behind a feature flag.

initvector:
Post notifications are currently queued up and sent in the middle of the request, after the post has been saved and before the post is rendered. This update moves executing of the `ActivityModel` queue, which includes post notifications, to the local scheduler. This scheduler processes its own queue after the response has been sent to the client. The intent is to remove sending notifications as a blocking operation for responses.

This is the first use of the scheduler, introduced into Vanilla via #9290.

### Testing

Make sure you've enabled the deferredNotifications feature flag in your config (`Feature.deferredNotifications.Enabled` set to `true`).

The only functional change here should be the order in which notifications are sent. To test, you can add a high-interval [`sleep`](https://www.php.net/sleep) call to the start of the `run` method of `ExecuteActivityQueue`, then create a discussion you expect to receive a notification for. You should notice the server's response is received before the expected completion of the `sleep` call. Alternatively, you can debug the code and add a breakpoint in the same method. Leave execution paused while you verify the response has been received as expected by the browser.

Closes #9286

### DIFF
--- a/applications/dashboard/library/Jobs/ExecuteActivityQueue.php
+++ b/applications/dashboard/library/Jobs/ExecuteActivityQueue.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+use Vanilla\Scheduler\Job\JobExecutionStatus;
+use Vanilla\Scheduler\Job\JobPriority;
+
+/**
+ * Execute the activity queue.
+ */
+class ExecuteActivityQueue implements Vanilla\Scheduler\Job\LocalJobInterface {
+
+    /**
+     * Execute all queued up items in the ActivityModel queue.
+     */
+    public function run(): JobExecutionStatus {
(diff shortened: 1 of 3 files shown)


initvector:
> I was wondering if would be of any help to use the Scheduler in the ConversationModel

Definitely. I'm going to be creating some additional issues for improving notifications. I'll add this suggestion.

dannimajdalani:
**Tested Locally**

created a discussion, with debug on.
Discussed was posted while https://github.com/vanilla/vanilla/blob/5d939c6d0a505d7d7b1ff56d7589c744f8cf0b9b/applications/dashboard/library/Jobs/ExecuteActivityQueue.php#L31 was executing.

PR #9337, opened 2019-09-26
Fix infinite loop in theme preload

charrondev:
Closes https://github.com/vanilla/support/issues/836

## Reproducing

- Create a theme with directory name `not-theme-a`.
- Set the theme key in the `addon.json` or `about.php` file to `theme-a`.
- Set the sites current theme to `theme-a`.
- Visit any page.
- See the infinite loop and subsequent max nesting depth or OOM error.

## The fix

I've wrapped our theme fetching in a catch statement, and am ensuring all functions return some default value if that could not be found. I'm then triggering a warning if the theme doesn't match.

### DIFF
--- a/library/Vanilla/Models/ThemePreloadProvider.php
+++ b/library/Vanilla/Models/ThemePreloadProvider.php
@@ -8,11 +8,13 @@
 namespace Vanilla\Models;
 
 use Garden\Web\Data;
+use Garden\Web\Exception\NotFoundException;
 use Garden\Web\RequestInterface;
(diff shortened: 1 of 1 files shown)


nervoustwit:
Some themes don't have addon.json. I am testing with Total War Forum which has an about.php file and it breaks without an infinite loop. But it breaks.

tburry:
Is this the root of the issue or is there a problem in functions.error that can be addressed to prevent this type of error from happening again? Are there any tests we can write?

PR #9348, opened 2019-09-30
Add fix rows method to discussion model

cchabilall83:
Adding a utility method to the DiscussionModel to be used to prevent double encoding data.

Blocking https://github.com/vanilla/internal/pull/1995

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -3388,7 +3388,7 @@ class DiscussionModel extends Gdn_Model {
      * DiscussionModel::calculate applies htmlspecialchars to discussion name which could conflict when
      * data is encoded on the view.  As result characters will display in their entity codes.  Removing the
      * htmlspecialchars in the calculate method could result in several XSS vulnerabilities in Vanilla.  This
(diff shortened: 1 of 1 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -3385,15 +3385,18 @@ public function joinDiscussionData(&$dataSet, string $discussionID, array $field
...
+     * data is encoded on the view.  As result characters will display in their entity codes.  Removing the
+     * htmlspecialchars in the calculate method could result in several XSS vulnerabilities in Vanilla.  This
+     * method is a utility method avoid dencoding data where it's not necessary.
+     *
+     * @param array $data The discussion record to fix.
      * @return array
      */
-    public function fixRows(array $data):array {
-
+    public function fixRows(array $data): array {
```
charrondev: Since this method is `fixShows()` shouldn't it take `$rows`? This looks like it only works with a single row.

bleistivt:
https://github.com/vanilla/vanilla/blob/588da6090294f17a6fd54d3b76bf2c76163f9854/applications/vanilla/controllers/class.postcontroller.php#L245

Shouldn't this method be used with the form data here?

PR #9354, opened 2019-10-01
Adds database connection smoothness

egarciajulia:
Add supports for database connection smoothness.

In the case that the connection cannot be done because a resources limit is reached (max connections, max user connections, user limits) we loop up to `SmoothTimeoutMillis` milliseconds and we retry the connection.

Every try sleeps for a random amount of time between SmoothWaitMinMillis & SmoothWaitMaxMillis and then attempts a new connection.

The defaults value are:
SmoothTimeoutMillis = 5000
SmoothWaitMinMillis = 250
SmoothWaitMaxMillis = 750
These values can be overwritten via Gdn::config('Database')

In the `query` method, the connection to the database was moved outside of the retry loop. Otherwise, the connection smoothness would be also influenced by the retry configuration.

### DIFF
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -159,6 +159,9 @@ class Gdn_Database {
      * `$timeoutAt` holds the timestamps at which no more retries are done. It's static as it's a fixed amount of time
      * we are willing to wait during the whole request processing time (Otherwise, waiting time could be accumulative in cases)
      *
+     * Beware that in the theoretical "worst" case scenario we wait for up to (SmoothTimeoutMillis +  SmoothWaitMaxMillis - 1),
+     * 5749ms for the defaults values.
+     *
      * @param $dsn
      * @param $user
      * @param $password
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -181,7 +181,6 @@ class Gdn_Database {
                 $pDO->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_SILENT);
 
                 return $pDO;
-
             } catch (Exception $ex) {
                 if ($ex instanceof PDOException &&
                     microtime(true) < $timeoutAt &&
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -182,7 +182,8 @@ class Gdn_Database {
             } catch (Exception $ex) {
                 if ($ex instanceof PDOException &&
                     microtime(true) < $timeoutAt &&
-                    ($ex->getCode() === 1203 || $ex->getCode() === 1040 || $ex->getCode() === 1226)) {
+                    in_array($ex->getCode(), [1203, 1040, 1226])
+                ) {
                     /*
                      * https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html
                      * ---
--- a/library/database/class.database.php
+++ b/library/database/class.database.php
@@ -70,6 +70,15 @@ class Gdn_Database {
     /** @var int Number of retries when the db has gone away. */
     public $ConnectRetries;
 
+    /** @var int Number of milliseconds we are willing to wait to make a successful connection to the DB */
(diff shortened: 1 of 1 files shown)


library/database/class.database.php
```diff
@@ -149,30 +166,55 @@ public function commitTransaction() {
...
+                     * ---
+                     * Error number: 1040; Symbol: ER_CON_COUNT_ERROR; SQLSTATE: 08004
+                     * Message: Too many connections
+                     * ---
+                     * Error number: 1226; Symbol: ER_USER_LIMIT_REACHED; SQLSTATE: 42000
+                     * Message: User '%s' has exceeded the '%s' resource (current value: %ld)
+                     * ---
+                     * In case connection limit is reached, we wait a random amount of time hoping the connection is released
+                     */
+                    usleep(rand($this->SmoothWaitMinMillis, $this->SmoothWaitMaxMillis) * 1000);
```
kaecyra: Do you care whether this could result in a theoretical "worst" case where we wait for up to 5749ms vs your desired max of 5000?

library/database/class.database.php
```diff
@@ -70,6 +70,15 @@ class Gdn_Database {
...
 
+    /** @var int Number of milliseconds we are willing to wait to make a successful connection to the DB */
+    protected $SmoothTimeoutMillis;
+
+    /** @var int Minimum number of milliseconds we want to wait before attempting a new DB connection*/
+    protected $SmoothWaitMinMillis;
+
+    /** @var int Maximum number of milliseconds we want to wait before attempting a new DB connection*/
+    protected $SmoothWaitMaxMillis;
+
```
kaecyra: I feel we should be using the new coding standards for new code, even inside old files that don't currently conform. What do you think?

library/database/class.database.php
```diff
@@ -149,30 +166,55 @@ public function commitTransaction() {
...
 
-            throw new Exception(errorMessage('An error occurred while attempting to connect to the database', $this->ClassName, 'Connection', $ex->getMessage()), 500);
-        }
+                return $pDO;
 
-        return $pDO;
+            } catch (Exception $ex) {
+                if ($ex instanceof PDOException &&
+                    microtime(true) < $timeoutAt &&
+                    ($ex->getCode() === 1203 || $ex->getCode() === 1040 || $ex->getCode() === 1226)) {
```
kaecyra: Maybe use `in_array()` to make it look nicer?

library/database/class.database.php
egarciajulia: Your observation is quite good. I added a comment about it.
I wouldn't change the current behaviour as it would add more logic that doesn't add any extra value. All the setting are configurable to the desired values and this is a fallback mechanism that would be triggered only in rare cases

library/database/class.database.php
```diff
@@ -149,30 +169,56 @@ public function commitTransaction() {
...
-            if ($timeout) {
-                throw new Exception(errorMessage('Timeout while connecting to the database', $this->ClassName, 'Connection', $ex->getMessage()), 504);
-            }
+                // We only throw exceptions during connect
+                $pDO->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_SILENT);
 
-            throw new Exception(errorMessage('An error occurred while attempting to connect to the database', $this->ClassName, 'Connection', $ex->getMessage()), 500);
-        }
+                return $pDO;
+
```
initvector: Linter is complaining about the empty line at the end of the control structure.

```suggestion
```

library/database/class.database.php
initvector: @egarciajulia What _do_ you think?

egarciajulia: I wouldn't touch it at this point.
I agree that we could improve/refactor the class and apply the new standard, however I would do it as a different PR to avoid any unnecessary side effects.

PR #9356, opened 2019-10-01
Track errors related to the state token.

nervoustwit:
This PR will close: https://github.com/vanilla/support/issues/833

It became apparent that this PR was necessary when debugging a customer's SSO when the State token failed to verify. It is not a good practice to expose the details of every error when it comes to user authentication for fear of giving hackers too much knowledge. This PR uses the Logger class to update the Event Log with details of why a State token has failed to be verified.

## Testing

Using JSConnect on a local connection. Turn on the DB Logger. Manipulate the State Token so that it fails for the following reasons:
 - No State Token
 - An expired State Token
 - A State Token with no Expiry date.
 - A mismatched State Token with the local State Token cookie.


library/core/helpers/class.ssoutils.php
```diff
@@ -36,14 +37,17 @@ class SsoUtils {
     /** @var string */
     private $stateToken;
 
+    private $logger;
+
     /**
      * SsoUtils constructor.
      */
-    public function __construct(Gdn_Configuration $config, Cookie $cookie, Gdn_Session $session) {
+    public function __construct(Gdn_Configuration $config, Cookie $cookie, Gdn_Session $session, Logger $logger) {
```
initvector: This kinda thing has the potential to break backwards compatibility. You're adding a new required parameter to the `SsoUtils` constructor. If it's coming from the container, it's no big deal. But if anyone, anywhere, is manually instantiating this class, this change will break that. This constructor signature has been the same for about two years. It's very likely it's been manually instantiated somewhere in that time.

```suggestion
    public function __construct(Gdn_Configuration $config, Cookie $cookie, Gdn_Session $session, ?Logger $logger = null) {
```

You can check to see if `$logger` is null and grab it from the container, manually, if it is.

initvector: I think you wanna switch up this type to `Psr\Log\LoggerInterface` too, to make this more flexible.

library/core/helpers/class.ssoutils.php
```diff
@@ -145,21 +149,62 @@ protected function consumeStateToken($stateToken) {
...
      * @param string $stateToken
+     * @param string $source For logging purposes, optionally pass which token is being validated.
      * @return bool true if the data is valid and false otherwise.
      */
-    protected function isStateTokenValid($stateTokenData, $stateToken) {
+    protected function isStateTokenValid($stateTokenData, $stateToken, $source = '') {
         // Validate expected data.
-        if (!is_array($stateTokenData) || empty($stateTokenData['stateToken']) || empty($stateTokenData['exp'])) {
+        if (!is_array($stateTokenData)) {
+            $this->logger->event(
```
initvector: You should be able to use the `error` method on the logger as a shortcut.

library/core/helpers/class.ssoutils.php
```diff
@@ -145,21 +149,62 @@ protected function consumeStateToken($stateToken) {
...
+                'state_token',
+                $this->logger::ERROR,
+                'Missing stateTokenData',
+                ['source' => $source, 'stateTokenData' => $stateTokenData ?? []]
```
initvector: If we're only in here because `$stateTokenData` isn't an array, does it make sense to coalesce the value to an array?

library/core/helpers/class.ssoutils.php
```diff
@@ -37,12 +37,13 @@ class SsoUtils {
...
     private $stateToken;
 
+    /** @var Logger  */
     private $logger;
 
     /**
      * SsoUtils constructor.
      */
-    public function __construct(Gdn_Configuration $config, Cookie $cookie, Gdn_Session $session, Logger $logger) {
+    public function __construct(Gdn_Configuration $config, Cookie $cookie, Gdn_Session $session, ?\Psr\Log\LoggerInterface $logger = null) {
```
initvector: You're still going to want to grab this out of the container, manually, inside the constructor if `$logger` is `null`. Otherwise, you could potentially have it not be set and your calls to its methods further down will fatal.

charrondev: @initvector That is not necessary in the latest versions of Vanilla. At least not since `garden-container` 3.0: https://github.com/vanilla/garden-container/releases/tag/v3.0.0

> Add support for null defaults for classes that don't exist.

initvector: > That is not necessary in the latest versions of Vanilla. At least not since garden-container 3.0

I'd mentioned in an earlier comment, but my concern wasn't instantiation through the container, but rather instantiation _without_ the container. This class is a couple years old now. I don't know what kinda usage there is in the wild. If there's anyone manually instantiating this thing, it's likely to break if we aren't compensating for this new dependency not automatically being injected.

PR #9357, opened 2019-10-02
Add a check for  record ownership before testing view permission when joining records

dannimajdalani:
closes https://github.com/vanilla/support/issues/547

### Details

in https://github.com/vanilla/vanilla/blob/a9b799697ff7242913eab6179ebc0c8f38bb6232/library/core/functions.general.php#L2284 when Joining records, I added a condition to skip checking of view permissions on the category if the record being tested is owned by the user.

Testing details are in the referenced support ticket.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -2280,7 +2280,8 @@ if (!function_exists('joinRecords')) {
 
             $record = $joinData[$recordType][$iD];
 
-            if ($checkCategoryPermission && $allowedCats !== true) {
+            // Make sure the user is not the owner of the record before unsetting record values.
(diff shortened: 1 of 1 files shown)


initvector:
On looking at this more, would this simply be fixed by updating the call site for this issue to opt out of `$checkCategoryPermission`?

dannimajdalani:
> On looking at this more, would this simply be fixed by updating the call site for this issue to opt out of `$checkCategoryPermission`?

I don't see why we would be needing to check category permissions in this case. 
So yes setting it to false would do the trick instead of having to modify the function being called.

PR #9358, opened 2019-10-02
Touch all feature flag config values

charrondev:
This will enable us to change the default value of one of these in the future so that it only affects new sites.

- Writes all current core feature flags into a site config on structure.
- Cleans up the ending section of the dashboard structure file to be a bit cleaner.


charrondev:
Cancelling, this doesn't need to be done. We can use the jobber to handle this when required.

PR #9372, opened 2019-10-05
Add Psr\SimpleCache support to the app

### LINKED ISSUE #9519: Add PSR-16 Psr\SimpleCache\CacheInterface support
Make an instance of `Psr\SimpleCache\CacheInterface` available in the container for use within the app.

tburry:
This PR adds PSR caching support to Vanilla. There is a simple adapter class that acts as a container factory.

closes #9519.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -54,10 +54,14 @@ $dic->setInstance(Garden\Container\Container::class, $dic)
     ->setConstructorArgs([null, null])
 
     // Cache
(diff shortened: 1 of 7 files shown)


tburry:
Big oops on this one. I forgot about the cache prefix. Will need to re-work.

tburry:
I now realize we should just use a real adapter to `Gdn_Cache` that doesn't involve changing the class. This overly cute PR completely forgot to take into account cache prefixes.

Issue #9374, opened 2019-10-06
Don't cache 302 redirects


tburry:
I had an issue on my phone where I was clicking into a private community and would continuously get the sign in page even after I signed in. Here is what I think is happening:

1. Click a discussion link.
2. 302 to sign in page (response cached).
3. Sign in page (response cached).

We recently made changes to cache all of /entry and I think that this is one of the repercussions. What I would like to try is either:

1. Set response headers to not cache any 302 redirect.
2. If the above is felt to be too risky then just don't cache the private community redirects.

It's my hope that not caching the redirects will help. If it doesn't it is still a positive change in its own right.

charrondev:
I had the same thing happen to me this morning.

PR #9375, opened 2019-10-07
Fix caching of temporary redirects

charrondev:
Fixes https://github.com/vanilla/vanilla/issues/9374

302 redirects are temporary and should have no cache headers.

Upon investigation I found we were not actually sending any additional headers when return 302 redirects.

## `redirectTo()` changes

- Ensure `CacheControlMiddleware::NO_CACHE` is sent for 302 redirects.
- Make sure `safeHeader` is used in `CacheControlMiddleware`. Not really sure if there would be an issue without it, but I don't know all the contexts, `redirectTo()` might be called.

## Signout view

During testing I found the signout view actually doesn't work with `DeferredLegacyScripts` on. Updating this minor script to not rely on jquery fixes it though.

## CacheControlMiddleWare

- Update the middleware itself to send the vary header. This was causing pages using the new controller that redirect to signin to not invalidate properly after the cookie was set.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3058,10 +3058,9 @@ if (!function_exists('redirectTo')) {
             CacheControlMiddleware::sendCacheControlHeaders(CacheControlMiddleware::NO_CACHE);
         }
 
-        $controller = Gdn::controller();
-        if ($controller !== null
-            && in_array($controller->deliveryType(), [DELIVERY_TYPE_ASSET, DELIVERY_TYPE_VIEW], true)
-            && $controller->deliveryMethod() === DELIVERY_METHOD_JSON) {
+        if (Gdn::controller() !== null
+            && in_array(Gdn::controller()->deliveryType(), [DELIVERY_TYPE_ASSET, DELIVERY_TYPE_VIEW], true)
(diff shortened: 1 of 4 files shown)


library/core/functions.general.php
```diff
@@ -3052,9 +3054,25 @@ function redirectTo($destination = null, $statusCode = 302, $trustedOnly = true)
...
+
+        // Send any headers that were put together.
+        if ($controller !== null) {
+            if ($statusCode === 302) {
+                // 302 redirects are TEMPORARY and we don't want them cached by the browser.
+                $controller->setHeader('Cache-Control', CacheControlMiddleware::NO_CACHE);
+            }
+
+            // Send the headers the controller has collected so far.
+            $controller->sendHeaders();
```
tburry: This is potentially a larger change with other unexpected side-effects. Do you think this is necessary?

charrondev: I did find this https://github.com/vanilla/vanilla/blob/2ceba39c4cf112af807c4559231118115195161a/library/core/class.controller.php#L2101-L2102

Currently I can't think necessarily think of a particular scenario in favour or against, but I'll remove it for now.

This seems very hacky to begin with so when we want something more considered we can do it in some new class.

PR #9376, opened 2019-10-07
Mock classes for site-section implementation

cchabilall83:
Add mock classes for site-section implementation.
- MockSiteSectionProvider
- MockSiteSection

require for https://github.com/vanilla/knowledge/pull/1251


tests/fixtures/src/MockSiteSectionProvider.php
```diff
@@ -0,0 +1,85 @@
...
+     * Create site-sections to a section group.
+     * @return array
+     */
+    public static function createSiteSections() {
+        $locales = ["en", "fr", "es", "ru"];
+        $siteSections = [];
+
+        foreach ($locales as $locale) {
+            $siteSectionPath = $locale.'/';
+            $siteSectionName = uniqid("siteSectionName");
```
charrondev: It's generally a good idea to not use random generators in unit tests, as they aren't consistent between builds. It would be better to just append something form the iteration (the locale or the index).

tests/fixtures/src/MockSiteSectionProvider.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Create site-sections to a section group.
+     * @return array
+     */
+    public static function createSiteSections() {
+        $locales = ["en", "fr", "es", "ru"];
+        $siteSections = [];
+
+        foreach ($locales as $locale) {
+            $siteSectionPath = $locale.'/';
```
charrondev: Is the trailing slash here normal? Most things in Vanilla making URLs don't have a trailing slash to start.

tests/fixtures/src/MockSiteSectionProvider.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function getCurrentSiteSection(): SiteSectionInterface {
+    }
+
+    /**
+     * Create site-sections to a section group.
+     * @return array
+     */
+    public static function createSiteSections() {
+        $locales = ["en", "fr", "es", "ru"];
```
charrondev: Something like this would be a little more flexible

```
public static function fromLocales(array $locales = ["en", "fr", "es", "ru"]) {
```

tests/fixtures/src/MockSiteSectionProvider.php
cchabilall83: in the SubcommunitySiteSection a `/` is appended to the folder

https://github.com/vanilla/multisite/blob/b21973009046574e34ff12f7f3b3e2cb8442abd8/plugins/subcommunities/models/SubcommunitySiteSection.php#L49

tests/fixtures/src/MockSiteSectionProvider.php
```diff
@@ -0,0 +1,85 @@
...
+        $siteSections = [];
+
+        foreach ($locales as $locale) {
+            $siteSectionPath = $locale.'/';
+            $siteSectionName = uniqid("siteSectionName");
+            $sectionGroup = "subcommunities-group-1";
```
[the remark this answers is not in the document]
cchabilall83: the site-section-group is the grouping of subcommunities with the same product id.  
no?

tests/fixtures/src/MockSiteSectionProvider.php
cchabilall83: done

tests/fixtures/src/MockSiteSectionProvider.php
cchabilall83: done

Issue #9390, opened 2019-10-10
UserID in afterRegister event may be a string (REDIRECT_APPROVE constant)


bleistivt:
`UserModel::validateSpamRegistration` may return the constant UserModel::REDIRECT_APPROVE instead of an ID:
https://github.com/vanilla/vanilla/blob/60fec99600beec2caf07908a88080485a0b6525e/applications/dashboard/models/class.usermodel.php#L3504

This is passed down all the way to the `afterRegister` event, where it is then passed as an event argument ("UserID").
https://github.com/vanilla/vanilla/blob/60fec99600beec2caf07908a88080485a0b6525e/applications/dashboard/models/class.usermodel.php#L2036

One might expect UserID to always be numeric, which in this case it is not.
This is confusing and undocumented, which leads me to believe, it is a bug.

Could there be plugins relying on this behaviour?
Otherwise I will supply a fix.

charrondev:
@bleistivt after a quick look through our full codebase, including our proprietary plugins, the only references to this constant (or the value it represents) are in:

- `UserModel`
- `EntryController`
- `UsersApiController`

charrondev:
I tracked down the commit where this was introduced https://github.com/vanilla/vanilla/commit/70b73e8c9eb0019bfe60d65ada9b973905d361a9

A string is definitely not appropriate because it may be cast to an int on a strongly typed method call, which would be an issue.

Maybe another constant to set as the userID which would be an int would work. Negative numbers tend to be used for this use case.

PR #9398, opened 2019-10-11
Fire AfterRegister event only if a UserID is present

bleistivt:
The UserID value may also be UserModel::REDIRECT_APPROVE, which is a string, see #9390
Event handlers usually want to act on an actual UserID.

This does not fix the non-strongly-typed return issue present throughout UserModel, as UsersApiController depends on UserModel::insertForBasic sometimes returning UserModel::REDIRECT_APPROVE.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -2031,7 +2031,7 @@ class UserModel extends Gdn_Model {
                 break;
         }
 
-        if ($userID) {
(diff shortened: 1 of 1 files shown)


charrondev:
I'm sorry, we seem to be having an issue with our CI scripts here & external fork/pull request contributions. I won't have a chance to look into it until next wednesday. I'll try to get that fixed up then & review the PR.

tburry:
I rebased and merged. See #9526.

PR #9427, opened 2019-10-18
Add the password requirements text to the invitation form

tburry:
This makes the invitation registration mirror the other registration methods. Closes https://github.com/vanilla/support/issues/568.


tburry:
Here is a screenshot of the result on the default theme.

T1

PR #9439, opened 2019-10-19
Remove Gdn_SqlDriver::_whereIn() empty set deprecation

tburry:
We were logging a deprecation warning which was filling up logs on certain servers. Furthermore, the deprecation warning was wrong because the method isn’t deprecated; just one code path was deprecated.

That being said, it shouldn’t really be an error to pass an empty array to a “in” clause. This change cleans up the logic a bit to just make the empty set cases a truthy or falsey expression rather than the `(null)` set.

I added some basic functional tests for the generated SQL. These will provide a good basis if there are any bugs found with this code.

### DIFF
--- a/library/database/class.sqldriver.php
+++ b/library/database/class.sqldriver.php
@@ -754,8 +754,10 @@ abstract class Gdn_SQLDriver {
      *
      * @param string $tableName The name of the table to delete from.
      * @param array $wheres An array of where conditions.
+     * @param int $limit Limit the number of records to delete.
+     * @return string Returns an DML statement.
      */
-    public function getDelete($tableName, $wheres = []) {
+    public function getDelete($tableName, $wheres = [], $limit = 0) {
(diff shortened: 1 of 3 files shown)


tburry:
Closes https://github.com/vanilla/support/issues/715.

library/database/class.sqldriver.php
```diff
@@ -754,8 +754,10 @@ public function getCountLike($table = '', $like = false) {
      *
      * @param string $tableName The name of the table to delete from.
      * @param array $wheres An array of where conditions.
+     * @param int $limit Limit the number of records to delete.
+     * @return string Returns an DML statement.
      */
-    public function getDelete($tableName, $wheres = []) {
+    public function getDelete($tableName, $wheres = [], $limit = 0) {
```
charrondev: Why are we adding this parameter to a method that just triggers an error?

tburry: So that the base class and subclass match definitions.

PR #9441, opened 2019-10-19
Fix date grammar on category list

### LINKED ISSUE #7252: Correct grammar for current date data
Link to the issue: https://github.com/vanilla/support/issues/30

A client recently noticed when a discussion is updated at the current date, the element **span.MostRecentOn** displays "on" instead of "at".

Unfortunately, the "on" string is outside the **Gdn_Format::date** logic: https://github.com/vanilla/vanilla/blob/master/applications/vanilla/views/categories/helper_functions.php#L121

tburry:
Two things:

- The most recent date used “most recent by user on date”. If the date is formatted as just a time then it should be “at”. This would be very tedious to determine with localization so I opted just to remove the word “on” entirely.
- The category list displays the date twice. I added an option to remove it from the string.

Closes https://github.com/vanilla/support/issues/30.
Closes #7252.

### DIFF
--- a/applications/vanilla/views/categories/helper_functions.php
+++ b/applications/vanilla/views/categories/helper_functions.php
@@ -95,8 +95,8 @@ endif;
 if (!function_exists('MostRecentString')):
     function mostRecentString($row, $options = []) {
         $options = (array)$options + [
-            'user' => true,
-            'date' => true,
+            'showUser' => true,
+            'showDate' => true,
         ];
 
         if (!$row['LastTitle']) {
@@ -112,7 +112,7 @@ if (!function_exists('MostRecentString')):
             $row['LastUrl'],
             'LatestPostTitle');
 
-        if ($options['user'] && val('LastName', $row)) {
+        if ($options['showUser'] && val('LastName', $row)) {
             $r .= ' ';
 
             $r .= '<span class="MostRecentBy">'.t('by').' ';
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/categories/helper_functions.php
```diff
@@ -93,9 +93,15 @@ function getOptions($category) {
 endif;
 
 if (!function_exists('MostRecentString')):
-    function mostRecentString($row) {
-        if (!$row['LastTitle'])
+    function mostRecentString($row, $options = []) {
```
charrondev: What's the reason for the array cast here? Is there something out there known to be passing a second parameter to this function?

applications/vanilla/views/categories/helper_functions.php
```diff
@@ -93,9 +93,15 @@ function getOptions($category) {
...
+        $options = (array)$options + [
+            'user' => true,
+            'date' => true,
```
charrondev: These aren't really named in a way that your would traditionally name a boolean option.

I would expect `user` to be an actual user record, and `date` to be a date object.

`displayUser` & `displayDate` would be more appropriate (or something else).

applications/vanilla/views/categories/helper_functions.php
tburry: There is not a known case, but PHP allows it.

PR #9442, opened 2019-10-19
Fix device detection for multiple device types

tburry:
The device detection library we use can return a string of “mobile;tablet”. In the past, we’ve modified that to return just “tablet”. This seems to be prone to regressions so I’m hardening the detection on the app-side.

Closes https://github.com/vanilla/support/issues/58.

### DIFF
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -3926,19 +3926,40 @@ if (!function_exists('userAgentType')) {
             $type = $value;
         }
 
+        if (defined('TESTMODE_ENABLED') && constant('TESTMODE_ENABLED')) {
+            $type = null;
+        }
+
         if ($type !== null) {
             return $type;
         }
 
+        // A function to make sure the type is one of our supported types.
+        $validateType = function (string $type): string {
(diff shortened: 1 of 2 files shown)


library/core/functions.general.php
```diff
@@ -3926,19 +3926,40 @@ function userAgentType($value = null) {
             $type = $value;
         }
 
+        if (defined('TESTMODE_ENABLED') && constant('TESTMODE_ENABLED')) {
+            $type = null;
+        }
```
charrondev: I'm not the biggest fan of doing explicit test checking in functions like this.

Since this function is just being called directly in the test could we use some parameter like `ignoreCache` or `skipCache` instead?

tburry: I tried your suggestion and it looks a lot more sloppy IMO and then adds a feature to the weird function that we would then have to support. The choice is to either do significantly more refactoring or just us the constant which is consistent with the way some other dev who will remain nameless did it.

tburry: Alright, I opted for **false** clearing the cache.

PR #9446, opened 2019-10-21
Add full regular expression patterns support for username validation

patches:
It is currently possible to add custom validation on usernames by adding the "ValidationRegex" and "ValidationLength" configs to the configuration file. The "ValidationRegex" is wrapped in a character class which makes the functionality very rigid.

For example, it is impossible to do negative look-around to prevent usernames from starting with a certain pattern.

The change was discussed in [this internal issue](https://github.com/vanillaforums/estimates/issues/708)

It will now be possible to add "ValidationRegexPattern" to your configuration file and write the full regular expression without any restrictions. This pattern will have priority on the other (ValidationRegex, ValidationLength) configurations.

### DIFF
--- a/library/core/functions.validation.php
+++ b/library/core/functions.validation.php
@@ -281,6 +281,8 @@ if (!function_exists('validateUsername')) {
         $validateUsernameRegex = validateUsernameRegex($isPartialRegex);
 
         if ($isPartialRegex) {
+            // Unescape the delimiters and escape them again to prevent double escaping.
+            $validateUsernameRegex = str_replace('`', '\\`', str_replace('\\`', '`', $validateUsernameRegex));
             $validateUsernameRegex = "`^({$validateUsernameRegex})?$`siu";
         }
 
--- a/library/core/functions.validation.php
+++ b/library/core/functions.validation.php
@@ -247,6 +247,13 @@ if (!function_exists('validateUsernameRegex')) {
         static $validateUsernameRegex;
 
         if (is_null($validateUsernameRegex)) {
+            // Get the full regular expression from the configuration file.
+            $validateUsernameRegex = c('Garden.User.ValidationRegexPattern');
+
+            if ($validateUsernameRegex) {
+                return $validateUsernameRegex;
+            }
+
             // Set our default ValidationRegex based on Unicode support.
             // Unicode includes Numbers, Letters, Marks, & Connector punctuation.
             $defaultPattern = (unicodeRegexSupport()) ? '\p{N}\p{L}\p{M}\p{Pc}' : '\w';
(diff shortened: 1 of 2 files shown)


patches:
@tburry Since we were discussing this together, I would like to know if you have something else in mind.

library/core/functions.validation.php
```diff
@@ -270,11 +270,16 @@ function validateUsernameRegex() {
      * @return bool Returns true if the value validates or false otherwise.
      */
     function validateUsername($value) {
-        $validateUsernameRegex = validateUsernameRegex();
+        $validationRegexPattern = c('Garden.User.ValidationRegexPattern');
```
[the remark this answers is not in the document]
patches: No there is not, this can be moved inside validateUsernameRegex.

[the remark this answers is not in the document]
patches: Currently trying to see if making this change in the validateUsernameRegex function would affect the Quotes addon.

tburry:
I've added some partial tests for your changes. I would like you to complete them and fix any bugs that they uncover. I'm available to help if you need it.

patches:
I completed the tests, let me know if there's anything you would like me to cover.

library/core/functions.validation.php
```diff
@@ -270,11 +275,18 @@ function validateUsernameRegex() {
...
      */
     function validateUsername($value) {
-        $validateUsernameRegex = validateUsernameRegex();
+        $isPartialRegex = true;
+
+        // $isPartialRegex will be set to false if Garden.User.ValidationRegexPattern is set in the config.
+        $validateUsernameRegex = validateUsernameRegex($isPartialRegex);
+
+        if ($isPartialRegex) {
+            $validateUsernameRegex = "`^({$validateUsernameRegex})?$`siu";
```
tburry: I see you changed to backticks, but then an admin could also put backticks in their regex. This is a difficult problem without an obvious solution. I've seen a trick where a non-printable character  as a delimiter. Perhaps @DaazKu has an idea.

DaazKu: I am a big fan of keeping `/` whenever possible as the regex delimiter.
That being said ```str_replace('`', '\\`', $validateUsernameRegex)``` should do the trick here.

tburry: The user can still escape the backtick in the config though. Example:

```php
preg_match('/\`/', '`') // returns 1
```

DaazKu: If you want a foolproof solution you unescape the delimiters and the escape them again. This will prevent double escaping stuff.

`str_replace('`', '\\`', str_replace('\\`', '`', $validateUsernameRegex))`

patches:
The logic to prevent the double escaping of backticks in partial regexes has been added to the `validateUsername` function. I also added 4 tests that covers possible cases.

PR #9470, opened 2019-10-24
Get rid of forcing tables to MyISAM when FULLTEXT key is used (rebased)

tburry:
Vanilla always checked if a table had a FULLTEXT key in one row and forced that table to MyISAM. The minimum required MySQL version for Vanilla already supports fulltext keys for InnoDB tables, therefore that logic isn't needed any more.

The Gdn_MySqlStructure class had a helper function "_supportsFulltext. Under the assumption that this will always return `true`, I've deleted all code lines that became unnessecary.

This will close 
https://github.com/vanilla/vanilla/issues/5555
https://github.com/vanilla/vanilla/issues/6029

### DIFF
--- a/library/database/class.mysqlstructure.php
+++ b/library/database/class.mysqlstructure.php
@@ -211,6 +211,93 @@ class Gdn_MySQLStructure extends Gdn_DatabaseStructure {
     }
 
     /**
+     * Generate the DDL for creating a table.
+     *
+     * @return string Returns a DDL statement.
+     */
+    final protected function getCreateTable(): string {
+        $primaryKey = [];
+        $uniqueKey = [];
+        $fullTextKey = [];
+        $indexes = [];
+        $keys = '';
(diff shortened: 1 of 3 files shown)


library/database/class.mysqlstructure.php
```diff
@@ -274,28 +268,10 @@ protected function _create() {
...
-                array_map('strtolower', $columnKeyTypes);
-                if (in_array('fulltext', $columnKeyTypes)) {
-                    $hasFulltext = true;
-                    break;
-                }
-            }
-            if ($hasFulltext) {
-                $this->_TableStorageEngine = 'myisam';
-            } else {
-                $this->_TableStorageEngine = c('Database.DefaultStorageEngine', 'innodb');
```
tburry: My reading of this seems to show that `Database.DefaultStorageEngine` is not respected anymore.

PR #9472, opened 2019-10-24
Add checkbox & isCurrent property to DropDownItemLink

slafleche:
Added subcommunity chooser to pages that aren't explicitely in a subcommunity.


slafleche:
@charrondev not sure how to fix this error, any suggestions?

PR #9474, opened 2019-10-24
Add UserProviderInterface & MockUserProvider

charrondev:
## Why?

With this new mock, we can now write unit tests against stuff that depends on user record/expansion without needing the a real DB for the actual `UserModel`.

## What is it?

- Add `Contracts\Models\UserProviderInterface`
  - Expands users.
  - Get user fragments by ID.
- Update `\UserModel` to implement this interface.
  - These are **new methods** that call out to old ones & do a little cleanup. The new methods don't have weird/unneeded parameters like the base user model, and doesn't have any 
  - Main application `bootstrap.php` and the test `Bootstrap` class have been updated with a definition for the interface/implementation.
- Add's a `MockUserProvider` that is configurable with fake users. No DB connection.
  - Configured automatically in the container setup of `MinimalContainerTestCase`

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -22,6 +22,15 @@ class UserModel extends Gdn_Model implements UserProviderInterface {
     /** @var int */
     const GUEST_USER_ID = 0;
 
+    /** @var int This happens to be the same as the guest ID because it's just been that way for so long. */
+    const UNKNOWN_USER_ID = 0;
+
+    /** @var string */
+    const GENERATED_FRAGMENT_KEY_UNKNOWN = "unknown";
+
+    /** @var string */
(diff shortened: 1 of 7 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -1173,16 +1196,13 @@ public function expandUsers(array &$rows, array $columns, array $options = []) {
                         // Add an alias to Photo. Currently only used in API calls.
                         setValue('PhotoUrl', $user, $photo);
                     } else {
-                        $user = [
-                            'userID' => 0,
-                            'name' => 'unknown',
-                            'email' => 'unknown@example.com'
-                        ];
-                        $user['photoUrl'] = self::getDefaultAvatarUrl($user);
```
charrondev: User was always undefined here anyways so not passing the parameter in `getUnknownFragment()` has no impact.

PR #9475, opened 2019-10-24
Get fresh user information in quote embeds when filtering

charrondev:
**In draft until https://github.com/vanilla/vanilla/pull/9474 and https://github.com/vanilla/vanilla/pull/9467 are merged**

Resolves https://github.com/vanilla/support/issues/923

## Changes

- Update embed filters to take an actual embed instance. This ensures they are normalized in the embed class before the filtering takes place.
- Use the filterer before rendering. This ensures fresh embed data is overlayed on top. Filters can add content that should be dynamic to an embed.
- Move test around stripping nested embeds into `QuoteEmbedFilterTest`.
	- Also add tests for the inverse case here when the full content option is passed.
- Use the user fragment provider to get fresh user data and overlay it on top of the embed's `insertUser`. 
- Did some refactoring to make the responsibility of the `QuoteEmbed` & `QuoteEmbedFilter` more clear

**The Embed**

- Normalizes fields. It does not do any HTML formatting anymore.
- Sets HTML fields that need to be set from a secure source to some known value ahead of time. The filter be responsible for generating secure content here.

**The Filter**

- Renders HTML into into the body through a secure source (the `FormatService`).
- Fetches user data from a secure source `UserProviderInterface`. This way our `label` will always have secure data.


cchabilall83:
Getting an error rendering the quote. (see image

)

cchabilall83:
works!

PR #9477, opened 2019-10-24
Implement ApplicationInterface and ApplicationProviderInterface toget…

alex-mtl:
Introduce `ApplicationInterface` and `ApplicationProviderInterface` together with their default implementations `Application` and `ApplicationProvider`

Those interfaces will allow to provide some information about apps installed to plugins which require that information.

For example to provide reserved root slugs which do not require to be redirected to default site section (site section agnostic endpoints).

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -70,6 +70,11 @@ $dic->setInstance(Garden\Container\Container::class, $dic)
     ->setClass(SingleSiteSectionProvider::class)
     ->setShared(true)
 
+    // Site applications
+    ->rule(\Vanilla\Contracts\Site\ApplicationProviderInterface::class)
+    ->setClass(\Vanilla\Site\ApplicationProvider::class)
+    ->setShared(true)
+
     // AddonManager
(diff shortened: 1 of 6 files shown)


bootstrap.php
```diff
@@ -70,6 +70,12 @@
     ->setClass(SingleSiteSectionProvider::class)
     ->setShared(true)
 
+    // Site applications
+    ->rule(\Vanilla\Contracts\Site\ApplicationProviderInterface::class)
+    ->setClass(\Vanilla\Site\ApplicationProvider::class)
+    ->addCall('add', [new \Vanilla\Site\Application('garden', ['api', 'entry', 'sso', 'utility'])])
```
charrondev: Normally when initializing container items we do them lazily by using `\Garden\Container\Reference`.

```suggestion
    ->addCall('add', [new \Garden\Container\Reference(
	    \Vanilla\Site\Application::class,
	    ['garden', ['api', 'entry', 'sso', 'utility']]
    )])
```

PR #9479, opened 2019-10-24
Remove discussion archiving

### LINKED ISSUE #2894: Archive function doesn't exclude threads from the bookmark counter
Currently, the Bookmark counter counts archived discussions. This is problematic as the bookmark page does not display the archived discussions, which can lead to an incorrect count of bookmarked discussions. Further, the pagination seems based on this count, so even if only five bookmarked discussions are displayed, if there are another 50 bookmarked discussions that are archived, pagination will allow for enough pages to display the total of 55 bookmarked discussions.

- Expected behaviour:Bookmark counter ignores archived discussions
- Pagination on bookmark page ignores archived discussions

**Reproduction Steps**

1. Create "Category A"
1. Create "Discussion 1" in "Category A"
1. Take note of current user's bookmark count.
1. Bookmark "Discussion 1"
1. Notice that bookmark counter goes up by 1.
1. Archive "Category A"

**Expected**

Bookmark counter goes down by 1.

**Actual**

Bookmark counter stays the same.

### LINKED ISSUE #3870: Remove auto-archiving exclude functionality
We found that this feature has little actual customer value and it is currently hidden from sites (although still present). The purpose of this issue is to fully remove it from the code base.

Upon investigation, I found some customers taking advantage of the auto-close functionality, but the exclude functionality is just too fragile to still support.

### LINKED ISSUE #5206: "Exclude archived discussions from the discussions list" is broken
In the advanced settings it is possible to set a date for archived discussions and a checkbox "Exclude archived discussions from the discussions list".

Looking at the DiscussionModel, that checkbox is handled in the `addArchiveWhere()` method. The recent discussion list get its data from:
~~~
    public function getWhereRecent($Where = [], $Limit = false, $Offset = false) {
        $result = $this->getWhere($Where, '', '', $Limit, $Offset);
        return $result;
    }
~~~

Archived discussions are not handled here. Adding `$this->addArchiveWhere($this->SQL);` as the first line to getWhereRecent() fixes that immediately, but still shows a false pagination since `getCount()` is used for getting the amount of recent discussions. 

Archived discussions are not only shown in /recent, but also in /categories, but not in /categories/discussions

I'm unsure what that checkbox should really do. Hiding archived discussions in each and every discussion list would be too much, I guess, but the way it is expressed right now leaves room for confusion. Can anybody explain how that checkbox should be understood?

### LINKED ISSUE #5294: Archiving has become a "hidden" feature
There is no link in the dashboard for "/vanilla/settings/archive" except when you already use that feature:

~~~
>addLinkIf(c('Vanilla.Archive.Date', false) &&  Gdn::session()->checkPermission('Garden.Settings.Manage'), t('Archive Discussions'), '/vanilla/settings/archive', 'forum.archive', 'nav-forum-archive', $sort)
~~~

Has it been on purpose that this feature has been removed for not knowing people?

Has it something to do with those inconsistencies?
https://github.com/vanilla/vanilla/issues/5206
https://github.com/vanilla/vanilla/issues/3297

tburry:
Discussion archiving has been unsupported and hidden for some time now. This removes it altogether as we are unlikely to be able to properly support it.

Closes #5294.
Closes #5206.
Closes #2894.
Closes #3870.


tburry:
It appears as though some of this functionality is still needed. Closing for now.

PR #9493, opened 2019-10-27
Add dateLastComment support to the discussions API

### LINKED ISSUE #9516: Add dateLastComment support to the discussions API
We should see `dateLastComment` on discussions, even though we can expand `lastPost` because it's a fairly fundamental column. Moreover, we should add a filter on this property to help people get around the page limitation.

tburry:
This adds a dateLastComment property and filter to the /api/v2/discussions resources.

Note: Unit testing uncovered a bug in `Gdn_MySQLDriver` so this PR also includes a fix and a test for that.

Closes #9516.


library/database/class.sqldriver.php
```diff
@@ -1567,9 +1567,7 @@ public function orWhereNotIn($field, $values) {
      * @return string The parsed expression.
      */
     protected function _parseExpr($expr, $name = null, $escapeExpr = false) {
-        $result = '';
-
-        $c = substr($expr, 0, 1);
```
tburry: This was causing a warning when `$expr` is not a string. We had made a change a while back to accept dates so it's a real possibility.

PR #9501, opened 2019-10-29
Use DateTimeImmutable to create timestamps for OAuth2 class.

nervoustwit:
Use the DateTimeImmutable class to create the expiry time for stashed sessions.

This is a follow up to this PR: https://github.com/vanilla/vanilla-patches/pull/636 in response to this issue: https://github.com/vanilla/support/issues/840

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -630,6 +630,7 @@ class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
             default:
 
                 // Save the access token and the profile to the session table, set expiry to 3 minutes.
+                $expiryTime = new \DateTimeImmutable('now + 5 minutes');
                 $stashID = $this->sessionModel->insert(
                     [
                         'Attributes' => [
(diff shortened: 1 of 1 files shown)


library/core/class.oauth2.php
```diff
@@ -630,14 +630,15 @@ public function entryEndpoint($sender, $code, $state = '') {
...
+                $expiryTime = new \DateTimeImmutable('now + 5 minutes');
                 $stashID = $this->sessionModel->insert(
                     [
                         'Attributes' => [
                                 'AccessToken' => $response['access_token'] ,
                                 'RefreshToken' => $response['refresh_token'],
                                 'Profile' => $profile,
                             ],
-                        'DateExpires' => date(MYSQL_DATE_FORMAT, strtotime('3 minutes')),
+                        'DateExpires' => $expiryTime->format(MYSQL_DATE_FORMAT),
```
tburry:
```suggestion
                        'DateExpires' => $expiryTime,
```

Can you see if this works?

nervoustwit: I tried it. The Session Insert class takes a formatted datetime string, $expiryTime is an object.

PR #9503, opened 2019-10-29
Set Google+ plugin to hidden

dannimajdalani:
This PR set Google+ plugin to hidden as requested in https://github.com/vanilla/vanilla-patches/issues/642

### DIFF
--- a/tests/APIv2/AddonsTest.php
+++ b/tests/APIv2/AddonsTest.php
@@ -14,7 +14,7 @@ class AddonsTest extends AbstractAPIv2Test {
     private $coreAddons = [
         'conversations', // applications
         'allviewed', 'emojiextender', 'facebook', 'flagging',
-        'googleplus', 'googleprettify', 'gravatar', 'indexphotos', 'profileextender', 'quotes',
+        'googleprettify', 'gravatar', 'indexphotos', 'profileextender', 'quotes',
         'splitmerge', 'stopforumspam', 'twitter', 'vanillainthisdiscussion', 'vanillastats', 'editor', 'oauth2',
         'recaptcha', 'stubcontent', 'vanillicon', 'googlesignin'// plugins
     ];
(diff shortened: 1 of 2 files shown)


dannimajdalani:
Failing test fixed.

tests/APIv2/AddonsTest.php
```diff
@@ -24,7 +24,7 @@ class AddonsTest extends AbstractAPIv2Test {
     ];
 
     private $hiddenAddons = [
-        'dashboard', 'vanilla', 'gettingstarted'
+        'dashboard', 'vanilla', 'gettingstarted', 'googleplus'
```
initvector: What do you think of avoiding including explicit checking for Google+, altogether? The other addons in `$hiddenAddons` are items we plan on keeping, but want to limit access to. Google+ is an addon we're limiting accesss to because we're trying to get rid of it. This seems like it'll be one more things we need to remember to cleanup when time comes to fully remove the addon.

tburry: Will we *have* to remember? The failing test does a pretty good job of reminding us.

That being said, I would be okay removing it, but I am guessing that would make the hidden addon test more difficult to write at this time whereas removing it from the test when we remove the addon would be easy.

initvector: > Will we _have_ to remember? The failing test does a pretty good job of reminding us.

What would fail, though? The tests here are basically ensuring all the "hidden" addons are treated as invisible/non-existent to the API. They're not supposed to show up in the index. They're supposed to 404 if you attempt to get them, directly. If a "hidden" addon really doesn't exist, I don't think the tests would fail.

tburry: I don't know. I haven't looked too closely at the test again to be honest. I guess I just figure it's not worth the time to refactor the test.  Usually, when I'm writing tests I'm testing something specific and after the fact I just trust that the tests are quietly doing good work for us.

In this particular case, it is probably fine to remove it. From a philosophical perspective I always want to err on the side of more tests/coverage.

PR #9508, opened 2019-10-30
Add ability to make Chromatic optional in build

initvector:
Vanilla's [Chromatic](https://www.chromaticqa.com) job runs every time there's a build. Forked repositories do not contain the necessary environmental variable (`CHROMATIC_APP_CODE`) required to perform this job, so builds based on those forks fail.

The build has been updated to check for the Chromatic app code in the environment to determine whether or not Chromatic tests should be run. If no Chromatic app code is detected, those tests should be skipped.

Making this happen required essentially wrapping the existing `chromatic/test` job in a custom `chromatic_test` job and putting in a conditional stop. The `core/checkout` step was not compatible with this method. [Yarn errors were triggered](https://app.circleci.com/jobs/github/vanilla/vanilla/21961) unless `checkout` was used, instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,7 @@ aliases:
 
                 # Run code sniffer on the diffed files.
                 ./.circleci/scripts/diff-standards.sh $CUSTOM_TARGET_BRANCH
+
 jobs:
     frontend_setup:
         executor: core/node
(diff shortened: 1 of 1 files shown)


charrondev:
@initvector Instead of checking for a "skip" environment var could we just check if the token is set?

For forked repos we can't allow access to our env vars (for security reasons) so their CI builds also fail on chromatic.

initvector:
> Instead of checking for a "skip" environment var could we just check if the token is set?

Makes good sense to me!

PR #9511, opened 2019-10-31
Do not add #latest anchor when AutoOffset is disabled

### LINKED ISSUE #2469: AutoOffset no longer acting as intended
Applied AutoOffset fales to config to stop jumping to last unread post:

```
$Configuration['Vanilla']['Comments']['AutoOffset'] = FALSE;
```

 Does not seem to be respected in latest master.

bleistivt:
fixes #2469 

While i appreciate closing off stale issues, this just came up recently:
https://open.vanillaforums.com/discussion/comment/258777/#Comment_258777
With the suggested fix of inserting JS to remove the anchor.

So why not fix what (for the time being) is still a feature.

### DIFF
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -108,7 +108,7 @@ if (!function_exists('writeComment')) :
 
                 <?php
                 // Write a stub for the latest comment so it's easy to link to it from outside.
-                if ($currentOffset == Gdn::controller()->data('_LatestItem')) {
+                if ($currentOffset == Gdn::controller()->data('_LatestItem') && c('Vanilla.Comments.AutoOffset')) {
                     echo '<span id="latest"></span>';
                 }
                 ?>
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -108,7 +108,7 @@ if (!function_exists('writeComment')) :
 
                 <?php
                 // Write a stub for the latest comment so it's easy to link to it from outside.
-                if ($currentOffset == Gdn::controller()->data('_LatestItem') && c('Vanilla.Comments.AutoOffset')) {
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -108,7 +108,7 @@ function writeComment($comment, $sender, $session, $currentOffset) {
 
                 <?php
                 // Write a stub for the latest comment so it's easy to link to it from outside.
-                if ($currentOffset == Gdn::controller()->data('_LatestItem')) {
+                if ($currentOffset == Gdn::controller()->data('_LatestItem') && Gdn::config('Vanilla.Comments.AutoOffset')) {
```
charrondev: I have a small issue with this: `Gdn::config` is pretty expensive call and this adds `n` calls to it where `n` is the number of comments on the page. A number of site use `100` which is quite a lot.

There other cases where this does happen but we're trying to remove those and not add new ones.

One way to workaround this would be set some value once in the controller's data and then check it here.

Eg.
```php
// In the controller
$this->Data['useAutoOffset'] = Gdn::config('Vanilla.Comments.AutoOffset'));

// In the view
$latestItem = $sender->Data['_LatestItem'] ?? null;
$useAutoOffset = $sender->Data['useAutoOffset'] ?? false;
if ($currentOffset == $latestItem && $useAutoOffset) {

}
```

`Gdn_Controller::data()` is almost as bad, isn't normally as bad of an impact expect in bigger loops like this.

Nowadays we just don't make our views like this anymore (see recent inclusion of twig & React), but for the old ones can tougher to work with.

bleistivt: With short circuit evaluation this should only be called once per page where
`$currentOffset == Gdn::controller()->data('_LatestItem')`. 

I don't really see a better way to do this apart from using the controller Data[] array directly which is not in line with other helper_functions.php.

charrondev: Ah that makes sense. Didn't think of that.

Just for future reference though, accessing `->Data` is definitely preferable to use `->data()`. Data is a just a `valr()` wrapper, which had a lot more utility before the `??` operator existed.

PR #9521, opened 2019-11-01
Fix invalid roles being included in user API responses

initvector:
The users resource of API v2 has a problem with invalid roles. If a user is assigned to a role that doesn't exist, output validation fails and the entire page of results is unable to be displayed. This can be caused by a user having a non-existent role association in the `UserRole` table.

### Why is this happening?
1. `UsersApiController::normalizeOutput` uses `UserModel::getRoles`.
1. `UserModel::getRoles` calls `RoleModel::roles`, passing `true` for its `$force` parameter which will leave invalid roles in the results returned. These invalid roles contain only an ID and an empty name.
1. When the output is validated, the schema requires a name for each role, so invalid roles having an empty name fails validation.

### How is it being fixed?
A new parameter has been added to `UserModel::getRoles`. It will allow explicitly opting out of including invalid roles in the return value. By default, the current behavior of including invalid roles will continue.

### Recommended Testing
1. Add a bogus role-to-user mapping in the `UserRole` table.
1. Get the user via the API (e.g. /users/api/v2/users/{userID}).
1. Before this fix, the previous call should fail with a validation error. On this branch, the user should be displayed without issue. The invalid role is not included in the output.


applications/dashboard/models/class.usermodel.php
```diff
@@ -1838,7 +1839,10 @@ public function getRoles($userID) {
 
         $result = [];
         foreach ($rolesDataArray as $roleID) {
-            $result[] = RoleModel::roles($roleID, true);
+            $role = RoleModel::roles($roleID, $includeInvalid);
```
tburry: I went to check on that second parameter for `RoleModel::roles()` and it just says `$force`. Would you mind filling out the docblock of `RoleModel::roles()` as part of this PR?

initvector: Done. I targeted the wrong branch (master) with the original docs commit, but they're here now.

PR #9522, opened 2019-11-01
Remove legacy Vanilla notifications queueing

initvector:
Many, many years ago, 181a6f6 introduced `Vanilla.QueueNotifications`. It was a way to offset the sending of notifications. It was used by (almost) no one. Today, we remove it from core.

All removed code is gated behind a truthy `Vanilla.QueueNotifications` config. I do not anticipate any collateral damage.


tburry:
Would it be possible to remove the legacy notification queueing altogether?

PR #9523, opened 2019-11-03
Fix some legacy spoiler handling in Markdown content

### LINKED ISSUE #5761: Markdown formatter can produce invalid html output with bbcode spoiler tags
The markdown editor is capable of producing invalid html markup. This can entirely break a pages flow if a browser parses it properly. This visual output may to appear completely different, but it depending on theme and how the browser parses this issue along with the rest of the page it can completely break the page flow. An example can be found on a live site [here](https://forums.marvelheroesomega.com/discussion/3919/improvements-to-loot-boxes-and-costume-changes-coming-soon/p23)

**Steps to Reproduce**

1. Open a vanilla forum with the advanced editor set to markdown
2. Create a post or comment with the following body
```markdown
> [spoiler]
>
> [/spoiler]
```
3. Observe the html output. The primary Chrome Dev tools view may attempt to correct the mistake, so you will need to look in the true source, which can be accessed in the chrome dev tools through the following menu
image

**Expected Output**
```html
<blockquote class="UserQuote">
    <div class="QuoteText">
        <p>[spoiler]</p>
    </div>
</blockquote>
<blockquote class="UserQuote">
    <div class="QuoteText">
    </div>
</blockquote>
<blockquote class="UserQuote">
    <div class="QuoteText">
        <p>[/spoiler]</p>
    </div>
</blockquote>
```

**Real ouput**
```html
<blockquote class="UserQuote">
    <div class="QuoteText">
       <p><div class="Spoiler"></p>
    </div>
</blockquote>

<p>&gt;</p>

<blockquote class="UserQuote">
    <div class="QuoteText">
        <p></div></p>
    </div>
</blockquote>
```

## Root Cause

Pretty sure this comes from our `legacySpoilers` formatting.

https://github.com/vanilla/vanilla/blob/7d8464da520795331cf84456466c952e90bcc046/library/core/class.format.php#L950-L958

tburry:
Our legacy spoiler can really mangle HTML. Ideally, I’d get rid of them, but I think that could lead to much controversy.

Instead, I’m making the checks a bit more strict. Legacy spoilers will only replace if:

1. They are marked up on a single line.
2. The tags are at the beginning of lines.
3. Process spoilers for the markdown before converting to markdown, not after. This removes some ugly line breaking cases.

This is not the best fix, and still leaves many edge cases un-addressed, but I think it’s a positive change IMO.

Closes #5761.


library/Vanilla/Formatting/Formats/MarkdownFormat.php
```diff
@@ -47,10 +47,30 @@ public function __construct(
...
     }
 
+    /**
+     * This copy of legacy spoilers processes first.
+     *
+     * @param string $html The HTML to process.
+     * @return string
+     */
+    private function parentLegacySpoilers(string $html): string {
+        return parent::legacySpoilers($html); // TODO: Change the autogenerated stub
```
initvector:
```suggestion
        return parent::legacySpoilers($html);
```

library/Vanilla/Formatting/Formats/HtmlFormat.php
```diff
@@ -225,4 +225,24 @@ protected function legacySpoilers(string $html): string {
...
+     * @return bool
+     */
+    private function hasLegacySpoilers(string $html): bool {
+        // Check for an inline spoiler.
+        if (preg_match('`(\[spoiler\])[^\n]+(\[\/spoiler\])`', $html)) {
+            return true;
+        }
+
+        // Check for a multi-line spoiler.
+        if (preg_match_all('`^\[\/?spoiler\]$`m', $html, $m)) {
```
initvector: Can this be simplified to a regular ole `preg_match`?

```suggestion
        if (preg_match('`^\[\/?spoiler\]$`m', $html)) {
```

library/Vanilla/Formatting/Formats/MarkdownFormat.php
```diff
@@ -47,10 +47,30 @@ public function __construct(
...
+     */
+    protected function legacySpoilers(string $html): string {
+        return $html;
+    }
+
     /**
      * @inheritdoc
      */
     public function renderHtml(string $value, bool $enhance = true): string {
+        $value = $this->parentLegacySpoilers($value);
```
initvector: Since `parentLegacySpoilers` is private, and we're currently only using it in this one place, what's the primary benefit of the separate method versus just making the call to `parent::legacySpoilers` here?

tburry: That is a brain fart. I'll change.

PR #9533, opened 2019-11-04
Prevent [deleted user] from appearing in activities

bleistivt:
When deleting newly registered users (e.g. spam registrations), [deleted user] appears heavily in the activity feed which looks unprofessional.

The reason for this is that when a users content is deleted, it is queried by InsertUserID which doesn't cover all of the users activity.
Especially the "USER hat joined." activity type can have InsertUserID == null (no session present) or the ID of the admin creating the profile for the user.
This fixes that by also deleting by ActivityUserID (which has an index).

Note that this will delete grouped activities where the ActivityUserID happens to belong to the user being deleted.
But this is already the case when InsertUserID == ActivityUserID (grouped activities do not appear on user feeds anyway if the userID is not ActivityUserID).
So this just makes the behaviour more consistent.

Because of those grouped activities (which contain the UserIDs in the JSON "Data" field that can hardly be queried), the second part of this fix involves ignoring deleted users when building activity headlines.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -3826,6 +3826,7 @@ class UserModel extends Gdn_Model implements UserProviderInterface {
 
         // Remove activities
         $this->getDelete('Activity', ['InsertUserID' => $userID], $content);
+        $this->getDelete('Activity', ['ActivityUserID' => $userID], $content);
 
         // Remove activity comments.
         $this->getDelete('ActivityComment', ['InsertUserID' => $userID], $content);
--- a/library/core/functions.general.php
+++ b/library/core/functions.general.php
@@ -1513,20 +1513,21 @@ if (!function_exists('_formatStringCallback')) {
                                     continue;
                                 }
 
+                                $separator = '';
                                 if ($i == $count - 1) {
-                                    $result .= ' '.t('sep and', 'and').' ';
+                                    $separator = ' '.t('sep and', 'and').' ';
                                 } elseif ($i > 0) {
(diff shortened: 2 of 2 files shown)


charrondev:
@bleistivt Sorry looks like I regressed the change that made the forked PRs work. The change I made broke our in branch PRs. I'm looking into it again, and I'll restart your build after I've fixed it.

tburry:
I don't think we can take this change as it is for the following reasons:

1. Calling `getDelete` can really be a problem with the scale of the activity table.
2. It heavy handedly deleting grouped activities may be looked at as a regression. Certainly, I would have to see some tests accompany the code in that case.

Would you consider doing a PR for the second part of your PR that fixes the formatting?

bleistivt:
It's already calling `getDelete` by InsertUserID. Deleting by ActivityUserID should be equally as fast (both indexed) and there are very few entries (mostly "user joined") left where `InsertUserID != ActivityUserID`.

But I understand that deleting grouped activities (which can already happen in some edge cases) is a regression. Here is the formatting fix:
#9560

In general, as activities can be an m:n realtionship, is it planned to break out the ActivityUsers into their own table anytime soon? The way grouped notifications work right now is really confusing. The newest user who is joining (or getting a badge etc.) "claims" the activity so that it appears on their wall. On busy boards it stays there for maybe a few seconds until the next user claims it.

I was thinking about a second PR that sets ActivityUserID to NULL when grouping activities so that those grouped activities are not owned by any particular user (then getDelete could actually be called on ActivityUserID safely). But I'm worried that may have unforeseen consequences on notifications.

PR #9537, opened 2019-11-04
Makes the Canonical Tag respect isHomepage

igraziatto:
**Estimate:** https://github.com/vanillaforums/estimates/blob/master/chipde/Chip.de%20-%20Homepage%20Canonical%20Tag%20Estimate.md
**Request:** htps://github.com/vanillaforums/estimates/issues/806

**Notes:**
The reason you might see more than one client mentioned is that this service is requested by a lot of clients.

**Related PR's:**

### DIFF
--- a/applications/vanilla/controllers/class.categoriescontroller.php
+++ b/applications/vanilla/controllers/class.categoriescontroller.php
@@ -580,7 +580,8 @@ class CategoriesController extends VanillaController {
         $this->addModule($CategoryFollowToggleModule);
         $this->addModule('TagModule');
 
-        $this->canonicalUrl(url('/categories', true));
+        $canonicalUrl = $this->getCanonicalUrl($this->Data);
+        $this->canonicalUrl($canonicalUrl);
 
         if ($this->View === 'all' && $displayAs === 'Flat') {
             $this->View = 'flat_all';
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -767,4 +770,14 @@ public function data($path, $default = '') {
...
         }
     }
+
+    /**
+     * Return URL based on 'isHomepage'
+     *
+     * @param array $data
+     * @return string
+     */
+    private function getCanonicalUrl($data) {
```
tburry: This isn't really a getter so should something like `calculateCanonicalUrl`.

applications/vanilla/controllers/class.categoriescontroller.php
```diff
@@ -767,4 +770,14 @@ public function data($path, $default = '') {
...
+        return $data['isHomepage'] ? url('/', true) : url('/categories', true);
```
tburry: Protect against `isHomepage` key not existing.

applications/vanilla/controllers/class.discussionscontroller.php
```diff
@@ -88,7 +88,10 @@ public function index($Page = false) {
         $this->fireEvent('AfterPageCalculation');
 
         // Set canonical URL
-        $this->canonicalUrl(url(concatSep('/', 'discussions', pageNumber($Offset, $Limit, true, false)), true));
+        $canonicalUrl = $this->Data['isHomepage'] ?
```
tburry: Protect against `isHomepage` key not existing.

charrondev: @tburry I'm pretty sure it always exists (ensured by `Gdn_Dispatcher` calling `->passData('isHomepage', $someValue)`)

I also don't think we have any intention of changing that due to the 100s of themes all rely on it being set all the time (and master views).

Issue #7769, opened 2018-09-13
Allow configuration of additional allowed URL schemes in user content 


baylf2000:
We run a cluster of gaming servers and would like to utilise Vanilla for our forums, however we have a small issue in that our users often need to post links to allow users to directly connect to their gaming servers.

These links look like "steam://connect/1.2.3.4:1234". These links can be used in any web browser to open the Steam client, run the appropriate game and connect to that server.

Steam (from Valve Corporation Ltd) is the world's largest digital distribution platform for PC games, and almost all PC games are distributed via this platform. See https://en.wikipedia.org/wiki/Steam_(software)

I would like to suggest the addition of "steam://" to the allowed url schemes for Vanilla.

Many thanks.

charrondev:
I definitely agree that it would be nice to allow such links. What I don't want to do is maintain a list of application specific URL schemes.

Would something like a configuration option for additional allowable URL schemes be sufficient? We don't currently have the bandwidth as a team to implement something like this as a priority right now, but if you are able I could point you in the right direction to make a contribution like this.

To summarize:

- Add a configuration option to the `VanillaHtmlFormatter` class. For consistency let's call it `'Garden.Html.CustomUrlSchemes`.
- Merge these the values of this array into the Htmlawed config.
- Due to the performance sensitive natures of this HTML filtering, be sure to implement the config lookup use the `StaticCacheConfigTrait` the `StaticCacheConfigTrait::c()` to lookup the value.

baylf2000:
A configuration option sounds like a great solution. I'm sure there are other url types out there that users may want to add themselves,

I'm far from a professional developer, but I'd be happy to at least take a look at anything you can suggest that might help me make such a contribution.

linc:
I'd be happy to allow `steam://` links by default.

Issue #4043, opened 2016-05-26
Cannot save array into config


rbrahmson:
I have a plugin (DiscussionAlert) that has a configuration screen that displays the defined categories and let the admin select categories via checkboxes.  The selection is saved as an array in config.php.  

_Works fine in 2.2, breaks in 2.3 - the values are not saved in the config.php file._

This is not an access issue - other variables are saved just fine.

**To reproduce:** look at the source of the plugin:
https://github.com/rbrahmson/VanillaPlugins/blob/master/DiscussionAlert/class.DiscussionAlert.plugin.php#L179-L223

The unsaved setting in question is 'Plugins.DiscussionAlert.CategoryNums'.

linc:
We did change this; we no longer allow arrays to be saved to the config. We should add proper documentation / error messages and let folks know their plugins need to be updated. It will still *read* an existing array afaik, just not write new ones. This allows plugin authors to seamlessly update their plugins.

rbrahmson:
Why would you make a change that breaks compatibility? 

Is there another way to save a list of elements into config?

linc:
* Performance.
* Moving to a JSON-formatted config in the future.
* Code maintenance (volume of code > value).
* Redundancy / Better ways to do the same thing (Our dot syntax allows using multiple sub-keys as an array when you pull it out).
* A rarely-used feature - I think all of Vanilla's code used it once or twice, ever.

Yes, save it as a comma-separated list, or use multiple keys.

rbrahmson:
The issue is not how I save is in config but _how Vanilla saves it_.  I am using "$configurationModule->initialize(" with a " 'Control' => 'CheckBoxList',".

I feel that if you want to change how it is saved in config then you need to do it so that the Checkboxlist is saving and reading it differently (however you like - I didn't care much that it was an array), transparent to the caller (so as to not break plugins/apps).

R-J:
I'm used to create simple setting screens like that:

~~~
$categories = CategoryModel::categories();
unset($categories[-1]);

$configurationModule = new ConfigurationModule($sender);

$configurationModule->initialize(
    [
        'Plugin.CategoryIDs' => [
            'Control' => 'CheckBoxList',
            'LabelCode' => 'Categories',
            'Items' => $categories,
            'Options' => ['ValueField' => 'CategoryID', 'TextField' => 'Name']
        ]
    ]
);

$configurationModule->renderAll();
~~~

[...]

PR #9538, opened 2019-11-04
Allow custom url schemes

bleistivt:
I followed the suggestions of #7769

The only difference is, that `Garden.Html.CustomUrlSchemes` replaces the default config in my implementation. This is because afaik array configs are somewhat deprecated and because of HtmLawed's special scheme syntax which doesn't really map to an array. String concatenation would be another option.


bleistivt:
Based on this comment, but it is quite old:
https://github.com/vanilla/vanilla/issues/4043#issuecomment-222366946

The problem is that the HtmLawed config defines multiple groups:

    classid: clsid;
    href: aim, feed, file, ftp, gopher, http, https, irc, mailto, news, nntp, rapidminer, sftp, ssh, telnet;
    style: nil;
    *: file, http, https'

http://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/htmLawed_README.htm#s3.4.3

Do you only want the `href` group to become `Garden.Html.AllowedUrlSchemes`?

charrondev:
> Do you only want the href group to become Garden.Html.AllowedUrlSchemes?

That's what I was thinking. Do you have any need of specifying the other ones? I feel like as a config setting it would be more useful if it didn't have to be based around `Htmlawed`'s specific syntax.

Rich editor for example, would like to use a shared config setting, and that doesn't use `Htmlawed` at all.

Issue #7957, opened 2018-10-22
Sign In page has no OpenGraph description, leading to unpredictable sharing


linc:
When you share a private discussion link it defaults to the Sign In page in link previews. 

The Sign In page has no OpenGraph description tag (`og:description`).

On Facebook (and possibly elsewhere) when a page is shared with no OpenGraph description it appears it will default to the first paragraph tag on the page. On the Sign In page, this is likely to be an arbitrary string of text in the header or footer, which can be confusing or inappropriate.

We should add the site's description as the OpenGraph description data for this page.

PR #9539, opened 2019-11-04
Add meta description to the signin page

bleistivt:
see #7957


charrondev:
For other reviewers:

- The part of the CI that isn't running is due to this being a PR from a fork. A fix is in progress w/ https://github.com/vanilla/vanilla/pull/9508 so ignore it for now.

PR #9545, opened 2019-11-05
Add the $me smart ID

### LINKED ISSUE #7344: Add the $me smartID
Smart IDs currently translate secondary keys to primary keys. One smart ID that would be very nice would be to have `$me` translate into the currently authenticated user's ID.

- If there is no signed in user then an exception should be thrown.
- The smart ID code might require a colon which would need to be changed.
- Since there is a custom `UserSmartIDResolver` the code can be added there.
- The smart ID docs need to be updated.
- Unit test.

tburry:
Closes #7344. See that issue for details.

### DIFF
--- a/library/Vanilla/Web/UserSmartIDResolver.php
+++ b/library/Vanilla/Web/UserSmartIDResolver.php
@@ -20,6 +20,20 @@ class UserSmartIDResolver {
     private $viewEmail = false;
 
     /**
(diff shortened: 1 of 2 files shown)


tburry:
The `$me` smart tag will work for anything such as specifying `?insertUserID=$me` for filters.

PR #9547, opened 2019-11-05
Embedded Comments: Prevent discussion creation when Body is empty.

### LINKED ISSUE #5561: [Embed] Discussion should not be created if comment is invalid
When using embedding for comments, if you click on creating a new comment but do not fill the body box you get the error "Body is required" which is fine but you also have the discussion, with no comment, created on the forum.

The discussion should not be left on the forum if the comment's creation failed.
I say left on the forum because I was told that we need to create the discussion to properly validate the comment (To be validated)

bleistivt:
fixes #5561
similar to this fix: https://github.com/vanilla/vanilla/commit/1c55708b9f520cffcb9f248e0dfa470eff3c7f69


bleistivt:
Since the original issue was marked with "Won't fix" (*The severity and priority of this issue do not warrant the time or complexity needed to fix it*), I'll close this two-line fix and patch it using a plugin to prevent my forum being spammed with empty comment threads.

PR #9550, opened 2019-11-05
Update comment and discussion embed factories to support site sections

charrondev:
Requires https://github.com/vanilla/multisite/pull/301 to actually work

- Updates `CommentsEmbedFactory` and it's tests.
  - To support site sections.
  - Tests now use a data provider.
- Updates `DiscussionEmbedFactory` and it's tests.
  - To support site sections.
  - Tests now use a data provider.

**Expected**

`/api/v2/media/scrape` now works on discussions/comment URLs that have a site section in them.

### DIFF
--- a/library/Vanilla/EmbeddedContent/AbstractEmbedFactory.php
+++ b/library/Vanilla/EmbeddedContent/AbstractEmbedFactory.php
@@ -62,7 +62,7 @@ abstract class AbstractEmbedFactory implements EmbedCreatorInterface {
             $pathMatches = $this->canHandleEmptyPaths;
         } else {
             $regex = $this->getSupportedPathRegex($domain);
-            $pathMatches = (bool) preg_match($regex, $path);
+            $pathMatches = preg_match($regex, $path) === 1;
         }
 
         return $schemeMatches && $pathMatches && $domainMatches;
--- a/library/Vanilla/EmbeddedContent/AbstractEmbedFactory.php
+++ b/library/Vanilla/EmbeddedContent/AbstractEmbedFactory.php
@@ -61,7 +61,8 @@ abstract class AbstractEmbedFactory implements EmbedCreatorInterface {
         if ($path === null) {
             $pathMatches = $this->canHandleEmptyPaths;
         } else {
-            $pathMatches = (bool) preg_match($this->getSupportedPathRegex($domain), $path);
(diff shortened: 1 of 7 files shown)


library/Vanilla/EmbeddedContent/AbstractEmbedFactory.php
```diff
@@ -61,7 +61,8 @@ public function canHandleUrl(string $url): bool {
         if ($path === null) {
             $pathMatches = $this->canHandleEmptyPaths;
         } else {
-            $pathMatches = (bool) preg_match($this->getSupportedPathRegex($domain), $path);
+            $regex = $this->getSupportedPathRegex($domain);
+            $pathMatches = (bool) preg_match($regex, $path);
```
alex-mtl: I don't see why we need to change anything here.
But if we do - I would prefer to avoid casting types - it is tricky and suspicious.
```suggestion
            $pathMatches = preg_match($regex, $path) === 1;
```

library/Vanilla/EmbeddedContent/Factories/AbstractOwnSiteEmbedFactory.php
```diff
@@ -0,0 +1,72 @@
...
+use Vanilla\EmbeddedContent\AbstractEmbed;
+use Vanilla\EmbeddedContent\AbstractEmbedFactory;
+use Vanilla\EmbeddedContent\Embeds\QuoteEmbed;
+use Vanilla\Site\SiteSectionModel;
+use Vanilla\Web\Asset\SiteAsset;
+
+/**
+ * Embed factory with utilities for matching our own site.
+ */
+abstract class AbstractOwnSiteEmbedFactory extends AbstractEmbedFactory {
```
alex-mtl: Why do we need this class to  be `abstract`?

charrondev: Because it doesn't implement all methods from `AbstractEmbedFactory`. Notably `getSupportedPathRegex()`

library/Vanilla/EmbeddedContent/AbstractEmbedFactory.php
charrondev: The change was actually to assist with debugging (so I could put a breakpoint and look at the generated regex).

PR #9553, opened 2019-11-05
Implement organize categories page warning message for non source locales

NeenaDev:
Show full width message as per the mockup:
https://app.zeplin.io/project/5c65db78a981283c0b8d956e/screen/5ced59ab17ac961dceaed3a2

Relates: https://github.com/vanilla/knowledge/pull/1340


library/src/scripts/messages/messageStyles.ts
```diff
@@ -95,6 +95,10 @@ export const messagesClasses = useThemeCache(() => {
         zIndex: 20,
     });
 
+    const fullWidth = style("fullWidth", {
```
slafleche: Instead of using these styles, I'd rather we optionally used the <Container> component, which already limits the width of the contents.

library/src/scripts/messages/messageStyles.ts
```diff
@@ -174,6 +178,17 @@ export const messagesClasses = useThemeCache(() => {
...
     });
 
+    const errorIcon = style("errorIcon", {
+        ...absolutePosition.middleLeftOfParent(),
+        maxWidth: percent(100),
+        transform: translate(`-100%`),
+        marginLeft: unit(-14),
+        $nest: {
+            "&&": {
+                color: "#555A62",
```
slafleche: Don't hard code colors, use globalVariables.

slafleche:
Also, your text is mis aligned vertically.

library/src/scripts/messages/messageStyles.ts
NeenaDev: Fixed- Wrap message component inside container component.

library/src/scripts/messages/messageStyles.ts
NeenaDev: Fixed- used globalVariables now

library/src/scripts/messages/messageStyles.ts
```diff
@@ -95,6 +95,24 @@ export const messagesClasses = useThemeCache(() => {
...
+                maxWidth: "max-content",
+            },
+        },
+    });
+
+    const noPadding = style("noPadding", {
+        $nest: {
+            "&&": {
+                padding: 0,
+                //position: "fixed",
```
charrondev: Commented line here.

```suggestion
```

NeenaDev:
> Also, your text is mis aligned vertically.

Done- aligned text vertically

library/src/scripts/messages/messageStyles.ts
```diff
@@ -95,6 +95,24 @@ export const messagesClasses = useThemeCache(() => {
         zIndex: 20,
     });
 
+    const innerWrapper = style("innerWrapper", {
+        $nest: {
+            "&&": {
+                flexDirection: "unset",
```
charrondev: Was there a concrete reason for going w/ "unset" here, instead of "initial" or "row"?

`unset`'s behaviour varies based on the parent definition. Do we actually want that?

library/src/scripts/messages/messageStyles.ts
```diff
@@ -95,6 +95,24 @@ export const messagesClasses = useThemeCache(() => {
...
+                maxWidth: "max-content",
```
charrondev: This isn't an actual finalized CSS feature. It seems to only be part of a draft specification.

https://drafts.csswg.org/css-sizing-3/#width-height-keywords

library/src/scripts/messages/messageStyles.ts
```diff
@@ -182,6 +211,8 @@ export const messagesClasses = useThemeCache(() => {
         justifyContent: "flex-start",
     });
 
+    const messageContainer = style("messageContainer", {});
```
charrondev: This seems to be entirely unused.

charrondev:
Just a note of process:

Generally we title out pull requests as a phrase that would complete a sentence and is an action. Eg.

"If I merge this pull request, it will __YOUR TITLE__".

For something like this "Implement organize categories page warning" would be more appropriate.

library/src/scripts/messages/messageStyles.ts
```diff
@@ -95,6 +95,25 @@ export const messagesClasses = useThemeCache(() => {
         zIndex: 20,
     });
 
+    const innerWrapper = style("innerWrapper", {
+        $nest: {
+            "&&": {
+                flexDirection: "row",
+                maxWidth: 920,
```
charrondev: Where does this number come from? Is it something that should be a calculated value (maybe based off of our column widths)?

In any case it needs to be a variable.

We want to ensure that if we change something like the main column width, that stuff like this is affected too.

NeenaDev: removed max-width: static number from container css

library/src/scripts/messages/messageStyles.ts
NeenaDev: removed unused variables

library/src/scripts/messages/messageStyles.ts
NeenaDev: deleted this styling, thanks for clarification

library/src/scripts/messages/messageStyles.ts
NeenaDev: removed  flexDirection: "unset" to innerWrapper

PR #9555, opened 2019-11-06
Add callback scheduled job

initvector:
A new type of scheduled job has been added, allowing for a simple callback to be executed at end of the request.

### Example Usage
```php
$scheduler->addJob(
    Vanilla\Scheduler\Job\CallbackJob::class,
    [
        "callback" => "implode",
        "parameters" => ["one", "two", "three"],
    ]
);
```

Any valid [`callable` in PHP](https://www.php.net/manual/en/language.types.callable.php) can be provided as the callback. The parameters array will be passed to the callable with each element representing a separate parameter to the function/method.

### DIFF
--- a/library/Vanilla/Scheduler/Job/CallbackJob.php
+++ b/library/Vanilla/Scheduler/Job/CallbackJob.php
@@ -17,8 +17,14 @@ class CallbackJob implements LocalJobInterface {
     /** @var callable */
     private $callback;
 
+    /** @var int */
+    private $delay;
+
     /** @var array */
     private $parameters = [];
+
+    /** @var JobPriority */
+    private $priority;
 
     /**
      * Execute the configured callback function.
@@ -59,6 +65,7 @@ class CallbackJob implements LocalJobInterface {
      * @return void
      */
     public function setPriority(JobPriority $priority) {
+        $this->priority = $priority;
     }
 
     /**
@@ -68,5 +75,6 @@ class CallbackJob implements LocalJobInterface {
      * @return void
      */
     public function setDelay(int $seconds) {
+        $this->delay = $seconds;
     }
 }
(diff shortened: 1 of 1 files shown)


library/Vanilla/Scheduler/Job/CallbackJob.php
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Execute a callback as a job.
+ */
+class CallbackJob implements LocalJobInterface {
+
+    /** @var callable */
+    private $callback;
+
+    /** @var array */
+    private $parameters = [];
```
tburry: Do I need parameters here or can I just use closures or higher order functions? Consider if your class just took a callback. I could make the following higher order function:

```php
public function callWithArgs(callable $callback, array $args) {
    return function () use ($callback, $args) {
        $callback(..$args);
    }
}
```

Moreover, I could just inline a closure fairly easily to whatever is deferring execution:

```php
new CallbackJob(function () use ($discussionID) {
    $this-processNotifications($discussionID);
});
```

library/Vanilla/Scheduler/Job/CallbackJob.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->parameters = $parameters;
+    }
+
+    /**
+     * Set job priority
+     *
+     * @param JobPriority $priority
+     * @return void
+     */
+    public function setPriority(JobPriority $priority) {
```
tburry: Are these methods supposed to be backed by properties?

library/Vanilla/Scheduler/Job/CallbackJob.php
initvector: The parameters here are available to be passed directly to a function, but they're not required. You can certainly create a `CallbackJob` and specify no parameters.

library/Vanilla/Scheduler/Job/CallbackJob.php
initvector: Not exactly. The interface requires setters, but no getters. We have a standard way of getting the data in, but no standard way of getting it out. Priorities and delays do not appear to be implemented in the scheduler at this time. That said, for the sake of not having empty methods, I have implemented the properties.

PR #9557, opened 2019-11-06
Add Scheduler to Vanilla's testing bootstrap

initvector:
If tests in Vanilla happen to touch any code that makes use of the scheduler, it will currently fail. This is due to the scheduler not being properly configured in the test bootstrap. That is resolved with this update.

### DIFF
--- a/tests/Bootstrap.php
+++ b/tests/Bootstrap.php
@@ -317,7 +317,7 @@ class Bootstrap {
             ->setShared(true)
 
             ->rule(Vanilla\Scheduler\SchedulerInterface::class)
-            ->setClass(Vanilla\Scheduler\DummyScheduler::class)
+            ->setClass(VanillaTests\Fixtures\Scheduler\InstantScheduler::class)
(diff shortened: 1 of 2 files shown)


tests/Bootstrap.php
```diff
@@ -315,6 +315,13 @@ public function initialize(Container $container) {
             ->rule('HtmlFormatter')
             ->setClass(\VanillaHtmlFormatter::class)
             ->setShared(true)
+
+            ->rule(Vanilla\Scheduler\SchedulerInterface::class)
```
tburry: Unless the dummy scheduler can execute jobs as soon as they are scheduled then it's probably not right for testing. I would like unit tests to run jobs synchronously so that we can get coverage on the code in the jobs too.

initvector: I don't _think_ there's a scheduler that currently does this, but I'm looking into it. We might just need to add one as part of this PR.

tburry: I was wondering if we can subclass the dummy and have its schedule method just call dispatchAll. Definitely need @egarciajulia's input here.

PR #9561, opened 2019-11-07
Displaying data in translationGrid on language selection

NeenaDev:
This PR is for -
Add TranslationLanguageHandler comp with language dropdown
Show translation on gris based on language.


library/src/scripts/content/translationGrid/TranslationLanguageHandler.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const classes = translationGridClasses();
+    let selectedIndex = 0;
+    let { newtranslationData } = props;
+
+    const handleOnChange = (name: string) => {
+        const filteredData = newtranslationData.filter(v => v.locale === name);
+        setData(filteredData);
+    };
+    useEffect(() => {
+        const filteredData = newtranslationData.filter(v => v.locale === "ca");
```
charrondev: Why are we filtering by a specific locale here?

library/src/scripts/content/translationGrid/TranslationLanguageHandler.tsx
```diff
@@ -0,0 +1,118 @@
...
+    translation: string;
+    translationStatus: string;
+    multiLine?: boolean;
+    maxLength?: number;
+}
+export interface ITranslationLanguageHandler {
+    data: ITranslation[];
+    inScrollingContainer?: boolean;
+    otherLanguages: ILanguageItem[];
+    i18nLocales: ILocale[];
```
charrondev: Why are we passing this as a prop instead of using the `useLocaleInfo()` hook?

charrondev: We're also taking `otherLanguages` here, which we shouldn't need at the same time. We should be able to use the locale info to make our items.

library/src/scripts/content/translationGrid/TranslationLanguageHandler.tsx
```diff
@@ -0,0 +1,118 @@
...
+    dateUpdated?: string;
+    newtranslationData: ITranslationDummy[]; // for storybook
```
charrondev: Casing on variables like this is traditionally `camelCase`. Eg. `newTranslationData`

library/src/scripts/forms/select/SelectBox.tsx
```diff
@@ -34,6 +35,7 @@ interface IProps {
     openAsModal?: boolean;
     selectedIndex?: number;
     renderLeft?: boolean;
+    handleOnChange?: (...args: any[]) => any;
```
charrondev: I'd recommend naming this differently. Normally a prop doesn't have `handle` at the beginning of it. The hander would.

Eg.

```
<SelectBox onChange={handleChange} />
```

Also we are we using `any` here? This should have types for the arg array and for the return type (does it even have a return type? I feel like the return type is normally `void` for this type of thing).

library/src/scripts/forms/select/SelectBox.tsx
```diff
@@ -79,36 +81,70 @@ export default class SelectBox extends React.Component<ISelfLabelledProps | IExt
             selectedIndex: index,
             selectedItem,
         });
+        this.props.handleOnChange(selectedItem.name);
     };
 
+    private renderChild = (child, selected, classes) => {
```
charrondev: This seems like something that should just be its own component. While we have done this type of thing in the past it's a bit of an anti-pattern.

library/src/scripts/content/translationGrid/TranslationLanguageHandler.tsx
NeenaDev: It was just for storybook but now changed to first option value in dropdown.

library/src/scripts/content/translationGrid/TranslationLanguageHandler.tsx
NeenaDev: Done.
Removed hardcode data and use useLocaleInfo() to get locales

library/src/scripts/content/translationGrid/TranslationLanguageHandler.tsx
NeenaDev: Fixed.

library/src/scripts/forms/select/SelectBox.tsx
NeenaDev: Made changes-
add onChange() props to "DropdownItemButton" in SelectBox component.
removed "LanguageDropdown" component.

charrondev:
Went with https://github.com/vanilla/internal/pull/2074 instead.

PR #9582, opened 2019-11-09
Prevent useless user preference queries

bleistivt:
Whenever a signed in user visits a page that is the controllers `index()` with non-empty arguments (e.g. a standard discussion page like /discussion/123/slug), `UserModel::clearSectionNavigationPreference()` is called.

This is because the user may have been redirected from a falsely set `DashboardNav.DashboardLandingPage` in their preferences. This logic is flawed, but since with the current architecture there is not way to "test" a dispatch to a particular route, it is probably the best that can be done.
See #4497 and #4734 for reference.

Unfortunately `clearSectionNavigationPreference()` calls `saveToSerializedColumn()` twice leading to 2 entirely useless and duplicated write queries on the user table even if there is no actual change in the data. 

This PR checks for an actual change in the preference data to prevent those queries.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -5229,8 +5229,10 @@ class UserModel extends Gdn_Model implements UserProviderInterface {
         }
 
         $user = $this->getID($userID);
-        $preferences = val('Preferences', $user, []);
-        $landingPages = val('DashboardNav.SectionLandingPages', $preferences, []);
+        $preferences = $user->Preferences ?? [];
+        $landingPages = $preferences['DashboardNav.SectionLandingPages'] ?? [];
+        $sectionPreference = $preferences['DashboardNav.DashboardLandingPage'] ?: '';
+        $sectionReset = false;
 
         // Run through the user's saved landing page per section and if the url matches the passed url,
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -5229,22 +5229,27 @@ public function clearSectionNavigationPreference($url = '', $userID = '', $reset
         }
 
         $user = $this->getID($userID);
-        $preferences = val('Preferences', $user, []);
-        $landingPages = val('DashboardNav.SectionLandingPages', $preferences, []);
+        $preferences = $user->Preferences ?? [];
+        $landingPages = $preferences['DashboardNav.SectionLandingPages'] ?? [];
+        $sectionPreference = $preferences['DashboardNav.DashboardLandingPage'] ?: '';
```
charrondev: What this a conscious choice to make this one `?:` instead of `??`

Is there some reason we might want the looser check?

bleistivt: Not really, I originally wanted to prevent falsy values getting changed unnecessarily, but after looking through the code there isn't really a place where it could be anything other than a string or unset.

PR #9586, opened 2019-11-09
Deprecate markString()

tburry:
We currently use `markString()` to add `<mark>` tags to search results. This presents a few issues:

1. The function only works on plain text, but is not documented as such. It ends up being called incorrectly.
2. We don’t often use plain text at all. Usually, when we say plain text we mean text that has passed through `htmlspecialchars()`. Currently, `markString()` has several edge case bugs with html entities.
3. Since we won’t always use plaintext it will make our functionality appear inconsistent.
4. I suspect there are also hidden XSS vulnerabilities waiting to be uncovered in this function too. We must remain vigilant.

### DIFF
--- a/library/core/functions.deprecated.php
+++ b/library/core/functions.deprecated.php
@@ -565,6 +565,42 @@ if (!function_exists('parseUrl')) {
     }
 }
 
+if (!function_exists('markString')) {
+    /**
+     * Wrap occurrences of {@link $needle} in {@link $haystack} with `<mark>` tags.
+     *
+     * This method explodes {@link $needle} on spaces and returns {@link $haystack} with replacements.
+     *
+     * @param string|array $needle The strings to search for in {@link $haystack}.
+     * @param string $haystack The string to search for replacements.
+     * @return string Returns a marked version of {@link $haystack}.
+     * @deprecated
+     */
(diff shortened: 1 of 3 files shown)


charrondev:
@tburry You're ok with this essentially 100% removing our string marking? Or is that happening somewhere else as well? Do we have an issue to re-introduce it back?

tburry:
That's right. I gave my reasons, but here are some more:

1. Our KB search doesn't do string marking right now anyway, does it? If we keep it in then it won't be long before the inconsistency is brought up by customers.
2. We want to try doing as little HTML manipulation as possible to allow for React auto-escaping.
3. We have wasted so many hours trying to get this functionality to work and we haven't succeeded. It's just a huge time sink.
4. If we are to do this kind of stuff correctly then we would really need to do proper DOM parsing and that's true of much of what we do. As it stands now, that's too much of an effort for the payoff.
5. We cannot properly integrate marking properly with the search engine's stemming algorithm. This leads to more inconsistent results.
6. Google stopped marking their search results years ago.

Really, this is a perfect example of tech debt: We added a feature that was easy to get 80% right, but near impossible to get 100% right.

I added a line about it in the next release notes. We'll see who complains, but really we just don't have the resources.

PR #9588, opened 2019-11-10
Update PHPUnit version

### LINKED ISSUE #9678: Update PHPUnit version in preparation for PHP version bump

tburry:
This PR updated PHPUnit to the max version allowed by PHP ~~7.1~~ 7.2. Keeping our dependencies up-to-date is a basic exercise in codebase hygiene. 

PHPUnit in particular is very aggressive with PHP support which means that we are usually on an old version with outdated dependencies. This makes it more difficult to add other packages as we start to get conflicts. I found this very thing when trying to add some static analysis to our code.

Given I had to update some of our tests in the PR there is the likelihood that some builds will break in our other repos.

Closes #9678.


tests/Library/Vanilla/EmbeddedContent/Embeds/FileEmbedTest.php
```diff
@@ -45,7 +46,7 @@ public function testLegacyDataFormat() {
 JSON;
 
         $oldData = json_decode($oldDataJSON, true);
-        $dataEmbed = new FileEmbedTest($oldData);
-        $this->assertInstanceOf(FileEmbedTest::class, $dataEmbed);
+        $dataEmbed = new FileEmbed($oldData);
```
tburry: @charrondev can you confirm that this is a correct change? The other way looks like a typo.

charrondev: Yup, you’ve got it the right way.

PR #9589, opened 2019-11-10
Fix EmbedService unmet dependency by checking if the Addon is enabled

egarciajulia:
Release/2019.016 introduced new dependencies for the EmbedService.

Classes `DiscussionEmbedFactory` and `CommentEmbedFactory` are part of the `application/vanilla` and they are not available in the Container if that addon is not enabled. As a result, the bootstrap process fails. A clear example of this condition is when you try to install the product in a clean environment.

The fix introduces a conditional where these factories are included only when `application/vanilla` is enabled.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1247,16 +1247,16 @@
         },
         {
             "name": "vanilla/safecurl",
-            "version": "v0.9",
+            "version": "v0.9.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/vanilla/safecurl.git",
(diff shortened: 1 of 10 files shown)


charrondev:
@tburry You merged over my feedback here, but I had made a PR a couple days ago doing this properly https://github.com/vanilla/vanilla/pull/9654

tburry:
I merged because the system was broken, it was blocking me, and this PR fixed it.

charrondev:
Ok. I'm reverting in favour of a different fix in https://github.com/vanilla/vanilla/pull/9654. I should have linked that over here first.

PR #9593, opened 2019-11-11
Updated react dashboard form components and modal for KB forms.

slafleche:
This PR is getting pretty big, so I'm splitting it. This PR is for the new React Add/Edit Knowledge form. **The form is not yet styled and not yet functional.** To test it, you can replace everything in knowledgesettings/index.twig with `<div data-react="knowledge-bases-manage"></div>` and then uncomment https://github.com/vanilla/knowledge/compare/feature%2Fkb-add-edit-modal?expand=1#diff-f61c7c49bb10b2d46567e841b32867fcR30

Requires: https://github.com/vanilla/knowledge/pull/1349

### DIFF
--- a/applications/dashboard/src/scripts/components/DashboardHeaderBlock.tsx
+++ b/applications/dashboard/src/scripts/components/DashboardHeaderBlock.tsx
@@ -4,7 +4,7 @@
  */
 import React from "react";
 import { LeftChevronCompactIcon } from "@library/icons/common";
-import { withRouter, RouteComponentProps } from "react-router";
+import { RouteComponentProps, useHistory } from "react-router";
 import Button from "@library/forms/Button";
 import { ButtonTypes } from "@library/forms/buttonStyles";
 
@@ -15,11 +15,12 @@ interface IProps extends Partial<RouteComponentProps<{}>> {
 }
 
 export function DashboardHeaderBlock(props: IProps) {
+    const history = useHistory();
     return (
         <header className="header-block">
             <div className="title-block">
-                {props.showBackLink && props.history && (
-                    <Button baseClass={ButtonTypes.ICON} aria-label="Return" onClick={props.history.goBack}>
+                {props.showBackLink && history && (
+                    <Button baseClass={ButtonTypes.ICON} aria-label="Return" onClick={history.goBack}>
                         <LeftChevronCompactIcon />
                     </Button>
                 )}
@@ -29,5 +30,3 @@ export function DashboardHeaderBlock(props: IProps) {
         </header>
     );
 }
-
-export default withRouter(DashboardHeaderBlock as any); // Typescript is complaining that RouteComponentProps are optional, but they are if you don't import the default.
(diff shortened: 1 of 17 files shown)


applications/dashboard/src/scripts/components/DashboardHeaderBlock.tsx
```diff
@@ -35,4 +30,4 @@ function DashboardHeaderBlock(props: IProps) {
     );
 }
 
-export default withRouter(DashboardHeaderBlock);
+export default withRouter(DashboardHeaderBlock as any); // Typescript is complaining that RouteComponentProps are optional, but they are if you don't import the default.
```
charrondev: We can resolve this pretty easily actually, now that we have the latest version of react router.

They have a hook.

```ts
    const history = useHistory();
	// Use this instead of through props.
```

Then we can remove the `withRouter()` wrapper and the `extends RouteComponentProps`.

applications/dashboard/src/scripts/forms/DashboardInput.tsx
```diff
@@ -7,18 +7,25 @@ import React from "react";
 import { useFormGroup } from "@dashboard/forms/DashboardFormGroup";
 import classNames from "classnames";
 import { DashboardLabelType } from "@dashboard/forms/DashboardFormLabel";
+import InputTextBlock, { IInputProps, IInputTextProps } from "@library/forms/InputTextBlock";
+import merge from "lodash/merge";
 
-interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {}
+interface IProps extends IInputTextProps {}
+
+// interface IProps extends React.InputHTMLAttributes<HTMLInputElement> {}
```
charrondev:
```suggestion
```

library/src/scripts/forms/InputTextBlock.tsx
```diff
@@ -19,23 +19,25 @@ export enum InputTextBlockBaseClass {
...
-        placeholder?: string;
-        valid?: boolean;
-        required?: boolean;
-        disabled?: boolean;
-        inputRef?: React.RefObject<HTMLInputElement | HTMLTextAreaElement>;
-        multiline?: boolean;
-        maxLength?: number;
-        className?: string;
-    };
+    inputProps: IInputProps;
```
charrondev: We should make this optional and just default to any empty object.

library/src/scripts/forms/RadioButton.tsx
```diff
@@ -31,33 +32,45 @@ interface IState {
 export default function RadioButton(props: IProps) {
     const labelID = useUniqueID("radioButton-label");
     const classes = checkRadioClasses();
-    const { isHorizontal } = props;
+    const { isHorizontal, note } = props;
+
+    const noteID = useMemo(() => {
+        return uniqueIDFromPrefix("radioButtonNote");
+    }, []);
```
charrondev: This can be replaced with `useUniqueID('radioButtonNote')`.

PR #9599, opened 2019-11-12
Update prettier to 1.19.1

charrondev:
**Why?**

- Prettier <= 1.19 breaks on nullish coalescing operators.

**Why are there changes?**

It looks like prettier 1.18 adjusted 2 things in how it formats stuff.

- HOC components will prefer to be on a single line when below the line length.
- Spreads w/ typecasts now get parantheses. Eg. `{(...thing as T)}` vs `{...thing as T}`


slafleche:
Does this also get rid of those warnings we discussed yesterday that are bad for performance?

charrondev:
@slafleche It does not. I have another branch that is WIP to fix those.

Issue #9499, opened 2019-10-29
Error in LocalesApiController: Class 'Locale' not found


R-J:
A fatal error is thrown in LocalesApiController:

```
Fatal Error in LocalesApiController.{closure}();
Class 'Locale' not found
The error occurred on or near: /var/www/vanilla/applications/dashboard/controllers/api/LocalesApiController.php

 99: 
100:         $populate = function (array &$row, array $locales) {
101:             $displayNames = [];
102:             foreach ($locales as $locale) {
103:                 $displayName = \Locale::getDisplayLanguage($row["localeKey"], $locale);
104: 
105:                 // Standardize capitalization
106:                 $displayName = mb_convert_case($displayName, MB_CASE_TITLE);
```

The method `getDisplayLanguage()` indeed is used twice, but not defined in the code base of the release 3.3

R-J:
Thanks, I already found help on the forum! 👍

initvector:
The `Locale` class requires PHP's [intl extension](https://www.php.net/manual/en/book.intl.php) to be enabled. Previous usages of this class were [wrapped in a `class_exists`](https://github.com/vanilla/vanilla/blob/696ae3668c702d583b22525ebc3523cd05118b6e/library/core/class.smarty.php#L92). Vanilla doesn't currently cite the intl extension in its requirements. To truly resolve this issue, I think we'll need to update Vanilla's requirements or find a way to gracefully handle the absence of this class.

charrondev:
We can stub it out in the API now with a class exists check, and can make Intl a requirement in 4.0 with some clear messaging.

linc:
This still causes the 3.3 release to fatally crash any forums without that extension installed on their server when they go to upgrade.

Like mine.

garygreen:
I'm also having the same issue with recent PHP upgrade - we don't use any locales except english, so why MUST this be a requirement for the intl php extension? For those on shared hosting who cannot easily install/enable that module it just seems a bit silly to me.

PR #9604, opened 2019-11-12
Add mysqlnd and intl as PHP requirements

### LINKED ISSUE #7770: Reliance on PDO returning INT column as PHP int
I believe that **mysqlnd** is the only PDO MySql driver that can be relied upon to return an INT column as a PHP int - other drivers may return a PHP string. mysqlnd is not listed as a requirement, but there is code that relies on its behaviour.

If the CategoryID passed to `categoryModel->get()` is a string, it is assumed that it is a slug rather than a primary key, and so the category lookup fails. This is an issue when incrementing discussion counts and 'most recent discussion' links in a category when a new discussion is added, as the CategoryID obtained from the DIscussion record is passed as a string. See also https://open.vanillaforums.com/discussion/35885/post-count-not-updating
 
In this case, either the CategoryID needs to be cast to an int when it's retrieved from the DB (eg, in `CategoryModel->incrementLastDiscussion()` and `CategoryModel->updateLastPost()`), or `categoryModel->get()`  needs changing to use `is_numeric` rather than `is_int`. The latter would break any forum that has all-numeric category slugs -- I don't know if that a possibility?

bleistivt:
Additionally upgrade links to HTTPS.

Fixes #7770 and #9499

### DIFF
--- a/applications/dashboard/controllers/class.setupcontroller.php
+++ b/applications/dashboard/controllers/class.setupcontroller.php
@@ -312,14 +312,15 @@ class SetupController extends DashboardController {
         // - GD
         // - Multibyte String
         // - Internationalization Functions
+        $phpManual = 'https://php.net/manual/en/';
         $requiredPHPModules = [
-            'pdo' => 'You must have the <a href="https://php.net/manual/en/pdo.installation.php">PDO module</a> enabled in PHP in order for Vanilla to connect to your database.',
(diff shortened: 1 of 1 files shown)


charrondev:
You seem to have triggered the PHP linter here w/ a couple of very long lines. It applies the newest coding standard to any touched lines, so you'll need to break those strings up onto multiple lines (or use a heredoc).

Issue #9481, opened 2019-10-25
Translation data API


tburry:
Developers should have an API to load/save translations so that content and translation customizations can be maintained in the database.

## Acceptance Criteria

- Translating is gated by a translator permission that defaults to community managers.

- I should be able to maintain separate resources. A resource is a collection of translation strings.

    - I need to be able to add/edit/delete resources.
    - I should be able to load a resource by a string slug even if we use integer IDs.
    - I should be able to add new string keys.

- I should be able to load all of the translations for a resource.

    - A resource is is represented as a plain string key/value array. There are more complex representations that can be looked into.
    - Something like: `function getStrings(string $resource, string $locale): array`.

- I should be able to load all translations for a record of a particular type and ID within a resource. In this case a resource contains many record.

    - If a string isn't translated then it should fallback to the source language as an option.
    - Something like: `function getRecordStrings(string $resource, string $locale, string $recordType, int $recordID): array`.

- An API endpoint for a resource should have an easy way of fetching and applying translations for its fields on the backend so that the frontend doesn't have to do any special translation handling beyond supplying the correct locale to the API.

    - ```
        // GET /translations/:resource?recordType=knowledgeBase&recordID=123&locale=fr
        [
        	{
        		"resource": ":resource"
        		"key": "knowledgeBase.123.name",
        		"source": "kb",
        		"locale": "fr",
        		"translation": "le kb",
        		"translationStatus": "translated" // later
        	},
        	{
        		"resource": ":resource",
        		"key": "knowledgeBase.123.description",
        		"locale": "fr",
        		"source": "Help me",
        		"translation": "Aide moi"
        		"translationStatus": "translated" // later
        	}
        ]
        
        // PATCH|POST /translations/:resource
        [
          {
            "key": "knowledgeBase.123.description",
            "locale": "fr"
            "translation": "Aide toi"
          }
        ]
        
        OR
        
        // PATCH|POST /translations/:resource
        [
          {
          	"recordType": "knowledgeBase",
          	"recordID": 123,
            "key": "description",
            "locale": "fr"
            "translation": "Aide toi"
          }
        ]
        ```

- When I'm requesting resources I should have the option to return source strings alongside translations.

## Later

- I should be able to generate a basic text file from a resource so that it can be sent to a translation service. There are a few formats, but a specific output format is TBD.
- If I update a source translation I should have some way of knowing that translations need to be updated.

```
// GET /knowledgeBases/123?locale=fr
{
[...]

PR #9605, opened 2019-11-12
Declare translation interface, model and default provider

alex-mtl:
Relates to: #9481


library/Vanilla/Contracts/Site/TranslationProviderInterface.php
```diff
@@ -0,0 +1,52 @@
...
+ * Interface TranslationProviderInterface
+ * @package Vanilla\Contracts\Site
+ */
+interface TranslationProviderInterface {
+    /**
+     * Return true when translation model supports user content translation
+     *
+     * @return bool
+     */
+    public function supportContentTranslation(): bool;
```
charrondev:
```suggestion
    public function supportsContentTranslation(): bool;
```

Grammer

library/Vanilla/Site/TranslationProvider.php
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public function supportContentTranslation(): bool {
+        return false;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function translate(string $propertyKey, $sourceValue): string {
+        return Gdn::translate($propertyKey, $sourceValue);
```
charrondev:
```suggestion
        return $this->locale->translate($propertyKey, $sourceValue);
```

We should be able to access the same method just by dependency injecting `LocaleInterface`.

Issue #4520, opened 2016-09-22
Empty discussion title should be fixed in model, not views


DaazKu:
We currently fix empty discussions names in "some" view which is problematic everywhere else.
We should fix this directly in the model when we fetch them.

See https://github.com/vanilla/vanilla/pull/4504#issuecomment-248943636

PR #9612, opened 2019-11-13
Correct empty discussion titles at the model level

tburry:
The discussions APIv2 was failing with empty discussion titles. This fix is primarily for that defect, but also includes the enhancement to correct the title at the model level so that other code that may benefit from it.

This change also renames the translation for untitled discussions to a simple “(Untitled)” that is more generic for other types of data.


bleistivt:
this closes #4520

PR #9617, opened 2019-11-14
update release version

dannimajdalani:
update release version.

### DIFF
--- a/environment.php
+++ b/environment.php
@@ -22,7 +22,7 @@ if (!defined('APPLICATION')) {
 if (!defined('APPLICATION_VERSION')) {
     // Rules for the versioning
     // {OSS version}-{Cloud release version}-{? SNAPSHOT if it's a dev build}
-    define('APPLICATION_VERSION', '3.3-2019.016-SNAPSHOT');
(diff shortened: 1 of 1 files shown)


environment.php
```diff
@@ -22,7 +22,7 @@
 if (!defined('APPLICATION_VERSION')) {
     // Rules for the versioning
     // {OSS version}-{Cloud release version}-{? SNAPSHOT if it's a dev build}
-    define('APPLICATION_VERSION', '3.3-2019.016-SNAPSHOT');
+    define('APPLICATION_VERSION', '3.3-2019.017');
```
initvector: This is ultimately going to be 3.4, so let's go ahead and bump that in here too.

```suggestion
    define('APPLICATION_VERSION', '3.4-2019.017');
```

dannimajdalani: Shouldn't this track the current release version?

initvector: This branch will be released as 3.4, which is why it makes sense to me to increment the version to 3.4 here. To me, it's a little more confusing to indicate 017 is 3.3, when 3.4 will be the exact same branch.

PR #9627, opened 2019-11-14
Backport: Fix fatal error when rendering unknown format

initvector:
Backporting #9625 

> `FormatInterface` was updated with #9554. Most of core Vanilla's formats seem to have been updated, with the exception of `NotFoundFormat`. The method is added here.


initvector:
Jumped the gun a little bit here. 😄

PR #9634, opened 2019-11-15
Add state token to OAuth2 registrations

nervoustwit:
This PR will close this issue: https://github.com/vanilla/support/issues/987

We have recently added the State Token to Authorization request. Most OAuth2 providers do not use a separate URL for users to register so we weren't adding the Token to the Registration request. 

However, some customers ***do*** have a register button and would like to return the user back to the forum after registering them. This requires a State Token. This PR will modify the Registration URL by passing it through the same authorization redirect endpoint as the Authorization URL to generate and save  the State Token and attach it to the request.

### Testing
 - Turn on the OAuth2 Plugin.
 - Put a valid URL in the RegisterURL field (you could put the same URL as the Authorization).
 - Click on Register.
 - Observe that the State Token is being added to the request.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -24,9 +24,6 @@
  */
 class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
 
-    /** @var string context is registration workflow */
-    const CONTEXT_REGISTER = 'register';
-
     /** @var string token provided by authenticator  */
     protected $accessToken;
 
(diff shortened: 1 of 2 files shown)


library/core/class.oauth2.php
```diff
@@ -349,12 +352,14 @@ protected function getSettingsFormFields() {
      * Redirect to the OAuth redirect page with a verification nonce.
      *
      * @param EntryController $sender The controller initiating the request.
-     * @param  string $state The state to pass along the OAuth2 flow.
+     * @param string $state The state to pass along the OAuth2 flow.
+     * @param string $context The context of the redirect either register or sign in.
      */
-    public function entryRedirectEndpoint(\EntryController $sender, $state = '') {
+    public function entryRedirectEndpoint(\EntryController $sender, $state = '', $context = null) {
```
initvector: Is it safe to slap a type on this new `$context` parameter?

library/core/class.oauth2.php
```diff
@@ -455,11 +471,9 @@ public function authorizeUri($state = []) {
      *
      * @return string Endpoint of the provider.
      */
-    protected function realAuthorizeUri(array $state = []): string {
+    protected function realAuthorizeUri(string $uri, array $state = []): string {
```
initvector: Won't adding a new parameter in front of an existing parameter break backwards compatibility with classes that have extended this one?

nervoustwit: This method has only been available for a 5 months. We have not extended it in any of our plugins. Do you think it could be a risk for open source?

initvector: That is a very real possibility, but it's usually a good habit to avoid this type of thing, in general.

library/core/class.oauth2.php
tburry: I would rather you just put the context inside the state.

library/core/class.oauth2.php
tburry: I agree with Ryan. Usually the refactor is to do something like this.

1. Make a `realAuthorizeUriInternal` method with the signature like you have here (please mark this final).
2. Make `realAuthorizeUri` a 1-liner that calls `realAuthorizeUriInternal`.
3. Make another 1-liner `realRegisterUri` that calls `realAuthorizeUriInternal`.

PR #9639, opened 2019-11-16
Filter the rich format on API output

tburry:


library/Vanilla/Formatting/Formats/RichFormat.php
```diff
@@ -271,4 +271,53 @@ private function logBadInput(string $input) {
...
+    }
+
+    /**
+     * Strip sensitive user info from a rich string and rewrite it.
+     *
+     * @param string $input The rich text input.
+     * @return string The string.
+     */
+    private static function stripSensitiveInfoRich(string $input): string {
+        if (strpos($input, "password") === false) {
```
kaecyra: Should this be `stripos` ?

tburry: In this case I think it's okay because APIv2 camel cases everything.

PR #9642, opened 2019-11-17
Add the POST /api/v2/tokens/oauth endpoint

tburry:
This endpoint allows the user to exchange an OAuth2 SSO token for a Vanilla access token.

This is a fairly large and seemingly daunting PR, but most of its code is YAML and tests. The major changes to our core are the following:

1. `UserModel::connect()` has been given an option not to sync existing users upon connection.
2. `Gdn_OAuth` has added the `POST /api/v2/tokens/oauth` endpoint.

Closes https://github.com/vanilla/internal/issues/1976.

## Caveats

- I added a workaround for the problem of multiple addons that extend `Gdn_OAuth`. I can walk through the code with reviewers as it may seem odd just on inspection.
- Any changes to the `Gdn_OAuth` class should be carefully looked at since we have so many subclasses. Furthermore, some subclasses may not be compatible with this endpoint if they implement `getProfile()` in an odd way.
- I am opting out of some dependency injection in `Gdn_OAuth` which is typically an anti-pattern. However, since the plugin is always constructed every request I would rather not construct objects that are only used in a tiny fraction of requests. Since we don't have good alternatives I'm just making the classes through the global container.

## Testing Suggestion

The addon has unit tests for most of its code coverage. To test manually you can do the following:

1. Enable and configure the OAuth addon.
2. Get an access token. You can do this by putting a breakpoint in the code to where the access token is issued and copy it. Some providers will let you issue yourself an access token yourself.
3. Use that access token in Postman or something like that to test the endpoint.


applications/dashboard/models/class.usermodel.php
```diff
@@ -872,6 +872,14 @@ public function syncUser($currentUser, $newUser, $force = false, $isTrustedProvi
      */
     public function connect($uniqueID, $providerKey, $userData, $options = []) {
         trace('UserModel->Connect()');
+
```
tburry: I've just moved some of the defaults up top to make the options easier to find.

applications/dashboard/models/class.usermodel.php
```diff
@@ -897,7 +905,9 @@ public function connect($uniqueID, $providerKey, $userData, $options = []) {
 
         if ($userID) {
             // Save the user.
-            $this->syncUser($userID, $userData, false, $isTrustedProvider);
+            if ($options['SyncExisting']) {
```
tburry: This is a new option to this method.

library/core/class.oauth2.php
```diff
@@ -413,7 +436,11 @@ public function settingsEndpoint($sender, $args) {
 
         $formFields = $formFields + $this->getSettingsFormFields();
 
-        $formFields['IsDefault'] = ['LabelCode' => 'Make this connection your default signin method.', 'Control' => 'checkbox'];
+        $formFields['AllowAccessTokens'] = [
+            'LabelCode' => 'Allow this connection to issue API access tokens.',
+            'Control' => 'toggle'
+        ];
+        $formFields['IsDefault'] = ['LabelCode' => 'Make this connection your default signin method.', 'Control' => 'toggle'];
```
tburry: I changed this to a toggle because the checkbox repeats the label twice.

library/core/class.oauth2.php
```diff
@@ -1045,4 +1079,156 @@ protected function decodeState(string $state): array {
...
+    final private function sso(array $payload): int {
+        unset($payload['UserID']); // safety precaution due to Gdn_UserModel::connect() behaviour
+
+        /* @var \UserModel $userModel */
+        $userModel = $this->container->get(\UserModel::class);
+
+        $userID = $userModel->connect(
+            $payload['UniqueID'] ?? '',
+            static::PROVIDER_KEY,
+            $payload,
```
nervoustwit: This const does not exist in the Oauth2 class. The provider key was never a constant in this context because we hoped to have multiple OAuth2 connections at the same time.

tburry: Great catch!

nervoustwit:
This is great. I tested it for retrieving the accesstoken of an existing user, for creating a new user and getting his accesstoken and retrieving the token of user connecting through a custom subclass.

For subclasses it is necessary that the subclass have the `AllowAccessTokens` in the provider for it to work.

tburry:
@nervoustwit I re-requested your review, but I really just changed that one point you made.

PR #9652, opened 2019-11-18
Refactored categoryUrl input to manage long urls

slafleche:
Refactored the styles for the categoryUrl field. The elements are now all "inline", so that they wrap with the text.

Closes: https://github.com/vanilla/support/issues/171


charrondev:
2 Notes

- There's a commented line in the PR
- The URL seems to be duplicated:
image

slafleche:
> 2 Notes
> 
> * There's a commented line in the PR
> * The URL seems to be duplicated:
>   image

Fixed

PR #9654, opened 2019-11-18
Fix forum classes being referenced in the core library

charrondev:
Fixes https://github.com/vanilla/dev-inter-ops/issues/36

The Vanilla application is optional, so I've removed references to it from the `EmbedService`.

- `DiscussionEmbedFactory` and `CommentEmbedFactory` moved to `applications/vanilla`.
- Added a `bootstrap.php` file with registration of these factories only when the forum app is enabled.
- Moved the tests to the forum (vanilla) app.
- Removed a reference to the `EmbedManager` in the main bootstrap. This class doesn't exist anymore for a few releases, and was removed in favour of the `EmbedService`.

### DIFF
--- a/library/Vanilla/EmbeddedContent/Factories/AbstractOwnSiteEmbedFactory.php
+++ b/library/Vanilla/EmbeddedContent/Factories/AbstractOwnSiteEmbedFactory.php
@@ -61,7 +61,7 @@ abstract class AbstractOwnSiteEmbedFactory extends AbstractEmbedFactory {
             }
         }
         $siteSectionRegex = count($allowedSlugs) > 0 ?
-            '(' . implode('|', $allowedSlugs) . ')'
+            '(' . implode('|', $allowedSlugs) . ')?'
             : '';
         $root = $this->request->getAssetRoot() . $siteSectionRegex;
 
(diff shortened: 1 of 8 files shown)


cchabilall83:
when i'm trying to embed a discussion I get a 400 error

Screen Shot 2019-11-19 at 2 21 47 PM

charrondev:
@cchabilall83 I've fixed it (although it looked like it was an issue before).

The regex was setup so that site sections were required in matching urls (if you had more than 1) or it would fallback to a scrape. We've done hardening that prevents scraping of internal IP addresses so we get the 400 error.

charrondev:
After https://github.com/vanilla/vanilla/pull/9589 I had to add a revert for it (it was an alternative, but worse fix for this).

library/Vanilla/EmbeddedContent/Factories/AbstractOwnSiteEmbedFactory.php
```diff
@@ -61,7 +61,7 @@ protected function getRegexRoot(): string {
             }
         }
         $siteSectionRegex = count($allowedSlugs) > 0 ?
-            '(' . implode('|', $allowedSlugs) . ')'
```
tburry: @charrondev  this should be run through `preg_quote`.

Issue #9440, opened 2019-10-19
Rich Editor slows down to zero when there is a code paragraph


R-J:
On Windows with Firefox (not tested if this also happens on other browsers)
I have written a small post, three lines of unformatted text and 12 lines of code, and the Rich Editor slows down so that my browser asks me if I want to stop the execution of the script.
Here is the resulting message from the console:

```
Error: Script terminated by timeout at:
R/e.className<@https://open.vanillaforums.com/dist/forum/async~highlightJs.min.js?84c934150f3d85eceda1:1:13861
R@https://open.vanillaforums.com/dist/forum/async~highlightJs.min.js?84c934150f3d85eceda1:1:13974
e/<@https://open.vanillaforums.com/dist/forum/shared.min.js?h=5daa022f:375:352
l@https://open.vanillaforums.com/dist/forum/vendors.min.js?h=5daa022f:58:7421
u/i._invoke</<@https://open.vanillaforums.com/dist/forum/vendors.min.js?h=5daa022f:58:7209
S/</e[t]@https://open.vanillaforums.com/dist/forum/vendors.min.js?h=5daa022f:58:7843
r@https://open.vanillaforums.com/dist/forum/vendors.min.js?h=5daa022f:6:58597
s@https://open.vanillaforums.com/dist/forum/vendors.min.js?h=5daa022f:6:58799

```

R-J:
Some other user voices:
[ps.. i'm noticing that the text entry box, as shown here.. eats a hell of a lot of cpu while typing on firefox.](https://open.vanillaforums.com/discussion/comment/258926/#Comment_258926)

[Off topic- this forum is hanging very badly, I can't write anything after 3-4 lines, the page becomes unresponsive for several minutes.](https://open.vanillaforums.com/discussion/comment/258950/#Comment_258950)

Really, really annoying. I prefer using any other text editor and paste the text into the Rich Editor => I avoid it whenever possible.

PR #9659, opened 2019-11-19
Fix editor syntax highlighting moving cursor and spiking CPU

charrondev:
Fixes https://github.com/vanilla/vanilla/issues/9440
Fixes https://github.com/vanilla/knowledge/issues/1347

There were a few issues causing the cursor to jump around and CPU usage to spike.

1. When initializing a synchronized editor `useInitialValue()` was re-initializing too often due to some faulty logic. Additionally it was not clearing history or setting the initial selection. This could cause the selection to jump when first editing a code block.

2. `SyntaxModule` has some logic to preserve the selection after a highlight. Unfortunately we made `highlightText` asynchronous for a dynamic import. This method **has to work synchronously**.
    - If we awaited the async value we would likely be setting back the old selection value or content.

3. Our editor UI synchronizer was listening to `Quill.sources.SILENT` events. This was a mistake, because being silent, they should not cause any UI updates.

## The fixes

- Update `useInitialValue()` to clear history, and reset only when the `reinitialize` value has changed from `false|undefined` -> `true` (or starts as true).
- Create a synchronous highlighter function `highlightTextSync()`.
    - Uses an already downloaded highlightJS to highlight code synchronously.
    - Starts loading highlightJS if it's not loaded yet.
    - Returns null if the library is not loaded.
- Update `CodeBlockBlot.highlight()` to have a synchronous control flow.
    - Makes no changes while the highlight function returns null.
- Update the editor UI synchronizer to ignore silent updates.

## Testing

1. Load up rich editor on the forum or knowledge base.
2. Create a code block.
3. Start putting a bunch of text in it.
4. Nothing should "jump around".

I also did some testing in firefox to make sure https://github.com/vanilla/vanilla/issues/9440 was resolved.

### DIFF
--- a/library/src/scripts/content/code.ts
+++ b/library/src/scripts/content/code.ts
@@ -5,6 +5,7 @@
 
 import { onContent } from "@library/utility/appUtils";
 import { globalVariables } from "@library/styles/globalStyleVars";
+import { ILoadable, LoadStatus } from "@library/@types/api/core";
 
 type HLJS = typeof import("highlight.js");
 
@@ -19,6 +20,8 @@ export async function highlightCodeBlocks(domNode: HTMLElement = document.body) 
     blocks.forEach(hljs.highlightBlock);
 }
 
+let hljsCache: HLJS | null = null;
+
 /**
  * Highlight some text and return HTML for it.
  *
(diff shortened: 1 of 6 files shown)


charrondev:
Marking as hold while I investigate another possible improvement.

charrondev:
I made some pretty signficant changes here after the 2 reviews, so I'm re-requesting.

- Syntax highlighting is now triggered only by text changes and is not on a timer/loop anymore.
- Fixed a bug where pressing enter in the middle of a code block would jump you to the end.

Issue #8314, opened 2019-02-06
Rich Editor: collapsable code blocks


adrianspeyer:
**Describe the Feature**

When dealing with more than 10 lines of code, it would be interesting to make it collapsible (like spoiler), so the discussion post or comment is not so long

**User Story**

Scott from RapidMiner deals with code over 100-200 lines, and is needing to use combo of code and spoiler tags, but it's hard to teach the user community to do the same.

**Acceptance Criteria**

A list of the minimum requirements that a pull request must meet in order to call the feature complete.

Set a limit on lines of code or blockquote before it's collapsible

charrondev:
So I will say there is a relatively easy workaround that we have implemented in our new base themes/keystone etc. and that is setting a max-height on the code block element using CSS.

That doesn't invalidate this as a feature, (which involve this max-height + a button to remove the limit), but they aren't stuck without having a limit.

adrianspeyer:
Good solution if on keystone, so thanks for that. In this case, theme change is not an option.

charrondev:
It seems I was mistaken. Keystone does not implement this at all, so this actually does become a little bit more important. I'm thinking it's open source forum that properly implements this then.

tburry:
In general, I think a better approach to supporting hundreds of lines of code would be to support gist embeds. Showing small code snippets inline is nice, but when you need to type a bunch of stuff a dedicated IDE is better.

PR #9661, opened 2019-11-19
Apply a max-height to code blocks

charrondev:
Closes https://github.com/vanilla/vanilla/issues/8314

- Applies a max-height to code blocks w/ overflow scrolling. The variable defaults to 600, which seems to be a sane maximum.
- This max-height is similar to what Github uses.
- The code block styles seem to be split into 2 places. Not sure what's up with that. I applied to both. @slafleche Maybe you can clarify?

### DIFF
--- a/library/src/scripts/content/userContentStyles.ts
+++ b/library/src/scripts/content/userContentStyles.ts
@@ -61,6 +61,7 @@ export const userContentVariables = useThemeCache(() => {
         lineHeight: 1.45,
         fg: blocks.fg,
         bg: blocks.bg,
+        maxHeight: 500,
     });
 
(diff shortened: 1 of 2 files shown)


slafleche:
Why do we want to do this? If it's too long, wouldn't it better to maybe have toggle?

I think we cannot, have this in edit mode. The position of the paragraph menu does not respect the scrolling of the code block.

slafleche:
@charrondev the 2 styles sheets are for edit or article mode.

charrondev:
I missed the issue in edit mode. I don't particularly feel like handling it now, so I'm just going to close this out.

PR #9665, opened 2019-11-20
Fix alt install tests

### LINKED ISSUE #9144: Update reports successful when errors encountered
*Note: My PR is for a bug fix, but should also close this issue. I sort of hijacked this issue in order to get my PR reviewed.*

If errors are encountered during a utility-update, the site still reports a "successfully updated" response, unless debug mode is enabled. This appears to be due to [`UpdateModel::runStructure`](https://github.com/vanilla/vanilla/blob/16f43e0e179d22b6c5a804b419cdda8b61e0a63a/applications/dashboard/models/class.updatemodel.php#L660) catching all exceptions from addon structure methods and only re-throwing them if the site's debug mode is enabled. This is problematic, because beyond simply hiding specifics of the errors, the response indicates everything was successful.

tburry:
The test was failing due to our session restoring functionality during the alt install. That has been checked as a special case.

In addition to the fix I added 500 response codes when utility/update fails to hopefully help troubleshoot errors more. This should close #9144.

### DIFF
--- a/applications/dashboard/models/class.updatemodel.php
+++ b/applications/dashboard/models/class.updatemodel.php
@@ -667,7 +667,7 @@ class UpdateModel extends Gdn_Model {
         } finally {
             if ($userID && $userID !== Gdn::session()->UserID) {
                 Gdn::session()->start($userID, false, false);
-            } elseif (!$userID) {
+            } elseif (!$userID && null !== c('Garden.Installed', false)) {
                 Gdn::session()->end();
             }
         }
--- a/applications/dashboard/models/class.updatemodel.php
+++ b/applications/dashboard/models/class.updatemodel.php
@@ -668,6 +668,8 @@ class UpdateModel extends Gdn_Model {
             if ($userID && $userID !== Gdn::session()->UserID) {
                 Gdn::session()->start($userID, false, false);
             } elseif (!$userID && null !== c('Garden.Installed', false)) {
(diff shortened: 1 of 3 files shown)


charrondev:
Re: failing build here https://app.circleci.com/jobs/github/vanilla/vanilla/24109

applications/dashboard/models/class.updatemodel.php
```diff
@@ -667,7 +667,7 @@ public function runStructure($captureOnly = false) {
         } finally {
             if ($userID && $userID !== Gdn::session()->UserID) {
                 Gdn::session()->start($userID, false, false);
-            } elseif (!$userID) {
+            } elseif (!$userID && null !== c('Garden.Installed', false)) {
```
charrondev: I don't really understand this statement. Looking at it, this should never be true.

`Garden.Installed` will always be `true | false` because of the fallback, and null is never strictly to either of those values.

tburry: We have an alternative way of installing Vanilla in which case `Garden.Installed === null`. These are the tests for that method.

charrondev:
```suggestion
            } elseif (!$userID && null !== c('Garden.Installed', false)) {
            // Vanilla has an alternate install method where this config value is set to null
            // When this using this alternate install method we don't have authenticators configured and can't end the session.
```

Issue #8713, opened 2019-04-17
Imports must not allow blank discussion titles


linc:
Allowing blank discussion titles via import breaks the APIv2 validation schema, blocking access to the resource entirely. The importer must replace any blank discussion titles with default text.

tburry:
The API can add a title programmatically before validating the output.

dannimajdalani:
Fixed in https://github.com/vanilla/vanilla/pull/9612

PR #9666, opened 2019-11-20
[Apiv2] Discussions: add discussion title before validating schema ou…

dannimajdalani:
closes https://github.com/vanilla/vanilla/issues/8713

### Details

When an imported discussion does not have a title, api/v2/discussion fails with the following error:
Screen Shot 2019-11-20 at 12 16 56 PM
we are doing the validation here [DiscussionsApiController.php#L588](https://github.com/vanilla/vanilla/blob/81535af8c0c7507afb754a5579a72cb4960781af/applications/vanilla/controllers/api/DiscussionsApiController.php#L588)


tburry:
Can you check my PR here: https://github.com/vanilla/vanilla/pull/9612 and make sure this one is still needed?

PR #9691, opened 2019-11-25
Add SEO attribute values to rich embeds

initvector:
Rich embed `rel` attributes are being updated with values to improve SEO.


charrondev:
Tests are probably failing due to some assertions that we have a specific HTML output for these. You'll need to update those.

tests/Library/Vanilla/EmbeddedContent/Embeds/QuoteEmbedFilterTest.php
```diff
@@ -117,7 +117,7 @@ public function testFullContentFiltering() {
     class="js-embed embedResponsive"
     data-embedJson="{&quot;url&quot;:&quot;http:\/\/test.com&quot;,&quot;embedType&quot;:&quot;testEmbedType&quot;}"
 >
-    <a href="http://test.com">
+    <a href="http://test.com" rel="nofollow noreferrer ugc">
```
charrondev: This doesn't seem right.

Quote embed shouldn't have these at all (because they refer to our own site). That's part of why they have a separate view `QuoteEmbed.twig`.

initvector: There's definitely some pre-existing issue in that test. The proper Twig view is used on the site, but not in this test. I've reverted the "fix" to expose the issue again.

initvector: To recap: those attribute values _should_ be there, because we're evaluating the inner quote, not the outer quote. We'd expect the other view to be used for the outer quote. The inner quote is purely the quotes internal content.

Issue #9664, opened 2019-11-19
Content-Security-Policy header is blank


kaecyra:
I visited a few pages on our infrastructure, including some knowledge base pages, expecting to see a Content-Security-Policy header being returned. Instead, it seems to be set to a blank string, despite code existing in GDN_Controller that seems to try to set it. Not sure what's going on.

https://github.com/vanilla/vanilla/blob/master/library/core/class.controller.php#L286

charrondev:
Just to be clear on this the expected policies are:

**Gdn_Controller pages**

Should have a `frame-ancestors` policy **only**.

**Vanilla\Web\Page pages**

Should have

- `frame-ancestors` allowing trusted domains, and Embed.RemoteUrl, etc when embedding is enabled.
- `script-source` (w/ nonce) and the following exclusions
  - Our own scripts
  - embed whitelist
  - trusted domains
  - google fonts / analytics

charrondev:
Followup:

The bug is here: https://github.com/vanilla/vanilla/blob/89caa5cff5b6755cfed236d54d0348b58d6101d7/library/Vanilla/Web/ContentSecurityPolicy/ContentSecurityPolicyModel.php#L66

- It should check if `$filter` is all, not `$directive`

tburry:
I would like an RCA of how this regression happened in addition to tests for the regression aspects of it.

charrondev:
Investigation will be put into the next sprint. Issue here https://github.com/vanilla/support/issues/1101

PR #9692, opened 2019-11-25
Fix CSP headers

alex-mtl:
Fix ContentSecurityPolicyModel  wasn't generating proper headers.

Add unit test.

Closes: #9664


alex-mtl:
> * Applying even when embedding is disabled.

Your request conflicts with our implementation:
https://github.com/vanilla/vanilla/blob/380a0fe4594980a1f54c9907a9ce06c5d100c151/library/Vanilla/Web/ContentSecurityPolicy/DefaultContentSecurityPolicyProvider.php#L34

PR #9698, opened 2019-11-27
Add translation debug configuration

cchabilall83:
This allows developers to see if the content strings have been in translated in their respective locales.  Changes in the localApiController check if the en _localkey_ exists in the enabled local packs, if not add it, this allow for the tx-source pack to be read as the default en.

Steps to enable

1. Symlink tx-source dir from the locales repo to your vanilla/locales dir
2.  Add following keys to your config
`$Configuration['EnabledLocales']['tx-source'] = 'en';`
`$Configuration['TranslationDebug'] = true;`

requires https://github.com/vanilla/locales/pull/237

## Other Notes

- Static calls to `t()` have never worked. They now have a warning if they are called before the translations are setup. All existing calls have been updated.
- The test config had to be updated to properly configure the test environmental variable.


packages/vanilla-i18n/src/translationStore.ts
```diff
@@ -41,6 +59,11 @@ export function translate(str: string, defaultTranslation?: string): string {
         return translationStore[str];
     }
 
+    // report any untranslated strings when in translation debug mode
+    if (defaultTranslation === undefined && translationDebug(getMeta("context.translationDebug"))) {
```
charrondev: Please use the `internalTranslationDebugValue` here. These packages should never have a dependency on library.

packages/vanilla-i18n/src/translationStore.ts
```diff
@@ -1,3 +1,5 @@
+import { getMeta } from "@vanilla/library/src/scripts/utility/appUtils";
```
charrondev: This import should be removed. See below.

library/src/scripts/bootstrap.ts
```diff
@@ -20,6 +21,9 @@ import { bootstrapLocales } from "@library/locales/localeBootstrap";
 const debugValue = getMeta("context.debug", getMeta("debug", false));
 debug(debugValue);
 
+const translationDebugValue = getMeta("context.translationDebug", getMeta("translationDebug", false));
```
charrondev: What's the reason for the nesting fallback? This value should always be in the same place.

packages/vanilla-i18n/src/translationStore.ts
cchabilall83: removed the getMeta call

packages/vanilla-i18n/src/translationStore.ts
cchabilall83: done

library/src/scripts/bootstrap.ts
cchabilall83: removed second getMeta call

packages/vanilla-i18n/src/translationStore.ts
```diff
@@ -41,6 +57,11 @@ export function translate(str: string, defaultTranslation?: string): string {
         return translationStore[str];
     }
 
+    // report any untranslated strings when in translation debug mode
+    if (defaultTranslation === undefined && translationDebug()) {
```
charrondev: I'd even go a bit further. I think we should wrap the result even with the "default" provided.

```js
    let result = defaultTranslation !== undefined ? defaultTranslation : str;

    if (translationDebug()) {
    	result = "☢️☢️☢️" + result + "☢️☢️☢️";
    }
	return result;
```

PR #9715, opened 2019-11-28
Fix BBcode spacing issue

cchabilall83:
~Removes the LIBXML_NOEMPTYTAG from HtmlFormat::cleanupEmbeds.  This was causing and extra <br> to appended to the body content causing formatting issues.  Tests will be written ensure this doesn't occur.~

~https://www.php.net/manual/en/libxml.constants.php~
~` LIBXML_NOEMPTYTAG (integer) Expand empty tags (e.g. <br/> to <br></br>)`~

filter out `</br>` from content, this breaks formatting.

closes https://github.com/vanilla/support/issues/1106


alex-mtl:
Many tests are failing. I think that is a signal that probably we can not go with this fix.
But not sure....

Issue #9702, opened 2019-11-27
Embed display regression


tburry:
The embeds have a bad top padding and their text is cut off. See this screenshot:

<img width="758" alt="Accessibility_in_Product_Design_-_Vanilla_Success" src="https://user-images.githubusercontent.com/97432/69744597-79dd9300-110e-11ea-8bbf-42e920c28422.png">

PR #9716, opened 2019-11-28
Fix/Adjusting margin-top for embed header

NeenaDev:
Fix - https://github.com/vanilla/vanilla/issues/9702

Adjust margin-top for embed header

### DIFF
--- a/library/src/scripts/styles/globalStyleVars.ts
+++ b/library/src/scripts/styles/globalStyleVars.ts
@@ -172,7 +172,7 @@ export const globalVariables = useThemeCache(() => {
         },
         alignment: {
             headings: {
-                capitalLetterRatio: 0.715, // Calibrated for Open Sans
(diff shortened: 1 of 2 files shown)


library/src/scripts/styles/globalStyleVars.ts
```diff
@@ -172,7 +172,7 @@ export const globalVariables = useThemeCache(() => {
         },
         alignment: {
             headings: {
-                capitalLetterRatio: 0.715, // Calibrated for Open Sans
+                capitalLetterRatio: 0.73, // Calibrated for Open Sans
```
charrondev: @slafleche Did you suggest this change? It seems a bit arbitrary and I'd like to have some idea fo what adjusting our global offset is meant to accomplish here.

slafleche: I did. There's often extra space between the line height and the actual text. The text was very slightly getting cut off (about 1 px).

PR #9718, opened 2019-11-28
Fixed overflow issue in rich quotes

slafleche:
Fixes bbcode alignment. 

I also noticed that in some themes, the quotes were overflowing, so I put a box-sizing on the content.

For https://github.com/vanilla/support/issues/1108


slafleche:
@charrondev I added the bb code alignment to Storybook, but the build is broken. It seems like one of our dependencies is broken. I tried to upgrade, but with no luck. Can you take a look?

PR #9719, opened 2019-11-28
Revert "Fix BBcode spacing issue"

slafleche:
Reverts vanilla/vanilla#9715


cchabilall83:
We will go with fix, so no revert is needed
see https://github.com/vanilla/vanilla/pull/9715

PR #9721, opened 2019-11-28
Fix confusing message and workflow in SSO connections.

nervoustwit:
After discovering a bug in which user's could not input a different user name than the one that was passed over SSO (see https://github.com/vanilla/support/issues/906) we decided to revisit the workflow of the Connect page. This PR will change a little bit of the wording and clean up the logic of the Connect method in the Entry Controller. 

This PR will: 
 - Create 3 distinct code blocks for the 3 distinct scenarios of connection (Name exists, Email exists, create a new User).
 - Change some of the wording on the Connect page. (see images below).
 - Not overwrite the ConnectName with the Name passed over SSO (which was causing the bug to allow user's to change their name if it existed.

## Before:
Screen Shot 2019-11-28 at 4 41 19 PM

## After
Screen Shot 2019-11-28 at 4 39 50 PM

### DIFF
--- a/applications/dashboard/views/entry/connect.php
+++ b/applications/dashboard/views/entry/connect.php
@@ -123,7 +123,7 @@ if (!$hasUserID) {
                     if (count($ExistingUsers) >= 1 && !$NoConnectName){
                         echo $this->Form->label('Username', 'ConnectName');
                         echo \Gdn::translate('ConnectWithExistingUser', 'One or more users with your name already exist, would you like to connect as them?');
-                        $connectNameMessage = (!$allowConnect) ? '' : ' <span class="FinePrint">('.\Gdn::translate('Requires a password').')</span>';
+                        $connectNameMessage = (!$allowConnect) ? '' : ' <span class="FinePrint">'.\Gdn::translate('Requires a password').'</span>';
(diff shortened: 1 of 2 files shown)


applications/dashboard/views/entry/connect.php
```diff
@@ -117,24 +117,31 @@
...
-                            foreach ($ExistingUsers as $Row) {
-                                echo wrap($this->Form->radio('UserSelect', $Row['Name'], ['value' => $Row['UserID']]), 'div');
-                            }
-                            echo wrap($this->Form->radio('UserSelect', t('Other'), ['value' => 'other']), 'div');
+                    // Found User's Name in GDN_User.
+                    if (count($ExistingUsers) >= 1 && !$NoConnectName){
+                        echo $this->Form->label('Username', 'ConnectName');
+                        $connectInstructions = \Gdn::translate('One or more users with your name already exist, would you like to connect as them?');
+                        echo $connectInstructions;
+                        $connectNameMessage = (!$allowConnect) ? '' : \Gdn::translate(' <span class="FinePrint">(Requires a password)</span>');
```
tburry: Jut make the translation string "(Requires a password)" and wrap it in the HTML.

PR #9723, opened 2019-11-29
Fix comment watch status in visited discussions

initvector:
Since #9488, visiting a new discussion fails to create a new "watch" status. This status is used by Vanilla to determine which discussions have new content since the user last viewed them. Existing "watch" status records are still being updated. Only _new_ discussion "watch" statuses are being missed. This was caused by mistakenly removing the insert routine for these statuses, along with the archiving code it was adjacent to.

The code for inserting new discussion "watch" statuses has been restored and a simple (yet verbose) test has been added to ensure the relevant records continue to be added and updated.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -861,6 +861,18 @@ class CommentModel extends Gdn_Model {
                     ]
                 );
             }
+        } else {
+            // Insert watch data.
+            $this->SQL->options('Ignore', true);
+            $this->SQL->insert(
+                'UserDiscussion',
+                [
+                    'UserID' => $userID,
+                    'DiscussionID' => $discussion->DiscussionID,
+                    'CountComments' => $countWatch,
+                    'DateLastViewed' => Gdn_Format::toDateTime()
+                ]
+            );
         }
 
         /**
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.commentmodel.php
```diff
@@ -861,6 +861,18 @@ public function setWatch($discussion, $limit, $offset, $totalComments) {
...
+        } else {
+            // Insert watch data.
+            $this->SQL->options('Ignore', true);
+            $this->SQL->insert(
+                'UserDiscussion',
+                [
+                    'UserID' => $userID,
+                    'DiscussionID' => $discussion->DiscussionID,
+                    'CountComments' => $countWatch,
+                    'DateLastViewed' => Gdn_Format::toDateTime()
```
charrondev: Any particular reason we opted for using this deprecated method? Suggested alternative now is `DateTimeFormatter::timeStampToDateTime()`.

initvector: > Any particular reason we opted for using this deprecated method?

This was restoring erroneously-deleted code, verbatim. However, since I have technically touched the code, I have updated the deprecated method usage.

PR #9741, opened 2019-12-02
Added config setting to bypass js-focus-visible plugin

slafleche:
Added config (true by default) to use the focus visible plugin. Some clients may not want this style and it can now be bypassed with a config.

Note that there are more styles in our core theme that is problematic. For the moment if a client wants it, they will need to style the outline. They can do a feature request to remove the core styles. It's not so easy, since we need to worry about backwards compatibility. 

@irina-vanillaforums

### DIFF
--- a/library/src/scripts/bootstrap.ts
+++ b/library/src/scripts/bootstrap.ts
@@ -17,7 +17,7 @@ import { _mountComponents } from "@library/utility/componentRegistry";
 import { blotCSS } from "@rich-editor/quill/components/blotStyles";
 import { bootstrapLocales } from "@library/locales/localeBootstrap";
 
-if (!gdn.getMeta("Feature.UseFocusVisible.Enabled", true)) {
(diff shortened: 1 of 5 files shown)


library/src/scripts/bootstrap.ts
```diff
@@ -17,6 +17,10 @@ import { _mountComponents } from "@library/utility/componentRegistry";
 import { blotCSS } from "@rich-editor/quill/components/blotStyles";
 import { bootstrapLocales } from "@library/locales/localeBootstrap";
 
+if (gdn.getMeta("Feature.UseFocusVisible.Enabled", true)) {
```
charrondev: I think this is named incorrectly.

```suggestion
if (gdn.getMeta("featureFlags.useNativeFocus.Enabled", true)) {
```

Also please add the feature and it's description to https://staff.vanillaforums.com/kb/articles/170-feature-flags

PR #9746, opened 2019-12-03
Removed "color" style on legacy blockquotes

slafleche:
~~When quotes are too deeply nested, they become impossible to read. The background color has a transparency, so it gets darker and darker the more you nest the quotes. I've capped it to 4 and then it's a solid color. Unfortunately, there's no magic bullet here for all themes because the specificity of deeply nested classes will be higher than styles overwriting the base styles. Fixes will need to go through services.~~

~~I checked the core themes and the only theme it didn't work for has a PR open.~~

Issue https://github.com/vanilla/support/issues/1107

The root of the issue was setting a color on the legacy blockquotes. We've moved that style to KB only.

### DIFF
--- a/library/src/scripts/content/UserContentLegacy.story.tsx
+++ b/library/src/scripts/content/UserContentLegacy.story.tsx
@@ -15,15 +15,15 @@ storiesOf("User Content", module)
        <h2>Legacy Code - BB Code</h2>
                        
         <div class="bbcode_left">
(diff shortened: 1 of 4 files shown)


slafleche:
I don't agree. The text color was there, I just made it darker, but it has the same specificity. Any themes on top should be able to keep their overwrites no problem.

PR #9748, opened 2019-12-03
Fixing adding GET params to existing URLs in OAuth2.

nervoustwit:
This PR is in response to this issue: https://github.com/vanilla/support/issues/1088

We have a customer whose Register URL in OAuth2 connection has query parameters attached to it which makes for double '?' when we add our GET parameters. 

This PR will create a static public method (so we can use it elsewhere) that will add an array of queries to an existing URL regardless of if the URL already has parameters.

## Testing

 - Turn on the OAuth2 plugin.
 - Configure the plugin so that the Register URL and the Authorize URL has a query parameter (e.g. `https://myauthserver.com/authorize?forward=true`).
 - Try to connect.
 - Observe borking.
 - Checkout this branch.
 - Now try to correct. 
 - It may still bork because of the extra parameters you passed but if the URL contains only one '?' then we are good.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -94,7 +94,7 @@ class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
      *
      * @param string $uri The URL with or without a query string already attached.
      * @param array $get Array of key/value pairs to be passed as GET params.
-     * @return string URL with
+     * @return string URL with or without param string attached.
      */
     public static function concatUriQueryString(string $uri, array $get = []): string {
         if (!$get) {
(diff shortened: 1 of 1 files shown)


library/core/class.oauth2.php
```diff
@@ -89,6 +89,20 @@ public function __construct($providerKey, $accessToken = false) {
         }
     }
 
+    /**
+     * Add a query to a URL without checking if there is already a query attached.
+     *
+     * @param string $uri The URL with or without a query string already attached.
+     * @param array $get Array of key/value pairs to be passed as GET params.
+     * @return string URL with
```
tburry: Looks like you stopped mid-thought here.

Issue #9713, opened 2019-11-28
Implement separare KB & Forum sections in the `<UserDropdown />` of mebox


charrondev:
## Expected changes

- Add a new `Articles` section above the `Discussions` section.
- This should contain a link to the article drafts w/ the article draft count.
- The existing `drafts` link the `Discussions` section should display the forum draft count and link the forum drafts page.
- Each section should have it's own permission checks.
- The "Articles" section should only appear when the KB is enabled.
- The "Discussions section should only appear the forum is enabled.

## Current Design

![image](https://user-images.githubusercontent.com/1770056/69819131-3fd2c680-11cc-11ea-8cbe-2eac1cdd9f93.png)

PR #9749, opened 2019-12-03
Split sections: discussion and article in user menu

alex-mtl:
Add section `Articles` with menu item `Drafts` with proper implemented link to `/kb/drafts/` and counter.

Update section `Discussion` with proper link for discussions drafts `/drafts` and counter.

Closes: #9713


library/src/scripts/headers/mebox/pieces/UserDropDownContents.tsx
```diff
@@ -41,14 +41,23 @@ function UserDropDownContents(props: IProps) {
             <DropDownUserCard className="userDropDown-userCard" />
             <DropDownItemSeparator />
             <DropDownItemLink to="/profile/edit" name={t("Edit Profile")} />
+            <DropDownSection title={t("Articles")}>
+                <Permission permission="articles.add">
```
slafleche: The permission should be on the outside of the section, or else it will still render HTML.

PR #9751, opened 2019-12-03
Adds support for static-asset handling

egarciajulia:
Adds support for static-asset handling. It allows better handling of Cache header

#### library/Vanilla/Contracts/Web/AssetInterface.php
+ Adds method `isStatic(): bool` to the interface

#### library/Vanilla/Web/Asset/SiteAsset.php
+ Implements default isStatic as false
+ `makeAssetPath` and `makeWebPath` now handle path as relative
+ fix small phpdocs

#### library/Vanilla/Web/Asset/ExternalAsset.php
+ Implements isStatic as false

#### library/Vanilla/Web/Asset/PolyfillAsset.php 
+ Implements isStatic as true

#### library/Vanilla/Web/Asset/LegacyAssetModel.php
+ fix `style-compat.css` path as relative

#### library/Vanilla/Web/Asset/LocaleAsset.php
+ fix  `self` usage for `$this`

#### library/Vanilla/Web/Asset/ThemeScriptAsset.php 
+ Implements isStatic as true
+ fix  `self` usage for `$this`

#### library/Vanilla/Web/Asset/WebpackAsset.php 
+ Implements isStatic as true

#### library/Vanilla/Web/Page.php
+ Adds EventManager as a dependency
+ Fires a new Event `BeforeRenderMasterView` before rendering the Twig template

#### library/Vanilla/Web/ThemedPage.php
+ Adds EventManager as a dependency
+ Removes some unused imports

#### applications/dashboard/modules/class.headmodule.php
+ `addScript` and `addCss` methods assume that a new added asset is static unless the asset is added with `$options['static'] = false`
+ A new event called `HeadTagsBeforeRender` is fired to allow manipulation of the tags before rendering.

#### library/core/class.controller.php 
+ adds `static` option when head->addScript is used

#### library/Vanilla/Models/SiteMeta.php
+ Adds `cacheBuster` and `staticPathFolder` properties
+ Fixes a small issue with getLogo returning null

#### build/entries/public-path.ts
+ Adds support for meta `context.staticPathFolder`

#### build/scripts/configs/makeBaseConfig.ts
+ Adds missing `chunkhash`

#### applications/dashboard/controllers/api/LocalesApiController.php
+ fix an apparently broken functionality regarding `Cache-Control` header. The new header is set to `1800` seconds (half an hour)

#### library/core/class.gdn.php
+ remove a static reference to the Gdn_Configuration object
+ The configuration object is taken from the container when needed.
+ This resolves an issue when the configuration object needs to be decorated or reimplemented in the container

### DIFF
--- a/applications/dashboard/controllers/api/LocalesApiController.php
+++ b/applications/dashboard/controllers/api/LocalesApiController.php
@@ -129,13 +129,6 @@ class LocalesApiController extends Controller {
     public function index_translations(string $locale, array $query = []) {
         $this->permission();
 
-        $in = $this->schema([
-            'etag:s?' => 'Whether or not output is cached.'
-        ], 'in');
-        $out = $this->schema([':o'], 'out');
-
-        $query = $in->validate($query);
-
         $this->locale->set($locale);
 
         // Don't bother validating the translations since they are a free-form array.
@@ -145,10 +138,7 @@ class LocalesApiController extends Controller {
             $translations = (object)[];
         }
         $r = new Data($translations);
-        if (!empty($query['etag'])) {
-            $r->setHeader('Cache-Control', 'public, max-age=604800');
-        }
-
+        $r->setHeader('Cache-Control', 'public, max-age=1800');
         return $r;
     }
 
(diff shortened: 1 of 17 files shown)


egarciajulia:
Wrong base

PR #9754, opened 2019-12-04
Add API format tests and fixes

tburry:
Currently, our app supports only a subset of formats, but some API endpoints don't check against formats. This PR adds some base tests for formats and makes fixes where appropriate.

Note: I've intentionally left some of the formats we will allow (html, ipb) out of the OpenAPI docs because I want to consider them deprecated.

### DIFF
--- a/applications/dashboard/openapi/schemas.yml
+++ b/applications/dashboard/openapi/schemas.yml
@@ -3,6 +3,18 @@ info:
 paths:
 components:
   schemas:
+    Format:
+      description: The format of the body used to convert it to HTML.
+      type: string
+      enum:
+        - rich
+        - markdown
+        - text
+        - textex
+        - wysiwyg
+        - bbcode
+        - display
+      example: markdown
     RoleFragment:
       type: object
       properties:
--- a/applications/dashboard/openapi/schemas.yml
+++ b/applications/dashboard/openapi/schemas.yml
@@ -13,7 +13,6 @@ components:
         - textex
         - wysiwyg
         - bbcode
-        - display
       example: markdown
     RoleFragment:
       type: object
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -1273,11 +1273,11 @@ class CommentModel extends Gdn_Model {
                     $this->fireEvent('AfterSaveComment');
                 }
             }
-        }
 
         // Update discussion's comment count.
         if (isset($formPostValues['DiscussionID'])) {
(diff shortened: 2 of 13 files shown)


tburry:
@charrondev I would really rather not make our formats data driven as you are suggesting. Every time we've added a new format we've ended up having to do a tonne of changes to our code. It is highly unlikely that we have a system that supports adding ad-hoc formats that won't cause regression tests.

What I would rather do is standardize our format support in our core repo and not allow new formats to be added as a rule unless we do it ourselves.

Furthermore, adding a deep dependency chain to the format schema is another tech debt inducing change. I would rather just add "ipb" to that format and be done with it.

Us knowing that these our our formats and that's it is a decent constraint that is easily understood and will challenge devs to find solutions within the current system. It will allow us to make our current code more robust rather than adding low value additions that will themselves destabilize the system.

tburry:
After speaking with Adam, we decided to put the format schema validation behind some feature flags so that sites that have dirty formats can still function until they remove offending data/code.

applications/vanilla/models/class.commentmodel.php
```diff
@@ -1273,11 +1273,11 @@ public function save($formPostValues, $settings = false) {
...
-        }
 
-        // Update discussion's comment count.
-        if (isset($formPostValues['DiscussionID'])) {
-            $this->updateCommentCount($formPostValues['DiscussionID'], ['Slave' => false]);
+            // Update discussion's comment count.
+            if (isset($formPostValues['DiscussionID'])) {
+                $this->updateCommentCount($formPostValues['DiscussionID'], ['Slave' => false]);
+            }
         }
```
charrondev: Was this change intended to make it into this PR?

applications/dashboard/openapi/schemas.yml
```diff
@@ -3,6 +3,18 @@ info:
...
+      description: The format of the body used to convert it to HTML.
+      type: string
+      enum:
+        - rich
+        - markdown
+        - text
+        - textex
+        - wysiwyg
+        - bbcode
+        - display
```
charrondev: You'd documented `display` here, but it's not actually in the main schema.

Additionally until we actually create some `DisplayFormat` and register it with the format service, that format would display errors in a bunch of places (search, activity, notifications, etc).

applications/vanilla/models/class.commentmodel.php
tburry: It was sort of related. The code before was updating the count regardless of whether or not the comment validated. There were some issues with invalid comments calling this code and making some odd errors. However, I'm guessing after further changes in my PR that no longer happened. Regardless, you can see the if statement I'm putting this in seems very sensible.

PR #9764, opened 2019-12-04
Fix scroll bar on desktop

irina-vanillaforums:
Vertical scroll bar partially overlays Save Preferences button on Notifications setting page.
Issue: https://github.com/vanillaforums/smartsheet/issues/131


themes/theme-boilerplate/src/scss/pages/_profile.scss
```diff
@@ -449,7 +449,10 @@ body.Section-EditProfile {
     }
 
     .Preferences > form > div {
-        overflow-x: scroll;
+
+        @include maxWidth {
```
slafleche: This doesn't work, see screen cap.

Screen Shot 2019-12-04 at 15 40 56

I really don't think it's worth adding the media query here. it should be:

```
    .Preferences > form > div {
        overflow-x: auto;
	}
```

First of all, your break point is much too wide to target only mobile devices and second of all, you won't see the scroll bar on mobile, because it'll be under the fold.

PR #9765, opened 2019-12-04
Fix options not being passed to visible colour picker form input

charrondev:
Our `Gdn_Form::color()` creates a visible text input and some hidden ones, along with a color chooser.

We weren't passing the full options to visible input though, so it was no possible to set certain attributes on it, such as `placeholder`.


charrondev:
Updated to add a test.

PR #9771, opened 2019-12-05
Add afterSave as props to IContentTranslatorProps 

NeenaDev:
Fix: https://github.com/vanilla/knowledge/issues/1372

Add afterSave() as props to IContentTranslatorProps to get navigationFlat data after saving translations for organize categories page.

Requires: 
https://github.com/vanilla/knowledge/pull/1432,
https://github.com/vanilla/internal/pull/2127

### DIFF
--- a/packages/vanilla-i18n/src/ContentTranslator.tsx
+++ b/packages/vanilla-i18n/src/ContentTranslator.tsx
@@ -14,7 +14,7 @@ export interface IContentTranslatorProps {
     title: string;
     sourceLocale?: string | null;
     activeLocale?: string | null; // for organizecategories
-    renderNavigationData?: () => void;
+    afterSave?: () => void;
 }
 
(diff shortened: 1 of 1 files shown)


packages/vanilla-i18n/src/ContentTranslator.tsx
```diff
@@ -14,6 +14,7 @@ export interface IContentTranslatorProps {
     title: string;
     sourceLocale?: string | null;
     activeLocale?: string | null; // for organizecategories
+    renderNavigationData?: () => void;
```
charrondev: I'd encourage you to think more about the naming of props like this.

This is a component interface for a **translation** package. It has nothing to do with navigation and no knowledge of it.

Something like `afterUpdate` or `afterSave` would be more appropriate.

PR #9776, opened 2019-12-06
Add translation debug for server rendered content

cchabilall83:
Enhancement for translationDebug tool https://github.com/vanilla/vanilla/pull/9698


library/core/class.configuration.php
```diff
@@ -376,6 +376,9 @@ public function get($name, $defaultValue = false) {
             if (is_array($value) && array_key_exists($keys[$i], $value)) {
                 $value = $value[$keys[$i]];
             } else {
+                if ($this->Data["TranslationDebug"] ?? false) {
```
charrondev: This looks like the right place, but I think we could give it a better name.

Maybe we can just make a private property + setter & call it the container.

```
                if ($this->fallbackDecorator ?? false) {
                	$defaultValue = $this->fallbackDecorator . $defaultValue . $this->fallbackDecorator;
                }
```

library/core/class.locale.php
```diff
@@ -379,6 +382,7 @@ public function unload() {
         $this->LocaleContainer = new Gdn_Configuration();
         $this->LocaleContainer->splitting(false);
         $this->LocaleContainer->caching(false);
+        $this->LocaleContainer->Data['TranslationDebug'] = $this->gdnConfig->get("TranslationDebug");
```
charrondev: Maybe something like this.

```
        if ($this->config->get("TranslationDebug")) {
            $this->LocaleContainer->setFallbackDecorator("☢️☢️");
        }
```

library/core/class.locale.php
```diff
@@ -36,6 +36,9 @@ class Gdn_Locale extends Gdn_Pluggable implements LocaleInterface {
      */
     private $addonManager = null;
 
+    /** @var \Vanilla\Contracts\ConfigurationInterface $gdnConfig */
+    private $gdnConfig;
```
charrondev: `$gdnConfig` is a pretty odd name for something that might not be a `Gdn_Config` at all.


Maybe just `config`?

PR #9801, opened 2019-12-10
More title bar variations in Storybook

NeenaDev:
Requires : https://github.com/vanilla/vanilla/pull/9792

Changes-
add navigationLinks as props for extra links to tileBar
add logoURL as props for custom logos to titleBar
Add Globe icon and dropdown.


slafleche:
There shouldn't be back buttons for "hamburger" style menus.

PR #9809, opened 2019-12-11
fix relative url path for multisite

egarciajulia:
The PR fixes an issue with url-paths in multisite environments.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -823,7 +823,7 @@ jQuery(document).ready(function($) {
 
     gdn.stats = function() {
         // Call directly back to the deployment and invoke the stats handler
-        var StatsURL = gdn.url(gdn.getMeta('context')["dynamicPathFolder"] + '/settings/analyticstick.json');
+        var StatsURL = gdn.getMeta('context')["dynamicPathFolder"] + gdn.url('/settings/analyticstick.json');
         var SendData = {
             'TransientKey': gdn.definition('TransientKey'),
             'Path': gdn.definition('Path'),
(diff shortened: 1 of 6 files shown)


charrondev:
> I disagree with this implementation.

To detail a little more, there was some in-person discussion related to confusing looking nature of appending a prefix to the result of `gdn.url()` in global.js

I definitely agree that it behaves differently from our other URL generation functions (which tend to generate an absolute URL), but I personally don't want to risk updating the `gdn.url` function.

egarciajulia:
@alex-mtl would you mind to elaborate? Do you have a counter-proposal?

js/global.js
```diff
@@ -823,7 +823,7 @@ jQuery(document).ready(function($) {
 
     gdn.stats = function() {
         // Call directly back to the deployment and invoke the stats handler
-        var StatsURL = gdn.url(gdn.getMeta('context')["dynamicPathFolder"] + '/settings/analyticstick.json');
+        var StatsURL = gdn.getMeta('context')["dynamicPathFolder"] + gdn.url('/settings/analyticstick.json');
```
alex-mtl: @egarciajulia  The code I see in this line breaks single responsibility principle.
Counter proposal here would be to either remove call to `gdn.url()`
```
var StatsURL = gdn.getMeta('context')["dynamicPathFolder"] + '/settings/analyticstick.json';
```
or pass this `dynamicPathFolder` to the function as second argument like:
```
var StatsURL =  gdn.url('/settings/analyticstick.json', gdn.getMeta('context')["dynamicPathFolder"]);
```

alex-mtl: Another thing to mention. I don't think that `gdn.getMeta('context')["dynamicPathFolder"] + '` is safe call to some predefined (we think) array key.

In case it is not defined, this code will produce string
```
StatsURL = 'undefined/settings/analyticstick.json'
```
with no error message or any exception thrown.

PR #9810, opened 2019-12-11
Added new media query for title bar only

slafleche:
This PR add a new context for the title bar only. It doesn't really make sense to be using the media queries for the layout for the title bar. Their layouts have nothing in common.

I do have an idea for a much better way to implement the title bar's breakpoints, but I won't have time this sprint.

### DIFF
--- a/library/src/scripts/layout/TitleBarContext.tsx
+++ b/library/src/scripts/layout/TitleBarContext.tsx
@@ -1,0 +1,81 @@
+/*
+ * @author Stéphane LaFlèche <stephane.l@vanillaforums.com>
+ * @copyright 2009-2019 Vanilla Forums Inc.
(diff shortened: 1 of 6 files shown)


library/src/scripts/layout/TitleBarContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export interface ITitleBarDeviceProps {
+    device: TitleBarDevices;
+}
+
+const TitleBarDeviceContext = React.createContext<TitleBarDevices>(TitleBarDevices.FULL);
+export default TitleBarDeviceContext;
+
+export function useDevice() {
```
charrondev:
```suggestion
export function useTitleBarDevice() {
```

PR #9828, opened 2019-12-12
Add threshold checking to MySQL index modifications

initvector:
Vanilla's alter table thresholds are not currently respected for index or "additional" queries being run. Since these operations can also have performance penalties, we're implementing threshold checking to these changes as well. All MySQL table modification queries should now be covered.

In addition, we're implementing default options for index operations: [inplace](https://dev.mysql.com/doc/refman/5.6/en/alter-table.html#alter-table-performance) and [no-lock](https://dev.mysql.com/doc/refman/5.6/en/alter-table.html#alter-table-concurrency). If the index operation fails, another attempt will be made without these options.

### Testing
1. Set your `Database.AlterTableThreshold` config to something small, like 1.
1. Delete an index from your discussions table.
1. Run a utility-update. You should see an error indicating your index could not be recreated due to the alter table threshold being exceeded.
1. Delete the `Database.AlterTableThreshold` config.
1. Run another utility-update. Your index should be created without error.


library/database/class.mysqlstructure.php
```diff
@@ -654,16 +668,24 @@ protected function _modify($explicit = false) {
         // Modify all of the indexes.
         foreach ($indexSql as $name => $sqls) {
             foreach ($sqls as $sql) {
-                if (!$this->executeQuery($sql)) {
-                    throw new Exception(sprintf(t('Error.ModifyIndex', 'Failed to add or modify the `%1$s` index in the `%2$s` table.'), $name, $this->_TableName));
+                try {
+                    $sqlWithOptions = $this->indexSqlWithOptions($sql);
+                    if (!$this->executeQuery($sqlWithOptions)) {
+                        throw new Exception(sprintf(t('Error.ModifyIndex', 'Failed to add or modify the `%1$s` index in the `%2$s` table.'), $name, $this->_TableName));
```
tburry: Can you make this a more specific exception?

library/database/class.mysqlstructure.php
```diff
@@ -337,9 +340,20 @@ protected function getTableInfo($tableName) {
     }
 
     /**
+     * Given an SQL representing a single basic alter-table query to add indexes, append the default index options.
      *
+     * @param string $sql
+     * @return string
+     */
+    private function indexSqlWithOptions(string $sql): string {
+        $result = preg_replace('/;?(?=\n*$)/', ", " . self::INDEX_OPTIONS . ";", $sql, 1);
```
tburry: Is there any way to restructure things to not have to do this regex kludge or is that too risky? Like you can modify the index SQL to not end with a semi colon? Otherwise I would do something similar to what Eduardo suggested, but maybe instead: `rtrim(rtrim($sql), ';')`.

initvector: Modifying the existing SQL is what I was hoping to avoid doing here. If we can avoid that, I would definitely prefer it.

> Otherwise I would do something similar to what Eduardo suggested, but maybe instead: `rtrim(rtrim($sql), ';')`.

No love for my surgical-precision regex? 😩

nervoustwit:
Testing locally. I am having trouble creating a table with enough records to fail. I will work on it again later.

egarciajulia:
@nervoustwit take a look at the base PR. There are some instruction for your issue.
https://github.com/vanilla/vanilla/pull/9828

tburry:
I see this has two approvals, but I would like to re-iterate my main ask: I would really like to see some tests for this. All of this is code that has proven problematic in terms of regressions so we really should have some coverage.

PR #9829, opened 2019-12-12
Remove fulltext indexes from Vanilla tables

initvector:
Full-text indexes of discussion and comment tables have not been used, and are not necessary, so they are being removed here.


alex-mtl:
Fulltext indexes are very useful if we do any searches with kind like  pattern `%text%` right now we use `LIKE` keyword. but `MATCH`  would bring the performance. 
So I would vote to keep these indexes and just start using them.

alex-mtl:
Fulltext indexes are very useful if we do any searches with kind like  pattern `%text%` right now we use `LIKE` keyword. but `MATCH`  would bring the performance. 
So I would vote to keep these indexes and just start using them.

initvector:
> So I would vote to keep these indexes and just start using them.

We discussed adding a way to opt into these indexes, so they'll be off by default but you can switch them on to potentially gain the performance benefits you mention. That'll likely be discussed further after we release v4 to the OSS community. Before _we_ can start using them, though, I think we need some additional internal prep. Until then, removing them is addressing an immediate issue without any performance penalty for hosted communities.

PR #9839, opened 2019-12-17
Fix media API GIF uploads losing animation

initvector:
The media API resource runs image uploads through `ImageResizer`. This class rewrites all images. However, if animated GIFs are rewritten, they lose their animation. Animated GIF uploads that are within the height, width and file size restrictions will no longer be rewritten.

### Testing
1. Using Rich Editor, upload an animated GIF. The image should display as a static image.
1. Switch to this branch.
1. Perform the same upload as step one. The image should be animated.


bootstrap.php
```diff
@@ -66,6 +66,9 @@
     ->addAlias('Config')
     ->addAlias(Contracts\ConfigurationInterface::class)
 
+    ->rule(\Vanilla\ImageResizer::class)
+    ->addCall('setAlwaysRewriteGif', [false])
```
tburry: You can set this from the config by using a reference. Frankly, I'm okay with what you did here though.

tburry:
FYI: https://images.v-cdn.net/progress.gif

bootstrap.php
initvector: I'm all about only adding new configs as we need them. We have the ability to add one, easily, but I would like to avoid doing so until we know we need it.

oliveratgithub:
Thanks for this fix! I added the changes manually to my Vanilla 3.3 and it works 😃

PR #9847, opened 2019-12-17
Add themes index endpoint

cchabilall83:
This adds  the `/Themes` endpoint that will return all available themes with a set variables that can generate a preview.

related to https://github.com/vanilla/internal/pull/2149
closes #9847


library/Vanilla/Theme/JsonAsset.php
```diff
@@ -41,9 +41,18 @@ public function asArray(): array {
...
      *
-     * @return string
+     * @param string $type
+     * @return mixed
      */
-    public function getData(): string {
-        return $this->data;
+    public function getData(string $type = 'string') {
+        if ($type === 'array') {
+            return json_decode(json_encode($this->data), true);
```
alex-mtl: I am not sure why we need to `encode` and `decode`, could we at least try to return it as is if that is an array already?
```suggestion
            if (is_array($this->data)) {
                return $this->data;
            } else {
                return json_decode(json_encode($this->data), true);
            }
```

charrondev: Another note, any time we are encoding/decoding we should be passing `JSON_UNESCAPED_UNICODE` so that any unicode characters don't get messed up.

library/Vanilla/Models/FsThemeProvider.php
```diff
@@ -289,4 +306,39 @@ private function getDefaultAssets(): array {
...
+            foreach ($sites as $s) {
+                if ($s === $clientName || fnmatch($s, $clientName)) {
+                    $hidden = false;
+                    break;
+                }
+            }
+        }
+
+        if ($hidden && !in_array(strtolower($themeInfo['key']), $alwaysVisibleThemes)) {
+            return[];
```
alex-mtl:
```suggestion
            return [];
```

library/Vanilla/Addon.php
```diff
@@ -167,6 +167,11 @@ private function scanInfo() {
                 }
             }
 
+            // ensure that the theme key is always lowercase.
+            if ($info['type'] === 'theme') {
+                $info['type'] = strtolower($info['key']);
```
alex-mtl:
```suggestion
                $info['key'] = strtolower($info['key']);
```

library/Vanilla/Theme/JsonAsset.php
```diff
@@ -41,9 +41,22 @@ public function asArray(): array {
...
+            } else {
+                return json_decode(json_encode($this->data, JSON_UNESCAPED_UNICODE), true);
+            }
+        }
+
+        if ($type === 'class') {
+            return (object)$this->data;
+        }
+
+        return json_encode($this->data, JSON_UNESCAPED_UNICODE);
```
charrondev: Can we leave this alone entirely and just a new method that would return the array?

We don't really care about the other use cases.

- `getData(): string` and keep your JSON encoding.
- `getVariables(): array` and just return the array directly.

library/Vanilla/Models/ThemeModel.php
```diff
@@ -233,4 +250,24 @@ public function deleteAsset(string $themeKey, string $assetKey) {
...
+        if ($variables) {
+            $preview['global.mainColors.primary'] = $variables['global']['mainColors']['primary'] ?? null;
+            $preview['global.mainColors.bg'] = $variables['global']['mainColors']['bg'] ?? null;
+            $preview['global.mainColors.fg'] = $variables['global']['mainColors']['fg'] ?? null;
+            $preview['titleBar.colors.bg'] = $variables['titleBar']['colors']['bg'] ?? null;
+            $preview['titleBar.colors.fg'] = $variables['titleBar']['colors']['fg'] ?? null;
+            $preview['splash.outerBackground.image'] = $variables['splash']['outerBackground']['image'] ?? null;
+        }
+        return $preview;
+    }
```
charrondev: This is failing on my machine with the `foundation` theme. For some reason it comes up with no variables asset.

image

Issue #9825, opened 2019-12-12
Render Foundation on the twig default master and create view for home page


slafleche:
- [ ] Create view for home page that will by default keep the current behaviour, but allow for customizations
- [ ] Load all the assets from the new theming system into it (JS/CSS, header/footer/fonts, etc)

PR #9848, opened 2019-12-18
Add foundation theme

charrondev:
Closes https://github.com/vanilla/vanilla/issues/9825

- Add `ThemeFeature` module.
  - `NewFlyouts` -> Already existed as a feature flag.
  - `ProfileHeader` -> Adds mobile profile header common in boilerplate and keystone.
  - `SharedMasterView` -> Enables shared KB/Forum master view for a forum theme.
- Refactor `PageHead` out of `Page` class.
  - Created interface and `PageHeadProxyTrait` so existing calls to the page work.
  - Put rendering into its own view.
- Create dedicated `MasterViewRenderer` class to render either a `Page` or a `Gdn_Controller` on the new master view.
  - `Move `default-master.twig` to `Vanilla\Web\MasterView.twig`
- Add numerous additional utilities & data to `TwigEnhancer` and `SiteMeta`
- Update `InstanceValidatorSchema` to allow multiple classes.
- Update themingAPI to allow `TwigAsset` for header and footer.
- Add foundation theme
  - Uses new twig template master view and subtemplates.
  - All current theme features enabled.
  - Built on boilerplate.
  - Uses build in build process instead of vanilla-cli.

## Known foundation specific theme bugs

We should iterate on these in future PRs

- KbIcons font doesn't load from boilerplate. I don't think we should use this anyways.


library/Vanilla/Theme/ThemeFeatures.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function __construct(Addon $theme, ConfigurationInterface $config) {
+        $this->theme = $theme;
+        $this->config = $config;
+    }
+
+    /**
+     * Get all of the current theme features.
+     */
+    public function allFeatures() {
```
alex-mtl:
```suggestion
    public function allFeatures(): array {
```

charrondev: Added

PR #9856, opened 2019-12-18
Add tests for Data class methods

blutarch:
I added tests that cover the Data class as well as the Meta trait and JsonFilter trait.

I don't know why the codecov test claims it will decrease code coverage.


charrondev:
@blutarch I think it might be some combination of how we merge the target branch into the testing branch before collecting coverage. I'm going to try and merge master into here and see if that changes anything.

PR #9860, opened 2019-12-19
Expose quill parser format and blot classes

bleistivt:
I hit a roadblock, trying to modify a quill format like this:

    public function container_init(\Garden\Container\Container $dic) {
        $dic
            ->rule(\Vanilla\Formatting\Quill\Parser::class)
            ->addCall('addFormat', [\Bleistivt\Formats\ImprovedLink::class]);
    }

This only lets me add a format to the end of the chain, which will then be wrapped in the old format I am actually trying to overwrite.
(I currently have to use reflection to do this instead.)

For real extensibility (e.g. setting the precedence as well), formatClasses and blotClasses would have to be exposed.

### DIFF
--- a/library/Vanilla/Formatting/Quill/Parser.php
+++ b/library/Vanilla/Formatting/Quill/Parser.php
@@ -312,4 +312,40 @@ class Parser {
             $lastOp["insert"] = preg_replace('/\s+$/', "\n", $lastOp["insert"]);
         }
     }
+
+    /**
+     * Get all registered blot classes.
+     *
+     * @return array
+     */
+    public function getBlotClasses(): array {
+        return $this->blotClasses;
+    }
+
+    /**
+     * Replace all registered blot classes.
+     *
+     * @param array
+     */
+    public function setBlotClasses(array $blotClasses) {
+        $this->blotClasses = $blotClasses;
+    }
+
+    /**
+     * Get all registered format classes.
+     *
+     * @return array
+     */
+    public function getFormatClasses(): array {
+        return $this->formatClasses;
+    }
+
+    /**
+     * Replace all registered format classes.
+     *
+     * @param array
+     */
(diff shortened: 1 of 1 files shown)


charrondev:
@bleistivt I support this, but probably in a more limited way.

I don't think we want to open up wholesale replacing all formats, because it will result in fragile implementations as we add additional ones.

I'd prefer to some implementation like the following:

```php
public function replaceFormat(string $newFormatClass, string $existingFormatClass) {
   // Ensure the new format is subclass the old one. Otherwise throw an error.

   // Replace our existing class definition and only that one.
}

public function removeFormat(string $existingFormatClass) {
   // Ensure the new format is subclass the old one. Otherwise throw an error.

   // Replace our existing class definition and only that one.
}
```

I don't like the wholesale replacement access, because at some point in the near future we might have different sets of formats for different parts of the site, or for different permission levels.

bleistivt:
I implemented your suggestion for both format and blot classes.

Is a subclass check really necessary? The class would have to be instantiated for this as far as I understand:
https://www.php.net/manual/en/function.is-subclass-of.php

charrondev:
@bleistivt From the PHP manual

> 
> ## Parameters ¶
> **object**
> A class name or an object instance. No error is generated if the class does not exist.
> 
> **class_name**
> The class name
> 
> **allow_string**
> If this parameter set to false, string class name as object is not allowed. This also prevents from calling autoloader if the class doesn't exist.

bleistivt:
Thanks, I must've missed that.

bleistivt:
A use case for this recently came up here: https://open.vanillaforums.com/discussion/comment/262655#Comment_262655
I'm still closing this, because just as Vanilla doesn't seem to be interested in this, I'm not interested in nannying these pull requests so they're not swallowed by the stale bot.

PR #9868, opened 2019-12-21
Fix some lint errors and code coverage

tburry:
Now that our code coverage of functions.general.php is in good shape I’ve added some ignore directives for areas that we just can’t hit.

In doing that I wanted to clean up some lint errors because I find it easier to not introduce new ones if an entire file is in a clean state.

Our old `Gdn_Autoloader` class has not been used for some time so I’m marking it as deprecated.

### DIFF
--- a/library/Garden/ArrayContainer.php
+++ b/library/Garden/ArrayContainer.php
@@ -44,8 +44,7 @@ class ArrayContainer extends \ArrayObject implements ContainerInterface {
      *
      * @param string $id Identifier of the entry to look for.
      *
-     * @throws NotFoundException  No entry was found for this identifier.
-     * @throws ContainerException Error while retrieving the entry.
+     * @throws ContainerNotFoundException  No entry was found for this identifier.
      *
      * @return mixed Entry.
      */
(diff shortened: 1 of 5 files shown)


library/Garden/ArrayContainer.php
```diff
@@ -44,8 +44,7 @@ private function normalizeID($id) {
      *
      * @param string $id Identifier of the entry to look for.
      *
-     * @throws NotFoundException  No entry was found for this identifier.
-     * @throws ContainerException Error while retrieving the entry.
+     * @throws ContainerNotFoundException  No entry was found for this identifier.
```
tburry: The previous two hints were pointing at a class that didn't exist. I put the correct class in here.

library/core/functions.deprecated.php
```diff
@@ -917,6 +917,23 @@ function viewLocation($view, $controller, $folder) {
...
+     */
+    function tagFullName($row) {
+        $result = val('FullName', $row);
+        if (!$result) {
+            $result = val('Name', $row);
+        }
+        return $result;
+    }
+}
+
```
tburry: This was moved here from functions.general.php.

Issue #9869, opened 2019-12-22
Good classes for new tests


tburry:
Here are some classes that are good candidates for unit tests:

- [x] `Vanilla\Utility\SchemaFilters`
- [x] `Vanilla\Web\ApiFilterMiddleware`: Add a getter for the blacklisted fields and fill in the remaining tests for the rest of the lines.
- [x] `Vanilla\Web\CacheControlMiddleware`: You can add tests for everything but `sendCacheControlHeaders()`.

PR #9877, opened 2020-01-02
Add tests for SchemaFilters class

blutarch:
Added tests for the methods in the SchemaFilters class.

Partially addresses #9869

### DIFF
--- a/library/Vanilla/Utility/SchemaFilters.php
+++ b/library/Vanilla/Utility/SchemaFilters.php
@@ -43,8 +43,10 @@ class SchemaFilters {
             if ($value === null) {
                 try {
                     $value = unserialize($value);
+                    // @CodeCoverageIgnoreStart
                 } catch (\Exception $e) {
                     $value = null;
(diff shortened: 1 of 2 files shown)


library/Vanilla/Utility/SchemaFilters.php
```diff
@@ -43,8 +43,10 @@ public static function decodeValue($value, ValidationField $field) {
             if ($value === null) {
                 try {
                     $value = unserialize($value);
+                    // @CodeCoverageIgnoreStart
```
initvector: Why do we want to ignore coverage here? Also, I don't know if it actually matters, but this tag usually starts with a lowercase C (e.g. [\@codeCoverageIgnoreStart](https://phpunit.de/manual/3.7/en/code-coverage-analysis.html#code-coverage-analysis.ignoring-code-blocks))

blutarch: Todd said he didn't think that part of the code was hittable, so we should ignore the code coverage. I made it camel case and am pushing the change.

PR #9881, opened 2020-01-03
Add discussion thread change log page

igraziatto:
This PR adds a page to output the discussion thread (discussions and its comments) changelog on the dashboard. This change is necessary for this service: https://github.com/vanillaforums/mse/issues/64
Service Estimate: https://github.com/vanillaforums/estimates/blob/master/mse/MSE%20-%20Change%20Log%20Customization%20Estimate.md


applications/dashboard/controllers/class.logcontroller.php
```diff
@@ -241,6 +244,47 @@ public function record($recordType, $recordID, $page = '') {
         $this->render();
     }
 
+    /**
+     * Access the log history of a specific discussion thread
+     *
+     * @param string $recordID
+     * @param string $page
+     */
+    public function discussion(string $recordID, string $page = '') {
```
initvector: I know there's some existing bluriness between applications, but since discussions are part of Vanilla, not Dashboard, I'd really like to see this functionality in that application, if at all possible. I'm not opposed to still using the Dashboard's log controller, but we should be able to extend it with Vanilla.

initvector: This shares an awful lot in common with `LogController::record`. Is there any way to reduce the code duplication being introduced?

applications/dashboard/models/class.logmodel.php
```diff
@@ -394,7 +394,7 @@ public function getIDs($iDs) {
      * @param bool $limit The database limit.
      * @return array Returns a data set.
      */
-    public function getWhere($where = false, $orderFields = '', $orderDirection = 'asc', $offset = false, $limit = false) {
+    public function getWhere($where = false, $orderFields = '', $orderDirection = 'asc', $offset = false, $limit = false, $orWhere = false) {
```
initvector: I would like to get away from using `false` as the default for these types of things. I know it is consistent with old code, but a boolean `false` has a very specific meaning. I think `null` is a more appropriate default. It also has a specific meaning: nothing. I think that's what we're trying to say with the default value of this parameter: we don't have any or-where conditions. Nullable parameters will also play nicely with type declarations.

applications/dashboard/models/class.logmodel.php
```diff
@@ -404,16 +404,21 @@ public function getWhere($where = false, $orderFields = '', $orderDirection = 'a
...
             ->from('Log l')
             ->join('User ru', 'l.RecordUserID = ru.UserID', 'left')
             ->join('User iu', 'l.InsertUserID = iu.UserID', 'left')
-            ->where($where)
-            ->limit($limit, $offset)
-            ->orderBy($orderFields, $orderDirection)
-            ->get()->resultArray();
+            ->where($where);
+
+        if ($orWhere) {
```
initvector: If we're expecting an array, I think we should explicitly test for an array, instead of any truthy value.

applications/dashboard/models/class.logmodel.php
```diff
@@ -430,19 +435,25 @@ public function getWhere($where = false, $orderFields = '', $orderDirection = 'a
      * Get the count of log entries matching a query.
      *
      * @param array $where The filter.
+     * @param bool|array $orWhere
      * @return int Returns the count.
      */
-    public function getCountWhere($where) {
+    public function getCountWhere($where, $orWhere = false) {
```
initvector: The changes we're making here are very similar to what we're making in `getWhere`, so I have the same concerns (`null` versus `false` and is-array versus is-truthy).

applications/dashboard/views/log/record.php
```diff
@@ -14,7 +14,7 @@ class="Info"><?php echo t('Edits and deletions are recorded here. Use &lsquo;Res
         echo anchor(t('Restore'), '#', ['class' => 'RestoreButton btn btn-primary']);
         echo anchor(t('Delete Forever'), '#', ['class' => 'DeleteButton btn btn-primary']);
         ?>
-        <?php PagerModule::write(['Sender' => $this, 'Limit' => 10, 'View' => 'pager-dashboard']); ?>
+        <?php PagerModule::write(['Sender' => $this, 'Limit' => $this->data('Limit'), 'View' => 'pager-dashboard']); ?>
```
initvector: This is a nice change, but we've gone from always having a value for `Limit` here to potentially having nothing (e.g. `Limit` isn't set in the data array). Maybe we should include a default. Otherwise, I think we'll either be using whatever the default of `PagerModule` is (probably not 10) or the default of `data` (an empty string?) will lead to an error when rendering.

igraziatto: I thought this was used as the default https://github.com/vanilla/vanilla/blob/7a85198092eb0b40e1577e97d7013c6350556fce/applications/dashboard/modules/class.pagermodule.php#L100 when no `$Limit` is provided.

initvector: > I thought this was used as the default [...] when no $Limit is provided.

It is. That's what I was referring to here:

> Otherwise, I think we'll either be using whatever the default of PagerModule is (probably not 10) [...]

I'm pretty sure that default isn't going to be 10, which was the limit we were using here before. It looks like it might actually be 30. Was altering the default limit on this page an intended change?

igraziatto:
Closing this PR because the service will be refactored: https://github.com/vanillaforums/estimates/issues/906#issuecomment-572596566

PR #9884, opened 2020-01-06
Fix issues with hidden social addons

### LINKED ISSUE #9885: Hidden social addons cannot be disabled
On the social connections page, when I try and disable a hidden social addon I get an "addon already enabled" error. Also, hidden social addons show up in the social connections page even when they are disabled.

tburry:
- Hidden social addons could not be disabled because they were being checked after the hidden filter was applied, making that app think they were enabled.
- Hidden social addons were never hidden from the social connections page.

Closes #9885.


blutarch:
Reproduced error on master and tested fix locally.

Issue #9826, opened 2019-12-12
Make new UI components for theme asset editor


charrondev:
## White TitleBar

![image](https://user-images.githubusercontent.com/1770056/70741158-af76a480-1ce8-11ea-9243-ade707bd7422.png)

- Currently have some implementation in knowledge base, in `EditorHeader`, `NavigationManagerMenu`, `DraftsHeader`.

Extract the common layout styles, into a new component and add variations in storybook.

## Tabs

Make the tabbing component and add it storybook.

Investigate using https://reacttraining.com/reach-ui/tabs/ as a base.

## Text Editor component

- Use monaco https://microsoft.github.io/monaco-editor/
- Official React wrapper here https://www.npmjs.com/package/@monaco-editor/react

- Put 2 variants in storybook, dark & light w/ button to toggle them.

![image](https://user-images.githubusercontent.com/1770056/70741442-45123400-1ce9-11ea-99e0-6998aa0a6bea.png)

PR #9886, opened 2020-01-06
Tabs and TextEditor Components 

cchabilall83:
This add the Tabs and TextEditor components along with a storybook story for them.

closes https://github.com/vanilla/vanilla/issues/9826


charrondev:
It looks like this PR causes some unwanted titlebar regressions:

https://www.chromaticqa.com/snapshot?appId=5d5eba16c782b600204ba187&id=5e13bb08c952af00208b0d19

library/src/scripts/sectioning/Tabs.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface IData {
+    label: string;
+    panelData: string;
+    contents: React.ReactNode;
+}
+interface IProps {
+    data: IData[];
+}
+
+export function DataTabs(props: IProps) {
```
charrondev: It's a bit odd to call this component `DataTabs` when the name of the file is `Tabs`. We should call it `Tabs` and adjust the import above.

Eg.:

```ts
import { Tabs as ReachTabs } from "@reach/tabs";
```

library/src/scripts/sectioning/Tabs.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Tabs
+            className={classes.root}
+            onChange={index => {
+                setActiveTab(index);
+            }}
+        >
+            <TabList className={classes.tabList}>
+                {data.map((tab, index) => {
+                    const isActive = activeTab === index;
+                    const style = {
```
charrondev: This should be moved into `tabClasses()`

charrondev:
There's a couple of styling issues here to match the mockups:

image

- Font weight should semibold.
- Active tab should not have text color changed.

PR #9893, opened 2020-01-07
Fix getting the permissions for a category with an invalid permission category ID

dannimajdalani:
closes https://github.com/vanilla/support/issues/1234

### Details
 in [class.categorymodel.php#L2540](https://github.com/vanilla/vanilla/blob/92bfca055e907b38d30f65a1512b7a3307246f86/applications/vanilla/models/class.categorymodel.php#L2540)

we are calling self::calculate($category); without checking if we have a category returned from  `$category = self::categories(val('PermissionCategoryID', $category)); `this is resulting in a fatal error when categoryModel::Calculate is called that requires $category to be an array
Screen Shot 2020-01-07 at 5 53 32 PM

### DIFF
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -2527,8 +2527,10 @@ class CategoryModel extends Gdn_Model {
      * @since 2.2
      */
     public static function permissionCategory($category) {
-        if (empty($category) || empty(((object)$category->PermissionCategoryID))) {
-            return self::categories(-1);
+        if (!is_integer($category)) {
+            if (empty($category) || empty(((object)$category->PermissionCategoryID))) {
+                return self::categories(-1);
+            }
         }
 
         if (!is_array($category) && !is_object($category)) {
--- a/applications/vanilla/models/class.categorymodel.php
+++ b/applications/vanilla/models/class.categorymodel.php
@@ -2534,10 +2534,13 @@ class CategoryModel extends Gdn_Model {
         if (!is_array($category) && !is_object($category)) {
             $category = self::categories($category);
         }
-
         $category = self::categories(val('PermissionCategoryID', $category));
+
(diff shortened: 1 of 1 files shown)


applications/vanilla/models/class.categorymodel.php
```diff
@@ -2534,10 +2534,13 @@ public static function permissionCategory($category) {
         if (!is_array($category) && !is_object($category)) {
             $category = self::categories($category);
         }
-
         $category = self::categories(val('PermissionCategoryID', $category));
+
         // Ensure all of our values are processed properly.
-        self::calculate($category);
+        if (isset($category)) {
```
initvector: A few lines up, we're returning the default category for permissions if the target category,`$category`, doesn't exist (is "empty"). What do you think about returning the default category for permissions if the actual "permission category" doesn't exist?

initvector:
`CategoryModel` is a blind spot in our test coverage. We should discuss how we want to tackle that.

applications/vanilla/models/class.categorymodel.php
```diff
@@ -2527,18 +2527,18 @@ protected static function _MakeTreeChildren($category, $categories, $depthAdj =
      * @since 2.2
      */
     public static function permissionCategory($category) {
-        if (empty($category)) {
+        if (empty($category) || empty($category['PermissionCategoryID'])) {
```
initvector: A couple lines down, it looks like this function is supposed to support arrays, objects and even integers. Won't this break that support?

dannimajdalani:
> Untested, but the code looks good to me.
> 
> We definitely need coverage for this model, but I'm not willing to block what is effectively a three-line PR to introduce a new test suite. It would be great to have that, if you have the time and desire to add it, but otherwise this PR can be merged without the code coverage requirements being met.

I'll look into that this sprint.

PR #9894, opened 2020-01-08
Updates to KB tiles, background images for home page and design tweaks

slafleche:
Companion PR: https://github.com/vanilla/knowledge/pull/1466

Renamed subcommunity tiles to just tiles.

Updated background image options. We can now have a different background image for the home page.

Added option to have a background for the left column in the article.

Fixed some alignment issues.

______________
To test, either hard code panelBackground.config.render to true, or set it in theme to see background.

To test global background image, set options in theme for global.body.backgroundImage. Note the options are described in IBackground interface.

To test home page background image, set options in theme for homePage.backgroundImage. Note the options are described in IBackground interface.

To test number of tiles, either hard code columns in Tiles.tsx (<KnowledgeBaseList columns=x) OR hard code value in tilesVariables.options.columns OR set in theme.


library/src/scripts/theming/ThemeProvider.tsx
```diff
@@ -92,7 +92,7 @@ export const ThemeProvider: React.FC<IProps> = (props: IProps) => {
 
     return (
         <>
-            <Backgrounds />
+            <Backgrounds isHomePage={true} />
```
slafleche: @charrondev I'm not sure how I can detect the home page here. What do you suggest?

slafleche:
See note about detecting the home page.

PR #9895, opened 2020-01-08
Add props to TextEditor 

NeenaDev:
Changes - 
Using  `<ControlledEditor />`  instead `<TextEditor/>`
UI fixes to `<ThemePreviewCard />`


library/src/scripts/textEditor/TextEditor.tsx
```diff
@@ -25,11 +28,12 @@ export default function TextEditor(props: ITextEditorProps) {
             <button onClick={toggleTheme} className={classes.themeToggleIcon} disabled={!isEditorReady}>
                 <ToggleIcon />
             </button>
-            <Editor
+            <ControlledEditor
                 theme={intialTheme}
                 language={language}
                 editorDidMount={handleEditorDidMount}
```
cchabilall83: do we still need editorDidMount prop?

NeenaDev: yes, if we remove this, it affect theme toggle button event.

library/src/scripts/textEditor/textEditorStyles.ts
```diff
@@ -32,13 +40,25 @@ export const textEditorClasses = useThemeCache(() => {
             height: viewHeight(90),
             backgroundColor: theme === "dark" ? colorOut("#1f2024") : colorOut(globalVars.elementaryColors.white),
             position: "relative",
+
+            $nest: {
+                "&& .monaco-editor": {
+                    paddingTop: unit(vars.editorPadding.padding.top),
+                    paddingLeft: unit(vars.editorPadding.padding.left),
```
charrondev: I'm pretty sure this is the cause of the editor menu being offset.

You'll need to add it to a container further outside the editor root like in the example here. https://wangchujiang.com/react-monacoeditor/

charrondev:
@NeenaDev You've also got merge conflicts.

PR #9905, opened 2020-01-09
Translate "roles" into "Roles" when passing roles over SSO.

patches:
# SSO roles adjustments

A client is currently unable to pass roles via SSO because Vanilla's SSO is using "Roles" and does not accept "roles" (in lower case).

I initially hid this feature behind a feature flag, however, I felt it was a bit extreme. In this change, I am also making sure that "roles" does not take over "Roles" if both are set.

Estimate request: https://github.com/vanillaforums/estimates/issues/886
Project: https://github.com/vanillaforums/sonicwall/issues/10

## Requirements

To test this change, you will need to have SSO setup locally. I followed the directives from both repositories below.

https://github.com/vanilla/vanilla-docker
https://github.com/vanilla/stub-sso-providers

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -503,6 +503,8 @@ class EntryController extends Gdn_Controller {
         }
 
         $isTrustedProvider = $this->data('Trusted');
+        $saveRoles = false;
+        $saveRolesRegister = false;
 
         // Check if we need to sync roles
         if (($isTrustedProvider || c('Garden.SSO.SyncRoles'))) {
@@ -535,9 +537,6 @@ class EntryController extends Gdn_Controller {
 
                 $this->Form->setFormValue('RoleID', $roleIDs);
             }
-        } else {
-            $saveRoles = false;
-            $saveRolesRegister = false;
         }
 
         $userModel = Gdn::userModel();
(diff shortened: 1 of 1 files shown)


initvector:
I'm gonna tag in the SSO guru, @nervoustwit, to review this, as well as request some support from @dannimajdalani.

applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -503,30 +503,40 @@ public function connect($method) {
...
             $roles = $this->Form->getFormValue('Roles', null);
-            $roles = RoleModel::getByName($roles);
-            $roleIDs = array_keys($roles);
 
-            // Ensure user has at least one role.
-            if (empty($roleIDs)) {
-                $roleIDs = $this->UserModel->newUserRoleIDs();
+            // Allow the roles index to be lower case.
+            // This feature was requested and approved here https://github.com/vanillaforums/estimates/issues/886
+            $loweredRoles = $this->Form->getFormValue('roles', null);
```
dannimajdalani: We can probably use   
$roles = $this->Form->getFormValue('Roles',  $this->Form->getFormValue('roles', null));

patches: I think you are right, I did this over-complicated logic to prevent from affecting any existing plugin. However, I don't think the suggestion you made will affect anything.

PR #9909, opened 2020-01-10
Adjust ThemeModel to provide necessary data to SiteMeta and Preload provider

alex-mtl:
Relates to: https://github.com/vanilla/internal/issues/2167


library/Vanilla/Models/ThemeModel.php
```diff
@@ -203,8 +203,34 @@ public function setCurrentTheme($themeID): array {
...
+                throw $e;
+            }
+            //do nothing if db provider does not exist
+        }
+
+        if (is_null($current)) {
+            $provider = $this->getThemeProvider("FILE");
+            $current = $provider->getCurrent();
+        }
+        return $current;
```
charrondev: Looking at this, I think that the theme model knows too much about the implementation details of the its providers. Maybe we we could fan it by adding an `isCurrent(): boolean` method to the providers and looping over them?

Maybe we can't, but I feel like we should be able to come up with a cleaner implementation than this.

charrondev: After talking in person it seems it's worth adding additional complexity. This odd split is mostly due to the location of the DB provider being out of repo.

Issue #9903, opened 2020-01-09
Create a published foundation package


charrondev:
## Acceptance criteria

- Foundation is published as an NPM package is equivalent to theme-boilerplate.
- It can scaffold out a new theme based on itself (using an npx command).
- It is a dependency that can be built into the package.

PR #9918, opened 2020-01-11
Merge pull request #9918 from vanilla/feature/addon-load-themes

charrondev:
Prerequisite for https://github.com/vanilla/vanilla/issues/9903
Companion PR here to allow CustomizeTheme to work on themes in the addons directory https://github.com/vanilla/internal/pull/2176

## Changes

1. Allows plugins & themes to be loaded from the `/addons` directory.
2. Adds the `/addons` directory to the yarn workspace.

## Why?

In order to fulfill https://github.com/vanilla/vanilla/issues/9903 we need to get some themes into our yarn workspace. Unfortunately there are a couple restrictions on adding packages to the workspace.

1. The package's entire parent directory must be part of the workspace.
2. Every package in the workspace has to have valid package.json
  - All packages must install on modern systems.
  - All packages must have: `name`, `version`, and `public: true | false` set.

Unfortunately we have 100s of themes that don't meet this criteria, spread across multiple repos. The prevalence of old native modules from the 2013-2017 era spread across all the repos means **adding the existing `themes` directory to the workspace will cause the `yarn install` command to fail under many circumstances (we have at least 40+ themes that would need to update packages that are 4-5 years old).**

As a result I've created a new `/addons` directory that is part of the workspace. This way we can ignore most of the cruft that was never meant to be built in the workspace.

**But why "/addons"?**

We call them addons in all of our documentation and user facing text. It's mostly been laziness that has caused our new addons continue to be installed in to the "plugins" and "themes" directories.

We had some intention of making

## Testing

This PR is mostly just adding a bunch of extra directories to our lookup paths. I felt actually moving anything during this PR would make things a lot harder to review, so that will come in a followup to fulfill https://github.com/vanilla/vanilla/issues/9903

In order to test that this works, I recommend doing the following temporarily.

- Move your current theme from the `/themes` directory into the `/addons` directory.
- Move your knowledge base symlink from the `/plugins` directory into the `/addons` directory.
- Clear your addon cache.
- Restart your dev build.
- Notice the theme and knowledge base still work as they did before.


charrondev:
> I think there are going to be some fundamental issues with mixing themes and addons in the same directory. Currently, the AddonManager::getSingleIndex() scans just a directory and takes everything there to get a list of themes or locales. It specifically doesn't check addon.json files as an optimization there.

I'll add some tests around this and see what I can do. 

> I think in the past we've also made a habit of naming a customer theme and addon the same thing, but just putting them in separate directories.

Yes, this would potentially be an issue if we were moving all existing themes into this new directory (and into the new build, which fails as well with duplicate addon keys).

However the intention here is to **not** have all of our existing themes put here. Only new stuff that is included into the core build.

My followup PR to this includes 3 themes in that directory.

- `theme-boilerplate`
- `theme-keystone`
- `theme-foundation`

Keystone and foundation both had to get moved because they share the core package `@vanilla/theme-core` and as such all need to be in the workspace.

charrondev:
@tburry I've added extra additional tests and everything still seems to be working as expected.

I see the potential issue that could occur if a single index was used for this mixed directory, but through some testing it seems a multi-index is used everytime, which doesn't have the issue.

Do you want me to still update the single index creation?

I'm thinking I could just force themes to always use the multi-cache. At that point locales would be the only one using the single cache.

charrondev:
Alright @tburry I've updated the implementation and split the new directories to prevent cache building issues.

The new directories are now `/addons/themes` & `/addons/addons`.

applications/dashboard/settings/class.hooks.php
```diff
@@ -298,9 +298,12 @@ public function base_beforeFetchMaster_handler($sender) {
...
         $isDefaultMaster = $sender->MasterView == 'default' || $sender->MasterView == '';
         if ($theme != '' && $isDefaultMaster) {
-            $htmlFile = paths(PATH_THEMES, $theme, 'views', 'default.master.tpl');
-            if (file_exists($htmlFile)) {
-                $sender->EventArguments['MasterViewPath'] = $htmlFile;
+            $themeHtmlFile = paths(PATH_THEMES, $theme, 'views', 'default.master.tpl');
+            $themeAddonHtmlFile = paths(PATH_ADDONS_THEMES, $theme, 'views', 'default.master.tpl');
+            if (file_exists($themeHtmlFile)) {
+                $sender->EventArguments['MasterViewPath'] = $themeHtmlFile;
+            } elseif (file_exists($themeAddonHtmlFile)) {
```
tburry: Duplicate themes will result in some very off bugs. Is there something we can do to check for that? Maybe not for now.

charrondev: https://github.com/vanilla/vanilla/issues/9934

PR #9920, opened 2020-01-13
Apply more case-insensitive format checks for rich

charrondev:
With https://success.vanillaforums.com/kb/articles/144-release-2020-001#api-v2 going out we are likely to begin seeing more lowercase usages of Rich throughout existing parts of the application.

Unfortunately not everything was:

- Using the constant.
- Doing a case-insensitive comparison.

I've updated some case-sensitive checks to now be case-insensitive & used the constant in a few places it wasn't in use.

### DIFF
--- a/plugins/rich-editor/src/scripts/mountEditor.tsx
+++ b/plugins/rich-editor/src/scripts/mountEditor.tsx
@@ -22,9 +22,9 @@ export default function mountEditor(containerSelector: string | Element) {
         throw new Error("Could not find the BodyBox to mount editor to.");
     }
 
-    const initialFormat = bodybox.getAttribute("format") || "Rich";
+    const initialFormat = bodybox.getAttribute("format");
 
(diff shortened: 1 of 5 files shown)


plugins/rich-editor/src/scripts/mountEditor.tsx
```diff
@@ -22,9 +22,9 @@ export default function mountEditor(containerSelector: string | Element) {
         throw new Error("Could not find the BodyBox to mount editor to.");
     }
 
-    const initialFormat = bodybox.getAttribute("format") || "Rich";
```
alex-mtl: Don't we still require coalescing default here?

charrondev: In person discussion summary:

We don't actually want it anymore. We'd rather treat no format as a plain text box than as a rich editor.

PR #9924, opened 2020-01-14
Add basic class for simple resource events

initvector:
This update adds a very simple class that will be used for creating [PSR-14-style events](https://www.php-fig.org/psr/psr-14/) for basic resource operations (e.g. adding a discussion, deleting a comment, registering a user).

### DIFF
--- a/library/Garden/Events/ResourceEvent.php
+++ b/library/Garden/Events/ResourceEvent.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * @copyright 2009-2020 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace Garden\Events;
+
+/**
+ * An event affecting a specific resource.
+ */
+abstract class ResourceEvent {
+
+    /** A resource has been removed. */
+    const ACTION_DELETE = "delete";
+
+    /** A resource has been created. */
+    const ACTION_INSERT = "insert";
+
+    /** An existing resource has been updated. */
+    const ACTION_UPDATE = "update";
+
+    /** @var string */
+    protected $action;
+
+    /** @var array */
+    protected $payload;
+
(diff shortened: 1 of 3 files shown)


library/Garden/Events/ResourceEvent.php
```diff
@@ -0,0 +1,86 @@
...
+    public function getPayload(): ?array {
+        return $this->payload;
+    }
+
+    /**
+     * Get the event type.
+     *
+     * @return string|null
+     */
+    public function getType(): ?string {
```
tburry: I think this should just be `string` to avoid edge cases. Also, the constructor makes it a string.

library/Garden/Events/ResourceEvent.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->payload = $payload;
+        $this->type = $this->typeFromClass();
+    }
+
+    /**
+     * Get the event action.
+     *
+     * @return string|null
+     */
+    public function getAction(): ?string {
```
tburry: I think this should just be `string` to avoid edge cases. Also, the constructor makes it a string.

tburry:
You can add some basic tests for this class to ensure that the arguments passed to the constructor come back in the getters. This is something @blutarch can do if you wish.

library/Garden/Events/ResourceEvent.php
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Garden\Events;
+
+/**
+ * An event affecting a specific resource.
+ */
+abstract class ResourceEvent {
+
+    /** A resource has been removed. */
+    const ACTION_DELETE = "delete";
```
tburry: Now that we are PHP 7.1 it is a best practice to declare these as `public`.

PR #9929, opened 2020-01-14
Fix display of discussion buttons

blutarch:
Fixes vanilla/support#639

The Ask a Question button was behaving erratically when it was configured to be its own button.

This change fixes that issue AND allows you to configure any discussion button to be its own button.

# Instructions for implementing the fix

You have to add the following settings to the config, setting them to true or false depending on your need:
```
NewDiscussionModule.Types.Discussion.AsOwnButton = false;
NewDiscussionModule.Types.Poll.AsOwnButton = false;
NewDiscussionModule.Types.Question.AsOwnButton = true;
NewDiscussionModule.Types.Idea.AsOwnButton = false;
```
The default value if the settings are not included is `false`.

Here are some various scenarios (with three buttons):

### No settings in the config (same as all buttons set to false)
Screen Shot 2020-01-15 at 12 20 03 PM

### Question button ['AsOwnButton'] set to true; all others set to false:
* Dropdown closed:
Screen Shot 2020-01-15 at 12 22 26 PM
* Dropdown open:
Screen Shot 2020-01-15 at 12 24 31 PM

### All buttons set to true:
Screen Shot 2020-01-15 at 12 29 07 PM

## Setting the text on a group button

In addition, you can designate a default setting in the config by setting Vanilla.DefaultNewButton to the desired button's url code (e.g., `'post/question'`). If nothing is entered, the text on a dropdown menu will default to whatever button comes first in the array. For instance, if you do not set the default button to `'post/question'` and have a grouped button with question and poll, the button will look like this:
Screen Shot 2020-01-15 at 12 44 30 PM
because the 'New Poll' button is earlier in the buttonGroup array.
If you add `Vanilla.DefaultNewButton = 'post/question'` to the config, the button will look like this:
Screen Shot 2020-01-15 at 12 46 48 PM

I tested this in the following scenarios:
* All buttons are individual buttons
* One dropdown button
* One individual button and one dropdown with two items
* Changing the default button name for the dropdown button


applications/vanilla/modules/class.newdiscussionmodule.php
```diff
@@ -61,11 +61,12 @@ public function assetTarget() {
...
      *
-     * @param $text
-     * @param $url
+     * @param string $text
+     * @param string $url
+     * @param bool $asOwnButton Whether to display as a separate button or not.
      */
-    public function addButton($text, $url) {
-        $this->Buttons[] = ['Text' => $text, 'Url' => $url];
+    public function addButton($text, $url, $asOwnButton) {
```
tburry: This new parameter needs a default of false or else it breaks BC.

applications/vanilla/modules/class.newdiscussionmodule.php
```diff
@@ -123,7 +125,10 @@ public function toString() {
                 $url = $this->GuestUrl.'?Target='.$url;
             }
 
-            $this->addButton(t(val('AddText', $type)), $url);
+            // Check whether to display in dropdown or as a separate button.
+            $asOwnButton = $buttonsConfig[$type['Singular']]['AsOwnButton'] ?? false;
+
+            $this->addButton(t(val('AddText', $type)), $url, $asOwnButton);
```
tburry: The `val()` function is now deprecated. Can you change to `$type['AddText'] ?? ''`?

applications/vanilla/modules/class.newdiscussionmodule.php
```diff
@@ -135,4 +140,26 @@ public function toString() {
 
         return parent::toString();
     }
+
+    /**
+     * Groups buttons according to whether they are standalone or part of a dropdown.
+     *
+     * @return array Returns buttons grouped by whether they are standalone or part of a dropdown.
+     */
+    public function getButtonGroups() {
```
tburry: Type hint the return to an array.

tests/fixtures/src/MockNewDiscussionModule.php
```diff
@@ -0,0 +1,57 @@
...
+ * @license GPL-2.0-only
+ */
+
+namespace VanillaTests\Fixtures;
+
+/**
+ * A mock new discussion module to test NewDiscussionModule methods.
+ */
+
+class MockNewDiscussionModule {
```
tburry: In this case you aren't really testing the module. If the module is changed this test won't help.

Maybe you can try `class MockNewDiscussionModule extends NewDiscussionModule` and then override the constructor to not do anything?

applications/vanilla/modules/class.newdiscussionmodule.php
```diff
@@ -135,4 +140,26 @@ public function toString() {
...
+        $this->reorder = true;
```
tburry: This line seems wrong. What is your intention here?

PR #9938, opened 2020-01-15
Display categories with no discussions on "Mixed" layout

igraziatto:
Closes https://github.com/vanilla/support/issues/275

The issue It's not keystone related actually, it's just how this view works (disable the theme and you will see the same behavior).

The view doesn't render categories who have no discussions. Since you can't add discussions to categories type "Nested", it ignores them: https://github.com/vanilla/vanilla/blob/9dda0f328015a74cd88a3a54f0a66ad5addd76e8/applications/vanilla/views/categories/discussions.php#L17

**Notes:** The view seems to flat the category tree when rendering so I've just followed the same logic.

### DIFF
--- a/applications/vanilla/views/categories/discussions.php
+++ b/applications/vanilla/views/categories/discussions.php
@@ -36,7 +36,38 @@ $ViewLocation = $this->fetchViewLocation('discussions', 'discussions');
 
                 </div>
 
-            <?php endif; ?>
+            <?php else:
+
+                if ($Category->CountCategories > 0) { ?>
+
+                    <div class="CategoryBox Category-<?php echo $Category->UrlCode; ?>">
+                        <?php echo getOptions($Category); ?>
+                        <h2 class="H">
+                            <?php
+                                echo anchor(htmlspecialchars($Category->Name), categoryUrl($Category));
+                                Gdn::controller()->EventArguments['Category'] = $Category;
+                                Gdn::controller()->fireEvent('AfterCategoryTitle');
+                            ?>
+                        </h2>
+                    </div>
+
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/categories/discussions.php
```diff
@@ -36,7 +36,38 @@
...
+                    <div class="CategoryBox Category-<?php echo $Category->UrlCode; ?>">
+                        <?php echo getOptions($Category); ?>
+                        <h2 class="H">
+                            <?php
+                                echo anchor(htmlspecialchars($Category->Name), categoryUrl($Category));
+                                Gdn::controller()->EventArguments['Category'] = $Category;
+                                Gdn::controller()->fireEvent('AfterCategoryTitle');
+                            ?>
+                        </h2>
+                    </div>
```
tburry: This looks like the exact same code. Can you refactor so it's just one block? It seems alike you can take the main if statement out altogether and then just wrap an if statement around the table output no?

applications/vanilla/views/categories/discussions.php
```diff
@@ -34,12 +36,15 @@
...
-            <?php endif; ?>
+                <?php else: ?>
+                    <div class="Empty"><?php echo t('No discussions were found.'); ?></div>
+                <?php endif; ?>
+            </div>
 
         <?php endforeach; ?>
-    <?php else: ?>
-        <div class="Empty"><?php echo t('No categories were found.'); ?></div>
+    <?php else:
```
tburry: This seems wrong. You are outside of your loop. It seems like you still need that "No categories were found" from above.

PR #9944, opened 2020-01-16
Fix logic for determining if a Discussion has been read by a user.

nervoustwit:
This will close https://github.com/vanilla/support/issues/559

We have some logic issues when it comes to determining if a user has actually read all the most recent comments in a discussion. The problem happens when users delete comments. We update the Discussion table's Number of Comments in the Discussion but we do not (and cannot for scaling reasons) update the UserDiscussion table. So if I have read 100 Comments in a Discussion and then some users delete 10 Comments, we will compare the number of Comments I have read with the number of Comments available to be read, we see that I have read 100 comments and there are 90 available. Now as more users add comments it still looks like I have read them all.

This PR:
* Extracts the logic for determining if a user really has read all the comments into a two simple testable methods.
    * One determines the user's correct "Last Viewed" date for that discussion. 
    * The second determines whether the discussion should be marked read or not. If there is a discrepancy between the read status and the number of comments marked read, this will resolve it. However, in the case that the comment count is incommensurable with the read status, it will set unread comments to 0 if the conversation is determined to be read and unread comments to 1 if the conversation is determined to be unread. This last case is not ideal (there may in fact have been more than one unread comment), but it is simpler and more scalable than trying to guess the unread count by checking the dates of the last view against all of the comments' dates. It also may be less confusing to the user than guessing the number.
* Adds some logic to setWatch() in commentmodel to update the userDiscussion table and the discussion object with the correct number of comments read by the user.

## Testing
 - Create a fresh discussion.
 - Add 3 comments.
 - View the 3 comments as a different user.
 - Delete one Comment.
 - Make sure that the Discussion appears to be UnRead when you see it on the Discussions page.
 - Click on the discussion as the different user.
 - Confirm that the CountComments column in the userDiscussion table has a value of 2.


applications/vanilla/controllers/class.discussioncontroller.php
```diff
@@ -188,6 +175,18 @@ public function index($DiscussionID = '', $DiscussionStub = '', $Page = '') {
         // Load the comments
         $this->setData('Comments', $this->CommentModel->getByDiscussion($DiscussionID, $Limit, $this->Offset));
 
+        $LatestItem = $this->Discussion->CountCommentWatch;
```
initvector: Was this done in order to keep the comment-oriented operations together or is there a functional reason?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1110,39 +1110,32 @@ public function calculate(&$discussion) {
...
+            if (!is_null($discussion->DateLastViewed) && is_numeric($discussion->CountUnreadComments)) {
+                $updatedReadStatusAndUnreadCount = self::determineReadStatusAndUnreadCount(
+                    $discussion->CountComments,
+                    $discussion->DateLastComment,
+                    $discussion->CountCommentWatch,
+                    $discussion->DateLastViewed
+                );
+                $discussion->Read = $updatedReadStatusAndUnreadCount[0];
+                $discussion->CountUnreadComments = $updatedReadStatusAndUnreadCount[1];
+//                $discussion->CountCommentWatch = $discussion->CountComments - $discussion->CountUnreadComments;
```
initvector: Can we nuke this?

blutarch: Yes!

applications/vanilla/controllers/class.discussioncontroller.php
blutarch: @tburry, can you speak to this?

tburry: This was done when we thought we were going to add another correction for `$LatestItem` based on the comments that were fetched from the DB. That ended up not being the case, but the move was not undone.

I would say though that the correction (comparing the last page of comments to date last viewed) is sensible and may be done in the future if we can figure a way around the UX issues with it. In that case, leaving the code here is still sensible.

blutarch:
@initvector, please don't review this yet. I have to add one more test and add a couple of comments.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1167,6 +1159,67 @@ public function calculate(&$discussion) {
...
+        if ($userLastReadDate > $discussionLastCommentDate && $unreadCommentCount > 0) {
+            $unreadCommentCount = 0;
+        }
+
+        if (!$userReadComments) {
+            $unreadCommentCount = $discussionCommentCount;
+        }
+
+        if ($unreadCommentCount < 0 && !$isRead) {
+            $unreadCommentCount = 1;
```
tburry: You are missing coverage for this case.

I would also add comments to this and the next case. This case is a kludge and if you don't comment it another developer may remove it.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1110,39 +1110,31 @@ public function calculate(&$discussion) {
...
-
-                if ($discussion->Read) {
-                    $discussion->CountUnreadComments = 0;
-                }
-            }
-
-            // Discussions are always unread to guests.
-            if (!Gdn::session()->isValid()) {
-                $discussion->Read = false;
+                $discussion->DateLastViewed = self::determineLastViewedDate($discussion->DateLastComment, $category['DateMarkedRead']);
```
tburry: This is wrong. If I haven't session'd then I don't have a `$category['DateMarkedRead']`.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1110,39 +1110,31 @@ public function calculate(&$discussion) {
...
+        // Discussions are always unread to guests. Otherwise check Read status and Unread count.
+        if (!Gdn::session()->isValid()) {
+            $discussion->Read = false;
+        } else {
+            if (!is_null($discussion->DateLastViewed) && is_numeric($discussion->CountUnreadComments)) {
+                $updatedReadStatusAndUnreadCount = self::determineReadStatusAndUnreadCount(
+                    $discussion->CountComments,
+                    $discussion->DateLastComment,
+                    $discussion->CountCommentWatch,
+                    $discussion->DateLastViewed
```
tburry: This is the max of `$category['DateMarkedRead']` and `$discussion->DateLastViewed`. Are you taking that into account? If so, can you clean out some of the `$category['DateMarkedRead']` stuff above?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1167,6 +1159,67 @@ public function calculate(&$discussion) {
...
+     */
+    public static function determineReadStatusAndUnreadCount(
+        int $discussionCommentCount,
+        $discussionLastCommentDate,
+        int $userReadComments,
+        $userLastReadDate
+    ) {
+        $isRead = true;
+        $unreadCommentCount = $discussionCommentCount - $userReadComments;
+        if ($discussionLastCommentDate > $userLastReadDate) {
```
tburry: Your function says you are taking `mixed` but you are using simple `>` comparisons. In practice you are comparing strings with greater than which I'm not super happy with. I recommend making the function officially take `DateTimeInterface` instances or strings, but then deal with the strings properly by converting them into dates.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1167,6 +1159,67 @@ public function calculate(&$discussion) {
...
+
+    /**
+     * Decide which date we want to consider as the date last viewed, either when the
+     * user has marked the Category read or when the user last viewed the Discussion.
+     *
+     * @param mixed $discussionLastViewedDate Date discussion was last viewed.
+     * @param mixed $categoryLastViewedDate Date category was marked viewed.
+     * @return mixed Returns most recent date.
+     */
+    public static function determineLastViewedDate($discussionLastViewedDate, $categoryLastViewedDate) {
```
tburry: Again, these are rated as mixed. You need to be very careful about protecting your data.

Also, your function names are too specific. You want to think about how your function can be useful beyond your use. How about something like `maxDate`? Isn't that what this function does?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1167,6 +1159,67 @@ public function calculate(&$discussion) {
...
+        if ($discussionLastViewedDate === null || $discussionLastViewedDate < $categoryLastViewedDate) {
```
tburry: `empty($discussionLastViewedDate)` is more robust. Also, you want to account for either date being empty.

tests/Models/DiscussionModelTest.php
```diff
@@ -208,4 +208,121 @@ public function testCanCloseDiscussionObject() {
...
+        );
+        $this->assertEquals($expected, $actual);
+    }
+
+    /**
+     * Provide test data for testDetermineReadStatusAndUnreadCount().
+     *
+     * @return array Returns an array of test data.
+     */
+    public function provideTestDetermineReadStatusAndUnreadCount() {
```
tburry: You need test cases where `$userLastReadDate` is empty.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1167,6 +1159,67 @@ public function calculate(&$discussion) {
...
+     * is more likely to be reliable because when Comments are deleted we update the Discussion table
+     * but not the UserDiscussion table.
+     *
+     * @param int $discussionCommentCount Number of Comments according to the Discussion table.
+     * @param mixed $discussionLastCommentDate Date of last Comment according to the Discussion table.
+     * @param int $userReadComments Number of Comments the user has read according to the UserDiscussion table.
+     * @param mixed $userLastReadDate Date of last Comment read according to the UserDiscussion table.
+     * @return array
+     */
+    public static function determineReadStatusAndUnreadCount(
```
tburry: This function name is too verbose. We also use "calculate" frequently in our code rather than "determine".

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1167,6 +1159,67 @@ public function calculate(&$discussion) {
...
+    /**
+     * Calculate if the user has read all the Comments in a discussion. The data in the Discussion table
+     * is more likely to be reliable because when Comments are deleted we update the Discussion table
+     * but not the UserDiscussion table.
+     *
+     * @param int $discussionCommentCount Number of Comments according to the Discussion table.
+     * @param mixed $discussionLastCommentDate Date of last Comment according to the Discussion table.
+     * @param int $userReadComments Number of Comments the user has read according to the UserDiscussion table.
+     * @param mixed $userLastReadDate Date of last Comment read according to the UserDiscussion table.
+     * @return array
```
tburry: Given that this returns an array with a specific format, you should document the return.

applications/vanilla/models/class.discussionmodel.php
blutarch: It turns out that this case is actually already covered earlier in the method:
```
      if ($discussionLastCommentDate > $userLastReadDate) {
            $isRead = false;
            if ($discussionCommentCount > 0 && $userReadComments >= $discussionCommentCount) {
                $unreadCommentCount = 1;
            }
        }
```
However, we don't have a case for `if ($unreadCommentCount === 0 $$ !isRead)`

Is it ok to just return 0 for the unreadCount and false for $isRead?

tburry:
Hey @blutarch, can you do some manual testing to make sure the set comment watch code does not update the database every page request? Can you report back here what test scenarios you did?

blutarch:
@tburry I tested this locally to make sure it didn't put extra stress on the db. 
I tested the following scenarios:
* New discussion, never read before
    * Hits this chunk:
```
            // Insert watch data.
            $this->SQL->options('Ignore', true);
            $this->SQL->insert(
                'UserDiscussion',
                [
                    'UserID' => $userID,
                    'DiscussionID' => $discussion->DiscussionID,
                    'CountComments' => $countWatch,
                    'DateLastViewed' => DateTimeFormatter::timeStampToDateTime(time())
                ]
            );
```
* Discussion with unread comments
    * Hits this chunk:
```
$this->SQL->put(
                    'UserDiscussion',
                    [
                        'CountComments' => $countWatch,
                        'DateLastViewed' => Gdn_Format::toDateTime()
                    ],
                    [
                        'UserID' => $userID,
                        'DiscussionID' => $discussion->DiscussionID
                    ]
                );
```
* Read discussion, no new comments
    * Doesn't touch the db.

This seems to be working as intended.

tburry:
## User Tests

- [x] Delete GDN_UserDiscussion; everything should appear new.
- [x] Read a discussion; should mark read.
- [x] Read 1st page of multi-page discussion.
- [x] Click in to a partially read discussion; should offset.
- [x] Comment on read discussion; should appear new and offset.
- [x] GDN_UserDiscussion goes past discussion list, but new comment.
  - [x] Should appear new.
  - [x] When I click in the count should be fixed.

PR #9946, opened 2020-01-16
Theme preview implementation using theme providers

alex-mtl:


library/src/scripts/features/toaster/themePreview/ThemePreviewToastReducer.ts
```diff
@@ -0,0 +1,80 @@
...
+import { ICoreStoreState } from "@library/redux/reducerRegistry";
+
+export interface IPreviewToasterState {
+    applyStatus: ILoadable<{ themeID: number | string }>;
+    cancelStatus: ILoadable<{ themeID: number | string }>;
+}
+
+export interface IPreviewToasterStoreState extends ICoreStoreState {
+    themePreviewToaster: IPreviewToasterState;
+}
```
charrondev: Could we rename this file and move it up into `/theming`? I'm thinking `themePreviewReducer`

Nothing here is really specific to the toast itself.

library/src/scripts/theming/ThemesActions.ts
```diff
@@ -0,0 +1,93 @@
+/**
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license Proprietary
+ */
```
charrondev: Incorrect license for this repo.

library/src/scripts/theming/ThemesActions.ts
```diff
@@ -0,0 +1,93 @@
...
+    );
+
+    public static readonly putCurrentThemeACs = actionCreator.async<
+        { themeID: number | string },
+        IManageTheme,
+        IApiError
+    >("PUT_CURRENT");
+
+    public static readonly putPreviewThemeACs = actionCreator.async<IPutCurrentThemeRequest, IManageTheme, IApiError>(
+        "PUT_PREVIEW",
```
charrondev: We have a `ThemeActions` right next to this one. It seems really confusing to have both. Could we merge this into that one?

alex-mtl:
> This seems to be missing some reducer registrations in order to function without the new plugin. Essentially your stuff from the "common" entry in the API plugin probably needs to be moved up here.

I've tested on my localhost with plugin disabled. It works to me. Do you see something missing in particular?

slafleche:
I can't get this to run. Even with both branches up to date, i get an error.

charrondev:
The issue @slafleche had should be resolved by the merge to master that was done here.

PR #9949, opened 2020-01-17
Add edit log access from within a discussion

### LINKED ISSUE #9948: Allow access into the change log for revision history and deleted comments
1. You can view the edit history of individual discussions or comments by clicking on a link on each individual discussion/comment. This is how most every application works that shows edit histories. Seeing the edit history of an individual post, ordered by date is much easier for a moderator to read and understand the context of what a user was doing.

2. From a discussion, moderators will gave a link to view all deleted comments from that discussion. This will go to the log showing the deleted comments which would include the last text from each comment. In the case that the comments were deleted for offensive or abusive text then this view in and of itself will be useful for moderators.

tburry:
Closes #9948. See the linked issue for details.

### DIFF
--- a/applications/vanilla/views/discussion/helper_functions.php
+++ b/applications/vanilla/views/discussion/helper_functions.php
@@ -355,13 +355,47 @@ if (!function_exists('getDiscussionOptionsDropdown')):
 
         $dropdown->addLinkIf($canDismiss, t('Dismiss'), "vanilla/discussion/dismissannouncement?discussionid={$discussionID}", 'dismiss', 'DismissAnnouncement Hijack')
             ->addLinkIf($canEdit, t('Edit').$timeLeft, '/post/editdiscussion/'.$discussionID, 'edit')
+            ->addLinkIf($canTag, t('Tag'), '/discussion/tag?discussionid='.$discussionID, 'tag', 'TagDiscussion Popup');
+
+        if ($canEdit && $canAnnounce) {
+            $dropdown->addDivider();
+        }
+
+        $dropdown
             ->addLinkIf($canAnnounce, t('Announce'), '/discussion/announce?discussionid='.$discussionID, 'announce', 'AnnounceDiscussion Popup')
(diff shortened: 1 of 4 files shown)


applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -355,21 +355,55 @@ function getDiscussionOptionsDropdown($discussion = null) {
...
             ->addLinkIf($canClose, t($discussion->Closed ? 'Reopen' : 'Close'), '/discussion/close?discussionid='.$discussionID.'&close='.(int)!$discussion->Closed, 'close', 'CloseDiscussion Hijack')
             ->addLinkIf($canRefetch, t('Refetch Page'), '/discussion/refetchpageinfo.json?discussionid='.$discussionID, 'refetch', 'RefetchPage Hijack')
-            ->addLinkIf($canMove, t('Move'), '/moderation/confirmdiscussionmoves?discussionid='.$discussionID, 'move', 'MoveDiscussion Popup')
-            ->addLinkIf($canTag, t('Tag'), '/discussion/tag?discussionid='.$discussionID, 'tag', 'TagDiscussion Popup')
-            ->addLinkIf($canDelete, t('Delete Discussion'), '/discussion/delete?discussionid='.$discussionID.'&target='.$categoryUrl, 'delete', 'DeleteDiscussion Popup');
+            ->addLinkIf($canMove, t('Move'), '/moderation/confirmdiscussionmoves?discussionid='.$discussionID, 'move', 'MoveDiscussion Popup');
+
+        $hasDiv = false;
+        if ($canEdit && !empty(val('DateUpdated', $discussion))) {
```
initvector: I'll be able to see the "Revision History" item as a regular user who can edit their post, but I'll run into a permission error page when trying to access the log page, since it's gated behind moderator permissions.

applications/vanilla/views/discussion/helper_functions.php
```diff
@@ -454,6 +488,14 @@ function getCommentOptions($comment) {
                 'Url' => '/post/editcomment/'.$comment->CommentID,
                 'EditComment'
             ];
+
+            if (!empty(val('DateUpdated', $comment))) {
```
initvector: Same kinda permission issue with the discussion link. A user with the ability to edit their post will see the "Revision History" link, but it'll only lead them to a permission error page.

PR #9952, opened 2020-01-17
Update to allow post\patch endpoints to allow assets parameter to be array

cchabilall83:
This updates the POST\PATCH themes endpoints so that assets can be passed as an array or string. This lines up with the GET response.

required for https://github.com/vanilla/internal/pull/2184

related to https://github.com/vanilla/internal/issues/2140

### DIFF
--- a/applications/dashboard/controllers/api/ThemesApiSchemes.php
+++ b/applications/dashboard/controllers/api/ThemesApiSchemes.php
@@ -96,11 +96,11 @@ trait ThemesApiSchemes {
                     'description' => 'Parent theme template name.',
                 ],
                 'parentVersion:s' => [
-                   'description' => 'Parent theme template version/revision.',
+                    'description' => 'Parent theme template version/revision.',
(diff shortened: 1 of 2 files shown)


applications/dashboard/controllers/api/ThemesApiSchemes.php
```diff
@@ -96,11 +96,11 @@ private function themePostSchema(string $type = 'in'): Schema {
...
                 ],
                 'parentVersion:s' => [
-                   'description' => 'Parent theme template version/revision.',
+                    'description' => 'Parent theme template version/revision.',
                 ],
                 'assets?' => Schema::parse([
-                    "header:s?",
-                    "footer:s?",
+                    "header?" => $this->assetsPutArraySchema(),
+                    "footer?" => $this->assetsPutArraySchema(),
```
alex-mtl: Don't we need to update openapi yml config then as well?

PR #9955, opened 2020-01-18
Add tests for misc garden classes

tburry:


library/Garden/Web/Cookie.php
```diff
@@ -235,16 +237,6 @@ public function makeCookieHeader() {
...
-    /**
-     * Return the cookies that will be set in a format suitable for "Set-Cookie" HTTP headers.
-     *
-     * @return array Returns an array of "Set-Cookie" header values.
-     * @throws \Exception TODO.
-     */
-    public function makeSetCookieHeader() {
-        throw new \Exception('Not implemented.', 501);
-    }
-
```
tburry: This has been in a todo state for a long time. I would rather just remove it since it doesn't seem to be needed.

tests/Library/Garden/Web/DataTest.php
```diff
@@ -416,4 +416,30 @@ public function testMergeMetaArray() {
...
+     * Trying to get a data item from a non-array data is an exception.
+     */
+    public function testGetDataItemException() {
+        $data = new Data('foo');
+        $this->expectException(\Exception::class);
+        $data->getDataItem('0');
+    }
+
+    /**
+     * Trying to get a data item from a non-array data is an exception.
```
blutarch: I think the comment here should read:
```
/**
* Trying to set a data item on a non-array data is an exception.
*/

Comment (blutarch) [inline tests/Library/Garden/Web/DataTest.php]:
```suggestion
     * Trying to set a data item on a non-array data is an exception.
```

tburry:
It looks fine to be @blutarch. Thanks.

PR #9959, opened 2020-01-21
Fix permission check when using quotes

dannimajdalani:
closes https://github.com/vanilla/support/issues/195#issuecomment-576496613

When member level users quote inside a group, they are being caught by 'Vanilla.Discussions.View' permission on the 'Social Groups' category.

This PR overrides that permission.

### To Test

- Create a group/discussion (with group view permission revoked for members)
- as a member, quote a comment.

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -2986,6 +2986,24 @@ class DiscussionModel extends Gdn_Model {
     }
 
     /**
+     * Tests whether a user/admin has permission to view a specific discussion.
+     *
+     * @param object $discussion
+     * @return bool whether the user can view a discussion.
+     */
+    public function canViewDiscussion($discussion): bool {
+        $canViewDiscussion = true;
+        $session = $this->sessionInterface;
+        $userID = $session->UserID;
+        $canView = $this->canView($discussion, $userID);
+        $isModerator = $session->checkRankedPermission('Garden.Moderation.Manage');
+        if (!$canView && !$isModerator) {
+            $canViewDiscussion = false;
+        }
+        return $canViewDiscussion;
+    }
+
+    /**
      * Tests whether a user has permission for a discussion by checking category-specific permissions.
      *
      * Fires an event that can override the calculated permission.
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -2992,13 +2992,13 @@ class DiscussionModel extends Gdn_Model {
      * @return bool whether the user can view a discussion.
      */
     public function canViewDiscussion($discussion): bool {
-        $canViewDiscussion = true;
+        $canViewDiscussion = false;
         $session = $this->sessionInterface;
         $userID = $session->UserID;
         $canView = $this->canView($discussion, $userID);
         $isModerator = $session->checkRankedPermission('Garden.Moderation.Manage');
-        if (!$canView && !$isModerator) {
-            $canViewDiscussion = false;
+        if ($canView || $isModerator) {
+            $canViewDiscussion = true;
         }
         return $canViewDiscussion;
     }
--- a/plugins/Quotes/class.quotes.plugin.php
+++ b/plugins/Quotes/class.quotes.plugin.php
@@ -34,7 +34,8 @@ class QuotesPlugin extends Gdn_Plugin {
     /**
      * Set some properties we always need.
      */
-    public function __construct() {
+    public function __construct(DiscussionModel $discussionModel) {
+        $this->discussionModel = $discussionModel;
         parent::__construct();
 
         if (function_exists('ValidateUsernameRegex')) {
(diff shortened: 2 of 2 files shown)


plugins/Quotes/class.quotes.plugin.php
```diff
@@ -42,6 +45,8 @@ public function __construct() {
 
         // Whether to handle drawing quotes or leave it up to some other plugin
         $this->HandleRenderQuotes = c('Plugins.Quotes.RenderQuotes', true);
+        $discussionModel = new DiscussionModel();
+        $this->discussionModel = $discussionModel;
```
initvector: We should be able to inject this via the addon's constructor.

plugins/Quotes/class.quotes.plugin.php
```diff
@@ -422,29 +427,21 @@ protected function formatQuote($type, $id, &$quoteData, $format = false) {
...
         }
 
-        $discussionModel = new DiscussionModel();
         $type = strtolower($type);
         switch ($type) {
             case 'comment':
                 $commentModel = new CommentModel();
                 $data = $commentModel->getID($id);
-                $discussion = $discussionModel->getID(val('DiscussionID', $data));
+                $discussion = $this->discussionModel->getID(val('DiscussionID', $data));
```
initvector: I think you're fine here, but as a cautionary note (in case you weren't already considering it): you're going from a one-off instance to a persistent instance. This means the state of that instance will persist between usages across the various methods in this class, which can introduce unexpected consequences.

plugins/Quotes/class.quotes.plugin.php
```diff
@@ -544,6 +541,20 @@ protected function formatQuote($type, $id, &$quoteData, $format = false) {
         Emoji::instance()->enabled = $emojiEnabled;
     }
 
+    /**
+     * Checks if the user can view discussion.
+     *
+     * @param object $discussion
+     * @throws Exception If the user cannot view the discussion.
+     */
+    private function canViewDiscussion(object $discussion): void {
```
initvector: This is more of a nitpick, but naming like this implies there will be a boolean return value. It's like you're asking a question: "Can this user view the discussion, yes or no?" and the answer would be the return value. This is reinforced by `DiscussionModel`'s own `canView` method, which returns a value. It might make more sense to phrase this in the imperative mood, like `verifyDiscussionViewPermissions` or something.

plugins/Quotes/class.quotes.plugin.php
```diff
@@ -544,6 +541,20 @@ protected function formatQuote($type, $id, &$quoteData, $format = false) {
...
+        $userID = Gdn::session()->UserID;
+        $canView = $this->discussionModel->canView($discussion, $userID);
+        $isAdmin = Gdn::session()->checkRankedPermission('Garden.Moderation.Manage');
```
initvector: `$isAdmin` for moderator permissions?

plugins/Quotes/class.quotes.plugin.php
```diff
@@ -544,6 +541,20 @@ protected function formatQuote($type, $id, &$quoteData, $format = false) {
         Emoji::instance()->enabled = $emojiEnabled;
     }
 
+    /**
+     * Checks if the user can view discussion.
+     *
+     * @param object $discussion
+     * @throws Exception If the user cannot view the discussion.
```
initvector: I know this is what `permissionException` documents as its return type. While technically true, you are actually throwing an instance of `Gdn_UserException`, specifically. We should document specific exceptions, whenever possible. It's probably worth updating the doc block of `permissionException` to also reflect the proper return value.

plugins/Quotes/class.quotes.plugin.php
initvector: Do you think some form of this method would be beneficial to add to `DiscussionModel`?

plugins/Quotes/class.quotes.plugin.php
dannimajdalani: it's a ranked permission. The user could be an admin.

initvector: They could be. But that's not what we're testing for, specifically. This code reads like "Are they an admin? Check to see if they have moderator permissions."

plugins/Quotes/class.quotes.plugin.php
dannimajdalani: I've already used it twice. It would be beneficial.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2985,6 +2985,24 @@ public function canView($discussion, $userID = 0) {
         return $canView;
     }
 
+    /**
+     * Tests whether a user/admin has permission to view a specific discussion.
```
initvector:
```suggestion
     * Tests whether a user has permission to view a specific discussion, with consideration for mod permissions.
```

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2985,6 +2985,24 @@ public function canView($discussion, $userID = 0) {
...
     }
 
+    /**
+     * Tests whether a user has permission to view a specific discussion, with consideration for mod permissions.
+     *
+     * @param object $discussion
+     * @return bool whether the user can view a discussion.
+     */
+    public function canViewDiscussion($discussion): bool {
+        $canViewDiscussion = true;
```
initvector: This is probably more of a style choice, but I do recommend you default to `false` and only flip it to `true` when you've confirmed the user has access to the discussion. If the default is `false`, we have to prove the user has that access.

dannimajdalani: I agree. Defaulting to false it is.

PR #9963, opened 2020-01-21
Fix theme preview with wrong preview theme stored in session

alex-mtl:


alex-mtl:
something wrong with this branch

PR #9964, opened 2020-01-21
Fix edge case when wrong theme preview stored in session data

alex-mtl:

Some extra fixes related to: https://github.com/vanilla/internal/pull/2183


slafleche:
Does this also handle the information saved in the database. @charrondev ?

PR #9965, opened 2020-01-21
Allow Rich content to be converted if Garden.ForceInputFormatter exists

cchabilall83:
Editing rich content when `Garden.ForceInputFormatter` is `true` causes unformatted json to be rendered in a discussion or comment if the InputFormatter is set to something other than `Rich`.  This detects if `Garden.ForceInputFormatter` and if the Forced InputFormat is different then the original contents allows for it to be modified.

To Test

1) Set Garden.ForceInputFormatter = true` in config.php
2) post rich discussion and comment
3) Set posting format to TextEx
4) Try edit the discussion and comment
5) Content should save and display properly 

closes https://github.com/vanilla/support/issues/1210


plugins/rich-editor/RichEditorPlugin.php
```diff
@@ -100,7 +111,29 @@ public function gdn_form_beforeBodyBox_handler(Gdn_Form $sender, array $args) {
...
+        } elseif(c('Garden.ForceInputFormatter')) {
+
+            $originalRecord = $sender->formData();
+            $newBodyValue = null;
+            $body = $originalRecord['Body'] ?? false;
+            $originalFormat = $originalRecord['Format'] ?? false;
+
+            if ($body && (c('Garden.InputFormatter') !== $originalFormat ) ) {
+                switch (strtolower(c('Garden.InputFormatter', 'unknown'))) {
+                    case Formats\TextFormat::FORMAT_KEY:
```
charrondev: Some docblock of why this is being done would be nice here.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2138,8 +2138,8 @@ public function save($formPostValues, $settings = false) {
                     $stored = $this->getID($discussionID, DATASET_TYPE_OBJECT);
 
                     // Block Format change if we're forcing the formatter.
```
charrondev: You should update this description too since we're updating the code.

cchabilall83: done

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2137,9 +2137,9 @@ public function save($formPostValues, $settings = false) {
                     // Updating
                     $stored = $this->getID($discussionID, DATASET_TYPE_OBJECT);
 
-                    // Block Format change if we're forcing the formatter.
-                    if (c('Garden.ForceInputFormatter')) {
-                        unset($fields['Format']);
+                    // Make sure that the discussion get formatted in the method defined by Garden.
+                    if (!val('Format', $fields) || c('Garden.ForceInputFormatter')) {
```
alex-mtl: `val()` function is deprecated we use coalescing  operator instead
```suggestion
                    if (!($fields['Format'] ?? false) || c('Garden.ForceInputFormatter')) {
```

plugins/rich-editor/RichEditorPlugin.php
```diff
@@ -18,11 +19,17 @@ class RichEditorPlugin extends Gdn_Plugin {
...
+
     /**
      * Set some properties we always need.
+     *
+     * @param \Vanilla\Formatting\FormatService $formatService
      */
-    public function __construct() {
+    public function __construct(\Vanilla\Formatting\FormatService $formatService) {
         parent::__construct();
+        $this->formatService = $formatService;
```
alex-mtl: I am not sure about the order. I have a feeling that it is better to have it reordered.
```suggestion
        $this->formatService = $formatService;
        parent::__construct();
```

But not sure about `self::$editorID++;` part

PR #9981, opened 2020-01-23
Insert logic to validate empty string where the default is not null

blutarch:

Along with vanilla/internal#2197, this will close issue vanilla/support#1239

The updated comboBreaker data was not being saved to the db. 
Because the badge status was an empty string and didn't have a default value of null, it was not passing validation. This fix adds logic to validateRequired() in functions.validation that allows it to pass validation.

### To Test:
* Open the GDN_UserBadge table
* Give a user a badge
* Verify that the Attributes field for the comboBreaker badge (BadgeID = 3) updates with a new event
* Give that user 4 more badges
* Verify that the status field for the comboBreaker badge switches from NULL to given

### DIFF
--- a/library/core/functions.validation.php
+++ b/library/core/functions.validation.php
@@ -63,11 +63,8 @@ if (!function_exists('ValidateRequired')) {
 
             // Checking for an Enum with an empty string.
             $hasEmptyEnum = false;
-            if ($field instanceof ArrayObject && $field->offsetExists('Enum')) {
-                $enum = $field->offsetGet('Enum');
-                if (is_array($enum) && in_array('', $enum, true)) {
-                    $hasEmptyEnum = true;
-                }
+            if (is_array($enum = val('Enum', $field, null))) {
+                $hasEmptyEnum = in_array('', $enum, true);
             }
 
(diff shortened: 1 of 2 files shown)


tests/Library/Core/ValidateRequiredTest.php
```diff
@@ -35,34 +35,64 @@ public function testValidateRequired($testValue, $testField, $expected) {
      */
     public function provideTestValidateRequiredArrays() {
         $r = [
-            'valueEmptyString' => [
+//            'valueEmptyString' => [
```
tburry: Whoops. Need to uncomment.

library/core/functions.validation.php
```diff
@@ -58,8 +58,16 @@ function validateRequired($value, $field = null) {
         }
 
         if (is_string($value)) {
```
tburry: Nothing wrong here, but can you update the file copyright to 2020.

dannimajdalani:
**Tested**
UserBadge.Attributes updating with the events
Screen Shot 2020-01-27 at 1 17 27 PM
Screen Shot 2020-01-27 at 1 18 56 PM

PR #9993, opened 2020-01-27
Fix SchemaFactory not firing events on schema objects

initvector:
A typo in `SchemaFactory::prepare` caused an event to not be dispatched when expected. The bug is fixed here by checking the proper variable and tests have been implemented to protect against regressions of the behavior.


initvector:
This issue is currently breaking some builds, so I'm going to flex and force a merge.

Issue #9972, opened 2020-01-23
Always display the logo in the mobile title bar


tburry:
As a brand, I want my logo always to be displayed because its important for people to know my company at all times on my site.

## Acceptance Criteria

- The logo will go on the title bar and replace the central text portion.
- There should be an option for a left aligned logo on mobile. In this case the logo will go after the hamburger menu with some acceptable padding.
- Editor pages don't have the logo. They stay as is.
- We currently have some KB pages that require a title. In this case the pages can move their title to an h1 on the page.
- There are some pages where the title acts as its own hamburger menu. In this case we have another PBI to move that navigation into the global hamburger menu.
- This feature will require some acceptance testing to make sure all of the pages in the KB still function properly. Grab Todd for this.

PR #10000, opened 2020-01-28
Add logo to mobile titleBar

NeenaDev:
Fix: https://github.com/vanilla/vanilla/issues/9972
Relates: https://github.com/vanilla/knowledge/pull/1495
Changes-
- Add Logo in mobile tileBar 
- Add align option for mobile logo- {mobileLogo.justifyContent} - default value is centre 

TEST - 
Try variable set:
`{
    "titleBar": {
        "mobileLogo": {
            "justifyContent": "left"  
        }
    }
}`

Closes: https://github.com/vanilla/vanilla/issues/9972


library/src/scripts/forms/buttonStyles.ts
```diff
@@ -50,6 +51,9 @@ export const buttonGlobalVariables = useThemeCache(() => {
         minHeight: formElVars.sizing.height,
         minWidth: 104,
         compactHeight: 24,
+        mobile: {
+            minHeight: 24,
+        },
```
charrondev: Could we not use a new variable here and just use the `compactHeight`? I think that's the purpose.

library/src/scripts/headers/titleBarStyles.ts
```diff
@@ -149,7 +149,8 @@ export const titleBarVariables = useThemeCache(() => {
     const endElements = makeThemeVars("endElements", {
         flexBasis: buttonSize * 4,
         mobile: {
-            flexBasis: button.mobile.width * 2,
+            //flexBasis: button.mobile.width * 2,
```
charrondev:
```suggestion
```

Commented line

library/src/scripts/forms/buttonStyles.ts
NeenaDev: Fixed - use compactHeight

library/src/scripts/headers/titleBarStyles.ts
NeenaDev: Delete commented line

library/src/scripts/headers/titleBarStyles.ts
```diff
@@ -205,6 +205,10 @@ export const titleBarVariables = useThemeCache(() => {
         tablet: {},
     });
 
+    const mobileLogo = makeThemeVars("mobileLogo", {
+        align: "center",
```
slafleche: Since you're using flexbox here and not text-align, I think it would be clearer to use: "justifyContent" as a variable name. I gave the same feedback to Chris for the nav alignment, so I think it would be better.

library/src/scripts/headers/TitleBar.tsx
```diff
@@ -76,12 +81,14 @@ export default function TitleBar(_props: IProps) {
...
     const isGuest = isUserGuest(currentUser.data);
-
     const vars = titleBarVariables();
     const classes = titleBarClasses();
+    const logoClasses = titleBarLogoClasses();
     const homeClasses = titleBarHomeClasses();
     const showSubNav = device === TitleBarDevices.COMPACT && props.hasSubNav;
     const meBox = isCompact ? !isSearchOpen && <MobileMeBox /> : <DesktopMeBox />;
+    const logoAlignment = titleBarVariables().mobileLogo.align;
+    const logoAlign = logoAlignment === "center" ? true : false;
```
slafleche: Use the prop directly in the class, no need for a condition here.

library/src/scripts/headers/titleBarStyles.ts
```diff
@@ -149,7 +149,7 @@ export const titleBarVariables = useThemeCache(() => {
     const endElements = makeThemeVars("endElements", {
         flexBasis: buttonSize * 4,
         mobile: {
-            flexBasis: button.mobile.width * 2,
+            flexBasis: button.mobile.width - 20,
```
slafleche: Where does this `20` come from?

library/src/scripts/headers/titleBarStyles.ts
charrondev: @slafleche I disagree entirely here. Just because we use flex for alignment today, doesn't mean that variable only directly translates to flex property. It may be used in multiple classes, or we may switch away from using flex for the alignment in the future.

charrondev: From the perspective of someone utililizing the variables, I shouldn't actually care whether it's flexbox or not.

PR #10004, opened 2020-01-28
Filter emojis when parsing Image Urls in HtmlFormatter

cchabilall83:
Filters out emoji images when parsing image urls.

closes vanilla/support#1236
related to https://github.com/vanilla/support/issues/1236

### DIFF
--- a/library/Vanilla/Formatting/Formats/HtmlFormat.php
+++ b/library/Vanilla/Formatting/Formats/HtmlFormat.php
@@ -185,6 +185,9 @@ class HtmlFormat extends BaseFormat {
         /** @var \DOMNode $domImages */
         foreach ($domImages as $domImage) {
             $src = $domImage->getAttribute('src');
+            if (preg_match("/emoji/", $src)) {
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


library/Vanilla/Formatting/Formats/HtmlFormat.php
```diff
@@ -185,6 +185,9 @@ public function parseImageUrls(string $content): array {
         /** @var \DOMNode $domImages */
         foreach ($domImages as $domImage) {
             $src = $domImage->getAttribute('src');
+            if (preg_match("/emoji/", $src)) {
```
slafleche: Do we know where in the path the "/emoji" will be? If we can be more specific, I think we should... just to make sure no outside paths have "emoji" in it.

charrondev: @cchabilall83 You should just be able to check the `class` attribute.

cchabilall83: done

Issue #9971, opened 2020-01-23
Center aligned menu option for title bars


tburry:
As a themer, I should be able to center my title bar menu because some designs feel more balanced that way.

## Acceptance Criteria

- After done, talk to Todd to update the docs.

PR #10010, opened 2020-01-28
Add  titleBarNavAlignment variable  

cchabilall83:
Allows title bar nav menu to be centered.  Current implementation supports 'center' and 'left'.
```
{
    "titleBar": {
        "navAlignment": {
            "alignment": "center"
        }
    }
}
```

Screen Shot 2020-01-28 at 5 52 58 PM

closes https://github.com/vanilla/vanilla/issues/9971


library/src/scripts/headers/titleBarStyles.ts
```diff
@@ -92,6 +92,10 @@ export const titleBarVariables = useThemeCache(() => {
         },
     });
 
+    const titleBarNavAlignment = makeThemeVars("titleBarNavAlignment", {
```
slafleche: You can remove the "titleBar" part, it's redundant, because it's already scoped in "titleBar". Look at the class generated, it'll be "titleBar-titleBarNavAlignment", i'd rather get "titleBar-navAlignment"

cchabilall83: done

library/src/scripts/headers/titleBarStyles.ts
```diff
@@ -92,6 +92,10 @@ export const titleBarVariables = useThemeCache(() => {
         },
     });
 
+    const titleBarNavAlignment = makeThemeVars("NavAlignment", {
+        justifyContent: "flex-start",
+    });
+
```
charrondev: - `NavAlignment` should be `navAlignment` for consistency.
- We variable should probably be `alignment: 'left' as 'left' | 'center'`. From their we can translate it into a flex justification.

When making variables it's best not to couple them to the current exact implementation. In the future, multiple properties might need to get set (not just the flex justification) or it might not use flex at all. By using a more generic variable name we are freerer to change that implementation in the future.

library/src/scripts/headers/titleBarStyles.ts
```diff
@@ -92,6 +92,10 @@ export const titleBarVariables = useThemeCache(() => {
         },
     });
 
+    const navAlignment = makeThemeVars("navAlignment", {
+        alignment: "left",
```
charrondev:
```suggestion
        alignment: "left" as "left" | "center",
```

Better to tighten up this string a bit.

PR #10014, opened 2020-01-30
Implement flexible DefaultController

alex-mtl:
Relates to: https://github.com/vanilla/multisite/issues/328

### DIFF
--- a/library/Vanilla/Site/SiteSectionModel.php
+++ b/library/Vanilla/Site/SiteSectionModel.php
@@ -28,13 +28,16 @@ class SiteSectionModel {
     /** @var SiteSectionInterface $currentSiteSection */
     private $defaultSiteSection;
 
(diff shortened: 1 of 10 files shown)


library/Vanilla/Site/SiteSectionModel.php
```diff
@@ -49,6 +52,41 @@ public function addProvider(SiteSectionProviderInterface $provider) {
...
+
+    /**
+     * Get layout options
+     *
+     * @return array
+     */
+    public function getLayoutOptions(): array {
+        $layouts = [
+            'discussions' => 'Discussions',
+            'categories' => 'Categories'
```
charrondev: I think these need to be translated.

alex-mtl: I think it is should be translated during rendering on back end or front end component, but not in core component.

PR #10035, opened 2020-02-05
fix adding required user fields when approving a post from moderation

dannimajdalani:
closes https://github.com/vanilla/support/issues/1309

### Issue 

When a post gets sent to moderation after being edited->saved, the post is missing InsertUserID and DateInserted.

### Details
whenever a plugin hooks into https://github.com/vanilla/vanilla/blob/03eba20c06cf9678556c4b87cc964f5ee1fdfb86/applications/vanilla/models/class.commentmodel.php#L1148, $commentData is missing InsertUserID and DateInserted. If a post goes to moderation and then it gets approved, the post gets overwritten here https://github.com/vanilla/vanilla/blob/cb23d99019c198ea2f7aeeb903f8a93eaa615b07/applications/dashboard/models/class.logmodel.php#L967 with the missing information.

There is also another bug that was introduced in https://github.com/vanilla/vanilla/pull/9944 when creating a discussion, if a discussion gets sent to moderation before being posted(doesn't have any comment DateLastComment is usually null https://github.com/vanilla/vanilla/blob/03eba20c06cf9678556c4b87cc964f5ee1fdfb86/applications/vanilla/models/class.discussionmodel.php#L1140 I set that paramter as optional.

### To test

Test posting comments and discussion, you can use this plugin https://github.com/vanillaforums/niantic/tree/master/plugins/nianticimagemoderation to send a post to moderation if it has an image
create a comment. edit it, add an image
as a mod approve the post, notice the fields InsertUserID and DateInserted.
Do the same as a discussion

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -1129,8 +1129,11 @@ class CommentModel extends Gdn_Model {
             if (!$insert || !$this->checkUserSpamming(Gdn::session()->UserID, $this->floodGate)) {
                 $fields = $this->Validation->schemaValidationFields();
                 unset($fields[$this->PrimaryKey]);
+                $comment = $this->getID($commentID, DATASET_TYPE_ARRAY);
+                $insertUserID = $comment['InsertUserID'] ?? null;
+                $dateInserted = $comment['DateInserted'] ?? null;
 
-                $commentData = $commentID ? array_merge($fields, ['CommentID' => $commentID]) : $fields;
+                $commentData = $commentID ? array_merge($fields, ['CommentID' => $commentID, 'InsertUserID' => $insertUserID, 'DateInserted' => $dateInserted]) : $fields;
                 // Check for spam
                 $spam = SpamModel::isSpam('Comment', $commentData);
                 if ($spam) {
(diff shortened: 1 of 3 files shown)


applications/vanilla/models/class.commentmodel.php
```diff
@@ -1129,8 +1129,11 @@ public function save($formPostValues, $settings = false) {
             if (!$insert || !$this->checkUserSpamming(Gdn::session()->UserID, $this->floodGate)) {
                 $fields = $this->Validation->schemaValidationFields();
                 unset($fields[$this->PrimaryKey]);
+                $comment = $this->getID($commentID, DATASET_TYPE_ARRAY);
+                $insertUserID = $comment['InsertUserID'] ?? null;
```
tburry: Can't you just unset InsertUserID from `$fields` and not fetch the record?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1184,7 +1184,7 @@ public function calculate(&$discussion) {
      */
     public function calculateCommentReadData(
         int $discussionCommentCount,
-        string $discussionLastCommentDate,
+        ?string $discussionLastCommentDate,
```
tburry: This needs a unit test.

applications/vanilla/models/class.commentmodel.php
dannimajdalani: InsertUserID is not set on the $fields array we need to add it before firing AfterValidateComment.

PR #10036, opened 2020-02-05
Removed customizations on Foundation, it should be "vanilla"

slafleche:
Removed custom styles in Foundation, as it should be the defaults.


charrondev:
@slafleche Is this still needed?

cchabilall83:
as discussed with @charrondev @slafleche this is no longer needed

PR #10039, opened 2020-02-06
Add a sensible default Smarty template directory

tburry:
Currently, Smarty uses the directory of the template being rendered as the root path for included partials. This currently has two issues:

1. If the template is specified using the `file:` protocol then the function will fail. **This is fixed with a specific logic check.**
2. If some other protocol is used then a nonsense path is used. **This is fixed by falling back to the theme’s view path.

### DIFF
--- a/library/core/class.smarty.php
+++ b/library/core/class.smarty.php
@@ -33,10 +33,10 @@ class Gdn_Smarty implements \Vanilla\Contracts\Web\LegacyViewHandlerInterface {
         }
 
         // Get an ID for the body.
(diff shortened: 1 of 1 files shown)


nervoustwit:
When I test this I still get the same error. When I put break points in my code there is not time when $smarty->render() receives a $path  that starts 'file:'. It is much later, in class Smarty_Internal_Template that we parse out the partials. Maybe I'm missing something.

tburry:
@nervoustwit, this fix doesn't address all build errors in and of itself. There is another PR that works in conjunction with this one. The `file:` I put in because a) It is the correct logic, and b) I am doing an explode anyway.

PR #10052, opened 2020-02-07
Fix package licenses

charrondev:
Add licenses for all NPM packages.

### DIFF
--- a/packages/vanilla-dom-utils/package.json
+++ b/packages/vanilla-dom-utils/package.json
@@ -3,6 +3,7 @@
     "main": "./src/index.ts",
     "module": "./src/index.ts",
     "version": "3.0.1",
+    "license": "gpl-2.0-only",
     "dependencies": {
         "@vanilla/utils": "^3.0.1",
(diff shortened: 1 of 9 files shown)


charrondev:
@tburry All of the addons have "private: true" in their configs. They do not get published.

https://github.com/vanilla/vanilla/blob/474d413c44bb2d1f13c4ed9f909f59983dd8f2db/plugins/rich-editor/package.json#L4

I'm not sure it's necessary when the repo itself contains a license declaration, all of the files have it in their headers, and their addon.json already has one.

I can still proceed with that, just not sure what benefit we get from it.

tburry:
It's for tooling that checks licenses. I guess we can have a thing where if a package is private AND it has no dependencies it doesn't need a license. I'm just not sure if it's easier to just tell people to always add a license.

tburry:
Okay I'm fine leaving the licenses out of the other packages, but can you use the same casing as SPDX? (i.e. GPL-2.0-only)

PR #10061, opened 2020-02-09
Relax DateTimeFormatter::dateTmeToTimesStamp()

### LINKED ISSUE #9489: Refactor DateTimeFormatter::dateTimeToTimeStamp() RFC
I noticed `DateTimermatter::dateTimeToTomeStamp()` triggering its warning in my output logs and found a few issues with the function. I think this function needs a bit of refactoring. I'm going to say what I think should be done and them my reasons why.

First, the function in question:

https://github.com/vanilla/vanilla/blob/50f75e3b87306a544ba93260140e27cf397e0ee5/library/Vanilla/Formatting/DateTimeFormatter.php#L223-L231

## Changes

1. The `$dateTime` parameter should not have a default. The default serves no purpose and leads to a warning anyway.
2. Just make the `$fallback` parameter the fallback. The caller can pass `time()` if they want that as the fallback.
3. Don't trigger a warning on fallback. Just return the fallback.
4. It's `Timestamp`, not `TimeStamp`.

Triggering a warning is at odds with the seeming purpose of this method, which is to cast a value into a timestamp or use the fallback I supplied. A warning on an expected output means that the programmer is going to have to do special handling. I would think the purpose of this method is to remove the need for special handling. Moreover, there are already other functions that also require special handling.

- I could use `strtotime()` and check against `false`.
- I could use the `DateTime` constructor and check for an exception.

In your case, I just want to return my fallback.

tburry:
- Trigger notices instead of warnings.
- Add a third parameter for empty/null fallbacks.
- Allow the method to return null instead of just an integer.

Closes #9489

### DIFF
--- a/library/Vanilla/Formatting/DateTimeFormatter.php
+++ b/library/Vanilla/Formatting/DateTimeFormatter.php
@@ -218,14 +218,18 @@ class DateTimeFormatter {
      * @param string $dateTime The Mysql-formatted datetime to convert to a timestamp. Should be in one
      * of the following formats: YYYY-MM-DD or YYYY-MM-DD HH:MM:SS.
      * @param mixed $fallback The value to return if the value couldn't be properly converted.
-     * @return int A timestamp or now if it couldn't be parsed properly.
+     * @param mixed $emptyFallback The fallback for an empty value. If not supplied then the `$fallback` will be used.
+     * @return int|null A timestamp or now if it couldn't be parsed properly.
      */
-    public static function dateTimeToTimeStamp(?string $dateTime = '', $fallback = null): int {
-        if (($testTime = strtotime($dateTime)) !== false) {
+    public static function dateTimeToTimeStamp(?string $dateTime, $fallback = false, $emptyFallback = false): ?int {
+        if (empty($dateTime)) {
+            $emptyFallback = $emptyFallback !== false ? $emptyFallback : $fallback;
+            return $emptyFallback !== false ? $emptyFallback : time();
+        } elseif (($testTime = strtotime($dateTime)) !== false) {
             return $testTime;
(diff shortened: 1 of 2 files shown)


library/Vanilla/Formatting/DateTimeFormatter.php
```diff
@@ -218,14 +218,18 @@ public function formatSeconds(int $seconds): string {
      * @param string $dateTime The Mysql-formatted datetime to convert to a timestamp. Should be in one
      * of the following formats: YYYY-MM-DD or YYYY-MM-DD HH:MM:SS.
      * @param mixed $fallback The value to return if the value couldn't be properly converted.
-     * @return int A timestamp or now if it couldn't be parsed properly.
+     * @param mixed $emptyFallback The fallback for an empty value. If not supplied then the `$fallback` will be used.
+     * @return int|null A timestamp or now if it couldn't be parsed properly.
      */
-    public static function dateTimeToTimeStamp(?string $dateTime = '', $fallback = null): int {
-        if (($testTime = strtotime($dateTime)) !== false) {
+    public static function dateTimeToTimeStamp(?string $dateTime, $fallback = false, $emptyFallback = false): ?int {
```
initvector: We did go from making the first parameter optional, to making it required. Was this intentional?

tburry: That was intentional, yes. It makes no sense to supply no arguments to a method that's entire purpose is to operate on the first parameter. The method as it was before also had a default that was a warning.

tburry: As a side note. I don't see any calls that don't supply any arguments across our codebase.

PR #10067, opened 2020-02-10
Fix SiteTotalsModule by implementing lock and job scheduling

igraziatto:
Closes https://github.com/vanilla/support/issues/1405

### DIFF
--- a/applications/dashboard/modules/class.sitetotalsmodule.php
+++ b/applications/dashboard/modules/class.sitetotalsmodule.php
@@ -8,22 +8,31 @@
  * @since 2.0
  */
 
+use Vanilla\Scheduler\Job\CallbackJob;
+use Vanilla\Scheduler\SchedulerInterface;
+
 /**
  * Class SiteTotalsModule
  */
 class SiteTotalsModule extends Gdn_Module {
 
     /** @var int */
-    const CACHE_TTL = 3600;
+    const CACHE_TTL = 43200;
 
     /** @var int */
-    const RECALCULATE_INTERVAL = 300;
+    const RECALCULATE_INTERVAL = 900;
+
+    /** @var int */
+    const LOCK_INTERVAL = 30;
 
     /** @var string */
     const CACHE_KEY = 'module.sitetotals';
 
     /** @var string */
     const COUNTS_KEY = self::CACHE_KEY.'.counts';
+
+    /** @var string */
+    const LOCK_KEY = self::CACHE_KEY.'.lock';
 
     /** @var string */
     const RECALCULATE_KEY = self::CACHE_KEY.'.recalculate';
@@ -52,7 +61,7 @@ class SiteTotalsModule extends Gdn_Module {
         // check recalculate flag
         $recalculateFlag = Gdn::cache()->get(self::RECALCULATE_KEY);
 
-        if ($recalculateFlag !== Gdn_Cache::CACHEOP_FAILURE) {  // expired
+        if ($recalculateFlag === Gdn_Cache::CACHEOP_FAILURE) {  // expired
             $this->tryRecalculate();
         }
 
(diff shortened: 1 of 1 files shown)


applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,121 @@
...
  */
 
 /**
- * Site-wide stats.
+ * Class SiteTotalsModule
  */
 class SiteTotalsModule extends Gdn_Module {
 
+    /** @var int */
+    const CACHE_TTL = 3600;
```
kaecyra: Let's make this longer. 43200 (half a day) might be a good start.

applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,121 @@
...
+
+    /** @var int */
+    const RECALCULATE_INTERVAL = 300;
```
kaecyra: Lets make this 900 (15 minutes)

applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,121 @@
...
+     * cache key already exists, which would mean the lock is already in place.
+     */
+    private function tryRecalculate() {
+        $lockKey = mt_rand(0, 9999999);
+        $added = Gdn::cache()->add(self::RECALCULATE_KEY, $lockKey, [Gdn_Cache::FEATURE_EXPIRY => self::RECALCULATE_INTERVAL]);
+
+        if ($added) {
+            /** @var Vanilla\Scheduler\SchedulerInterface $scheduler */
+            $scheduler = Gdn::getContainer()->get(Vanilla\Scheduler\SchedulerInterface::class);
+            $scheduler->addJob($this->getAllCounts());
```
kaecyra: This isn't how closures work. I *think* this will execute `$this->getAllCounts()` right there, rather than later.

You probably want something more like:

```
        $scheduler->addJob(
            Vanilla\Scheduler\Job\CallbackJob::class,
            [
                "callback" => function () {
                    $this->getAllCounts();
                }
            ]
        );
```

applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,121 @@
...
-        $count = Gdn::cache()->get($key);
-        if ($count !== Gdn_Cache::CACHEOP_FAILURE) {
-            return $count;
+    /**
+     * Attempt to implement lock system. A failure likely means the
+     * cache key already exists, which would mean the lock is already in place.
+     */
+    private function tryRecalculate() {
+        $lockKey = mt_rand(0, 9999999);
+        $added = Gdn::cache()->add(self::RECALCULATE_KEY, $lockKey, [Gdn_Cache::FEATURE_EXPIRY => self::RECALCULATE_INTERVAL]);
```
kaecyra: This implementation means that if something fails during recalculation, we'll only try again in 900 seconds (self::RECALCULATE_INTERVAL). My original spec called for a 3rd key, used for locking the recalculation operation, with a much shorter lifetime of 30 seconds.

`tryRecalculate()`
* start by `Gdn::cache()->get()` checking for self::RECALCULATE_LOCK_KEY. 
* Not found? `Gdn::cache()->add(self::RECALCULATE_LOCK_KEY, mt_rand(0,999999), [Gdn_Cache::FEATURE_EXPIRY => self::RECALCULATE_LOCAL_TTL])`
* $added? queue job to run `$this->getAllCounts()`

`getAllCounts` should `store()` the self::RECALCULATE_KEY once it succeeds.

applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,144 @@
...
-        // Try and get the count from the cache.
-        $key = "$table.CountRows";
-        $count = Gdn::cache()->get($key);
-        if ($count !== Gdn_Cache::CACHEOP_FAILURE) {
-            return $count;
+    /**
+     * Attempt to implement lock system. A failure likely means the
+     * cache key already exists, which would mean the lock is already in place.
+     */
+    private function tryRecalculate() {
```
kaecyra: This method is returning a boolean on some case, nothing in others. If you refactor it slightly to make the "happy path" lead to one place without branching and `return true` in there, you can put a trailing `return false` at the end and make it a `: boolean` signature method.

So do:

```
$lock = Gdn::cache()->get(self::LOCK_KEY);

if ($lock !== Gdn_Cache::CACHEOP_SUCCESS) { 
    $added = Gdn::cache()->add(self::LOCK_KEY, mt_rand(0, 999999), [Gdn_Cache::FEATURE_EXPIRY => self::LOCK_INTERVAL]);

    if ($added) {
        /** @var Vanilla\Scheduler\SchedulerInterface $scheduler */
        $scheduler = Gdn::getContainer()->get(Vanilla\Scheduler\SchedulerInterface::class);
        $scheduler->addJob(
            Vanilla\Scheduler\Job\CallbackJob::class,
            [
                "callback" => function () {
                    $this->getAllCounts();
                }
            ]
        );
        return true;
    }
}

return false;
```

applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,144 @@
...
+    private function getAllCounts() {
+        $counts = ['User' => 0, 'Discussion' => 0, 'Comment' => 0];
+
+        foreach ($counts as $name => $value) {
+            $counts[$name] = $this->getCount($name);
         }
 
-        // The count wasn't in the cache so grab it from the table.
+        // cache counts
+        Gdn::cache()->store(self::COUNTS_KEY, $counts, [Gdn_Cache::FEATURE_EXPIRY => self::CACHE_TTL]);
```
kaecyra: Before storing this, you probably want to iterate over `$counts` and check each value for being `=== null` or `=== false` and then don't overwrite the cache if any are not valid. Otherwise you'll break the cached data.

So you'd bail out of this method and not overwrite the COUNTS_KEY and not re-set the RECALCULATE_KEY.

applications/dashboard/modules/class.sitetotalsmodule.php
```diff
@@ -2,58 +2,144 @@
...
+        Gdn::cache()->store(self::RECALCULATE_KEY, time(), [Gdn_Cache::FEATURE_EXPIRY => self::RECALCULATE_INTERVAL]);
+    }
+
+    /**
+     * Query the DB for count
+     *
+     * @param string $table
+     * @return mixed
+     */
+    protected function getCount($table) {
```
kaecyra: Just want to confirm that this will return `false` or `null` if it fails for some reason? If so, good. And you should document that in the docblock.

PR #10068, opened 2020-02-10
Adjust home widget styles

charrondev:
- Fix empty spacers not having min-width.
  - Add responsive breakpoints to this story.
- Fix alignment of inner-bg to spec discusssed with elisa.
- Remove console log statements that were left lying around.
- Add ESLint config to error on most console, debugger, and alert statements.


library/src/scripts/__tests__/testStoreState.ts
```diff
@@ -21,6 +21,7 @@ export function testStoreState(state: DeepPartial<ICoreStoreState>) {
 
 export function resetStoreState(state: DeepPartial<ICoreStoreState>) {
     const store = getStore();
+    // eslint-disable-next-line no-console
```
slafleche: clean up comments

PR #10070, opened 2020-02-11
Fix system profile URLs

charrondev:
Closes https://github.com/vanilla/support/issues/1384

Added to structure:

- Check if our system user has an `http` scheme URL on our own domain.
- If it does and we can use SSL, update the scheme to `https`


charrondev:
> Maybe I'm missing something but when I test this, it never updates the url to https because in `setCalculatedFields()` in the userModel it finds that if a url does not start with https (isUrl()) it is not a URL and it recalculates the path as a relative path. If the path is to a CDN it doesn't execute because it is not `$ownSiteHttp`.

FWIW I only tested that with our default image URL that is inserted when structuring a new site. I'm assuming that all CDN links on our infra are already `https://`.

If you spawn up a new site in a local dir (https://github.com/vanilla/vanilla-docker/blob/master/docs/vanilla-localhost-dirs.md) does it works correctly for me, and it also works correctly if I reset my system user logo to be the default.

So this purely fixed `http://siteDomain.com/applications/dashboard/design/images/defaulticon.png` -> `https://siteDomain.com/applications/dashboard/design/images/defaulticon.png`

That's the one that still shows up even for new sites. Was there another common support case that was coming up?

PR #10075, opened 2020-02-11
Fix Stub Content re-adding content by re-enforcing the check for Garden.Installed

igraziatto:
Closes https://github.com/vanilla/support/issues/1412

The issue happens because **Stub Content** checks for the config `Garden.Installed` before adding the content, and all those communities (for whatever reason I couldn't find why) have this config set to `'1'` instead of `true`.

I've made a fix which reenforces this check so it doesn't happen again.

### DIFF
--- a/plugins/stubcontent/class.stubcontent.plugin.php
+++ b/plugins/stubcontent/class.stubcontent.plugin.php
@@ -243,7 +243,7 @@ class StubContentPlugin extends Gdn_Plugin {
                     if ($rowID) {
                         $receipt = $this->createReceipt($content, $rowID);
                     } else {
-                        Logger::event("stubcontent-insertfailed", "Failed to insert {type} ({content}): {error}", [
+                        Logger::event("stubcontent-insertfailed", Logger::WARNING, "Failed to insert {type} ({content}): {error}", [
                             'type' => $content['type'],
                             'content' => $contentID,
                             'error' => print_r($model->validationResults(), true)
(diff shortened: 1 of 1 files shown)


nervoustwit:
Fixes https://github.com/vanilla/support/issues/1412

plugins/stubcontent/class.stubcontent.plugin.php
```diff
@@ -201,7 +201,7 @@ public function addStubContent($type) {
     public function insertContent($content) {
 
         // Don't affect installed forums
-        if (c('Garden.Installed', false) === true) {
+        if (c('Garden.Installed', false) == true) {
             return;
         }
```
nervoustwit: I see why you are doing this (==) because on infrastructure we use "1" and "0" but it feels like a step in the wrong direction.

plugins/stubcontent/class.stubcontent.plugin.php
```diff
@@ -201,7 +201,7 @@ public function addStubContent($type) {
     public function insertContent($content) {
 
         // Don't affect installed forums
-        if (c('Garden.Installed', false) === true) {
+        if (c('Garden.Installed', false) == true) {
```
initvector: For the sake of compatibility, this check should probably be the same as how Vanilla, itself, actually checks to see if it's installed: with a regular ole truthy check.

https://github.com/vanilla/vanilla/blob/a09bf1558ddbba3635a05f306c31faccfbdf9551/applications/dashboard/controllers/class.setupcontroller.php#L49

What we're doing here is basically a long-form version. We can shorten it a bit.

```suggestion
        if (c('Garden.Installed', false)) {
```

initvector: As an alternative, you could keep this strict, by matching on the default (and an explicit `false`).

```suggestion
        if (c('Garden.Installed', false) !== false) {
```

PR #10078, opened 2020-02-12
Add minimum length validation rule for the original post of a discussion

blutarch:
Closes vanilla/support#343

The minimum length configuration setting previously applied to comments, but not the original post of a discussion. Since users expect that that setting would apply to the original post as well as comments, this fix implements that.

### TO TEST
1. On master, set a minimum post length and start a discussion where the body is shorter than the min.
1. Observe that the discussion passes validation.
1. Checkout branch, repeat steps 1 and 2, and verify that the 'This discussion is x characters too short' error message displays.


dannimajdalani:
@blutarch do we have an issue created for this?

blutarch:
@dannimajdalani Yes, just linked it.

PR #10092, opened 2020-02-16
Fix API authentication by setting HTTP_AUTHORIZATION in .htaccess.dist

### LINKED ISSUE #10091: Authorization header-based API access fails
**Describe the bug**
Using the `Authorization: Bearer $TOKEN` header to access the v2 API, described [in the docs](https://success.vanillaforums.com/kb/articles/41-authentication-with-personal-access-tokens), fails because it relies on `$_SERVER['HTTP_AUTHORIZATION']` being set in the [check access token logic](https://github.com/vanilla/vanilla/blob/Vanilla_3.3/applications/dashboard/settings/class.hooks.php#L753). This does not appear to get set by default in apache.

**Vanilla info**
1. Are you sure this is a core Vanilla problem and not caused by any addon you are using? Yes
2. Did you follow our generic troubleshooting steps already? Yes
3. What Vanilla version are you using? 3.3
4. When did the issue start? N/A

**To Reproduce**
1. Modify `index.php` to add the following lines at the top:

```php
print_r(getallheaders());
print_r($_SERVER);
echo 'Auth: ' . $_SERVER['HTTP_AUTHORIZATION'];
die();
```

2. Send an http api request with the `Authorization` header present. You'll see it in the first block (the list of headers) but not the second (`$_SERVER`) and the `Auth:` line will be blank. This demonstrates that the `Authorization` is not being forwarded to `$_SERVER['HTTP_AUTHORIZATION']`.

Now, if you remove those lines and instead add:

```php
$_SERVER['HTTP_AUTHORIZATION'] = getallheaders()['Authorization'];
```

Suddenly the API requests work.

**Expected behavior**
API requests should work using the `Authorization` header as [described in the docs](https://success.vanillaforums.com/kb/articles/41-authentication-with-personal-access-tokens).

wilson29thid:
Fixes #10091 

`$_SERVER['HTTP_AUTHORIZATION']` is empty by default, even when `Authorization` it's present in apache request headers. This passes the `Authorization` header to it, so [access token checking](https://github.com/vanilla/vanilla/blob/Vanilla_3.3/applications/dashboard/settings/class.hooks.php#L753) will work.

Feel free to disregard if there's a better way to address #10091.

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -46,10 +46,12 @@
     RewriteCond %{ENV:REDIRECT_X_PATH_INFO} (.+)
     RewriteRule ^index\.php - [E=X_REWRITE:1,E=!REDIRECT_X_REWRITE,E=X_PATH_INFO:%1,E=!REDIRECT_X_PATH_INFO,L]
     
-    ####
-    # Pass Authorization header to php environment variable to support API authentication
-    ####
-    SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
+    <IfModule mod_setenvif.c>
+        ####
+        # Pass Authorization header to php environment variable to support API authentication
+        ####
+        SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
+    </IfModule>
 </IfModule>
 
 <IfModule mod_headers.c>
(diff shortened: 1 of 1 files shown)


wilson29thid:
I've wrapped it in an `<IfModule>` as well check just to be safe.

PR #10093, opened 2020-02-16
Add DropDownSwitchButton component

cchabilall83:
Add new DropDownSwitchButton Component.

required for https://github.com/vanilla/knowledge/issues/1512

blocking https://github.com/vanilla/knowledge/pull/1547


library/src/scripts/flyouts/DropDownSwitchButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <>
+            <span className={classes.itemLabel}>{label}</span>
+            <span className={classes.checkContainer}>
+                {isLoading ? (
+                    <ButtonLoader />
+                ) : (
+                    <CheckCompactIcon
+                        aria-hidden={true}
+                        className={classNames(
+                            { [visibilityClasses.visuallyHidden]: !status },
```
slafleche: Where's the "on"/"off" accessible text we discussed? Also, you're using the visuallyHidden class the wrong way. It's for making things visible to screen readers. You're making the icon hidden to sighted users, but visible to blind users. If it's not there, just don't render it. `!status && <CheckCompactIcon...`

PR #10095, opened 2020-02-18
Fix formatting always counting against post character limits

nervoustwit:
The PR will address directly https://github.com/vanilla/support/issues/1439 and https://github.com/vanilla/support/issues/1396 and will set the ground work to solve other issues with concerns validating content, including https://github.com/vanilla/support/issues/728 and possibly others.

What's at issue here is the fact that we verify string lengths on Discussions and Comments and other user generated content by counting the number of characters being put in the database. That content contains a lot of markup and meta data that the user does not know is being put in.

This PR will create a validateRaw function that will only count the visible characters on Comments and Discussions.


nervoustwit:
I will have to add tests for this code.

library/Vanilla/Formatting/Formats/HtmlFormat.php
```diff
@@ -92,6 +92,13 @@ public function renderPlainText(string $content): string {
         return $this->plainTextConverter->convert($html);
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function getVisibleTextLength(string $content): int {
+        return strlen($this->renderPlainText($content));
```
charrondev: You probably want to go with `mb_strlen` here instead.

library/Vanilla/Formatting/Formats/NotFoundFormat.php
```diff
@@ -57,6 +57,13 @@ public function renderPlainText(string $content): string {
         return $this->getErrorMessage();
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function getVisibleTextLength(string $content): string {
```
charrondev: Why does this one return a string?

library/core/functions.validation.php
```diff
@@ -528,6 +528,27 @@ function validateLength($value, $field) {
...
+        $stringLength = Gdn::FormatService()->getVisibleTextLength($value, $format);
+        $diff = $stringLength - $field->Length;
+        if ($diff <= 0) {
+            return true;
+        } else {
+            return sprintf(t('ValidateLength'), t($field->Name), $diff);
+        }
+    }
+}
+
```
alex-mtl: Are we sure we want to introduce new global function?

initvector: I'm not sure, but I think it was originally done this way because it's the way 98% of our validation functions are implemented. However, the other 2% demonstrate we do not need a global function. As long as the "rule" argument passed to `addRule` is a callable, it'll be invoked using the value. We can see an example over in `VanillaSettingsController`.

https://github.com/vanilla/vanilla/blob/e6df634ff288e1d9e5a82092a0e2230a835ab1bb/applications/vanilla/controllers/class.vanillasettingscontroller.php#L162

library/Vanilla/VisibleTextLengthValidator.php
```diff
@@ -0,0 +1,73 @@
...
+    private $maxTextLength = 8000;
+
+    /**
+     * VisibleTextLengthValidator constructor.
+     *
+     * @param int $maxTextLength Maximum length of text.
+     * @param FormatService $formatService Service to apply a formatter.
+     * @param LocaleInterface $locale For translating error messages.
+     */
+    public function __construct(int $maxTextLength, \Vanilla\Formatting\FormatService $formatService, \Vanilla\Contracts\LocaleInterface $locale) {
```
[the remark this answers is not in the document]
initvector: I recommend using `use` statements for these fully-qualified class names. You're gonna shave off about half of the length this line.

```suggestion
    public function __construct(FormatService $formatService, LocaleInterface $locale) {
```

library/Vanilla/VisibleTextLengthValidator.php
```diff
@@ -0,0 +1,73 @@
...
+            $fieldName = $this->locale->translate($field->Name ?? '');
+            return new Invalid(sprintf($validationMessage, $fieldName, abs($diff)));
+        }
+    }
+
+    /**
+     * Execute validate function for visible text validator.
+     *
+     * @param string $value User input content.
+     * @param string $field Field name where content is found.
```
initvector: Same kinda thing as above. Is this supposed to be the name of the field or is it an object describing it?

tests/Library/Vanilla/VisibleTextLengthValidatorTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $this->assertEquals($isValid, $resultType);
+    }
+
+    /**
+     * Provide Markdown text to test counting text length with formatting stripped out.
+     *
+     * @return array
+     */
+    public static function providePostContent() {
+        return [
```
initvector: This could use some other format examples, like Rich or Wysiwyg. I'd also add a test in here without the "Format" field.

tests/Library/Vanilla/VisibleTextLengthValidatorTest.php
```diff
@@ -0,0 +1,70 @@
...
+    public function testVisibleTextLengthValidator(array $post, bool $isValid) {
+        $formatService = \Gdn::formatService();
+        $locale = \Gdn::locale();
+        $validator = new VisibleTextLengthValidator(self::TEST_MAX_COMMENT_LENGTH, $formatService, $locale);
+        $result = $validator($post['Body'] ?? '', 'Body', $post);
+        $resultType = true;
+        if ($result instanceof Invalid) {
+            $resultType = false;
+        }
+        $this->assertEquals($isValid, $resultType);
```
initvector: You could probably smoosh this all down into a couple lines. Maybe something like:

```php
$actual = !($result instanceof Invalid);
$this->assertEquals($isValid, $actual);
```

library/Vanilla/VisibleTextLengthValidator.php
nervoustwit: It's just the name.

library/Vanilla/VisibleTextLengthValidator.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->locale = $locale;
+        $this->formatService = $formatService;
+        $this->maxTextLength = $maxTextLength;
+    }
+
+    /**
+     * Validate content length by stripping most meta-data and formatting.
+     *
+     * @param string $value User input content.
+     * @param string $field Field name where content is found.
```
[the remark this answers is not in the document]
nervoustwit: It is a string that is name of the field.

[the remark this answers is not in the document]
initvector: > It is a string that is name of the field.

If `$field` is a string, how are you doing the following, just a few lines down?

```php
$field->maxTextLength
```

```php
$field->Name
```

library/Vanilla/VisibleTextLengthValidator.php
initvector: > It's just the name.

Are you sure about that? It looks like `$fieldInfo`, [an object containing the field's config info](https://github.com/vanilla/vanilla/blob/3eb6b83cc7c29d882318444eee49d689f033f508/library/core/class.validation.php#L796), is passed as [the second parameter](https://github.com/vanilla/vanilla/blob/3eb6b83cc7c29d882318444eee49d689f033f508/library/core/class.validation.php#L838) when this class is "invoked". It would _contain_ the name of the field (e.g. `$field->Name`).

library/Vanilla/VisibleTextLengthValidator.php
```diff
@@ -0,0 +1,73 @@
...
+     * @param array $post POST array.
+     * @return mixed Either an Invalid Object or the value.
+     */
+    private function validate($value, $field, $post) {
+        $format = $post['Format'] ?? '';
```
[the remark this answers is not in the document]
nervoustwit: I have added validation for formatting. It will "pass" if you don't have validation but it shouldn't.

library/Vanilla/VisibleTextLengthValidator.php
```diff
@@ -0,0 +1,96 @@
...
+            return new Invalid(sprintf($validationMessage, $fieldName, abs($diff)));
+        }
+    }
+
+    /**
+     * Get the Maximum Text Length.
+     *
+     * @return int
+     */
+    public function getMaxTextLength() : int {
```
initvector:
```suggestion
    public function getMaxTextLength(): int {
```

library/Vanilla/VisibleTextLengthValidator.php
```diff
@@ -0,0 +1,96 @@
...
+        return $this->maxTextLength;
+    }
+
+    /**
+     * Set the Maximum Text Length.
+     *
+     * @param int $maxTextLength
+     * @return void
+     */
+    public function setMaxTextLength(int $maxTextLength) :void {
```
initvector:
```suggestion
    public function setMaxTextLength(int $maxTextLength): void {
```

PR #10102, opened 2020-02-18
Add "sort" parameter to /discussions api endpoint

igraziatto:
Closes https://github.com/vanilla/support/issues/1437

**WIP: missing tests**


initvector:
As of today, this hasn't seen any movement in three months. I'm gonna close and we can reopen as necessary.

PR #10103, opened 2020-02-18
Fix- Add Theme button size

NeenaDev:
Closes: https://github.com/vanilla/internal/issues/2209

Require: https://github.com/vanilla/internal/pull/2247

Fix-  Add theme button (+) icon size is remain constant even though it shifts to new line.

### DIFF
--- a/library/src/scripts/theming/addThemeStyles.ts
+++ b/library/src/scripts/theming/addThemeStyles.ts
@@ -66,12 +66,21 @@ export const addThemeClasses = useThemeCache(() => {
         },
     });
 
+    const noBoxshadow = style("noBoxshadow", {
+        $nest: {
+            "&&": {
+                boxShadow: "unset",
+            },
+        },
+    });
+
     const addThemeLink = style("addThemeLink", {});
 
     return {
         addTheme,
         button,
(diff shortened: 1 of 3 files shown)


NeenaDev:
> This doesn't seem fixed to me:
> 
> ## Card is missing padding (should be the size of the inner theme card, not the whole description and outer space).
> image
> 
> ## Edges get cut off on hover
> image

NeenaDev:
> This doesn't seem fixed to me:
> 
> ## Card is missing padding (should be the size of the inner theme card, not the whole description and outer space).
> image
> 
> ## Edges get cut off on hover
> image

library/src/scripts/theming/addThemeStyles.ts
```diff
@@ -66,12 +66,21 @@ export const addThemeClasses = useThemeCache(() => {
         },
     });
 
+    const noBoxshadow = style("noBoxshadow", {
+        $nest: {
+            "&&": {
+                boxShadow: "unset",
```
slafleche: I'd rather you add the shadow conditionally than unsetting it conditionally.

PR #10124, opened 2020-02-24
Fix sign-in link generation sometimes redirecting back to entry pages

blutarch:
Closes vanilla/support#1450.

This fixes an issue where users who are signed out with autosignout sign back in, they are redirected to the the same "You have been signed out" page. This PR redirects users to the home page.

TO TEST
1. Enable the autosignout plugin and set the minutes to 1.
1. After you are signed out, sign back in.
1. Verify that you are redirected to the home page rather than the "You have been signed out" page.

### DIFF
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -1734,6 +1734,10 @@ if (!function_exists('signInUrl')) {
             }
         }
 
+        if ($target = '/entry/autosignedout') {
+            $target = '/';
+        }
+
         return '/entry/signin'.($target ? '?Target='.urlencode($target) : '');
     }
 }
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -1734,7 +1734,7 @@ if (!function_exists('signInUrl')) {
             }
         }
 
-        if ($target = '/entry/autosignedout') {
+        if (strpos($target, 'entry') === 0) {
             $target = '';
         }
 
--- a/library/core/functions.render.php
+++ b/library/core/functions.render.php
@@ -1735,7 +1735,7 @@ if (!function_exists('signInUrl')) {
         }
 
         if ($target = '/entry/autosignedout') {
-            $target = '/';
+            $target = '';
         }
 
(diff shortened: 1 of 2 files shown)


library/core/functions.render.php
```diff
@@ -1734,6 +1734,10 @@ function signInUrl($target = '', $force = false) {
             }
         }
 
+        if ($target = '/entry/autosignedout') {
```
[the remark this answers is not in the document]
initvector: This could probably stand to be more generic (e.g. could also work for a sign-in link on the registration page).

library/core/functions.render.php
```diff
@@ -1734,6 +1734,10 @@ function signInUrl($target = '', $force = false) {
...
+            $target = '/';
```
initvector: If we're ditching the target, let's use the default value (`''`).

initvector:
This is a good candidate for some tests.

blutarch:
I think all the comments have been addressed and this is ready for review again.

library/core/functions.render.php
```diff
@@ -1734,6 +1734,10 @@ function signInUrl($target = '', $force = false) {
             }
         }
 
+        if (strpos($target, 'entry') === 0) {
```
initvector: This could probably be tightened up a little bit to force a match on the entry controller, instead of any controller (or path) that starts with the word "entry" (e.g. "entry-exam/p1").

```suggestion
        if (strpos($target, 'entry/') === 0) {
```

Maybe add another test case for that too.

blutarch:
> Has this been tested on a site hosted in a subdirectory? In a subcommunity? I imagine it'll be fine, but those are both things that can affect URLs. We should verify there's no issue there.

Just tested both. If you're in a subdirectory, it takes you back to the homepage. If you're in a subcommunity, it takes you to the subcommunity page (i.e., signing in from http://dev.vanilla.localhost/subcommunity/entry/autosignedout takes you to http://dev.vanilla.localhost/subcommunity).

This seems like the desired behavior, right?

I'll add to the tests.

PR #10126, opened 2020-02-24
Add PSR-14 events for comment CUD operations

dannimajdalani:
closes  https://github.com/vanilla/internal/issues/2141
add CUD comment operation with unit tests.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -8,9 +8,17 @@
  * @since 2.0
  */
 
+use Garden\Schema\Schema;
+use Vanilla\Attributes;
 use Vanilla\Formatting\DateTimeFormatter;
 use Vanilla\Formatting\FormatService;
+use Vanilla\Formatting\FormatFieldTrait;
 use Vanilla\Formatting\UpdateMediaTrait;
+use Vanilla\Models\UserFragmentSchema;
+use Vanilla\SchemaFactory;
+use Vanilla\Community\Events\CommentEvent;
+use Vanilla\Utility\CamelCaseScheme;
+use Vanilla\Utility\ModelUtils;
 
 /**
  * Manages discussion comments data.
@@ -20,6 +28,8 @@ class CommentModel extends Gdn_Model {
     use \Vanilla\FloodControlTrait;
 
     use UpdateMediaTrait;
+
+    use FormatFieldTrait;
 
     /** Threshold. */
     const COMMENT_THRESHOLD_SMALL = 1000;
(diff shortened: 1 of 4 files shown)


applications/vanilla/models/class.commentmodel.php
```diff
@@ -850,6 +860,24 @@ public function getCount($wheres = '') {
         return parent::getCount($wheres);
     }
 
+    /**
+     * @return mixed
```
initvector: Need to punch up this doc block a bit.

applications/vanilla/models/class.commentmodel.php
```diff
@@ -1207,10 +1235,52 @@ public function save($formPostValues, $settings = false) {
                 $this->updateCommentCount($formPostValues['DiscussionID'], ['Slave' => false]);
             }
         }
-
+        $comment = $commentID ? $this->getID($commentID, DATASET_TYPE_ARRAY): false;
```
initvector:
```suggestion
        $comment = $commentID ? $this->getID($commentID, DATASET_TYPE_ARRAY) : false;
```

applications/vanilla/models/class.commentmodel.php
```diff
@@ -1207,10 +1235,52 @@ public function save($formPostValues, $settings = false) {
...
+        if ($comment) {
+            $commentEvent = $this->eventFromRow(
+                (array)$comment,
```
initvector: We're already specifying this row should be represented as an array, a few lines up. We shouldn't need to cast it here.

```suggestion
                $comment,
```

tests/Models/CommentModelTest.php
```diff
@@ -29,35 +42,139 @@ public static function setupBeforeClass(): void {
...
+        $discussion = [
+            'CategoryID' => 1,
+            'Name' => 'test delete event',
+            'Body' => 'foo foo foo',
+            'Format' => 'Text',
+            'InsertUserID' => 1
+        ];
+        $discussionID = $this->discussionModel->save($discussion);
+        $commentID = $this->commentModel->save([
+            "Name" => __FUNCTION__,
```
initvector: Comments don't have names. 🤔

applications/vanilla/models/class.commentmodel.php
```diff
@@ -1207,10 +1237,52 @@ public function save($formPostValues, $settings = false) {
...
+    }
+
+    /**
+     * Given a database row, massage the data into a more externally-useful format.
+     *
+     * @param array $row
+     * @param array $expand
+     * @return array
+     */
+    public function normalizeRow(array $row, $expand = []): array {
```
initvector: What's `$expand` supposed to do here?

applications/vanilla/models/class.commentmodel.php
```diff
@@ -1207,10 +1237,52 @@ public function save($formPostValues, $settings = false) {
...
+        $this->formatField($row, "Body", $row["Format"]);
+        $row['Url'] = commentUrl($row);
+        $row['Attributes'] = new Attributes($row['Attributes']);
+        $scheme = new CamelCaseScheme;
```
initvector:
```suggestion
        $scheme = new CamelCaseScheme();
```

PR #10127, opened 2020-02-25
Color Inputs v1 and panel sections

slafleche:
This PR includes the v1 color picker as well as the sections for the panel.

@charrondev I use this library called "collit" for validating color strings. I was hoping the `color()` function would error out, but it's too lenient and corrects some mistakes. I couldn't get to the color functions with `import` so I had to use `require`. Do you know what I did wrong or if there's a more elegant solution? It's the only library I could find to have these functions.

Note that I need to check a few design issues with @ElisaAlmeida before this gets merged, but the code can be reviewed.

Closes: https://github.com/vanilla/knowledge/issues/1568
Closes: https://github.com/vanilla/knowledge/issues/1566


library/src/scripts/forms/themeEditor/ColorPicker.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { t } from "@vanilla/i18n/src";
+import { uniqueIDFromPrefix } from "@library/utility/idUtils";
+import { themeBuilderClasses } from "@library/forms/themeEditor/themeBuilderStyles";
+
+const Collit = require("collit");
+
+type IErrorWithDefault = string | boolean; // Uses default message if true
+
+export interface IColorPicker {
+    inputProps?: React.HTMLAttributes<Omit<Omit<Omit<HTMLInputElement, "type">, "id">, "tabIndex">>;
```
charrondev: These can rewritten as a single omit.

```suggestion
    inputProps?: Omit<React.HTMLAttributes<HTMLInputElement, "type" | "id" | "tabIndex">>>;
```

library/src/scripts/forms/themeEditor/ColorPicker.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { color, ColorHelper } from "csx";
+import { colorOut } from "@library/styles/styleHelpersColors";
+import { useField } from "formik";
+import Button from "@library/forms/Button";
+import { ButtonTypes } from "@library/forms/buttonStyles";
+import { t } from "@vanilla/i18n/src";
+import { uniqueIDFromPrefix } from "@library/utility/idUtils";
+import { themeBuilderClasses } from "@library/forms/themeEditor/themeBuilderStyles";
+
+const Collit = require("collit");
```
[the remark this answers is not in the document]
charrondev: Don't forget to remove the package for package.json

library/src/scripts/forms/themeEditor/ColorPicker.tsx
```diff
@@ -0,0 +1,154 @@
...
+    const builderClasses = themeBuilderClasses();
+
+    const errorID = useMemo(() => {
+        return uniqueIDFromPrefix("colorPickerError");
+    }, []);
+
+    const initialColor = props.defaultValue ? props.defaultValue : color("#fff");
+
+    const [lastValidColor, setLastValidColor] = useState(initialColor); // Always "Color" object
+    const [inputTextValue, setInputTextValue] = useState(initialColor.toString()); // Always string
```
charrondev: This value should be stored on `selectedColor.value` from formik already. I'm not sure we need to store it again.

library/src/scripts/forms/themeEditor/ColorPicker.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    className={classes.swatch}
+                    tabIndex={-1}
+                    baseClass={ButtonTypes.CUSTOM}
+                >
+                    <span className={visibility().visuallyHidden}>{formattedColor}</span>
+                </Button>
+            </span>
+            {props.errors && props.errors.length > 0 && (
+                <ul id={errorID} className={builderClasses.errorContainer}>
+                    {props.errors.map((error, i) => {
```
charrondev: About the error messages here:

- What's the scenario that we would have multiple errors on a single input? The API certainly won't be giving any back. The only that I could see happening are invalid colors which could be validated here.
- It would probably be better to store the errors in formik so they can also be used for top level form validation. You use it inside of an `onBlur` handler and also run it when the component first mounts. Something like this:

```
const validateColor = () => {
	//... Some conditional
	helper.setFieldError("Invalid Color");
}

useEffect(() => {
    validateColor()
    // eslint-disable-next-line react-hooks/exhaustive-deps
}, []);

<input

   // ... other props
	onBlur={(e) => {
		selectedColor.onBlur(e);
		validateColor();
    }}
>

```

slafleche:
I've implemented your feedback @charrondev . Just a note, the color picker input is annoying to set because it expects a specific format of RGB and doesn't support transparencies. I didn't bother to code it since we're planning to use a plugin. For now it's an uncontrolled input. The swatch is a background color and is unaffected.

PR #10129, opened 2020-02-25
Fix PermissionModel

alex-mtl:
Unset (reset) `$juncIDs` var inside the loop to avoid data pollution.

Relates to: https://github.com/vanilla/knowledge/issues/1583

### DIFF
--- a/applications/dashboard/models/class.permissionmodel.php
+++ b/applications/dashboard/models/class.permissionmodel.php
@@ -406,9 +406,10 @@ class PermissionModel extends Gdn_Model {
 
         // Select any that match $LimitToSuffix
         foreach ($permissionColumns as $columnName => $value) {
-            if (!empty($limitToSuffix) && substr($columnName, -strlen($limitToSuffix)) != $limitToSuffix) {
+            if (!empty($limitToSuffix) && substr($columnName, -strlen($limitToSuffix)) !== $limitToSuffix) {
                 continue; // permission not in $LimitToSuffix
-            }            $this->SQL->select('p.`'.$columnName.'`', 'MAX');
+            }
+            $this->SQL->select('p.`'.$columnName.'`', 'MAX');
         }
 
         // Generic part of query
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.permissionmodel.php
```diff
@@ -590,6 +591,7 @@ public function getJunctionPermissions($where, $junctionTable = null, $limitToSu
             $junctionColumn = $row['JunctionColumn'];
             unset($row['PermissionID'], $row['RoleID'], $row['JunctionTable'], $row['JunctionColumn'], $row['JunctionID']);
 
+            unset($juncIDs);
```
charrondev: It's a bit hard to review this without more context about what the data pollution issue was causing/vs what was previously expected out of this function, but from what I can see, the `$juncIDs` get set again just a few lines after this, so I'm not really sure if this fully solves whatever the issue is.

alex-mtl: We are looping through the list of `JunctionTable` variants which were only `Category` before.
Now we are implementing another table to join.

So for the _1st cycle_ of the loop let say we got condition 
```
if ($limitToDefault === false && $junctionColumn != $junctionTable.'ID') {
```
equals to **true** and went inside `if (...){ ... }` which bring up the variable `$juncIDs` with some value.

This value is used in line 659:
```
 } elseif (isset($juncIDs)) {
```
which is fine during the _1st cycle_ . But during next loop iteration:
```
if ($limitToDefault === false && $junctionColumn != $junctionTable.'ID') {
```
 can be negative. Which means by design we expect  `$juncIDs` to be not initialized. But it was during 1st loop iteration and keeps previous data. And this bring false positive (using this polluted variable value) on the condition in line 659.

To avoid this we should _reset_ variable to some initial state that will not bring false positive trigger .

cchabilall83:
Tested with a few categories and custom permissions.

PR #10130, opened 2020-02-25
Fix category dropdown options when moving a reported discussion using a checkbox

dannimajdalani:
closes https://github.com/vanilla/support/issues/1443

Moving a post out for reported posts is not currently allowed. In the drop-down menu, the only available option is 'Reported Posts' category. However, this is not the case when selecting the discussion with a checkbox. then moving it.

**Changes**

I've set the CategoryID and DiscussionType on the Data array similar to what we're doing when moving a discussion with the cog wheel.

### To Test

- Move a discussion using the cog wheel, you should have Reported Posts as the only available option

Screen Shot 2020-02-25 at 12 33 54 PM

- Move a discussion using the checkbox 
- Confirm that you have a similar behaviour.

### DIFF
--- a/applications/vanilla/controllers/class.moderationcontroller.php
+++ b/applications/vanilla/controllers/class.moderationcontroller.php
@@ -370,13 +370,13 @@ class ModerationController extends VanillaController {
 
         // Check for edit permissions on each discussion
         $AllowedDiscussions = [];
-        $DiscussionData = $DiscussionModel->SQL->select('DiscussionID, Name, DateLastComment, CategoryID, CountComments')->from('Discussion')->whereIn('DiscussionID', $DiscussionIDs)->get();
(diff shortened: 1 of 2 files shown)


applications/vanilla/controllers/class.moderationcontroller.php
```diff
@@ -375,6 +375,10 @@ public function confirmDiscussionMoves($DiscussionID = null) {
         $DiscussionData = Gdn_DataSet::index($DiscussionData->resultArray(), ['DiscussionID']);
         foreach ($DiscussionData as $DiscussionID => $Discussion) {
             $Category = CategoryModel::categories($Discussion['CategoryID']);
+            if ($Category['Type'] === 'Reporting' && !array_key_exists('DiscussionType', $this->Data)) {
```
initvector: Is this "Reporting" type from an addon or core Vanilla? If it's from an addon, we probably don't want to reference it like this in core.

initvector:
I haven't performed any additional review, but we already have two approvals and no further explicit requests for a third, so...merging!

PR #10132, opened 2020-02-25
Adjust padding on searchbar

NeenaDev:
Require: https://github.com/vanilla/internal/pull/2255
https://github.com/vanilla/internal/pull/2259

~Changes "disablePointerEvent" prop from Banner, IndependentSearch, SearchBar, TitleBar, CompactSearch components.~

**Test-**
Go to Theme editor page -> click Styles tab
- Links and Search on titleBar and banner should not be clickable. 

Go to forum/kb page
- Links and Search on titleBar and banner should be clickable.


library/src/scripts/features/search/IndependentSearch.tsx
```diff
@@ -80,6 +81,8 @@ export function IndependentSearch(props: IProps) {
                 hideSearchButton={props.hideSearchButton}
                 contentClass={props.contentClass}
                 valueContainerClasses={props.valueContainerClasses}
+                disableAutocomplete={props.disablePointerEvent ? true : false}
```
slafleche: You can do this in the prop, no need to put the logic in the parent. It makes it more "portable" that way.

NeenaDev: Fix- move logic to searchBar component

slafleche:
I think we want to leave as much interactivity as possible so they can see the hover/focus states.

Regardless, we can iterate on this. I need this PR to move forward and it works.

charrondev:
After discussing further this PR is not actually blocking, so I'd still like to see the implementation revisited. If we want pointer-events, but not click events then maybe we could do a single wrapper component kind of like this.

```
function ClickHijackWrapper(props: {onClick: (e) => void, children: React.ReactNode}) {
	return <div onClickCapture={(e) => {
		e.stopPropagation();
		e.preventDefault();
        props.onClick(e)
	}}></div>
}
```

If the event order still doesn't work, I'd recommend making a ref and manually assigning the event handler in `useEffect()`.

## Other options
Another alternative would be that if we specifically don't want links to work, and still want all mouse events to work, we just wrap the whole thing in a `<MemoryRouter />`.

NeenaDev:
> After discussing further this PR is not actually blocking, so I'd still like to see the implementation revisited. If we want pointer-events, but not click events then maybe we could do a single wrapper component kind of like this.
> 
> ```
> function ClickHijackWrapper(props: {onClick: (e) => void, children: React.ReactNode}) {
> 	return <div onClickCapture={(e) => {
> 		e.stopPropagation();
> 		e.preventDefault();
>         props.onClick(e)
> 	}}></div>
> }
> ```
> 
> If the event order still doesn't work, I'd recommend making a ref and manually assigning the event handler in `useEffect()`.
> 
> ## Other options
> Another alternative would be that if we specifically don't want links to work, and still want all mouse events to work, we just wrap the whole thing in a `<MemoryRouter />`.

Adam, made changes  you need to pull https://github.com/vanilla/internal/pull/2259

PR #10133, opened 2020-02-26
Fix "new" tags on discussions created after category has been marked read

blutarch:
Closes vanilla/support#1481.

This PR fixes a bug where discussions created after a category has been marked read would display the number of unread comments, even though the user had never viewed the discussion. With this fix, the 'new' tag displays without a number, as it should.

### TO TEST

1. As user A, mark a category as read.
1. As user B, create a new discussion in that category and add a comment to that discussion.
1. As user A, refresh the page and ensure that discussion's tag is 'new' and not '1 new'

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -1146,7 +1146,8 @@ class DiscussionModel extends Gdn_Model implements FormatFieldInterface {
         } else {
             // If the category was marked explicitly read at some point, see if that applies here
             if ($category && !is_null($category['DateMarkedRead'])) {
-                // If the discussion was created after the category was marked read and hasn't been viewed, reset DateLastViewed to null
+                // If the discussion was created after the category was marked read and it hasn't been viewed,
+                // leave CountCountCommentWatch and DateLastViewed null.
                 if (!is_null($discussion->CountCommentWatch)) {
                     // If it's not a newly created discussion, set DateLastViewed to whichever is most recent.
                     $discussion->DateLastViewed = self::maxDate($discussion->DateLastViewed, $category['DateMarkedRead']);
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -1146,9 +1146,10 @@ class DiscussionModel extends Gdn_Model implements FormatFieldInterface {
         } else {
             // If the category was marked explicitly read at some point, see if that applies here
             if ($category && !is_null($category['DateMarkedRead'])) {
-                // If the discussion was created after the category was marked read and it hasn't been viewed,
-                // leave CountCountCommentWatch and DateLastViewed null.
-                if (!is_null($discussion->CountCommentWatch)) {
+                // If the discussion hasn't been viewed or was created after the category was marked read,
+                // leave CountCountCommentWatch and DateLastViewed null. Otherwise, calculate the correct DateLastViewed.
+                if (!is_null($discussion->DateLastViewed) |
+                    self::maxDate($category['DateMarkedRead'], $discussion->DateInserted) === $category['DateMarkedRead']) {
                     // If it's not a newly created discussion, set DateLastViewed to whichever is most recent.
                     $discussion->DateLastViewed = self::maxDate($discussion->DateLastViewed, $category['DateMarkedRead']);
                 }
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1147,7 +1147,12 @@ public function calculate(&$discussion) {
             if ($category && !is_null($category['DateMarkedRead'])) {
                 // If the category was marked explicitly read at some point, see if that applies here
                 $discussion->DateLastViewed = self::maxDate($discussion->DateLastViewed, $category['DateMarkedRead']);
+                // If the discussion was created after the category was marked read and hasn't been viewed, reset DateLastViewed to null
+                if ($category['DateMarkedRead'] <= $discussion->DateInserted && is_null($discussion->CountCommentWatch)) {
+                    $discussion->DateLastViewed = null;
```
initvector: It looks like we're setting `$discussion->DateLastViewed`, a few lines up, and then potentially immediately resetting it here. Can we alter this logic so we only need to set the value once?

blutarch: I didn't like that either, but couldn't think of a better solution. I'm going to think harder.

blutarch: Think I figured it out.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1144,10 +1144,17 @@ public function calculate(&$discussion) {
             $discussion->Read = false;
             $discussion->CountUnreadComments = true;
         } else {
+            // If the category was marked explicitly read at some point, see if that applies here
             if ($category && !is_null($category['DateMarkedRead'])) {
-                // If the category was marked explicitly read at some point, see if that applies here
-                $discussion->DateLastViewed = self::maxDate($discussion->DateLastViewed, $category['DateMarkedRead']);
+                // If the discussion was created after the category was marked read and hasn't been viewed, reset DateLastViewed to null
+                if ($category['DateMarkedRead'] <= $discussion->DateInserted && is_null($discussion->CountCommentWatch)) {
```
initvector: At this point, I'm pretty sure `DateMarkedRead` and `DateInserted` are strings (or possibly `null` for the latter). We're comparing two strings, which will usually work, but not for the reasons we're expecting and can potentially be unpredictable. PHP doesn't have any special handling for date strings, specifically. We could translate the string date value into a numerical timestamp value, using [`strtotime`](https://www.php.net/strtotime), which should give us more confidence when comparing the two. We're generating a standard numerical representation of a date, which can then be compared against other standard numerical representations.

initvector: If `$discussion->CountCommentWatch` is `null`, then wouldn't `$discussion->DateLastViewed` also be `null` already?

initvector:
One more thing: we know the logic in here can be a little hairy, so let's add some tests.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1144,10 +1144,15 @@ public function calculate(&$discussion) {
             $discussion->Read = false;
             $discussion->CountUnreadComments = true;
         } else {
+            // If the category was marked explicitly read at some point, see if that applies here
             if ($category && !is_null($category['DateMarkedRead'])) {
-                // If the category was marked explicitly read at some point, see if that applies here
-                $discussion->DateLastViewed = self::maxDate($discussion->DateLastViewed, $category['DateMarkedRead']);
+                // If the discussion was created after the category was marked read and hasn't been viewed, reset DateLastViewed to null
```
initvector: > If the discussion was created after the category was marked read [..]

How are we accounting for that here?

> [...] and hasn't been viewed [..]

If we're checking for the user's last-viewed date, wouldn't we use the discussion row's existing `DateLastViewed` property?

blutarch: Oops. That comment was for the old logic.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -1144,10 +1144,17 @@ public function calculate(&$discussion) {
...
             $discussion->CountUnreadComments = true;
         } else {
+            // If the category was marked explicitly read at some point, see if that applies here
             if ($category && !is_null($category['DateMarkedRead'])) {
-                // If the category was marked explicitly read at some point, see if that applies here
-                $discussion->DateLastViewed = self::maxDate($discussion->DateLastViewed, $category['DateMarkedRead']);
+                // If the discussion hasn't been viewed or was created after the category was marked read,
+                // leave CountCountCommentWatch and DateLastViewed null. Otherwise, calculate the correct DateLastViewed.
+                if (!is_null($discussion->DateLastViewed) |
+                    self::maxDate($category['DateMarkedRead'], $discussion->DateInserted) === $category['DateMarkedRead']) {
```
tburry: The `DiscussionModel::maxDate()` function specifies that it can take and return null values, but you do the check here. Wouldn't it be better to do the fix within `maxDate()` itself?

blutarch: I don't think I understand this comment, @tburry. `maxDate()` will only return null if both values passed to it are null. But that's not the case here. We don't want a null value to get overwritten with a date in this case. Can you explain what I'm missing?

PR #10138, opened 2020-02-27
Add PSR-14 events for user  write operations

dannimajdalani:
closes https://github.com/vanilla/internal/issues/2143
Add PSR-14 events for user write operations with unittest.

### DIFF
--- a/applications/dashboard/controllers/api/UsersApiController.php
+++ b/applications/dashboard/controllers/api/UsersApiController.php
@@ -134,43 +134,16 @@ class UsersApiController extends AbstractApiController {
 
         $this->userModel->removePicture($id);
     }
+
     /**
      * Get a schema instance comprised of all available user fields.
      *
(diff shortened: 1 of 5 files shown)


applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -512,32 +485,15 @@ public function index(array $query) {
...
-        $schemaRecord = ApiUtils::convertOutputKeys($dbRecord);
-        return $schemaRecord;
+        $normalizedRow = $this->userModel->normalizeRow($dbRecord, []);
+        $options = [];
+        $result = $this->getEventManager()->fireFilter(
+            'usersApiController_normalizeOutput',
+            $normalizedRow,
+            $this,
+            $options
+        );
```
initvector: It doesn't look like we had this before. Let's not add it as part of this update.

```suggestion
```

PR #10159, opened 2020-03-04
Implement MultiRoleInput

alex-mtl:
Relates to: https://github.com/vanilla/knowledge/issues/1582


alex-mtl:
We went with another approach: https://github.com/vanilla/vanilla/pull/10161

PR #10162, opened 2020-03-04
Fix permission query when join JunctionTable data

alex-mtl:
We need to only join when `JunctionTable` filtered, if not we can attach wrong data and get wrong permissions.

Relates to: https://github.com/vanilla/knowledge/issues/1538


alex-mtl:
Probably we don't need this

PR #10180, opened 2020-03-09
Added titleBar options for gradient, "swoop", new type of "border" and extra padding

slafleche:
Added required options titleBar to match client mockups. 

Note that the "height" may or may not be accurate to describe the titleBar as a whole if extra spacing is added with `titleBar.spacing.padding.top` and `titleBar.spacing.padding.bottom`. There is, however, a new variable called `titleBar.fullHeight` which will calculate the full height properly. I couldn't find anywhere that really needed this variable, but it's worth watching out for.

Closes: https://github.com/vanilla/knowledge/issues/1620


library/src/scripts/headers/titleBarStyles.story.tsx
```diff
@@ -0,0 +1,685 @@
...
+                },
+            }}
+        >
+            <MemoryRouter>
+                <BannerContextProvider>
+                    <TitleBarDeviceProvider>
+                        <StoryHeading depth={2}>{props.title}</StoryHeading>
+                        <TitleBar useMobileBackButton={false} isFixed={true} />
+
+                        <p>Scrollable content</p>
```
charrondev: This feels a bit bloated to have in line by line. Maybe we could replace it with:

```
{Array.from(new Array(100)).map(i => <p key={i}>Scrollable Content</p>)}
```

PR #10182, opened 2020-03-10
Fix embedded quote click functionality

blutarch:
Closes vanilla/support#1460.

Previously, when warning a user, in the draft message, the buttons to expand the message or follow a link to the post did not function. These buttons were being blocked by overlaying elements. The fix adds two things: a `pointer-events: none` css property to allow the clicks to go through the blocking elements and a div in the QuoteEmbed html with an onClick handler set to preventDefault. These changes makes those buttons functional and also adds expand functionality to quoted comments in drafts.

### TO TEST

1. Checkout the branch and do a new build.
1. Find or create a long comment to flag.
1. Choose the warn option and click on the embedded quotation.
1. Verify that you can expand the quotation and that the "View Post" link works.


igraziatto:
Tested locally, works great.

PR #10186, opened 2020-03-10
New Banner variables - seach position, new background options

slafleche:
- Added new position for search bar `banner.options.searchPlacement` "middle" or "bottom" supported

- This "search strip" may have
	- its own background color/gradient `banner.searchStrip.bg`
	- a minHeight: `banner.searchStrip.minHeight`
	- vertical offset: `banner.searchStrip.offset`

Examples in Story book.


charrondev:
You've got a few that have extra bottoms added in the storybook

https://www.chromaticqa.com/snapshot?appId=5d5eba16c782b600204ba187&id=5e67f395e517a30022fbd9d3

Also can we pick a better color for the gradient example? The one in the story doesn't example make it seem like something anyone would ever want to do.

image

PR #10187, opened 2020-03-10
Allow plugins to add selects in getWhere function of the DiscussionModel.

patches:
A change was made in https://github.com/vanilla/vanilla/pull/10094 to optimize the getWhere function of the discussion model.

A custom plugin created a little while ago was hooking into the `beforeGet` event fired in this same function to add a "select" to the SQL object. This new change is now breaking the query into two parts and the select is now being added to the wrong query. The query then fails when the plugin is turned on.

https://github.com/vanillaforums/mse/issues/166

### The solution

Instead of firing an additional event, a new argument is passed to it allowing plugins to specify columns to add to the query.

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -860,14 +860,8 @@ class DiscussionModel extends Gdn_Model implements FormatFieldInterface {
         }
 
         // Add select of addition fields to the SQL object.
-        if (!empty($selects)) {
-            if (!is_array($selects)) {
-                $selects = [$selects];
-            }
-
-            foreach ($selects as $select) {
-                $sql->select($select);
-            }
+        foreach ($selects as $select) {
+            $sql->select($select);
         }
 
         $outerQuery = $sql->getSelect();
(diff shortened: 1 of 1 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -857,6 +859,17 @@ public function getWhere($where = false, $orderFields = '', $orderDirection = ''
                 ->select('w.Participated');
         }
 
+        // Add select of addition fields to the SQL object.
+        if (!empty($selects)) {
+            if (!is_array($selects)) {
+                $selects = [$selects];
+            }
+
```
alex-mtl:
```
```

alex-mtl: Why do we need to check type?
We just introduced new variable and defined initial value as empty array.
I feel pretty safe to run `foreach` without extra checks.
```suggestion
```

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -857,6 +859,17 @@ public function getWhere($where = false, $orderFields = '', $orderDirection = ''
...
+            foreach ($selects as $select) {
+                $sql->select($select);
+            }
+        }
```
alex-mtl:
```suggestion
```

applications/vanilla/models/class.discussionmodel.php
patches: I see your point. I saw this type of check somewhere else in the code and I assumed it would be flexible to allow assigning a value directly. I can remove it if it doesn't bring any value.

PR #10203, opened 2020-03-11
Prepare core for theme editor preview and form submissions

charrondev:
Working towards vanilla/vanilla#10203
Goes with https://github.com/vanilla/internal/pull/2280

- Fix a bug in the number input that caused step down to convert into a string.
- Update `ThemePreloadProvider` to allow forcing a non-active theme (for the theme preview page).
- Merge `ThemeActions` and `ThemesActions` together.
- Add new action for setting the some "forced" or "previewed" variables and invalidating the whole theme cache.
- Fix color input validation and display an error message on it if there is one.


library/src/scripts/forms/themeEditor/ColorPicker.tsx
```diff
@@ -110,28 +140,27 @@ export default function ColorPicker(props: IColorPicker) {
                         [builderClasses.invalidField]: hasError,
                     })}
                     placeholder={"#0291DB"}
-                    value={selectedColor.value}
+                    value={textValue}
+                    // onKeyDown={onTextChange}
```
slafleche: Stray comment

Issue #5428, opened 2017-04-12
Gdn_SQLDriver::noReset fails to set class property


dafriend:
Vanilla version 2.3
This line of code in the method `noReset` is setting a local variable and not the intended class property.

 ` $_NoReset = $NoReset ? ($OneTime ? 1 : 2) : 0;`

 The effect will be that the method `reset()` will always reset all the class properties. In other words, the method `noReset()` has no effect on the class.

The code  should be `$this->_NoReset = $NoReset ? ($OneTime ? 1 : 2) : 0;`

PR #10206, opened 2020-03-12
Remove Gdn_SQLDriver::noReset()

bleistivt:
See #5428
Looking back at the oldest commit, `Gdn_SQLDriver::noReset()` never worked because it never actually set the class property:

https://github.com/vanilla/vanilla/blob/474d00d3dadb287ae0b3ca4da2177745c32c4c3e/library/database/class.sqldriver.php#L1207

As this hasn't been needed for 11 years now, I propose to remove this non-functionality and deprecate its usage.


bleistivt:
fixed in master

PR #10207, opened 2020-03-12
Fix quoted text showing white on white.

nervoustwit:
This PR will close vanilla/support#1321

Recent changes to our EmbedStyles added a white background colour explicitly but did not specify the font colour. This meant, in certain themes it was rendered white on white. This PR will add the the foreground colour to the font.

### Testing
 - Find a theme which has white text as default (or set it in custom CSS).
 - Do `yarn build` in the root of vanilla.
 - Warn a user's post.
 - Go to the user's profile page.
 - See that the body text is invisible.
 - Pull this branch.
 - Repeat the above steps. 
 - See that the body text is visible.

### DIFF
--- a/library/src/scripts/embeddedContent/embedStyles.ts
+++ b/library/src/scripts/embeddedContent/embedStyles.ts
@@ -17,7 +17,7 @@ export const embedContainerVariables = useThemeCache(() => {
 
     const colors = makeThemeVars("colors", {
         bg: globalVars.mainColors.bg,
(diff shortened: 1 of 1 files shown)


slafleche:
Awesome! Looks good. You just need to run prettier. `yarn prettier --write "**/*.ts*"` 

I suggest setting it up as a file watcher if you use PhpStorm. https://docs.vanillaforums.com/developer/contributing/coding-standard-typescript/#validating-your-typescript

PR #10214, opened 2020-03-13
Refactor permission fetching

dannimajdalani:
related https://github.com/vanilla/multisite/issues/348
Start refactoring permission fetching.

### DIFF
--- a/applications/dashboard/models/class.permissionmodel.php
+++ b/applications/dashboard/models/class.permissionmodel.php
@@ -514,36 +514,91 @@ class PermissionModel extends Gdn_Model {
     }
 
     /**
-     *
+     * Get user permissions.
      *
      * @param null $userID
-     * @param null $roleID
+     * @param Mixed $roleID
      * @return array|null
      */
-    public function cachePermissions($userID = null, $roleID = null) {
+    public function cachePermissions($userID = null, ...$roleID) {
         if (!$userID) {
             $roleID = RoleModel::getDefaultRoles(RoleModel::TYPE_GUEST);
+            $permissions = $this->getRolesPermissionsDB($roleID);
+            return $permissions;
         }
 
-        // Select all of the permission columns.
-        $permissionColumns = $this->permissionColumns();
-        foreach ($permissionColumns as $columnName => $value) {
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.permissionmodel.php
```diff
@@ -514,36 +514,91 @@ public function getAllowedPermissionNamespaces() {
...
+        return $permissions;
+    }
 
+    /**
+     * Get cached role permissions.
+     *
+     * @param string $key The cache key
+     * @return array|bool $permissions Role permissions
+     */
+    protected function getCachedRoles($key) {
```
tburry:
```suggestion
    private function getCachedRoles($key) {
```

applications/dashboard/models/class.permissionmodel.php
```diff
@@ -514,36 +514,91 @@ public function getAllowedPermissionNamespaces() {
...
+        return $permissions;
+    }
+
+    /**
+     * Save role permissions in the cache.
+     *
+     * @param string $key The cache key
+     * @param array $permissions Role permissions
+     */
+    protected function setCachedRoles($key, $permissions) {
```
tburry: private

applications/dashboard/models/class.permissionmodel.php
```diff
@@ -514,36 +514,91 @@ public function getAllowedPermissionNamespaces() {
...
+        return $permissions;
+    }
+
+    /**
+     * Get the permissions for one or more roles.
+     *
+     * @param array $roleID The role to get the permissions for.
+     * @return array Returns a permission array suitable for use in a session.
+     */
+    protected function cachePermissions2($roleID): array {
```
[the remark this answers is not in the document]
tburry: So you want to move this to the location of that method and make this the implementation.

dannimajdalani:
debugging this, i've tested a user, the first time logging in, we extracted the permissions from the db however, the second time, it was from the db too.
in permissionModel::getPermoisionByRole()
$inc = Gdn::userModel()->getPermissionsIncrement(); is returning false. When false, we're getting the permissions from the db.
will need to look there.

PR #10232, opened 2020-03-17
Edit translation string t() args with shorter codes

blutarch:
Some translation string codes are shortened by vanilla/locales#266. This PR changes the args of the relevant t() function calls to match the shortened codes.

Addresses vanilla/locales#264

### DIFF
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -568,8 +568,8 @@ class SettingsController extends DashboardController {
                 'LabelCode' => t('Defer Javascript Loading'),
                 'Control' => 'toggle',
                 'Default' => true,
-                'Description' => t('Defer Javascript Loading description' .
-                    anchor(t('More information'), 'https://success.vanillaforums.com/kb/articles/140-defer-javascript-loading-feature')),
+                'Description' => t('Defer Javascript Loading description') .
+                    anchor(t('More information'), 'https://success.vanillaforums.com/kb/articles/140-defer-javascript-loading-feature'),
                 'Options' => [
                     'UseRealBoolean' => true
                 ]
--- a/applications/dashboard/controllers/class.settingscontroller.php
+++ b/applications/dashboard/controllers/class.settingscontroller.php
@@ -568,7 +568,7 @@ class SettingsController extends DashboardController {
                 'LabelCode' => t('Defer Javascript Loading'),
                 'Control' => 'toggle',
                 'Default' => true,
-                'Description' => t('Defer Javascript Loading description') .
+                'Description' => t('This setting loads the page before executing Javascript') .
                     anchor(t('More information'), 'https://success.vanillaforums.com/kb/articles/140-defer-javascript-loading-feature'),
                 'Options' => [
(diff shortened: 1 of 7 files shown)


applications/dashboard/locale/en.php
```diff
@@ -167,6 +167,7 @@ function formatPossessive($word) {
 $Definition['Test Email Message'] = '<p>This is a test email message.</p>'.
     '<p>You can configure the appearance of your forum\'s emails by navigating to the Email page in the dashboard.</p>';
 $Definition['oauth2Instructions'] = '<p>Configure your forum to connect with an OAuth2 application by putting your unique Client ID, Client Secret, and required endpoints. You will probably need to provide your SSO application with an allowed callback URL, in part, to validate requests. The callback url for this forum is <code>%s</code></p>';
-
+$Definition['External link alert details'] = 'Alert users if they click a link in a post that will lead them away from the forum.
```
tburry: Your key here doesn't have the same voice or meaning as the full string. Wouldn't something like "Alert users about external links." be better? Also note that if the translation string is a sentence then the key should be a sentence.

applications/dashboard/models/class.invitationmodel.php
```diff
@@ -160,7 +160,7 @@ public function save($formPostValues, $settings = false) {
                 try {
                     $this->send($invitationID);
                 } catch (Exception $ex) {
-                    $this->Validation->addValidationResult('Email', sprintf(t('Although the invitation was created successfully, the email failed to send. The server reported the following error: %s'), strip_tags($ex->getMessage())));
+                    $this->Validation->addValidationResult('Email', sprintf(t('Invitation email error description'), strip_tags($ex->getMessage())));
```
tburry: Same comment about your key. Also be careful about the double translation. Look at the docblock of `Gdn_Validation::addValidationResult()`. Please start looking into what methods take a translated string or do the translation themselves to familiarize yourself with our methods instead of depending learning this from a code review.

applications/dashboard/controllers/class.settingscontroller.php
```diff
@@ -568,9 +568,8 @@ public function branding() {
                 'LabelCode' => t('Defer Javascript Loading'),
                 'Control' => 'toggle',
                 'Default' => true,
-                'Description' => t('This setting loads the page before executing Javascript which can improve your SEO.<br>
-                    <strong>**Warning: Enabling this feature may cause Javascript errors on your site.**</strong> ' .
-                    anchor(t('More information'), 'https://success.vanillaforums.com/kb/articles/140-defer-javascript-loading-feature')),
+                'Description' => t('This setting loads the page before executing Javascript') .
```
tburry: Not a sentence when the translation is a sentence. You also need to put some whitespace between this and the concat.

blutarch: How much whitespace? There's currently 1 space between the two because the translation string has a trailing space (I don't know if that's kosher or not, but that's how it was originally).

Screen Shot 2020-03-18 at 9 45 08 AM

tburry: Yeah no trailing space in the translation. You can put a space outside of the translation though.

PR #10234, opened 2020-03-17
Create ExtendedFormatService to allow extended content to be rendered.

cchabilall83:
The ExtendedFormatService allows for formats the ability to exclude certain elements from being stripped out by the vanilla formatter.

required for https://github.com/vanilla/knowledge/issues/1664
accompanying PR https://github.com/vanilla/knowledge/pull/1667

### DIFF
--- a/cache/.htaccess
+++ b/cache/.htaccess
@@ -1,1 +1,0 @@
-Deny from all
--- a/cache/.htaccess
+++ b/cache/.htaccess
@@ -1,0 +1,1 @@
+Deny from all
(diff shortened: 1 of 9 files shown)


library/Vanilla/Formatting/Html/HtmlSanitizer.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Vanilla\Formatting\Html;
 
+use phpDocumentor\Reflection\Types\Boolean;
```
charrondev:
```suggestion
```

cache/.htaccess
```diff
@@ -1 +0,0 @@
-Deny from all
```
alex-mtl: I think this file was pushed by mistake!?

PR #10235, opened 2020-03-18
Fix mass deletion of discussions causing response timeouts

dannimajdalani:
closes https://github.com/vanilla/support/issues/974

Creating a job for batch deleting discussions.
### To Test
Create/Delete multiple discussions.(50 very large discussions, include images) You can increase the number of characters allowed on a discussion here
Screen Shot 2020-03-20 at 3 13 11 PM

Note the request delay (might be difficult to simulate on 10012 cluster with barely any traffic) but should still see a difference


dannimajdalani:
I've tested this on 10012

- Created a very large discussion with plenty of large images (50 of those discussions)
- Deleted the discussions without the PR (6.5seconds) 
- Deleted the discussions with the PR (48ms)

applications/vanilla/controllers/class.moderationcontroller.php
```diff
@@ -322,14 +322,14 @@ public function confirmDiscussionDeletes() {
         $this->setData('CountNotAllowed', $countNotAllowed);
 
         if ($this->Form->authenticatedPostBack()) {
-            // Delete the selected discussions (that the user has permission to delete).
-            foreach ($allowedDiscussions as $discussionID) {
-                $deleted = $discussionModel->deleteID($discussionID);
-                if ($deleted) {
-                    $this->jsonTarget("#Discussion_$discussionID", '', 'SlideUp');
-                }
+            $discussionArray = array('discussionID' => $allowedDiscussions);
```
initvector:
```suggestion
            $discussionArray = ['discussionID' => $allowedDiscussions];
```

applications/vanilla/controllers/class.moderationcontroller.php
```diff
@@ -322,14 +322,14 @@ public function confirmDiscussionDeletes() {
...
+                // Queue deleting discussions.
+                /** @var Vanilla\Scheduler\SchedulerInterface $scheduler */
+                $scheduler = Gdn::getContainer()->get(Vanilla\Scheduler\SchedulerInterface::class);
+                $scheduler->addJob(ExecuteBatchDeleteDiscussion::class, $discussionArray);
```
initvector: The intending seems a little off through here.

applications/vanilla/library/Jobs/ExecuteBatchDeleteDiscussion.php
```diff
@@ -0,0 +1,85 @@
...
+     * @return Schema
+     */
+    private function messageSchema(): Schema {
+        $schema = Schema::parse([
+            "discussionID" => [
+                [
+                    "type" => "integer"
+                ]
+            ]
+        ]);
```
initvector: Seems like too many nested arrays for this structure. You have the top-level for your fields. The `discussionID` field is defined by an indexed array with an associative array inside it. I don't think that's really defining anything. The schema class will probably not detect the type and allow this field to be any value, regardless of type.

applications/vanilla/library/Jobs/ExecuteBatchDeleteDiscussion.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Execute all queued up items in the ActivityModel queue.
+     */
+    public function run(): JobExecutionStatus {
+        if (!is_array($this->message)) {
+            return JobExecutionStatus::abandoned();
+        }
+        $this->discussionModel->deleteID($this->message);
```
initvector: This method is documented as only taking an integer for its first parameter (before you changed the doc in this PR). Here, it looks like we're passing it an associative array (e.g. ["discussionID" => [1, 2, 3]]). Is this something we're doing intentionally and it's working as it's supposed to? Or is this happening to work by chance, even though it's unintentional? I wouldn't want this to break in the future because the method is changed, not accounting for unintended usage. Are there many other calls to `deleteID` in the codebase that pass in an array, instead of an integer?

dannimajdalani: There isn't, but the idea was to pass an array instead of looping over an id. What do you suggest I do here?
It worked because sql where accepts an array.

applications/vanilla/library/Jobs/ExecuteBatchDeleteDiscussion.php
```diff
@@ -0,0 +1,87 @@
...
+                    "items" => [
+                        "type" => "integer"
+                    ]
+                ]
+        ]);
+        return $schema;
+    }
+
+    /**
+     * Execute all queued up items in the ActivityModel queue.
```
initvector: >  ... in the **ActivityModel** queue.

🤔

applications/vanilla/library/Jobs/ExecuteBatchDeleteDiscussion.php
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Execute the discussion delete queue.
+ */
+class ExecuteBatchDeleteDiscussion implements Vanilla\Scheduler\Job\LocalJobInterface {
+
+    /** @var DiscussionModel */
+    private $discussionModel;
+
+    /** @var array*/
```
initvector:
```suggestion
    /** @var array */
```

applications/vanilla/library/Jobs/ExecuteBatchDeleteDiscussion.php
```diff
@@ -0,0 +1,87 @@
...
+ * @copyright 2009-2020 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+use Garden\Schema\Schema;
+use Vanilla\Scheduler\Job\JobExecutionStatus;
+use Vanilla\Scheduler\Job\JobPriority;
+
+/**
+ * Execute the discussion delete queue.
```
initvector:
```suggestion
 * Delete one or more discussions.
```

applications/vanilla/library/Jobs/ExecuteBatchDeleteDiscussion.php
```diff
@@ -0,0 +1,87 @@
...
+ */
+class ExecuteBatchDeleteDiscussion implements Vanilla\Scheduler\Job\LocalJobInterface {
```
initvector: This is just a suggestion, but this class could probably be a little less...wordy.

```suggestion
class DeleteDiscussions implements Vanilla\Scheduler\Job\LocalJobInterface {
```

Also, I know some of our previous jobs don't have namespaces, but it would be good if we started using them (e.g. `Vanilla\Library\Jobs`).

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -3006,7 +3006,7 @@ public function delete($where = [], $options = []) {
      *
      * Events: DeleteDiscussion.
      *
-     * @param int $discussionID Unique ID of discussion to delete.
+     * @param mixed $discussionID Unique ID of discussion to delete.
```
initvector: If we're going to officially declare support for an array, let's be explicit.

```suggestion
     * @param int|array $discussionID Unique ID of discussion to delete or an array of discussion IDs.
```

PR #10236, opened 2020-03-18
Core support for theme editor

charrondev:
Goes see details on https://github.com/vanilla/internal/pull/2295

This is just the companion to that.


applications/dashboard/controllers/api/ThemesApiController.php
```diff
@@ -56,11 +56,20 @@ private function contentTypeByAsset(string $assetKey): string {
      * Get a theme assets.
      *
      * @param string $themeKey The unique theme key or theme ID.
+     * @param array $query
      * @return array
      */
-    public function get(string $themeKey): array {
+    public function get(string $themeKey, array $query): array {
```
tburry:
```suggestion
    public function get(string $themeKey, array $query = []): array {
```

I think the default is safer and may prevent some edge cases.

PR #10244, opened 2020-03-18
Fix firing PSR-14 user events for users with empty password fields

nervoustwit:
This PR will close vanilla/support#1582

Moving the user Schema from the API controller to the usermodel caused a problem when updating the User table when the Password field is empty. This PR will fix that by adding a subset of the full schema to the UserModel.


nervoustwit:
Tested it with Ranks and Online, two plugins that use the UserSchema, and it works.

PR #10263, opened 2020-03-20
Implement new "Content Banner"

charrondev:
Working towards https://github.com/vanilla/knowledge/issues/1548

- Implements a new more minimalistic banner variant.
- Disabled by default (doesn't appear unless theme variable is set).
- Title, description, and searchbar are always hidden.
- Banner height is clamped to between 80 & 180.
  - Default is 120.
  - Default mobile is 80.
- Also fixes a bug where search was getting cut off again. I've added a story where the search is forced open.
- Adjust the default search strip story to look like our mockup https://zpl.io/25vQdOe


slafleche:
For https://www.chromaticqa.com/snapshot?appId=5d5eba16c782b600204ba187&id=5e7a5a1ffe7b1e0022499938

Shouldn't it be even? Maybe the excess margin can be put on the title so you keep even padding on the top and bottom?

PR #10266, opened 2020-03-21
Add the HtmlUtils::formatTags() method

tburry:
This method provides a PHP implementation similar to our [Translate component](https://github.com/vanilla/vanilla/blob/fc69c3cc00298ef0b7711b634fbcef72731fed86/library/src/scripts/content/Translate.tsx#L51) so that we can use the same translation strings on the client and server.

### DIFF
--- a/tests/Library/Vanilla/Utility/HtmlUtilsTest.php
+++ b/tests/Library/Vanilla/Utility/HtmlUtilsTest.php
@@ -1,0 +1,121 @@
+<?php
+/**
+ * @author Todd Burry <todd@vanillaforums.com>
+ * @copyright 2009-2020 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace VanillaTests\Library\Vanilla\Utility;
+
+use PHPUnit\Framework\TestCase;
+use Vanilla\Utility\HtmlUtils;
+
+/**
+ * Tests for the `HtmlUtils` class.
+ */
+final class HtmlUtilsTest extends TestCase {
+    /**
+     * Tests for `HtmlUtils::classNames()`
+     *
+     * @param array $args
+     * @param string $expected
+     * @dataProvider provideClassNameTests
+     */
+    public function testClassNames(array $args, string $expected): void {
+        $actual = HtmlUtils::classNames(...$args);
+        $this->assertSame($expected, $actual);
+    }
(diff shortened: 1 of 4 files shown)


tests/Library/Vanilla/Utility/HtmlUtilsTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Formatting tags when the argument isn't supplied is a notice.
+     */
+    public function testFormatTagsInvalidArgNotice(): void {
+        $actual = @HtmlUtils::formatTags('Hello <0/>');
+        $this->assertSame('Hello ', $actual);
+
+        $this->expectNotice();
+        $this->expectExceptionMessage('<0/>');
```
charrondev: I'm not seeing how this actually works. It doesn't look like the the notice I see thrown above. Maybe this should be in the `expectNotice()`?

tests/Library/Vanilla/Utility/HtmlUtilsTest.php
```diff
@@ -0,0 +1,121 @@
...
+     * @return array
+     */
+    public function provideAttributesTests(): array {
+        $r = [
+            'empty' => [[], ''],
+            'one' => [['a' => 'b'], ' a="b"'],
+            'two' => [['a' => 'b', 'c' => 'd'], ' a="b" c="d"'],
+            'bool true' => [['type' => 'checkbox', 'checked' => true], ' type="checkbox" checked'],
+            'bool false' => [['type' => 'checkbox', 'checked' => false], ' type="checkbox"'],
+            'json data' => [['data-foo' => ['a' => 'b']], ' data-foo="{&quot;a&quot;:&quot;b&quot;}"'],
```
charrondev: I think you're missing a unicode test here. When encoding json if you want unicode I think you need the `JSON_UNESCAPED_UNICODE` flag otherwise you'll encode it all.

tests/Library/Vanilla/Utility/HtmlUtilsTest.php
tburry: I think it works because notices are exceptions in PHPUnit, however it is a typo so I'll change. As an FYI, when I changed I tried changing the message in the test and the test failed so it is getting called properly.

PR #10267, opened 2020-03-21
Move some deprecated functions and classes to their own folder

tburry:
This is a cleanup task. I want to have deprecated code physically removed from the code we plan to continue using.


tburry:
Hoping to get a quick approval on this. That code coverage report is also a LIE.

.codecov.yml
```diff
@@ -32,3 +32,4 @@ ignore:
   - "**/conf/**/*.php"
   - "**/*.css"
   - "**/*.scss"
+  - /library/deprecated/**.*.php
```
charrondev: Looking at the docs here https://docs.codecov.io/docs/ignoring-paths

This should not have a proceeding path.

And shouldn't it just be `library/deprecated`?

PR #10268, opened 2020-03-23
Foundation - Style updates to search results

slafleche:
Match Foundation styles to new mockups.

Closes: https://github.com/vanilla/internal/issues/2305


charrondev:
I'm seeing some extra padding on the bottom that I don't think should be there.

image

slafleche:
@charrondev fixed.

Issue #10220, opened 2020-03-16
Foundation: Popular tags issues


ElisaAlmeida:
Popular Tags issues
1. Padding issues and the tag should have bigger a padding or a border. I recommend a border.
2. The Padding between the title and the tags is too big. Use the same padding that the other titles have.

<img width="318" alt="Screen Shot 2020-03-11 at 3 13 31 PM" src="https://user-images.githubusercontent.com/45467951/76468170-b4b7be00-63c1-11ea-9e64-af34670aae7e.png">

PR #10269, opened 2020-03-23
Foundation - Tag styles

slafleche:
Fixed styles for tags to match mockups: https://app.zeplin.io/project/5e271637b9bef8985fd77a08/screen/5e73dd7eb601dab3510c6736

Note that the color and border are a tad different from the mockups, because I used the same ones we have as defaults instead of making new colors.

Closes: https://github.com/vanilla/vanilla/issues/10220


slafleche:
I found a bug, will remove WIP tag when done.

PR #10272, opened 2020-03-23
New Theme Choose Input component

cchabilall83:
This add adds new ThemeChooseInput with will display all available themes.

accompanying pr for https://github.com/vanilla/multisite/pull/351
required for https://github.com/vanilla/multisite/issues/349

### DIFF
--- a/library/src/scripts/theming/ThemeChooserInput.tsx
+++ b/library/src/scripts/theming/ThemeChooserInput.tsx
@@ -14,7 +14,7 @@ import { DashboardSelect } from "@dashboard/forms/DashboardSelect";
 interface IProps {
     value?: number | string | null;
     initialValue: number | string;
-    onChange?: (value: number | string | null) => void;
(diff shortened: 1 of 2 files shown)


library/src/scripts/theming/ThemeChooserInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { LoadStatus } from "@library/@types/api/core";
+import { DashboardSelect } from "@dashboard/forms/DashboardSelect";
+
+interface IProps {
+    value?: number | string | null;
+    initialValue: number | string;
+    onChange?: (value: number | string | null) => void;
+}
+
+export function ThemeChooserInput(props: IProps) {
```
charrondev: Could we get a docblock on this component? Just describe its purpose. The props don't need to be detailed because we have an interface for it.

cchabilall83: added

PR #10275, opened 2020-03-23
Fix theme API filtering

charrondev:
Relating to https://github.com/vanilla/knowledge/issues/1681

We had some issues with hidden themes appearing in results, so I've done some refactoring to the method and added some tests for it.

- Update AddonInterface and MockAddon with additional info methods.
- Move `FsThemeProvider::filterTheme` to `ThemeHelperModel::isThemeVisible`.
- The method now takes an addon instead of an array.
- Add unit tests for this method.
- Update `MinimalContainerTestCase` to allow setting session information.
- Fix the minHeight on the add theme item (visible when it is by itself).
- SysAdmin users can see all themes now, similar to our behaviour on the addons page. **_Because of this I had to set the user to a guest for the theme index test_**

_Note:_ I moved a "public" method on the `FsThemeProvider` but it had no other usages.

### DIFF
--- a/library/Vanilla/Models/ThemeModelHelper.php
+++ b/library/Vanilla/Models/ThemeModelHelper.php
@@ -62,7 +62,7 @@ class ThemeModelHelper {
             return true;
         }
 
-        $isAdmin = ($this->session->User->Admin ?? 0) !== 0;
+        $isAdmin = ($this->session->User->Admin ?? 0) === 2;
         if ($isAdmin) {
             // All theme visible for system admins.
             return true;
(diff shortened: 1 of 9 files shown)


library/Vanilla/Models/ThemeModelHelper.php
```diff
@@ -41,6 +43,68 @@ public function __construct(
...
+            $siteName = CLIENT_NAME;
+        }
+        $confVisible = $this->config->get('Garden.Themes.Visible', '');
+
+        if ($confVisible === 'all') {
+            // Config setup to show all themes.
+            return true;
+        }
+
+        if ($this->session->getPermissions()->isAdmin()) {
```
tburry: This will show too many themes. The user that starts an account is admin = 1.

charrondev: Ah ok. I'll check for 2 specifically then, which is our system user.

Issue #10230, opened 2020-03-17
Foundation bugs: Search field issues


GeetikaMuthyam:
- [x] When I click on the Search button 'Search field' and 'Go' button is not as expected, the padding is too close and the go button is too big. 

![image](https://user-images.githubusercontent.com/45467951/76466697-87b5dc00-63be-11ea-9bb9-a829ff13b286.png)

- [x] Search results for each search icon
1. When I click on the search icon on discussions page, for eg: when I type the discussions
 and click on the search, what I see is a bug, the results are awful

![image](https://user-images.githubusercontent.com/56363201/76523892-8ed0fe80-643f-11ea-82a9-85b0deed7753.png)

charrondev:
@slafleche Figure this one can be added too.

![image](https://user-images.githubusercontent.com/1770056/77369057-081bfb80-6d34-11ea-914b-c90359958cf0.png)

when button style is none, desktop sized search bar seems to not obey the border radius.

slafleche:
@charrondev fixed. I had to create 2 additional PRs:
https://github.com/vanilla/internal/pull/2319
https://github.com/vanilla/knowledge/pull/1691

PR #10279, opened 2020-03-24
Foundation - Hide mini search and fix recommended search results

slafleche:
After talking with Elisa, we decided to hide the mini search. There are already 2 other search boxes on the page.

The legacy styles have been fixed for the search results. @ElisaAlmeida note that we don't have exactly the same data/html in this case. Let me know if you're satisfied. There's an extra line of text, i used the same styling as the search results. 

Closes:
https://github.com/vanilla/vanilla/issues/10230


slafleche:
@charrondev fixed. I had to create 2 additional PRs:
https://github.com/vanilla/internal/pull/2319
https://github.com/vanilla/knowledge/pull/1691

PR #10281, opened 2020-03-24
Add ability to apply site-section theme

cchabilall83:
Add the ability to apply the selected them for a site section.

Accompanying PRs

https://github.com/vanilla/internal/pull/2320
https://github.com/vanilla/multisite/pull/353

Require for https://github.com/vanilla/multisite/issues/350


library/Vanilla/Theme/ThemeProviderInterface.php
```diff
@@ -100,9 +100,10 @@ public function setPreviewTheme($themeID): array;
     /**
      * Get "current" theme.
      *
+     * @param int/string $themeID
      * @return array
      */
-    public function getCurrent(): ?array;
+    public function getCurrent($themeID): ?array;
```
charrondev: I'm not really understanding why we've co-opted this. Don't we already have `getThemeWithAssets()`?

library/Vanilla/Site/DefaultSiteSection.php
```diff
@@ -123,4 +123,11 @@ public function setApplication(string $app, bool $enable = true) {
     public function getAttributes(): array {
         return [];
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function getSectionThemeID() {
+        // TODO: Implement getSectionThemeID() method.
+    }
```
charrondev: We should have this return null I think. Also add it to the interface declaration up top.

library/Vanilla/Models/SiteMeta.php
```diff
@@ -149,10 +155,27 @@ public function __construct(
...
+        $sectionThemeID = '';
+        $configParam = 'Garden.Theme';
+        if ($this->currentSiteSection->getSectionThemeID()) {
+            $sectionThemeID =  $this->currentSiteSection->getSectionThemeID();
+            $configParam = 'Garden.CurrentTheme';
+        }
+
+        $theme = $themeModel->getCurrentTheme($sectionThemeID);
+        $themeID = $theme['themeID'] ?? null;
+        $themeAddon = $themeModel->getCurrentThemeAddon($configParam);
```
charrondev: If we're just getting the current whats the scenario where we want current, but actually some other themeID instead. We should be able to just use `getCurrentThemeAddon()`.

Most of this logic should probably be over in the theme model, instead of here.

PR #10283, opened 2020-03-24
Increment product version

igraziatto:
Reference: https://github.com/vanilla/support/issues/1593

### DIFF
--- a/environment.php
+++ b/environment.php
@@ -22,7 +22,7 @@ if (!defined('APPLICATION')) {
 if (!defined('APPLICATION_VERSION')) {
     // Rules for the versioning
     // {OSS version}-{Cloud release version}-{? SNAPSHOT if it's a dev build}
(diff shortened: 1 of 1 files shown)


environment.php
```diff
@@ -22,7 +22,7 @@
 if (!defined('APPLICATION_VERSION')) {
     // Rules for the versioning
     // {OSS version}-{Cloud release version}-{? SNAPSHOT if it's a dev build}
-    define('APPLICATION_VERSION', '4.0-2020.005-SNAPSHOT');
+    define('APPLICATION_VERSION', '4.0-2020.006-SNAPSHOT');
```
initvector: Current progress on master is actually a snapshot of what will ultimately become 2020.007.

```suggestion
    define('APPLICATION_VERSION', '4.0-2020.007-SNAPSHOT');
```

PR #10284, opened 2020-03-24
Add functions for comparing ranked permissions

blutarch:
Partially addresses vanilla/support#1571

This PR adds functions that allow comparison between the current user's permissions and the permissions of a user the current user wants to perform some kind of administrative action on (ban, mark as a troll, etc.). These functions can be called by other classes to make sure a user isn't able to take an action against another user with the same or higher permissions.

The PR also implements this check in the ban method of the `usercontroller.php`

### TO TEST
1. Make sure the moderator role can warn users (enable `Garden.Moderation.Manage` for that role).
1. Make sure you have at least to moderator users.
1. As one moderator, try to ban the other. Verify that it works.
1. Check out this branch and repeat step 4, but verify that it doesn't work.

### DIFF
--- a/library/core/class.session.php
+++ b/library/core/class.session.php
@@ -26,6 +26,15 @@ class Gdn_Session {
     /** Maximum length of inactivity, in seconds, before a visit is considered new. */
     const VISIT_LENGTH = 1200; // 20 minutes
 
+    /** Array of ranked permissions. */
+    private const RANKED_PERMISSIONS = [
+        'Garden.Settings.Manage',
+        'Garden.Community.Manage',
+        'Garden.Moderation.Manage',
+        'Garden.Curation.Manage',
+        'Garden.SignIn.Allow',
+    ];
+
     /** @var int Unique user identifier. */
     public $UserID;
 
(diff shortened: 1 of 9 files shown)


library/core/class.session.php
```diff
@@ -73,28 +84,92 @@ public function addPermissions($perms) {
...
+     * the user any permissions below their level even in if the permission isn't explicitly assigned.
+     * Based on checkRakedPermission().
+     *
+     * @param string $permission
+     * @param int|string $userID
+     * @return bool
+     * @throws ContainerException Throws exception if there's a problem getting the container.
+     * @throws NotFoundException Throws exception if there's a problem getting the container.
+     */
+    public function checkUserRankedPermission(string $permission, $userID) {
```
tburry: This belongs in the user model.

library/core/class.session.php
```diff
@@ -73,28 +84,92 @@ public function addPermissions($perms) {
...
+        }
+    }
+
     /**
      * Check the given permission, but also return true if the user has a higher permission.
      *
      * @param bool|string $permission The permission to check.  Bool to force true/false.
      * @return boolean True on valid authorization, false on failure to authorize
      */
     public function checkRankedPermission($permission) {
```
tburry: I think this can be extracted to the `Permission` class so it can be used by the session or user model.

tburry:
In addition to my requested changes I would say that this really should have some tests.

library/core/class.session.php
```diff
@@ -73,28 +84,92 @@ public function addPermissions($perms) {
...
+     * Checks the session user's permission level against the permissions level of the user they want
+     * to perform some action on.
+     *
+     * @param string $sessionPermission
+     * @param int|string $userID
+     * @return bool
+     * @throws NotFoundException Throws exception if there's a problem getting the container.
+     * @throws ContainerException Throws exception if there's a problem getting the container.
+     */
+    public function hasHigherPermissionLevel(string $sessionPermission, $userID) {
```
tburry: We use the "rank" nomenclature so introducing "level" adds confusion. What about "hasHigherPermissionRank"?

library/core/class.session.php
```diff
@@ -26,6 +28,15 @@ class Gdn_Session {
...
     const VISIT_LENGTH = 1200; // 20 minutes
 
+    /** Array of ranked permissions. */
+    private const RANKED_PERMISSIONS = [
+        'Garden.Settings.Manage',
+        'Garden.Community.Manage',
+        'Garden.Moderation.Manage',
+        'Garden.Curation.Manage',
+        'Garden.SignIn.Allow',
+    ];
```
tburry: It's possible that this will make your logic easier:

```suggestion
    private const RANKED_PERMISSIONS = [
        'Garden.Settings.Manage' => 1,
        'Garden.Community.Manage' => 2,
        'Garden.Moderation.Manage' => 3,
        'Garden.Curation.Manage' => 4,
        'Garden.SignIn.Allow' => 5,
    ];
```

tburry:
@blutarch, I'm going to take this and get it building.

PR #10306, opened 2020-03-27
Fix path of new message button to be new message and not inbox

slafleche:
Seems like we've had 2 links to the messages inbox instead of 1 to new message and 1 to inbox in the mebox messages menu.

@hhovhannes


library/src/scripts/headers/mebox/pieces/MessagesContents.tsx
```diff
@@ -43,7 +43,7 @@ export class MessagesContents extends React.Component<IProps> {
                     <FrameHeaderWithAction title={title}>
                         <LinkAsButton
                             title={t("New Message")}
-                            to={"/messages/inbox"}
+                            to={"/messages/new"}
```
hhovhannes: It should be messages/add I think, I tried /new, it gives permission problem, while /add works fine

PR #10309, opened 2020-03-27
Add support for addons hooking in before user signin

blutarch:
Along with vanilla/internal#2331, this closes vanilla/support#1606

This PR, in conjunction with vanilla/internal#2331, fixes a problem where temporary bans triggered by amassing more than 5 active warning points would never expire because the session would end before there was a check for expired warnings. This fixes that by adding a beforeSignIn event to complement the afterSignIn event calls in the entry controller.

### TO TEST
1. Give a user 5 or more active warning points and verify that the user is banned.
1. Alter the times in the GDN_UserAlert table to make the ban expire.
1. Try to sign in as the banned user and verify that the ban has been lifted

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -567,6 +567,7 @@ class EntryController extends Gdn_Controller {
             }
 
             // Sign the user in.
+            Gdn::userModel()->fireEvent('BeforeSignIn', ['UserID' => $userID]);
             Gdn::session()->start($userID, true, (bool)$this->Form->getFormValue('RememberMe', c('Garden.SSO.RememberMe', true)));
             Gdn::userModel()->fireEvent('AfterSignIn');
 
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -1106,7 +1110,9 @@ public function signIn($method = false, $arg1 = false) {
                                 Gdn::userModel()->setField(val('UserID', $user), ['Password' => $pw, 'HashMethod' => 'Vanilla']);
                             }
 
+                            Gdn::userModel()->fireEvent('BeforeSignIn', ['UserID' => $user->UserID]);
                             Gdn::session()->start(val('UserID', $user), true, (bool)$this->Form->getFormValue('RememberMe'));
+                            Gdn::userModel()->fireEvent('AfterSignIn');
```
initvector: I don't think it's necessary to move this event, so let's drop it back where it originally was.

applications/dashboard/controllers/class.entrycontroller.php
```diff
@@ -1106,6 +1110,7 @@ public function signIn($method = false, $arg1 = false) {
                                 Gdn::userModel()->setField(val('UserID', $user), ['Password' => $pw, 'HashMethod' => 'Vanilla']);
                             }
 
+                            Gdn::userModel()->fireEvent('BeforeSignIn', ['UserID' => $user->UserID]);
```
initvector: Keeping this consistent with `Gdn_Session` is a little muddier, since we want to avoid using `val`. The next best thing is probably null coalescence.

```suggestion
                            Gdn::userModel()->fireEvent('BeforeSignIn', ['UserID' => $user->UserID ?? false]);
```

Issue #10231, opened 2020-03-17
Foundation Bugs on SignIn page


GeetikaMuthyam:
- [ ] ON Mobile, when I goto the sign-in page, I see the SAML-Over-OAuth button is wrong, 
Only when I do hover, it appears the text
1. Go to the Login page on MOBILE.
2. I see the SAML OAuth button, the text appears only when I do HOVER.

![Screen Shot 2020-03-11 at 2 54 01 PM](https://user-images.githubusercontent.com/56363201/76524382-5b42a400-6440-11ea-9309-1234ce3847a2.png)

PR #10316, opened 2020-03-30
Foundation - Style fixes for login page

slafleche:
This PR fixes visual issues on the sign in page. Any additional work on improving the design them past this point should be an iteration in another issue. 

Screen Shot 2020-03-30 at 08 11 58
Screen Shot 2020-03-30 at 08 11 45

I tried to make the components consistent between each other, but they have different markup and don't always have an icon. If @ElisaAlmeida has any suggestions? Maybe we can make a placeholder icon, like a key or something for ones without an icon.

Closes: https://github.com/vanilla/vanilla/issues/10231


charrondev:
It looks like this includes all of the stuff from the link state PR as well. I'll circle back to it once that is merged.

PR #10318, opened 2020-03-30
Adjust rich editor menu items for accesibility

alex-mtl:
I did improve the accessibility of the editor. Please note, however that "Read Aloud", the extension mentioned by Andrew is **NOT** appropriate to test the accessibility of our editor. We already had good accessibility with Voice Over, I made some improvements that should make it even more accessible. However, I do not have access to NVDA or JAWS (both requiring windows machines). So I cannot test. 

Relates to: https://github.com/vanilla/support/issues/1495


slafleche:
@cchabilall83 I think you need to approve this one, because alex and I worked on the PR and Adam approved it already.

PR #10325, opened 2020-03-30
Update MockSiteSectionProvider and fix themeModel infinite loop

cchabilall83:
Small changes to MockSiteSectionProvider.

Fix for infiniteloop when Garden.CurrentTheme and  Garden.Theme aren't available.

required for https://github.com/vanilla/internal/pull/2338

related to https://github.com/vanilla/knowledge/issues/1706


library/Vanilla/Models/FsThemeProvider.php
```diff
@@ -123,7 +123,6 @@ public function getName($themeKey): string {
     public function getThemeAddon($themeKey): AddonInterface {
         $theme = $this->addonManager->lookupTheme($themeKey);
         if (!($theme instanceof AddonInterface)) {
-            $theme = $this->addonManager->lookupTheme(ThemeModel::FALLBACK_THEME_KEY);
```
alex-mtl: Are we sure this patch works?

charrondev: Chris, I do believe this may cause some issues in its removed (when fetching individual assets). Do any of your test cases cover that with an invalid theme?

cchabilall83: I've reverted the change and added some tests.

library/Vanilla/Models/FsThemeProvider.php
```diff
@@ -133,6 +133,23 @@ public function getThemeAddon($themeKey): AddonInterface {
...
+     * @param string|int $themeKey
+     * @return bool
+     */
+    public function themeExists($themeKey): bool {
+        $themeExists = true;
+        $theme = $this->addonManager->lookupTheme($themeKey);
+
+        if (!($theme instanceof AddonInterface) || is_null($theme)) {
+            $themeExists = false;
+        }
```
alex-mtl: Can't we default to `$themeExists = false;` until we check it and have any proof?
```suggestion
        $themeExists = false;
        $theme = $this->addonManager->lookupTheme($themeKey);
        if ($theme instanceof AddonInterface) {
            $themeExists = true;
        }
```

library/Vanilla/Models/ThemeModel.php
```diff
@@ -405,7 +416,7 @@ public function getCurrentTheme(): ?array {
                 $current = $provider->getThemeWithAssets(self::FALLBACK_THEME_KEY);
             }
         } catch (\Exception $e) {
-            trigger_error($e->getMessage(), E_USER_ERROR);
+            trigger_error($e->getMessage(), E_USER_NOTICE);
```
charrondev: Why are we downgrading this to a notice? I figure it should be a warning at least if someone has an invalid theme configured.

cchabilall83: I set it to Warning

Issue #10302, opened 2020-03-26
Import Bug with \N (NULL) Values


MrCaspan:
I am exporting data from vBulletin 4 and NULLs are represented by \N in the export file. During the import I am receiving the following error

Incorrect integer value: '' for column 'Garden.Profiles.View' at row 1

This lead to to finding out that \N was being replaced by a blank field value ie '' which is a opening single tick followed by a closing single tick with noting in between represented in the rest of the bug report as '{blank}' where {blank} is nothing no value

The file was being read in properly and the \N was replaced with NULL in the array but then it runs through a trim call back function and the trim command is turning NULL values into '{blank}' 

**Vanilla info**
1. Are you sure this is a core Vanilla problem and not caused by any addon you are using?
YES, this is on a blank install of 3.3 then importing my data

2. Did you follow our generic troubleshooting steps already?
YES,

3. What Vanilla version are you using?
3.3

4. When did the issue start?
When I tried to do an import

**To Reproduce**
Export a site that has NULL values in the database.
Verify those values are null in the new database
You will notice that \N get turned into ''

If you debug the function fGetCSV2 in class.importmodel.php if you var_dump() the $results array just the line before it's returned to loadTableInsert() that called it, you will see that this function properly converts \N to a null value. This array is returned to the loadTableInsert() function into the While Loop and is saved as $row and one of the commands preformed on $row is

$row = array_map('trim', $row);

From what I understand this will take each element in the array send it to the trim command then place the result back into the array in the same place.

Here is where the issue is, If you look at the trim manual on PHP.net https://www.php.net/manual/en/function.trim.php#72306 the comment says 

"It may be useful to know that trim() returns an empty string when the argument is an unset/null variable."

so basically this trim CallBack function turned any array element with a value of NULL to '{blank}' and to a database it does not like '' value being passed to a field that is for instance tinyInt(4) NULL allowed as '' is not NULL or an INT. This throw errors when trying to import

**Expected behavior**
If a NULL is exported and NULL shoudl be inserted 

**Additional context**
You can fix this by not removing the \N in the fGetCSV2 function. After you implode the array in  the loadTableInsert() function do a str_replace for and search for '\\N', and replace it with null (No Quotes) 

Next to that replace \N in the fGetCSV2() with the string 'NULL' (With Quotes) so it will be treated like text. After you implode the array in the loadTableInsert() function do a str_replace for and search for 'NULL' (With Quotes) and replace it with NULL (No Quotes)

[...]

PR #10326, opened 2020-03-30
Fix Null Imports from Porter App

MrCaspan:

https://github.com/vanilla/vanilla/issues/10302

### DIFF
--- a/applications/dashboard/models/class.importmodel.php
+++ b/applications/dashboard/models/class.importmodel.php
@@ -664,7 +664,7 @@ class ImportModel extends Gdn_Model {
             if ($inEscape) {
                 // Check for an escaped null.
                 if ($c == 'N' && strlen($token) == 0) {
-                    $token = null;
+                    $token = 'NULL'; // This will be replaced later with proper null value before an insert
                 } else {
                     $token .= $c;
                 }
@@ -2228,6 +2228,7 @@ class ImportModel extends Gdn_Model {
                 $inserts .= ',';
             }
             $inserts .= '('.implode(',', $row).')';
+            $inserts = str_replace("'NULL',", "null", $inserts);  // Need to replace any 'NULL',  values with actual null value for database
 
             if ($count >= $chunk) {
                 // Insert in chunks.
--- a/applications/dashboard/models/class.importmodel.php
+++ b/applications/dashboard/models/class.importmodel.php
@@ -2228,7 +2228,7 @@ class ImportModel extends Gdn_Model {
                 $inserts .= ',';
             }
             $inserts .= '('.implode(',', $row).')';
-            $inserts = str_replace("'NULL',", "null", $inserts);  // Need to replace any 'NULL',  values with actual null value for database
+            $inserts = str_replace("'NULL',", "null,", $inserts);  // Need to replace any 'NULL',  values with actual null value for database
 
             if ($count >= $chunk) {
                 // Insert in chunks.


MrCaspan:
How do i get 2 approving reviews?

applications/dashboard/models/class.importmodel.php
```diff
@@ -2228,6 +2228,7 @@ public function loadTableInsert($tablename, $path, $skipHeader = true, $chunk =
                 $inserts .= ',';
             }
             $inserts .= '('.implode(',', $row).')';
+            $inserts = str_replace("'NULL',", "null,", $inserts);  // Need to replace any 'NULL',  values with actual null value for database
```
alex-mtl, line 2231: Was it ever tested? 
IMHO this line will do nothing. because `$inserts` string should contain string like `(var1,var2,var3,NULL,var4)`
no quotes until quotes (single or double) is a part of those variable values.

MrCaspan: You have to look at the fGetCSV2 function when it reads in the txt files with the data in them they could have \N to represent NULL value. when it reads in the file it properly creates an array of values for each filed form the txt file and the \N get converted to null properly in the array. Problem is when it is returned to the loadTableInsert an trim is done on the Array and it changes all the NULL values to.. so to get aroudn this i suggest we change \N to the string value 'NULL' then after the trim and implode we just replace the 'NULL', with   null, this was the only way I could get the process to function and to get my data imported into VB

Next to that just replace the subroutine in fGetCSV2 that replaces the \N with null and jsut leave the \N in there till just before the inset and replace \N with null

MrCaspan: Here is an example of an export that would just error out on import

```
RoleID:int,Garden.SignIn.Allow:tinyint,ForumPermissions:varchar(255),GenericPermissions:varchar(255),Garden.Profiles.View:tinyint,Garden.Activity.View:tinyint,Garden.Profiles.Edit:tinyint,Plugins.Signatures.Edit:tinyint(1),Vanilla.Discussions.View:tinyint,Vanilla.Discussions.Add:tinyint,Vanilla.Comments.Add:tinyint,Plugins.Attachments.Download:tinyint(1),Plugins.Attachments.Upload:tinyint(1),Garden.Moderation.Manage:tinyint,Vanilla.Discussions.Announce:tinyint,Vanilla.Discussions.Close:tinyint,Vanilla.Discussions.Delete:tinyint,Vanilla.Comments.Delete:tinyint,Vanilla.Comments.Edit:tinyint,Vanilla.Discussions.Edit:tinyint,Vanilla.Discussions.Sink:tinyint,Garden.Activity.Delete:tinyint,Garden.Users.Add:tinyint,Garden.Users.Edit:tinyint,Garden.Users.Approve:tinyint,Garden.Users.Delete:tinyint,Garden.Applicants.Manage:tinyint(1),Garden.Settings.View:tinyint,Garden.Settings.Manage:tinyint,Garden.Messages.Manage:tinyint(1),Vanilla.Spam.Manage:tinyint(1)
"1",0,"9048079","64",\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N,\N
```

alex-mtl: @MrCaspan It is clear what you are trying to achieve. My comment only relates to one particular line of code.
From what I can understand before your changes we have some array like:
```
["1",0,"9048079","64",\N,\N, ...]
```
you want to transform it to
```
["1",0,"9048079","64","NULLL","NULLL", ...]
```
and then  we implode it to:
```
(1,0,9048079,64,NULLL,NULLL, ...)
```
please notice that imploded string has no quotes!
https://ideone.com/i8HuS7

That means that you can't replace with `str_replace("'NULL',"`
because there is no such substring there

MrCaspan: that's almost correct except \N get read in as a string value so your first array shoudl be  
`["1",0,"9048079","64","\N","\N", ...]`
Then I replace the "\N" with "NULL"
Then just before the insert replace "NULL", with NULL,   i check for the comma in the text string just to make sure i have captured a NULL value and not someone typing NULL in a comment or something

Next to that you could not replace the \N in the fGetCSV2 function but leave it there till just after the implode and replace "\N", will NULL, at that point.. one less step.

MrCaspan: Good point on your implode.. let me double check that!!

MrCaspan: okay now i know why I had to do this... there is a subroutine in fGetCSV2 that takes all the Array elements and puts them in '' so that when its imploded there is still single quotes around text So in the array you will see NULL will be [8] => "'NULL'" that is why i have to search for 'NULL', and replace it with just NULL,

vredeling: I believe one of the problems is the line trimming the row values:
`$row = array_map('trim', $row);`.
This will transform `NULL` values to `''` strings.
Casting these _NULL_ values to a _'null'_ string before trimming will circumvent this issue. Like so:

```suggestion
           foreach ($row as &$col_value) {
                if (is_null($col_value)) {
                    $col_value = 'null';
                }
            }
            $row = array_map('trim', $row);
```
and
```suggestion
            $inserts = str_replace("'null'", 'NULL', $inserts);
```

PR #10330, opened 2020-03-31
Extend database Operation with mode: force, default.

alex-mtl:
Adjust CurrentDateFieldProcessor to reflect operation mode accordingly.

Relates to: https://github.com/vanilla/knowledge/issues/1671


library/Vanilla/Database/Operation/CurrentDateFieldProcessor.php
```diff
@@ -61,7 +61,13 @@ public function handle(Operation $databaseOperation, callable $stack) {
             $fieldExists = $databaseOperation->getCaller()->getWriteSchema()->getField("properties.{$field}");
             if ($fieldExists) {
                 $set = $databaseOperation->getSet();
-                $set[$field] = date("Y-m-d H:i:s");
+                if (empty($set[$field] ?? null) || $databaseOperation->getMode() === Operation::MODE_FORCE) {
```
charrondev, line 64: Is this conditional backwards? I'd imagine if the mode is force then we want to do the opposite behaviour.

In any case I think we could use a test for this in `CurrentDateFieldProcessTest`

PR #10334, opened 2020-03-31
Fix the error message for Config Garden.Cookie.Domain

egarciajulia:
Swapped the order of:
+ clearing the value of $this->CookieDomain
+ emitting the error
The current value is used in the description of the error.


egarciajulia:
Some issue with my last name is not allowing the tests to pass. I am closing the PR and I will re-create later

PR #10335, opened 2020-03-31
Set default label to currentTheme if not theme is selected

cchabilall83:
Sets label to `Default ({currrentTheme})` when initialValue is null.

relates to https://github.com/vanilla/multisite/issues/349

### DIFF
--- a/library/src/scripts/theming/ThemeChooserInput.tsx
+++ b/library/src/scripts/theming/ThemeChooserInput.tsx
@@ -3,7 +3,7 @@
  * @license GPL-2.0-only
  */
 
-import { IComboBoxOption } from "@library/features/search/SearchBar";
+import { IComboBoxOption, IComboBoxOptionwithNull } from "@library/features/search/SearchBar";
 import { t } from "@vanilla/i18n";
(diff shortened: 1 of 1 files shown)


library/src/scripts/theming/ThemeChooserInput.tsx
```diff
@@ -65,7 +67,7 @@ export function ThemeChooserInput(props: IProps) {
                 value: option.value,
             };
         }
-    }) ?? { label: t("Unknown"), value: props.initialValue };
+    }) ?? { label: `Default (${defaultTheme.name})`, value: props.initialValue };
```
slafleche: Shouldn't you be using `<Translate` with a placeholder here?

cchabilall83: should be good now

cchabilall83:
> A couple change requests here.
> 
> 1. We should get this as an actual option in the menu. One I can explicitly choose. Right now if I've set the subcommunity once, there is no clear item to set it back to the default.
> 2. We should probably have some parenthese in the translation string. Something like `Default (Foundation)`

@charrondev  should be addressed

PR #10336, opened 2020-03-31
Add a Roles as translatable field to OAuth2 feature.

### LINKED ISSUE #10265: Allow the roles key to be mapped on the OAuth addon
In the OAuth plugin you can rename all of the keys passed, except roles. We want to be able to rename the key that roles are mapped to on the following config page:

image (11)

nervoustwit:
The OAuth2 SSO method allows for admins to translate the keys of the profile response for authenticating a user so that it matches what our script expects. More and more organizations pass Roles over SSO. This PR will add Roles to the translation array and gives an interface for the admins to modify the name of the key.

### Testing.
(Backwards compatibility)
 - With an existing OAuth2 configuration:
    - Checkout this branch.
    - Connect over SSO.
 - With an existing OAuth2 configuration:
   - Check out this branch.
   - Modify the Roles field in the Dashboard Settings.
   - Connect over SSO
 - With an existing OAuth2 configuration:
   - Check out this branch.
   - Modify the Roles field in the Dashboard Settings to match roles being passed. e.g. put 'position' in the Roles field. (if you are unable to pass roles over SSO you can fudge it by inserting a role into the rawProfile array before it gets translated [see here](https://github.com/vanilla/vanilla/blob/d510e30533a4be57a8bae9c7815d7c65fef02604/library/core/class.oauth2.php#L840) e.g. `$rawProfile['position'] = 'Moderator';`. )
   - Connect over SSO.
   - Verify that the user is given the role of Moderator.
 - Repeat the above tests with a fresh configuration of the OAuth2

Closes #10265

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -791,12 +791,12 @@ class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
     public function translateProfileResults($rawProfile = []) {
         $provider = $this->provider();
         $translatedKeys = [
-            $provider['ProfileKeyEmail'] ?? 'email' => 'Email',
-            $provider['ProfileKeyPhoto'] ?? 'picture' => 'Photo',
-            $provider['ProfileKeyName'] ?? 'displayname' => 'Name',
-            $provider['ProfileKeyFullName'] ?? 'name' => 'FullName',
-            $provider['ProfileKeyUniqueID'] ?? 'user_id' => 'UniqueID',
-            $provider['ProfileKeyRoles'] ?? 'roles' => 'Roles'
+            ($provider['ProfileKeyEmail'] ?? 'email') => 'Email',
+            ($provider['ProfileKeyPhoto'] ?? 'picture') => 'Photo',
+            ($provider['ProfileKeyName'] ?? 'displayname') => 'Name',
+            ($provider['ProfileKeyFullName'] ?? 'name') => 'FullName',
+            ($provider['ProfileKeyUniqueID'] ?? 'user_id') => 'UniqueID',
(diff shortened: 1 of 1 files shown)


library/core/class.oauth2.php
```diff
@@ -788,13 +790,13 @@ public function requestAccessToken($code, $refresh=false) {
...
-            val('ProfileKeyPhoto', $provider, 'picture') => 'Photo',
-            val('ProfileKeyName', $provider, 'displayname') => 'Name',
-            val('ProfileKeyFullName', $provider, 'name') => 'FullName',
-            val('ProfileKeyUniqueID', $provider, 'user_id') => 'UniqueID'
+            $provider['ProfileKeyEmail'] ?? 'email' => 'Email',
+            $provider['ProfileKeyPhoto'] ?? 'picture' => 'Photo',
+            $provider['ProfileKeyName'] ?? 'displayname' => 'Name',
+            $provider['ProfileKeyFullName'] ?? 'name' => 'FullName',
+            $provider['ProfileKeyUniqueID'] ?? 'user_id' => 'UniqueID',
+            $provider['ProfileKeyRoles'] ?? 'roles' => 'Roles'
```
initvector, line 799: For the sake of improving readability...

```suggestion
            ($provider['ProfileKeyEmail'] ?? 'email') => 'Email',
            ($provider['ProfileKeyPhoto'] ?? 'picture') => 'Photo',
            ($provider['ProfileKeyName'] ?? 'displayname') => 'Name',
            ($provider['ProfileKeyFullName'] ?? 'name') => 'FullName',
            ($provider['ProfileKeyUniqueID'] ?? 'user_id') => 'UniqueID',
            ($provider['ProfileKeyRoles'] ?? 'roles') => 'Roles'
```

PR #10337, opened 2020-04-01
Attempt to debug the failed revert when the theme's font is in our de…

slafleche:
WIP : Haven't solved the issue yet.

For: https://github.com/vanilla/knowledge/issues/1695

It seems to work fine when the theme you're editing starts with a custom font. I suspect it's because it's fetching the last value from the variables stored. When you start with the default, that's value isn't necessarily explicitly set and it bugs out.


charrondev:
Closing for https://github.com/vanilla/vanilla/pull/10337

PR #10354, opened 2020-04-02
Fix check for image in BBCode parser

MrCaspan:
Currently the logic tests the Media ImageWidth value to see if it has any value other then 0 or NULL to conclude if it is an image. This logic would have to assume that the Media ImageWidth column is 100% correct. But there are cases where ImageWidth value could be NULL for an image (IE. After an import form another forum) So instead of the images being shown properly inline only a link to the image is shown. 

The changes made above tests the file before displaying it to make sure its a file this guarantees that images are displayed properly and non images are displayed as a link

### DIFF
--- a/library/core/class.bbcode.php
+++ b/library/core/class.bbcode.php
@@ -54,7 +54,8 @@ class BBCode extends Gdn_Pluggable {
             $src = htmlspecialchars(Gdn_Upload::url(val('Path', $media)));
             $name = htmlspecialchars(val('Name', $media));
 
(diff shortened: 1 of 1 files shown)


MrCaspan:
This is the issue
https://github.com/vanilla/vanilla/issues/10355 

I have tested this code o my site and this fixes the issue.

charrondev:
If this still doesn't work for your use case (like an import that needs images processed), I'd recommend writing a one time job to process all of your imports.

We certainly don't want to be doing it on every single render of some content.

PR #10358, opened 2020-04-02
Add DashboardRadioGroups grid story

igraziatto:
Add story for styles used on https://github.com/vanilla/internal/pull/2322

**This PR is a dependency for:**
https://github.com/vanilla/internal/pull/2322

### DIFF
--- a/applications/dashboard/src/scripts/forms/DashboardForm.story.tsx
+++ b/applications/dashboard/src/scripts/forms/DashboardForm.story.tsx
@@ -123,6 +123,14 @@ export function RadioGroups() {
                             <DashboardRadioButton value={"option3"} label="Option 3" disabled />
                         </DashboardRadioGroup>
                     </DashboardFormGroup>
+                    <DashboardFormGroup label="Radio Group Grid">
+                        <DashboardRadioGroup onChange={setGroup2} value={group2} isGrid>
+                            <DashboardRadioButton value={"option1"} label="Option 1" />
+                            <DashboardRadioButton value={"option2"} label="Option 2" />
+                            <DashboardRadioButton value={"option3"} label="Option 3" />
+                            <DashboardRadioButton value={"option4"} label="Option 4" disabled />
+                        </DashboardRadioGroup>
+                    </DashboardFormGroup>
                     <DashboardFormGroup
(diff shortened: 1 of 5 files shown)


applications/dashboard/src/scripts/forms/DashboardRadioGroups.tsx
```diff
@@ -46,7 +47,7 @@ export function DashboardRadioGroup(props: IProps) {
     return (
         <RadioGroupContext.Provider value={props}>
             <div
-                className={classNames(rootClass, { inline: props.isInline })}
+                className={classNames(rootClass, { inline: props.isInline }, { grid: props.isGrid })}
```
slafleche, line 50: This is not incorrect, but FYI, you could of combined the two conditional class objects into 1 here.

applications/dashboard/src/scripts/forms/DashboardForm.story.tsx
```diff
@@ -123,6 +123,14 @@ export function RadioGroups() {
                             <DashboardRadioButton value={"option3"} label="Option 3" disabled />
                         </DashboardRadioGroup>
                     </DashboardFormGroup>
+                    <DashboardFormGroup label="Radio Group Grid">
```
slafleche, line 126: Don't forget that storybook is also the place to make sure we don't have bugs. I'd like to see cases where the texts are different lengths, maybe some examples on smaller screens?

library/src/scripts/forms/checkRadioStyles.ts
```diff
@@ -224,6 +224,17 @@ export const checkRadioClasses = useThemeCache(() => {
         marginTop: unit(globalVars.spacer.size / 2),
     });
 
+    const grid = style("grid", {
```
slafleche, line 227: This will work when you have enough room and short enough text. I'm not overly concerned about the responsiveness for the dashboard, but you might want to make sure they stay aligned, you might want to opt for a css grid or an old school css table.

charrondev: You could probably still flex them, but just a flex basis is insufficient

image

charrondev:
Still not looking so hot from the storybook side

image

PR #10361, opened 2020-04-03
Fix regional locale variants displaying incorrectly

cchabilall83:
Fixes regional locale variants being displayed as `unknown`.  We were originally not trimming to the locale key to the first 2 characters which exclude the variant (ie. pt_BR would be just pt).  This has been changed and the region has been appended to display name if there is one.

closes https://github.com/vanilla/multisite/issues/354

### DIFF
--- a/applications/dashboard/controllers/api/LocalesApiController.php
+++ b/applications/dashboard/controllers/api/LocalesApiController.php
@@ -71,9 +71,12 @@ class LocalesApiController extends Controller {
         $hasEnLocale = false;
         $result = [];
         foreach ($locales as $localeID => $locale) {
+            $localKey = (strlen($locale['Locale']) === 5) ?
+                substr($locale['Locale'], 0, 5) :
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/api/LocalesApiController.php
```diff
@@ -71,9 +71,12 @@ private function getEnabledLocales(): array {
         $hasEnLocale = false;
         $result = [];
         foreach ($locales as $localeID => $locale) {
+            $localKey = (strlen($locale['Locale']) === 5) ?
+                substr($locale['Locale'], 0, 5) :
+                substr($locale['Locale'], 0, 2);
```
charrondev, line 76: I think we can just go with `$locale['Locale']`, in this case localeKey and regionalKey will both be the same, but it's ok.

We already do normalization of these by this point.

cchabilall83: i've set it to $locale['Locale']

PR #10370, opened 2020-04-04
Fix canonical URL generation

tburry:
Endpoints added by event handlers get passed an `args` that is an array of remaining path variables. This wasn’t being added properly to calculated canonical URLs and was causing a notice.

I'm also removing the canonical URL comparison since it isn't really used anymore.

### DIFF
--- a/library/core/class.dispatcher.php
+++ b/library/core/class.dispatcher.php
@@ -854,10 +854,6 @@ class Gdn_Dispatcher extends Gdn_Pluggable {
 
 
         $canonicalUrl = url($this->makeCanonicalUrl($controller, $method, $args), true);
-        $canonicalUrlOld = $this->makeCanonicalUrlOld($controller);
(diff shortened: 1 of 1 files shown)


library/core/class.dispatcher.php
```diff
@@ -970,8 +966,10 @@ protected function makeCanonicalUrl($controller, \ReflectionFunctionAbstract $me
         // Add args after known names to the querystring.
         $query = [];
         $split = false;
+        $hasSender = false;
```
initvector, line 969: What's the primary benefit of detecting something like this and adjusting `$args` in the callee versus making adjustments or indicating there's a sender from the caller?

tburry: It was kind of a heuristic there. I was more worried that other endpoints could have an `$args` parameter and so I wanted to match the event handler as closely as possible.

Issue #10347, opened 2020-04-02
Input & buttons don't have enough padding with large border radiuses


charrondev:
~- Crank the border radius of your inputs and buttons.~
~- Notice the text can feel a little off-center in the input. (This is just adding more padding so the text doesn't start too deep in the curve)~
- We should have some algorithm to force a minimum border padding based on the border radius.
  - Theme control will change buttons and search bar boarder radius
    - [ ] 20px button padding for buttons
    - [ ] 16px for inputs at max roundedness

GeetikaMuthyam:
**Testing Results:**
Bug : https://github.com/vanilla/knowledge/issues/1753

PR #10381, opened 2020-04-07
Dynamic padding for buttons/inputs based on border radius

slafleche:
- Created function to dynamically set padding of buttons/input based on their border radius.
- Changed variable and label used for the border radius control in the theme editor (as specified in demo)
- Refactored the border radius variables to get rid of useless nesting
- Fixed some drop down styles
- Fixed inconsistencies with the primary and main buttons, they were using the wrong border radius variable

Closes: https://github.com/vanilla/vanilla/issues/10347

Companion PR: https://github.com/vanilla/internal/pull/2369


library/src/scripts/forms/buttonStyles.ts
```diff
@@ -276,13 +283,92 @@ export const buttonVariables = useThemeCache((forcedVars?: IThemeVariables) => {
...
+    if (props && props.radius && props.extraPadding) {
+        const maxValue = parseInt(props.extraPadding.toString());
+        const rawRadius = props.radius.toString().trim();
+        const workingBorderRadius = parseFloat(rawRadius);
+        const halfHeight = height / 2;
+
+        logDebugConditionnal(debug, "maxValue: ", maxValue);
+        logDebugConditionnal(debug, "rawRadius: ", rawRadius);
+        logDebugConditionnal(debug, "workingBorderRadius: ", workingBorderRadius);
+
```
alex-mtl, line 313: Do we push such debug code to the repo?

library/src/scripts/forms/buttonStyles.ts
```diff
@@ -276,13 +283,78 @@ export const buttonVariables = useThemeCache((forcedVars?: IThemeVariables) => {
...
-    const borderWidth = formElementVars.borders ? formElementVars.borders : buttonGlobalVariables().border.width;
+export interface IPaddingBasedOnRadius {
+    radius?: number | string; // pixel value in integer or percentage in string
+    extraPadding?: number | string; // If undefined, return 0 for both sides
+    height?: number; // Min height of element
+    side?: "horizontal" | "left" | "right"; // defaults to horizontal
+    debug?: boolean;
+}
+
+export const getPaddingOffsetBasedOnBorderRadius = (props: IPaddingBasedOnRadius) => {
```
charrondev, line 294: Looking at this function and it's usage outside buttons, I think we should use move this it's own file (or maybe the input styles?)

charrondev:
@slafleche You're tests are failing here.

PR #8416, opened 2019-02-19
admin.css font-face source realtive to root not folder

x00:
https://github.com/vanilla/vanilla/issues/8370

Please make sure these points are completed:

* [ ] Title is formatted in present imperative (ex: "Fix X by doing Y").
* [ ] Title makes sense as a release note, not a literal technical description.
* [x] An issue was filed, and this links to it.
* [x] Any related PRs are cross-referenced via link.
* [ ] Your description tells the story of why _this_ is the right change.
* [ ] You list QA steps or testing criteria if appropriate.

PR #10396, opened 2020-04-09
Change Vanillicon font path reference from absolute to relative for forums in subfolders

R-J:
admin.css shows an absolute path for the vanillicon font which results in a broken link in forums in subfolders.
This PR replaces an [older one](https://github.com/vanilla/vanilla/pull/8416) which has been closed months ago...


charrondev:
I had to jump on here and fix our dashboard build after some recent changes so I could resolve the merge conflicts.

PR #10408, opened 2020-04-12
Add the ability for Gdn_Dispatcher to use the container to satisfy type-hinted args

tburry:
This PR implements a new `Dispatcher::reflectArgs()` method to replace the global `reflectArgs()` function. It is mostly the same, but allows for type-hinted class parameters to be satisfied by the container.

The use-cases for this are as follows:

- Some controllers may not want to inject an object for every single method, but rather only the method that needs it.
- For SSO, it’s sometimes easier to just use the `RequestInterface` rather than map arguments. This way you can implement against an exact spec, using a schema against the query parameters or something like that.

### DIFF
--- a/library/Garden/ArrayContainer.php
+++ b/library/Garden/ArrayContainer.php
@@ -25,7 +25,7 @@ class ArrayContainer extends \ArrayObject implements ContainerInterface {
      * @param bool $lazy Whether or not to lazy instantiate objects that aren't in the container.
      */
     public function __construct($lazy = false) {
-        parent::__construct(null, 0, 'ArrayIterator');
+        parent::__construct([], 0, 'ArrayIterator');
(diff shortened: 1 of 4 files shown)


library/Garden/ArrayContainer.php
```diff
@@ -25,7 +25,7 @@ class ArrayContainer extends \ArrayObject implements ContainerInterface {
      * @param bool $lazy Whether or not to lazy instantiate objects that aren't in the container.
      */
     public function __construct($lazy = false) {
-        parent::__construct(null, 0, 'ArrayIterator');
+        parent::__construct([], 0, 'ArrayIterator');
```
tburry, line 28: It looks like this code was never used. These are straight up bugs.

PR #10410, opened 2020-04-13
Fix banner search strip positioning and home widget background

charrondev:
Issue https://github.com/vanilla/knowledge/issues/1700

**NOTE: I highly recommend [viewing with whitespace changes hidden](https://github.com/vanilla/vanilla/pull/10410/files?diff=split&w=1)**

## Search Strip

The search strip was being positioned on top of the background image, even if there was no offset being applied. This was due to a bug after our previous round of overflow fixes. Basically a `position: relative` container was missing.

## Home Widget BG

Update the home widget inner container default color to match the body background instead of the main colors bg. The body BG defaults to the main color BG, but it the home widget is meant to blend with the body BG even if explicitly set.

### DIFF
--- a/library/src/scripts/banner/Banner.tsx
+++ b/library/src/scripts/banner/Banner.tsx
@@ -112,21 +112,59 @@ export default function Banner(props: IProps) {
                 - Spacer elements for all the main content, but no actual content.
                 - Overflow hidden can't be applied to the main content, because it has a search box that can't be cut off.
             */}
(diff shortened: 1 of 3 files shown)


charrondev:
Unrelated to this PR as it's like this on master. https://www.chromaticqa.com/component?appId=5d5eba16c782b600204ba187&name=Banner&buildNumber=3024&specName=Unified%20Border

I opened a new issue for it https://github.com/vanilla/vanilla/issues/10414

PR #10418, opened 2020-04-14
Fix conversation message MaxLength config

dannimajdalani:
closes https://github.com/vanilla/support/issues/1579#event-3232022436
### Details

Conversation message max length is configurable. You can do that by setting 
$Configuration['Conversations']['Message']['MaxLength']. 
Setting this config is not taking effect because of 
https://github.com/vanilla/vanilla/blob/1f988352b1888568f8160c5cc7a96c38f5cae860/applications/conversations/controllers/api/MessagesApiController.php#L121

We are looking for MaxLength config instead of Conversations.Message.MaxLength.

### DIFF
--- a/applications/conversations/controllers/api/MessagesApiController.php
+++ b/applications/conversations/controllers/api/MessagesApiController.php
@@ -118,7 +118,7 @@ class MessagesApiController extends AbstractApiController {
                 'messageID:i' => 'The ID of the message.',
                 'conversationID:i' => 'The ID of the conversation.',
                 'body:s' => [
-                    'maxLength' => $this->config->get('MaxLength', 2000),
+                    'maxLength' => $this->config->get('Conversations.Message.MaxLength', 2000),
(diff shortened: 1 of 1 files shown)


applications/conversations/controllers/api/MessagesApiController.php
```diff
@@ -118,7 +118,7 @@ private function fullSchema() {
                 'messageID:i' => 'The ID of the message.',
                 'conversationID:i' => 'The ID of the conversation.',
                 'body:s' => [
-                    'maxLength' => $this->config->get('MaxLength', 2000),
+                    'maxLength' => $this->config->get('Conversations.Message.MaxLength', 2000),
```
initvector, line 121: I think you're only going to want the length restrictions to be applied to the write schemas. If you're applying them to all the schemas, then you're running the risk of having existing conversation messages break validation in API responses after decreasing the maximum length.

PR #10423, opened 2020-04-14
Fix SSO provider can be default when disabled

### LINKED ISSUE #10331: Unset OAuth2 as the default sign in method when addon disabled.
When you set Oauth2 as your default method of signing in but then turn the addon off, it remains configured as the default method of signing in. This creates several unwanted and confusing behaviours.

Related: https://github.com/vanilla/addons/issues/788
Related: https://github.com/vanilla/internal/issues/2342

blutarch:
vanilla/addons#796 depends on this PR.
vanilla/internal#2399 depends on this PR.

Partially addresses vanilla/support#1618
Partially addresses vanilla/internal#2342.
Closes #10331

This PR fixes a problem where an SSO provider could be the default authentication provider even when disabled. This PR creates a class, SSOAddon, that extends Gdn_Plugin and includes a method that sets the `Gdn_UserAuthenticationProvider`'s 'IsDefault' value to 0 for any provider matching the 'AuthenticationSchemeAlias'.

It also adds an authenticationSheme variable to the Gdn_oauth2 class, which can be set in the Oauth2Plugin constructor.

### TO TEST
I have done functional tests with every SSO plugin I know about, but this probably should be tested with one or two client custom oauth plugins as well.

1. Enable an authentication provider that extends the SSOAddon class.
1. Set the provider's 'IsDefault' value in the `Gdn_UserAuthenticationProvider` table to 1.
1. Disable the addon and verify that the 'IsDefault' value is still 1.
1. Check out this branch.
1. Repeat steps 1-3, except this time verify that the 'IsDefault' value changes to 0.

### DIFF
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -440,6 +440,7 @@ class Gdn_OAuth2 extends SSOAddon implements \Vanilla\InjectableInterface {
         } else {
 
             $form->setFormValue('AuthenticationKey', $this->getProviderKey());
+            $form->setFormValue('AuthenticationSchemeAlias', $this->getAuthenticationScheme());
 
             $sender->Form->validateRule('AssociationKey', 'ValidateRequired', 'You must provide a unique AccountID.');
             $sender->Form->validateRule('AssociationSecret', 'ValidateRequired', 'You must provide a Secret');
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -27,13 +27,16 @@ use Vanilla\Permissions;
  * any of its methods of constants.
  *
  */
-class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
+class Gdn_OAuth2 extends SSOAddon implements \Vanilla\InjectableInterface {
 
     /** @var string token provided by authenticator  */
     protected $accessToken;
 
     /** @var array response to token request by authenticator  */
     protected $accessTokenResponse;
+
+    /** @var string AuthenticationSchemeAlias value */
+    protected $authenticationScheme = '';
 
     /** @var string key for GDN_UserAuthenticationProvider table  */
     protected $providerKey = null;
@@ -58,6 +61,18 @@ class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
 
     /** @var  @var string optional set the settings view */
     protected $settingsView;
+
+    /**
+     * Gets the $authenticaitonScheme variable
+     *
+     * @return string
+     */
+    protected function getAuthenticationScheme(): string {
+        if (!$this->authenticationScheme) {
+            return '';
+        }
+        return $this->authenticationScheme;
+    }
 
     /**
      * @var SsoUtils
--- a/library/core/class.oauth2.php
+++ b/library/core/class.oauth2.php
@@ -63,24 +63,6 @@ class Gdn_OAuth2 extends SSOAddon implements \Vanilla\InjectableInterface {
     protected $settingsView;
 
     /**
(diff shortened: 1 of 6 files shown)


library/core/class.oauth2.php
```diff
@@ -27,14 +27,17 @@
...
 
     /** @var string token provided by authenticator  */
     protected $accessToken;
 
     /** @var array response to token request by authenticator  */
     protected $accessTokenResponse;
 
+    /** @var string AuthenticationSchemeAlias value */
+    protected $authenticationScheme = '';
+
```
nervoustwit, line 40: This should be call authenticationSchemeAlias to be consistent.

library/core/class.oauth2.php
```diff
@@ -59,6 +62,18 @@ class Gdn_OAuth2 extends Gdn_Plugin implements \Vanilla\InjectableInterface {
...
+     *
+     * @return string
+     */
+    protected function getAuthenticationScheme(): string {
+        if (!$this->authenticationScheme) {
+            return '';
+        }
+        return $this->authenticationScheme;
+    }
+
```
nervoustwit, line 76: After testing this with other plugins that extend this class already, this should be:
```
protected function getAuthenticationScheme(): string {
    return $this->authenticationScheme ?: $this->providerKey;
}
```
because that is the logic we have been using.

plugins/googlesignin/class.googlesignin.plugin.php
```diff
@@ -22,6 +22,7 @@ class GoogleSignInPlugin extends Gdn_OAuth2 {
     public function __construct() {
         $this->setProviderKey('googlesignin');
         $this->settingsView = 'plugins/settings/googlesignin';
+        $this->authenticationScheme = 'googlesignin';
```
nervoustwit, line 25: We need to create a setter method in the OAuth2 class so that here we can optionally call it. If we don't call it here, as per my comment above, the getAuthenticationSchemeAlias() method will just set it to the same this as $this->providerKey. We could call it here like this:

```
$this->setAuthenticationSchemeAlias('googlesignin');
```
but I think we should just remove this whole change to GoogleSignIn.

plugins/oauth2/class.oauth2.plugin.php
```diff
@@ -17,5 +17,6 @@ class OAuth2Plugin extends Gdn_OAuth2 {
     public function __construct() {
         $this->setProviderKey('oauth2');
         $this->settingsView = 'plugins/settings/oauth2';
+        $this->authenticationScheme = 'oauth2';
```
nervoustwit, line 20: Let's actually put the call to the setter here because, in the not too distant future we will want to separate the provider key and the authentication scheme alias for this plugin.

nervoustwit:
GDN_Oauth2 class requires two things:

 1. a setAuthenticationSchemeAlias() method that other child classes can call to set the Scheme.
 2. to explicitly add the setAuthenticationSchemeAlias to the set up form.

plugins/Facebook/class.facebook.plugin.php
```diff
@@ -666,7 +677,12 @@ public function structure() {
...
         Gdn::sql()->replace(
             'UserAuthenticationProvider',
-            ['AuthenticationSchemeAlias' => 'facebook', 'URL' => '...', 'AssociationSecret' => '...', 'AssociationHashMethod' => '...'],
+            [
+                'AuthenticationSchemeAlias' => self::AUTHENTICATION_SCHEME,
+                'URL' => '...',
+                'AssociationSecret' => '...',
+                'AssociationHashMethod' => '...'
+            ],
             ['AuthenticationKey' => self::PROVIDER_KEY],
```
nervoustwit, line 686: Excellent. Leave the code better then the way you found it.

PR #10426, opened 2020-04-15
Fix double encoding in DiscussionsModule titles

initvector:
#8980 altered the way discussion titles are rendered in some modules. In the case of `DiscussionModule`, it ran titles through `htmlspecialchars`. Since those titles were already encoded, they are ultimately rendered as double-encoded on the site. This leads to seeing things like &amp;gt; instead of > or &amp;amp instead of & in discussion titles displayed by the module.

Title rendering has fallen back to displaying these titles using the practical replacement of the original `Gdn_Format::text`. No unnecessary HTML encoding is performed.

Closes vanilla/support#1727

### DIFF
--- a/applications/vanilla/views/modules/helper_functions.php
+++ b/applications/vanilla/views/modules/helper_functions.php
@@ -16,8 +16,12 @@ if (!function_exists('WriteModuleDiscussion')):
    </span>
 
             <div class="Title"><?php
-                echo anchor(htmlspecialchars($discussion->Name), discussionUrl($discussion).($discussion->CountCommentWatch > 0 ? '#Item_'.$discussion->CountCommentWatch : ''), 'DiscussionLink');
-                ?></div>
+                echo anchor(
+                    Gdn::formatService()->renderHTML($discussion->Name, \Vanilla\Formatting\Formats\TextFormat::FORMAT_KEY),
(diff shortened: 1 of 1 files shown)


applications/vanilla/views/modules/helper_functions.php
```diff
@@ -16,8 +16,12 @@ function writeModuleDiscussion($discussion, $px = 'Bookmark', $showPhotos = fals
    </span>
 
             <div class="Title"><?php
-                echo anchor(htmlspecialchars($discussion->Name), discussionUrl($discussion).($discussion->CountCommentWatch > 0 ? '#Item_'.$discussion->CountCommentWatch : ''), 'DiscussionLink');
-                ?></div>
+                echo anchor(
+                    Gdn::formatService()->renderHTML($discussion->Name, \Vanilla\Formatting\Formats\TextFormat::FORMAT_KEY),
```
tburry, line 20: I'm pretty sure this is not supposed to be encoded at all. DiscussionModel::calculate() calls htmlspecialchars().

initvector: Dropped running things through the format-to-text in favor of applying HTML sanitization via `HtmlSanitizer` (and ultimately htmLawed).

PR #10465, opened 2020-04-20
Check if forcedFormat is set when Force.InputFormatter is true

cchabilall83:
Allows addons to adjust the format if the `force.inputformatter` is true

required for https://github.com/vanilla/internal/pull/2422

relates to https://github.com/vanilla/support/issues/999


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -8,6 +8,7 @@
  * @since 2.0
  */
 
+use Garden\EventManager;
```
charrondev, line 11:
```suggestion
```

PR #10466, opened 2020-04-20
Sort APIv2 comments by dateInserted rather than commentID

### LINKED ISSUE #9673: Comments should be sorted by dateInserted on /api/v2/comments?discussionID=???
When calling `/api/v2/comments` we should sort the comments by `dateInserted`. It appears as though they are either sorted by `commentID` or just not sorted.

blutarch:
Closes #9673 

The API sorts comments by commentID, but we want it to sort them by dateInserted instead. This PR does that by adding a $sort param in the `commentmodel.php` lookup function that can takes the name of the column to sort by. The `CommentsApiController` index function then specifies that comments should be sorted according to the DateInserted column.

### TO TEST
1. Check out this branch. 
1. Find a discussion with multiple comments and change the dates of that discussion's comments in the `GDN_Comment table` so that the date order doesn't match the commentID order.
1. Make the api call to get the comments (`https://dev.vanilla.localhost/api/V2/comments?discussionID={discussionID}`).
1. Verify that the comments are ordered by date rather than commentID.

### DIFF
--- a/applications/vanilla/models/class.commentmodel.php
+++ b/applications/vanilla/models/class.commentmodel.php
@@ -504,12 +504,15 @@ class CommentModel extends Gdn_Model implements FormatFieldInterface {
             }
         }
 
+        $this->orderBy('c.'.$sort);
+        $orderBy = $this->orderBy();
+
         $query = $this->SQL
             ->select('c.*')
             ->select('d.CategoryID')
             ->from('Comment c')
             ->join('Discussion d', 'c.DiscussionID = d.DiscussionID')
-            ->orderBy('c.'.$sort, $order);
+            ->orderBy($orderBy[0][0], $order);
         if (!empty($where)) {
             $query->where($where);
         }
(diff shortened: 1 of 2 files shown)


applications/vanilla/models/class.commentmodel.php
```diff
@@ -482,9 +482,10 @@ public function getDefaultLimit() {
      * @param int|null $limit Max number to get.
      * @param int $offset Number to skip.
      * @param string $order Order comments ascending (asc) or descending (desc) by ID.
+     * @param string $sort The column to sort by.
      * @return Gdn_DataSet SQL results.
      */
-    public function lookup(array $where = [], $permissionFilter = true, $limit = null, $offset = 0, $order = 'desc') {
+    public function lookup(array $where = [], $permissionFilter = true, $limit = null, $offset = 0, $order = 'desc', string $sort = 'CommentID') {
```
igraziatto, line 488: Shouldn't the 'CommentID' be part of the `protected $_OrderBy` property?

blutarch: I believe the `protected $OrderBy` property is getting set in this method with the `$sort` and `$order` arguments on line 512, but perhaps I'm missing something?

blutarch: I think my last commit incorporates this correctly now.

PR #10471, opened 2020-04-20
Add Panopto video embeds support.

patches:
Estimate request: https://github.com/vanillaforums/estimates/issues/932
Issue: https://github.com/vanillaforums/panopto/issues/2

Add embed support for Panopto videos. You can find videos to test with [right here](https://howtovideos.hosted.panopto.com/Panopto/Pages/Folders/DepartmentHome.aspx?folderID=4b9de7ae-0080-4158-8496-a9ba01692c2e)

You can find the documentation I referred to in [a comment](https://github.com/vanillaforums/panopto/issues/2#issuecomment-599748154) I left in the github issue living in the customer's repo.

I looked at existing code we had for the react part, but it might not be the most efficient way to do it, I am open to any recommendations.

### DIFF
--- a/library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php
+++ b/library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php
@@ -15,12 +15,6 @@ use Vanilla\EmbeddedContent\Embeds\PanoptoEmbed;
  */
 class PanoptoEmbedFactory extends AbstractEmbedFactory {
 
-    /** @var int DEFAULT_HEIGHT */
-    private const DEFAULT_HEIGHT = 360;
-
-    /** @var int DEFAULT_WIDTH */
-    private const DEFAULT_WIDTH = 640;
-
     /** @var array DOMAINS */
     private const DOMAINS = [
         'hosted.panopto.com',
@@ -41,8 +35,6 @@ class PanoptoEmbedFactory extends AbstractEmbedFactory {
             'embedType' => PanoptoEmbed::TYPE,
             'domain' => $path,
             'url' => $url,
-            'height' => self::DEFAULT_HEIGHT,
-            'width' => self::DEFAULT_WIDTH,
             'sessionId' => $this->sessionIDFromUrl($url),
         ]);
     }
--- a/library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php
+++ b/library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace Vanilla\EmbeddedContent\Factories;
+
+use Vanilla\EmbeddedContent\AbstractEmbed;
+use Vanilla\EmbeddedContent\AbstractEmbedFactory;
+use Vanilla\EmbeddedContent\Embeds\PanoptoEmbed;
+
+/**
+ * PanoptoEmbedFactory class.
+ */
+class PanoptoEmbedFactory extends AbstractEmbedFactory {
+
+    /** @var int DEFAULT_HEIGHT */
+    private const DEFAULT_HEIGHT = 360;
+
+    /** @var int DEFAULT_WIDTH */
+    private const DEFAULT_WIDTH = 640;
+
+    /** @var array DOMAINS */
+    private const DOMAINS = [
+        'hosted.panopto.com',
+        'ca.panopto.com',
+        'cloud.panopto.eu',
+        'ap.panopto.com',
+    ];
+
+    /**
+     * Use the page scraper to scrape page data.
+     *
+     * @inheritdoc
+     */
+    public function createEmbedForUrl(string $url): AbstractEmbed {
+        $path = parse_url($url, PHP_URL_HOST);
+
+        return new PanoptoEmbed([
+            'embedType' => PanoptoEmbed::TYPE,
+            'domain' => $path,
+            'url' => $url,
+            'height' => self::DEFAULT_HEIGHT,
+            'width' => self::DEFAULT_WIDTH,
+            'sessionId' => $this->sessionIDFromUrl($url),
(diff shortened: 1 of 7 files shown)


library/src/scripts/embeddedContent/PanoptoEmbed.tsx
```diff
@@ -17,8 +17,12 @@ interface IProps extends IBaseEmbedProps {
     height: number;
 }
 
+/** @constant {string} */
 const PANOPTO_SCRIPT = "https://developers.panopto.com/scripts/embedapi.min.js";
 
+/** @constant {string} */
+const EMBED_LOADED_CLASS = "isLoaded";
+
```
charrondev: This is a bit odd. We tend not to use JSDoc for our types. Better to enforce with typescript if you want to manually impose some types.

```suggestion
const PANOPTO_SCRIPT: string = "https://developers.panopto.com/scripts/embedapi.min.js";

const EMBED_LOADED_CLASS: string = "isLoaded";

```

tests/Library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactoryTest.php
```diff
@@ -0,0 +1,87 @@
...
+            ['https://demo.ca.panopto.com/Panopto/Pages/Viewer.aspx?id=7604d3f5-c51d-4053-8c59-ab4400e0bbe6'],
+            ['https://demo.cloud.panopto.eu/Panopto/Pages/Viewer.aspx?id=2eb0bf71-e051-4386-8aa3-0c7cf8a28135'],
+            ['https://demo.ap.panopto.com/Panopto/Pages/Viewer.aspx?id=25329997-baa2-42cf-89c4-ab4400e080e3'],
+        ];
+    }
+
+    /**
+     * Test the Panopto Embed instantiation.
+     */
+    public function testCreateEmbedForUrl() {
```
charrondev, line 57: Structurally it's preferred to use a PHPUnit provider for this type of thing.

https://github.com/vanilla/vanilla/blob/0359e395815443b113a7f88a4b9ba8f02a5d8e25/tests/Library/Vanilla/EmbeddedContent/Factories/ScrapeEmbedFactoryTest.php#L41-L53

library/src/scripts/embeddedContent/PanoptoEmbed.tsx
patches: Oh ok I didn't know that, I'll make the change.

tests/Library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactoryTest.php
charrondev: On second thought, you're already using a data provider in your earlier test. Couldn't you wire this one up the same way?

patches: Yes I can do that.

patches:
@alex-mtl Forgot PRs must be reviewed by 2 devs in core. Can you also have a look at this PR when you have the chance? Thanks

library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * PanoptoEmbedFactory class.
+ */
+class PanoptoEmbedFactory extends AbstractEmbedFactory {
+
+    /** @var int DEFAULT_HEIGHT */
+    private const DEFAULT_HEIGHT = 360;
+
+    /** @var int DEFAULT_WIDTH */
+    private const DEFAULT_WIDTH = 640;
```
alex-mtl, line 22: I am not sure if the embed URL returns frame ratio.
But in our other embed factory implementations we have it set up that way:
```
            'height' => $response['height'] ?? null,
            'width' => $response['width'] ?? null,

File: library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function createEmbedForUrl(string $url): AbstractEmbed {
+        $path = parse_url($url, PHP_URL_HOST);
+
+        return new PanoptoEmbed([
+            'embedType' => PanoptoEmbed::TYPE,
+            'domain' => $path,
+            'url' => $url,
+            'height' => self::DEFAULT_HEIGHT,
+            'width' => self::DEFAULT_WIDTH,
```
alex-mtl, line 45: I am not sure if the embed URL returns frame ratio.
But in our other embed factory implementations we have it set up that way:
```
            'height' => $response['height'] ?? null,
            'width' => $response['width'] ?? null,

patches: I saw in other embed factories that we ofter send requests to some API to retrieve information around the embed. However, in this particular case, all the information we need to build the embed is part of the URL, but not these two values.

The embed is generated in Javascript and the height and width can be passed when instantiating an EmbedApi object.

I created these two constants forcing the height and width to respect the ratio of the videos.

Does that make sense?

patches:
@alex-mtl I added the script the the whitelist. I used the constant from the PanoptoEmbed class, however other entries in this array were hardcoded. I understand if we would prefer to directly put the string instead to respect the already establish "standard".

https://github.com/vanilla/vanilla/blob/db7b16026a92fa0ccec01dfa1e95f37f90467d2d/library/Vanilla/Web/ContentSecurityPolicy/EmbedWhitelistContentSecurityPolicyProvider.php#L22

library/Vanilla/EmbeddedContent/Factories/PanoptoEmbedFactory.php
alex-mtl: Partially, what will happen if we are in mobile mode?

patches: Good catch, I removed the width and height constants declared in the factory and added the proper classes on the embed itself.

PR #10474, opened 2020-04-21
Add prompt for user navigating away form theme editor

cchabilall83:
WIP


packages/vanilla-react-utils/src/useRouteChangePrompt.tsx
```diff
@@ -0,0 +1,28 @@
+/**
+ * @copyright 2009-2020 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+import { useCallback, useEffect } from "react";
+
+export function useRouteChangePrompt(message: string, disabled: boolean = false) {
```
charrondev, line 7: This should get a doc block of some kind.

```suggestion
/**
 * Hook for triggering a a browser confirmation prompt if the user tries to navigate away or close the page.
 *
 * It is only possible to do this using nature browser UI, which is only a "message" can be provided and not a react component.
 */
export function useRouteChangePrompt(message: string, disabled: boolean = false) {
```

Issue #9696, opened 2019-11-26
React Based "New Discussion" buttons


slafleche:
## Storybook

Create react component version of new discussion button.

- Some variations should be added to storybook.
- The component should accept action items. These items can be a link (with a url) or a button (with an action). These should appear visually the same.
- Items accept an icon and a label.
- There is some transition as the menu opens.
- The main menu button stays fixed to the bottom of the screen.

## Mockups

- https://zpl.io/bLQ6qM7 
- https://zpl.io/brQ9A4r

Where looking to implement some motion design similar to this https://dribbble.com/shots/3184306-FAB-Motion

PR #10481, opened 2020-04-21
Add an icon to icons/common

tuan-ng:
I needed a few icons for issue [#9696](https://github.com/vanilla/vanilla/issues/9696). Just added a story for an icon to get going.


library/src/scripts/icons/common.tsx
```diff
@@ -888,3 +888,22 @@ export function BookmarkIcon(props: { className?: string; "aria-hidden"?: areaHi
         </svg>
     );
 }
+
+export function NewActionIcon(props: { className?: string; "aria-hidden"?: areaHiddenType }) {
+    const classes = iconClasses();
+    return (
+        <svg
+            className={classNames(classes.newAction)}
```
slafleche, line 896: Too generic. What is the icon for?

charrondev:
A couple things I'm noting from a visual design point of view:

- The link color seems to really pop out. I'd recommend using our `clickableItemStates()` method to generate your styles for the links.
- If you look at the mockups there is a background overlay. The make something similar for our `<ModalView />` component and it looks like this.

```
            <EntranceAnimation
                fade
                isEntered={props.isVisible}
                className={classes.overlayScrim}
                onDestroyed={handleDestroy}
            >
                <span></span>
            </EntranceAnimation>
```

It might be worth extracting that into it's own component.

PR #10490, opened 2020-04-22
Various banner title changes

cchabilall83:
- Modifies the bannerStyles.title.text to take the siteTitle and fallback to default text.
- Modifies bannerModel to only pass title prop if there is a category title.
- Modifies banner to use bannerStyles.title.text if no title prop is passed.
- Add allowEmpty on ThemeBuilderContext to allow for empty strings to be saved passed from ThemeInputText

required for https://github.com/vanilla/internal/pull/2436
related issue https://github.com/vanilla/knowledge/issues/1752

### DIFF
--- a/library/src/scripts/banner/bannerStyles.ts
+++ b/library/src/scripts/banner/bannerStyles.ts
@@ -43,6 +43,7 @@ import { NestedCSSProperties, TLength } from "typestyle/lib/types";
 import { titleBarVariables } from "@library/headers/titleBarStyles";
 import { breakpointVariables } from "@library/styles/styleHelpersBreakpoints";
 import { t } from "@vanilla/i18n";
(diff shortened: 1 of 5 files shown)


library/src/scripts/banner/bannerStyles.ts
```diff
@@ -254,7 +255,7 @@ export const bannerVariables = useThemeCache((forcedVars?: IThemeVariables, altN
             top: 18,
             bottom: 8,
         },
-        text: "How can we help you?",
+        text: getMeta("ui.siteName", t("How can we help you?")),
```
slafleche, line 258: So how's this going to work for multilingual? Has this been discussed?

cchabilall83: Multi lingual will be handled in a different sprint.

Issue #10499, opened 2020-04-23
Foundation: Category descriptions do not display on category list


Bparm:
I am not sure if this is even a bug but I noticed it and decided to bring it up.

On Keystone category descriptions display on the list of categories:
https://forum.vertica.com/categories

On Foundation we do not:
https://brendan.vanillademo.com/product-one

Perhaps this was a design choice but perhaps someone can comment. I do not know if this has any SEO impact given that you can give a category description and it displays in the banner image so does it matter if it's visible or not? I do think some people will miss those displaying in the category list however.

PR #10501, opened 2020-04-23
Fix a bug preventing descriptions from appearing in foundation category list

charrondev:
Fixed https://github.com/vanilla/vanilla/issues/10499

`$row` was actually the incorrect variable. Looks like it was a copy/paste mistake. +10 points to adding some static checking into our PHP process.


applications/vanilla/views/categories/helper_functions.php
```diff
@@ -169,7 +169,7 @@ function writeListItem($category, $depth) {
                 Gdn::controller()->EventArguments['Category'] = &$category;
                 Gdn::controller()->fireEvent('BeforeCategoryItem');
                 $headingClass = "CategoryNameHeading";
-                if (empty($row['Description'])) {
+                if (empty($category['Description'])) {
```
alex-mtl, line 172: This is the only time I would vote for `val()` :-)
```suggestion
                if (empty(val('Description', $category))) {
```

PR #10503, opened 2020-04-23
Fix 'Mark Read' displaying in dropdown when All Viewed plugin is disabled

blutarch:
Closes vanilla/support#1796

This fixes a problem where the 'Mark Read' option displays in the category dropdown even when the 'All Viewed' plugin is disabled by adding a check to see if the plugin is enabled or not.

### TO TEST
Before checking out branch:
1. Disable the "All Viewed" plugin.
1. Confirm that the cog wheel dropdown for a category displays the "Mark Read" option.
After checking out branch:
1. Confirm that the "Mark Read" option is now gone from the category cog wheel dropdown.


blutarch:
Turns out we want to keep the category marked read functionality separate from the "All Viewed" plugin.

Issue #10363, opened 2020-04-03
Unable to set token inputs for automated tests


slafleche:
The input we use for the token input doesn't use the DOM to store data, and so there's no way to set it correctly with the automated script.

I suggest we update the component to store it's current values in a hidden input (like we do for the dropdown) and trigger an update on the component if the hidden input changes.

For testing purposes I think we should also add a mandatory unique class on these inputs so We can target them appropriately. Our IDs and classes are generated, which is fine, we don't want to break the tests if the generated CSS classes or IDs change. We could just force a unique name for the component and it will be a reliable target for QA.

PR #10504, opened 2020-04-24
Token Improvements for testing

slafleche:
This PR is to make it possible for automated tests on token inputs. Since the values were all in memory, there was no easy way to access them. What I've done is added a hidden input that reflects the data stored in React. It will also listen to changes, so it can be set manually with something like: `document.querySelectorAll(".js-tokens-tokenInput")[0].value = '[{"value":14,"label":"AAAA"},{"value":15,"label":"BBBB"},{"value":6,"label":"CCCC"}]';`

I've also added a class to help with targeting the hidden input with JS. Note that it does not guarantee a unique class, unlike the actual ID of the component.

Closes: https://github.com/vanilla/vanilla/issues/10363

### DIFF
--- a/library/src/scripts/forms/select/Tokens.tsx
+++ b/library/src/scripts/forms/select/Tokens.tsx
@@ -4,7 +4,7 @@
  * @license GPL-2.0-only
  */
 
-import React from "react";
+import React, { useRef } from "react";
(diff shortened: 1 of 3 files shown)


slafleche:
@GeetikaMuthyam see instructions up top on how to test. Poke me if you have more questions.

PR #10509, opened 2020-04-24
Fix sphinx trait 007

alex-mtl:


alex-mtl:
This change is not neede

PR #10515, opened 2020-04-27
Various Designer Feedback Fixes

slafleche:
- Fixed size of title on KB search page to be bigger than the panel's title
- Fixed the styles of the token input, the spacing was ugly and strange when the inner input text wrapped to a new line. It looks a lot more consistent now
- Improvements to forum's advanced search input. The arrow was positioned properly and the padding was adjusted as to not have text going under the arrow.

Companion PR: https://github.com/vanilla/knowledge/pull/1796


GeetikaMuthyam:
As this is a PR, it doesn't need "Needs QA" label

PR #10522, opened 2020-04-28
Fix drafts CategoryDropdown when Subcomunities is enabled

igraziatto:
Closes: https://github.com/vanilla/support/issues/1775

Relevant info: 
> The issue happens because drafts are not aware of subcommunities, when you access /drafts it returns all drafts made by that user, independent of which subcommunity you are.
When enabled, Subcommunities hook into the categoryDropdown and change the structure to fit the current subcommunity. It also shift the draft category ID to the subcommunity category, since this category is usually type "nested" it doesn't select that option in the categoryDropdown (because those options are disabled).
> 
> Since Drafts are supposed to be a shared content the Subcommunities is not supposed to intervene on the categoryDropdown when editing a draft.

**Reference:** https://github.com/vanilla/support/issues/1775#issuecomment-620731068

**Steps to reproduce:**
- Make sure you have subcommunities enabled and setup
- Make a discussion draft, make sure to select a category
- Navigate to `/drafts` and open your draft, you will note the categoryDropdown does not have the category selected.
- Checkout this branch and repeat the last step above.

**Notes:**
I took the liberty os changing the parameter name in `PostController->discussion()` from `$CategoryUrlCode` to `$categoryID` since it was misleading.

**Related PRs:**
https://github.com/vanilla/multisite/pull/364

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -111,11 +111,11 @@ class PostController extends VanillaController {
      *
      * @param int $categoryID Unique ID of the category to add the discussion to.
      */
-    public function discussion($categoryUrlCode = '') {
+    public function discussion($categoryID = '') {
         // Override CategoryID if categories are disabled
         $useCategories = $this->ShowCategorySelector = (bool)c('Vanilla.Categories.Use');
         if (!$useCategories) {
-            $categoryUrlCode = '';
+            $categoryID = '';
         }
 
         // Setup head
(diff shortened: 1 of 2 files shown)


dannimajdalani:
I don't think I understood the issue you described.
Is it

1) when we save a draft "with a selected category", then when we go again to that saved draft, the selected category is no longer there?
Or by
> Navigate to /drafts and open your draft, you will note the categoryDropdown is not set
2) you mean the entire dropdown is not set?

Looking at the referenced ticket, it looks more likely like 1.

Here is what I've done:

I created a draft, selected a category from the dropdown (no subcommunity)
Open the draft, the category is there

I create a draft, selected a category from the dropdown (in a subcommunity)
Opened the draft, the category is not selected.

I've checked out your branch,
Created a new draft, selected a category and saved
Opened the draft, the category is not selected.

You've mentioned that w'ere updating the categorydropdown when subcommunities are enabled, with the code you've added, how does that change?

>  if (property_exists($this, 'Draft') && is_object($this->Draft)) {
>             $options['DraftID'] = $this->Draft->DraftID;
>         }

its still getting called and modified here https://github.com/vanilla/vanilla/blob/5fec7a588f904773b81e9fbd5e214dcab340b060/library/core/class.form.php#L474

PR #10524, opened 2020-04-28
Fix permission check in categories that inherit permissions

blutarch:
Closes vanilla/support#962

This PR fixes a problem in the DiscussionsAPIController class where the categoryID is being used in the patch() method's fieldPermission calls. This creates a permission error when trying to patch a discussion in a nested category that inherits custom permissions from its parent by replacing the categoryID with the PermissionsCategoryID.

### TO TEST
Make a category with custom permissions that allows a moderator to sink the discussion. Then add a subcategory that doesn't have custom permissions. Add a discussion to that subcategory and try to sink that discussion through postman as a moderator.

### DIFF
--- a/applications/vanilla/controllers/api/DiscussionsApiController.php
+++ b/applications/vanilla/controllers/api/DiscussionsApiController.php
@@ -576,7 +576,11 @@ class DiscussionsApiController extends AbstractApiController {
         }
 
         $category = CategoryModel::categories($categoryID);
-        $permissionCategoryID = $category->PermissionCategoryID;
+        if ($category) {
+            $permissionCategoryID = $category['PermissionCategoryID'];
+        } else {
+            $permissionCategoryID = -1;
+        }
 
         $this->fieldPermission($body, 'closed', 'Vanilla.Discussions.Close', $permissionCategoryID);
         $this->fieldPermission($body, 'pinned', 'Vanilla.Discussions.Announce', $permissionCategoryID);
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -567,6 +567,8 @@ public function patch($id, array $body) {
         $discussionData = ApiUtils::convertInputKeys($body);
         $discussionData['DiscussionID'] = $id;
         $categoryID = $row['CategoryID'];
+        $category = $this->categoryModel->getID($categoryID);
```
dannimajdalani, line 570: We can use CategoryModel::categories($categoryID),
https://github.com/vanilla/vanilla/blob/ad5b3d07138fd32f0820d02cb7d0eaa983f8035e/applications/vanilla/models/class.categorymodel.php#L518
It uses cache around here 

https://github.com/vanilla/vanilla/blob/995750f30a8c77c4d01e85639b08aa4ceaed4971/applications/vanilla/library/class.categorycollection.php#L177

applications/vanilla/controllers/api/DiscussionsApiController.php
```diff
@@ -575,9 +575,12 @@ public function patch($id, array $body) {
             $categoryID = $discussionData['CategoryID'];
         }
 
-        $this->fieldPermission($body, 'closed', 'Vanilla.Discussions.Close', $categoryID);
-        $this->fieldPermission($body, 'pinned', 'Vanilla.Discussions.Announce', $categoryID);
-        $this->fieldPermission($body, 'sink', 'Vanilla.Discussions.Sink', $categoryID);
+        $category = CategoryModel::categories($categoryID);
+        $permissionCategoryID = $category->PermissionCategoryID;
```
dannimajdalani, line 579: isn't category an array? you're accessing it as an object.

also you want to check if you have a category before doing
 $permissionCategoryID = $category->PermissionCategoryID;
something like 

 if ($category) {
                $permissionCategoryID = ...
            } else {
                $permissionCategoryID = -1;
            }

PR #10526, opened 2020-04-29
Fix deleting discussion timeouts

dannimajdalani:
closes https://github.com/vanilla/support/issues/1709
closes https://github.com/vanilla/support/issues/1814
closes https://github.com/vanilla/support/issues/1804

### Details

When a category has hundreds of thousands of discussions, deleting a discussion is causing timeouts.

running select count(DiscussionID) on a table with 150+ records will take >30 seconds to complete.
Everytime we delete a discussion, w'ere calling https://github.com/vanilla/vanilla/blob/9d44b56e402ddc159012ece449ab307cd3bc71e9/applications/vanilla/models/class.discussionmodel.php#L2638
that updates the discussion count on a category.
updateDiscussionCount is taking more than 30 seconds(timeout) to complete.

I've created a job to update the count.

### To Test

On a category with >150K discussions, delete a discussion

### DIFF
--- a/applications/vanilla/library/Jobs/UpdateDiscussionCount.php
+++ b/applications/vanilla/library/Jobs/UpdateDiscussionCount.php
@@ -43,7 +43,7 @@ class UpdateDiscussionCount implements LocalJobInterface {
         $schema = Schema::parse([
             "categoryID" => ["type" => "integer"],
             "discussion:a|n?"
-            ]);
+        ]);
         return $schema;
     }
 
--- a/applications/vanilla/library/Jobs/UpdateDiscussionCount.php
+++ b/applications/vanilla/library/Jobs/UpdateDiscussionCount.php
@@ -12,12 +12,12 @@ use Vanilla\Scheduler\Job\JobPriority;
 use Vanilla\Scheduler\Job\LocalJobInterface;
 
 /**
- * Update category discussion count.
+ * Update category discussion and comment count.
  */
 class UpdateDiscussionCount implements LocalJobInterface {
 
-    /** @var \DiscussionModel */
-    private $discussionModel;
+    /** @var \CategoryModel */
+    private $categoryModel;
 
     /** @var int */
     private $categoryID;
@@ -28,10 +28,10 @@ class UpdateDiscussionCount implements LocalJobInterface {
     /**
      * Initial job setup.
      *
-     * @param \DiscussionModel $discussionModel
+     * @param \CategoryModel $categoryModel
      */
-    public function __construct(\DiscussionModel $discussionModel) {
-        $this->discussionModel = $discussionModel;
+    public function __construct(\CategoryModel $categoryModel) {
+        $this->categoryModel = $categoryModel;
     }
 
     /**
@@ -43,7 +43,8 @@ class UpdateDiscussionCount implements LocalJobInterface {
         $schema = Schema::parse([
             "categoryID" => ["type" => "integer"],
             "discussion:a|b"
-            ]);
+            ]
+        );
         return $schema;
     }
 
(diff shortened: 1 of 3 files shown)


applications/vanilla/models/class.discussionmodel.php
```diff
@@ -3066,7 +3066,9 @@ public function deleteID($discussionID, $options = []) {
         $this->SQL->delete('Discussion', ['DiscussionID' => $discussionID]);
 
         $this->SQL->delete('UserDiscussion', ['DiscussionID' => $discussionID]);
-        $this->updateDiscussionCount($categoryID);
+        /** @var Vanilla\Scheduler\SchedulerInterface $scheduler */
+        $scheduler = Gdn::getContainer()->get(Vanilla\Scheduler\SchedulerInterface::class);
+        $scheduler->addJob(Vanilla\Library\Jobs\UpdateDiscussionCount::class, ['categoryID' => $categoryID]);
```
charrondev, line 3071: Shouldn't this go inside of the existing `updateDiscussionCount` method?

Currently there may still be places that are updating it sychronously.

dannimajdalani: Makes sense. There are multiple places where we call this method. I'll update it.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2634,7 +2633,21 @@ public function updateDiscussionCount($categoryID, $discussion = false) {
...
+        }
+    }
+
+    /**
+     * Scheduled category discussion count update.
+     *
+     * @param int $categoryID Unique ID of category we are updating.
+     * @param array|false $discussion The discussion to update the count for or **false** for all of them.
+     */
+    public function scheduledUpdateCount($categoryID, $discussion = false) {
```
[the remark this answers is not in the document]
initvector: The method is `scheduledUpdateCount`, but I wouldn't think this had anything to do with a scheduled job based on its content. It just looks like a method I'd use to update a category's discussion and comment fields.

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2634,7 +2633,21 @@ public function updateDiscussionCount($categoryID, $discussion = false) {
...
+        $discussionID = $discussion['DiscussionID'] ?: false;
+        $this->SQL
                 ->select('d.DiscussionID', 'count', 'CountDiscussions')
                 ->select('d.CountComments', 'sum', 'CountComments')
```
initvector, line 2652: The indenting in this block seems a little off.

applications/vanilla/models/class.discussionmodel.php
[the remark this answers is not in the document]
initvector: I normally wouldn't nitpick, but since this is new: it feels like a method for updating category fields should be in the category model.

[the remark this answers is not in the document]
initvector: When refactoring old code like this into new methods, it's a good chance to improve on them.

1. We probably wouldn't allow `false` as a value for an array parameter, anymore. If it's truly optional, we can make it nullable.
1. We can use types for parameters. `$discussion` looks like it already assumes it'll be an array, if it's anything. `$categoryID` will only work as an integer.

It's probably a good idea to alter usage at the call site (or, better still, when you're scheduling the job) to account for the signature changes in the method. Since it's currently only really used in one place, it should be easy.

applications/vanilla/library/Jobs/UpdateDiscussionCount.php
```diff
@@ -0,0 +1,91 @@
...
+            ]);
+        return $schema;
+    }
+
+    /**
+     * Update category discussion count.
+     */
+    public function run(): JobExecutionStatus {
+        $isValidDiscussion = (is_array($this->discussion) || is_bool($this->discussion));
+        $isValidCategory = is_int($this->categoryID);
```
initvector, line 55: Shouldn't this already be handled by the schema validation we did in `setMessage`?

applications/vanilla/models/class.discussionmodel.php
```diff
@@ -2634,32 +2633,10 @@ public function updateDiscussionCount($categoryID, $discussion = false) {
...
-                    'LastCommentID' => null,
-                    'LastDateInserted' => val('DateInserted', $discussion)
-                ]);
-            }
-
-            $categoryModel = new CategoryModel();
-            $categoryModel->setField($categoryID, $cacheAmendment);
-            $categoryModel->setRecentPost($categoryID);
+            /** @var Vanilla\Scheduler\SchedulerInterface $scheduler */
+            $discussion = $discussion ?: null;
```
initvector, line 2637: The function is documented as only working with an array, but it looks like it would've also worked with an object. To be safe, you'll probably want to ensure it's an array here (casting objects to an array).

PR #10529, opened 2020-04-29
Add option to delete moderation notes

blutarch:
Along with vanilla/internal#2485, this closes vanilla/support#1455.

This PR adds a checkbox option when deleting a user's content to choose whether to delete any moderation notes pertaining to the user's posts.

### TO TEST
1. Create 2 users and add a moderation note for each one.
1. Check out both this branch and the `add/delete-notes-check` branch from vanilla/internal.
1. From `dashboard/user`, select the first user to delete using the trash can icon.
1. Choose 'Delete User Content'.
1. Leave the 'Delete moderation information' box unchecked and click 'Delete User Forever'.
1. Verify that the user is deleted, but the moderation note is still present (in the `Gdn_UserNote` table).
1. Repeat steps 2-4 with the second user.
1. This time, check the checkbox.
1. Verify that the moderation note is deleted.

### DIFF
--- a/applications/dashboard/controllers/class.usercontroller.php
+++ b/applications/dashboard/controllers/class.usercontroller.php
@@ -509,9 +509,6 @@ class UserController extends DashboardController {
     public function delete($userID = '', $method = '') {
         $this->permission('Garden.Users.Delete');
         $session = Gdn::session();
-
-        $formValues = $this->Form->formValues();
-
         if ($session->User->UserID == $userID) {
             trigger_error(errorMessage("You cannot delete the user you are logged in as.", $this->ClassName, 'FetchViewLocation'), E_USER_ERROR);
         }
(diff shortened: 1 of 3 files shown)


applications/dashboard/controllers/class.usercontroller.php
```diff
@@ -550,7 +553,7 @@ public function delete($userID = '', $method = '') {
             }
 
             if ($this->Form->authenticatedPostBack(true) && $method != '') {
-                $userModel->deleteID($userID, ['DeleteMethod' => $method]);
+                $userModel->deleteID($userID, ['DeleteMethod' => $method, 'DeleteNotes' => $formValues['DeleteNotes']]);
```
tburry: I think you can just do `$this->Form->getFormValue('DeleteNotes')`.

applications/dashboard/views/user/deleteconfirm.php
```diff
@@ -15,7 +15,8 @@
     else if ($this->Method == 'wipe')
         echo t("All of the user content will be replaced with a message stating the user has been deleted.");
     else
-        echo t("The user content will be completely deleted.");
+        echo t("The user content will be completely deleted."),
+        $this->Form->checkBox('DeleteNotes', t('Delete moderation notes as well'), ['value' => '1']);
```
tburry: Can we do `DeleteModerationInfo` as the field and make it "Delete moderation information." for the label.

PR #10533, opened 2020-04-30
Improve theme scaffolding

charrondev:
Closes https://github.com/vanilla/internal/issues/2483

## Scaffold Command

Our scaffold command went out of date. To remedy this, I've removed the separate scaffold directory. Not our themes scaffold directly based on foundation.

## JSON5 variables.

- Theme variables can now be json5. This allowed me to fill up foundations variables with a lot of comments.

## Asset changes

- I added comments about the most common and important variables in foundation. These are all commented so that the real defaults can continue to live in code (and be generated).
- Added an empty header asset.
- Added comments into the header and footer and style assets.
- Removed the complex extra build from foundations footer styles. It was complicated things unnecessarily.


addons/themes/theme-foundation/assets/footer.twig
```diff
@@ -1,3 +1,10 @@
+{#
+    This is an example footer.
+    You will likely want to modify to match you brand.
+    See https://success.vanillaforums.com/kb/articles/171-a-themes-structure#footer
+
+    Styles from the styles.css asset will apply to this header.
+ #}
 <!-- Main footer END -->
```
slafleche, line 8: Can you just fix this comment, the start is set to "END"

PR #10534, opened 2020-04-30
Fixed sticky panel calculation

slafleche:
The ref for the banner was undefined when `options.overlayTitleBar` was false, causing the calculation of the banner height to be 0.

Closes: https://github.com/vanilla/knowledge/issues/1808


slafleche:
https://github.com/vanilla/vanilla/pull/10534/files#diff-cb370c9f2246a026ec295ac7a800c7ddR59 ?

Issue #10536, opened 2020-05-01
Posting settings set to Text for dekstop/mobile on newly created site


Bparm:
Whenever a new site is created the posting settings in the dashboard appear set to 'text" for both the mobile and desktop inputs despite the fact that the Rich Editor appears to be in use when creating a post.

Reproduce:
-create a new site
-visit posting settings page
-view inputs set to text.

PR #10537, opened 2020-05-01
Fix rich post format getting lost

charrondev:
Closes https://github.com/vanilla/vanilla/issues/10536

- At some point we changed the rich format that gets applied from the branding page to `'rich`
- Config defaults (local and on infra) are 'Rich'.
- Value was historically 'Rich'

The casening was done in most places to some lowercase constant, and a lot of checks are case-insensitive, however getting the config values to be lowercase would be a lot riskier. As a result I'm reverting the config values to the old value for rich, and added something in structure to ensure it gets fixed for existing sites that may have set themselves to the lowercase value in the interim.

## Notes

This affected a couple of our own sites which raised it to my attention. Notably in the last few days, our demo, staff, and success sites all lost their formatter as someone configured something on the posting page.


charrondev:
Codecov bot is bugged. Definitely wrong about this PR.

PR #10538, opened 2020-05-01
Honour ProfileExtender text fiels options on registration page

DaazKu:
Related to https://github.com/vanilla/support/issues/952

Options were honoured properly in the profile view but not the registration view.

This caused an issue where `MaxLength` option would be respected on the profile page but not at registration.

This fix replicate exactly what is in the other view to minimize unforeseen problems. The logic seems very odd to me...

### DIFF
--- a/plugins/ProfileExtender/views/registrationfields.php
+++ b/plugins/ProfileExtender/views/registrationfields.php
@@ -12,6 +12,10 @@ if (is_array($Sender->RegistrationFields)) {
             $Options = array_combine($values, $labels);
         }
 
+        if ($Field['FormType'] == 'TextBox' && !empty($Field['Options'])) {
+            $Options = $Field['Options'];
+        }
+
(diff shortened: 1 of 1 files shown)


plugins/ProfileExtender/views/registrationfields.php
```diff
@@ -12,6 +12,10 @@
             $Options = array_combine($values, $labels);
         }
 
+        if ($Field['FormType'] == 'TextBox' && !empty($Field['Options'])) {
```
charrondev, line 15: Was there a specific reason we decided to do this only for text boxes and not other fields?

DaazKu: Just added a comment on the main issue:

> This fix replicate exactly what is in the other view to minimize unforeseen problems. The logic seems very odd to me...

charrondev: That's fair.

Issue #9004, opened 2019-07-04
API Page links should be full URLs


tburry:
**Describe the bug**

When making an API call to a paged resource, the "Link" header contains links to the previous, next, first, and last pages. However, these links are relative rather than absolute links.

**To Reproduce**

1. Make a call to `/api/v2/discussions`.
2. Look at the headers of the results.

**Expected behavior**

The "Link" header should contain full URLs for all of the URIs in the result.

**Example**

When I call the following on my local dev:

```http
GET http://dev.vanilla.localhost/api/v2/discussions
```

I get the following header:

```http
Link: </api/v2/discussions?page=1>; rel="first", </api/v2/discussions?page=2>; rel="next", </api/v2/discussions?page=16>; rel="last"
```

I should get:

```http
Link: <http://dev.vanilla.localhost/api/v2/discussions?page=1>; rel="first", <http://dev.vanilla.localhost/api/v2/discussions?page=2>; rel="next", <http://dev.vanilla.localhost/api/v2/discussions?page=16>; rel="last"
```

PR #10548, opened 2020-05-04
Add host to links in discussion list API header

blutarch:
Closes vanilla/vanilla#9004

This PR fixes a problem where, when making this API call: `https://dev.vanilla.localhost/api/v2/discussions/`
the "Link" headers containing links to next and previous pages were relative rather than absolute.

### TO TEST
Check out the branch and make this call: `https://dev.vanilla.localhost/api/v2/discussions/`
Verify that the links in the "Link" header are absolute.

### DIFF
--- a/library/Vanilla/ApiUtils.php
+++ b/library/Vanilla/ApiUtils.php
@@ -157,7 +157,7 @@ class ApiUtils {
         if (!empty($argsStr)) {
             $url .= (strpos($url, '?') === false ? '?' : '&').$argsStr;
         }
-        $url = \Gdn::request()->url($url, true);
+        $url = \Gdn::request()->getSimpleUrl($url);
         return $url;
     }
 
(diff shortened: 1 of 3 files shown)


library/Vanilla/ApiUtils.php
```diff
@@ -157,6 +157,7 @@ private static function pagerUrlFormat($url, array $query, Schema $schema) {
         if (!empty($argsStr)) {
             $url .= (strpos($url, '?') === false ? '?' : '&').$argsStr;
         }
+        $url = \Gdn::request()->url($url, true);
```
charrondev, line 160: This may end up adding double subcommunity slugs onto the URLs. (Best to always test URL handling in a sub-folder, with subcommunities enabled).

`$request->getSimpleUrl($url)` is probably what you're looking for.

PR #10549, opened 2020-05-04
Added new components for event list

slafleche:
Added new components to storybook for main events component. Note that the event list for the panel and the full page will be in a future PR. This is just the main, component styles.

Closes: https://github.com/vanilla/internal/issues/2479

Companion PRs:
https://github.com/vanilla/internal/pull/2492

Translation strings: https://github.com/vanilla/locales/pull/289/files


slafleche:
>   I'd expect it to go down to the panel layout style on mobile (or maybe elisa had something else in mind?). If this were to look at the Device you could then have the panel variation force a smaller device with an additional context wrapper.

@charrondev  That will be next PR, with the panel styles. I was trying to break it up, especially since it only lives in Storybook for now.

slafleche:
The two dates are supposed to be the same. I fixed a bug in the code.

library/src/scripts/events/EventsList.tsx
```diff
@@ -0,0 +1,60 @@
...
+    data: IEvent[];
+}
+
+/**
+ * Component for displaying an accessible nicely formatted time string.
+ */
+export function EventsList(props: IEventList) {
+    const classes = eventsClasses();
+    if (!props.data || props.data.length === 0) {
+        return <p className={classes.empty}>{t("This category does not have any events.")}</p>;
```
charrondev, line 28:
```suggestion
        return <p className={classes.empty}>{t("There are no events.")}</p>;
```

Sorry I missed this on the first pass, but we should either change this to be more generic, or make it passable from outside. This widget may be used in anything events attach to, which could be a category, but also could be a group currently.

PR #10558, opened 2020-05-05
Event List for Panel

slafleche:
This PR includes the components and styles for the panel's event list component. The logic for not displaying it if empty, the view all and the actual placement are not part of this PR. This is strictly styling for the main component.

Closes: https://github.com/vanilla/internal/issues/2478

### DIFF
--- a/library/src/scripts/events/AttendanceStamp.tsx
+++ b/library/src/scripts/events/AttendanceStamp.tsx
@@ -6,8 +6,8 @@
 
 import React from "react";
 import { eventsClasses } from "@library/events/eventStyles";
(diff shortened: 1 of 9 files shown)


library/src/scripts/events/AttendanceStamp.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
```
charrondev, line 1: Missing a file doc block here.

slafleche:
@charrondev doc blocks have been fixed. I checked with Elisa and she was happy with the responsiveness of the events list. Keep in mind that in storybook, we have a lot more margin than we'll have in the KB.

PR #10563, opened 2020-05-07
Improve user feedback upon login if there's a user match but the password is missing

vanilla-dbarbier:
Original issue: https://github.com/vanillaforums/sonicwall/issues/50

**Description**: Depending on the login scenario the messaging sent to the user is confusing & may not be applicable. We need supplemental messages that would better explain the situation and guide the user. The new message goes as follow: "_You are trying to connect with a username that is already assigned to a user. If this is you, please enter the password you used when you created this account on this forum._" and is shown when the user tries to login with an existing user & no password is provided.

**QA Steps**:
- Have OAuth2 login enabled on the platform.
- Try to authentify with a user that would suggest an existing user handle. Ex.: If there is a user using the user handle "info" try an OAuth2 login with an email which starts with "info", like "info@somethingwebsite.com".
- Once the OAuth2 authentification is successful, the platform will suggest to login with the matching existing "info" user, thus displaying the newly added message.

### DIFF
--- a/applications/dashboard/controllers/class.entrycontroller.php
+++ b/applications/dashboard/controllers/class.entrycontroller.php
@@ -870,6 +870,14 @@ class EntryController extends Gdn_Controller {
                             } catch (Gdn_UserException $ex) {
                                 $this->Form->addError($ex);
                             }
+                        } else {
+                            // If we have a user match & there is no password.
+                            $this->Form->addError(
+                                t(
+                                    'You are trying to connect with a username that is already assigned to a user. '.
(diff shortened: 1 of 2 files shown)


nervoustwit:
That string will need to get translated. I think the best bet would be to give it a manageable name like t('UserMatchNeedsPassword') and add it to the `site_core` file in locales. 

For more info, see this doc https://staff.vanillaforums.com/kb/articles/225-guide-for-adding-and-editing-translation-strings and maybe talk to @blutarch if you need more guidance.

PR #10566, opened 2020-05-08
Add a test trait to make testing roles and user more easy

charrondev:
The trait structure and setup is inspired by the laravel setup test.

- Make an easy way to apply traits to our API testcase that each hook into the setup. This is done by making a method of a particular name. They cannot specify just `setUp` because then it is impossible to use multiple of these traits together.
- Create a user/role test trait.
  - Create users.
  - Create roles.

### DIFF
--- a/applications/dashboard/models/class.rolemodel.php
+++ b/applications/dashboard/models/class.rolemodel.php
@@ -13,6 +13,13 @@
  */
 class RoleModel extends Gdn_Model {
 
+    const DEFAULT_GUEST_ID = 2;
+    const DEFAULT_UNCONFIRMED_ID = 3;
+    const DEFAULT_APPLICANT_ID = 4;
+    const DEFAULT_MEMBER_ID = 8;
+    const DEFAULT_ADMIN_ID = 16;
+    const DEFAULT_MOD_ID = 32;
+
     /** Slug for Guest role type. */
     const TYPE_GUEST = 'guest';
 
--- a/applications/dashboard/models/class.rolemodel.php
+++ b/applications/dashboard/models/class.rolemodel.php
@@ -13,12 +13,12 @@
  */
 class RoleModel extends Gdn_Model {
 
-    const DEFAULT_GUEST_ID = 2;
-    const DEFAULT_UNCONFIRMED_ID = 3;
-    const DEFAULT_APPLICANT_ID = 4;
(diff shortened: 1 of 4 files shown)


applications/dashboard/models/class.rolemodel.php
```diff
@@ -13,6 +13,13 @@
  */
 class RoleModel extends Gdn_Model {
 
+    const DEFAULT_GUEST_ID = 2;
+    const DEFAULT_UNCONFIRMED_ID = 3;
+    const DEFAULT_APPLICANT_ID = 4;
+    const DEFAULT_MEMBER_ID = 8;
+    const DEFAULT_ADMIN_ID = 16;
+    const DEFAULT_MOD_ID = 32;
```
alex-mtl, line 21: I think we can truncate names
```suggestion
    const GUEST_ID = 2;
    const UNCONFIRMED_ID = 3;
    const APPLICANT_ID = 4;
    const MEMBER_ID = 8;
    const ADMIN_ID = 16;
    const MOD_ID = 32;
```

tests/InternalClient.php
```diff
@@ -22,7 +24,7 @@ class InternalClient extends HttpClient {
     /**
      * @var int
      */
-    private $userID;
+    private $userID = self::DEFAULT_USER_ID;
```
alex-mtl, line 27:
```suggestion
    private $userID = self::USER_ID;
```

tests/UsersAndRolesApiTestTrait.php
```diff
@@ -0,0 +1,108 @@
...
+    protected $lastUserID = null;
+
+    /** @var int|null */
+    protected $lastRoleID = null;
+
+    /**
+     * Clear local info between tests.
+     */
+    public function setUpUsersAndRolesApiTestTrait(): void {
+        $this->api()->setUserID(InternalClient::DEFAULT_USER_ID);
```
alex-mtl, line 27:
```suggestion
        $this->api()->setUserID(InternalClient::USER_ID);
```

tests/InternalClient.php
```diff
@@ -14,6 +14,8 @@
 
 class InternalClient extends HttpClient {
 
+    const DEFAULT_USER_ID = 2;
```
alex-mtl, line 17:
```suggestion
    const USER_ID = 2;
```

Issue #10530, opened 2020-04-29
Create accessibility for "New Discussion" buttons


tuan-ng:
### Scope
The React-based component in issue [#9696](https://github.com/vanilla/vanilla/issues/9696) has been created. 

Link: [accessibility](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-actions-active-descendant.html)

### Acceptance Criteria
+ Add accessibility handling into the "New Discussion" react component.

PR #10567, opened 2020-05-08
Add accessibility to **New Discussion** button 

tuan-ng:
Closes [#10530](https://github.com/vanilla/vanilla/issues/10530)

This handles accessibility according to [here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-actions-active-descendant.html). 

Note

1. The animation has been made to be very robust (and correct) now. 

I think a test suit (for testing handling accessibility) should be followed.


library/src/scripts/flyouts/NewPostBackground.tsx
```diff
@@ -2,31 +2,30 @@ import React from "react";
...
-
-    const trans = useSpring({
-        backgroundColor: props.open ? colorOut(vars.container.color.open) : colorOut(vars.container.color.close),
-        from: { backgroundColor: colorOut(vars.container.color.close) },
-        config: { duration: vars.container.duration },
-    });
 
     return createPortal(
-        <animated.div className={classNames(classes.container)} style={trans} onClick={props.onClick}>
+        <animated.aside
```
slafleche, line 21: Why is this an aside? It's not semantic to have click and keyboard events on that element. This is a button if you can click on it.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -29,7 +34,7 @@ export interface ITransition {
     transform: string;
 }
 
-function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
+function ActionItem({ item, style, aid }: { item: IAddPost; style?: ITransition; aid?: string }) {
```
slafleche, line 37: Look at other examples of declaring parameter types. You shouldn't be using ":" after the object, it should be on each prop.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +64,302 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                if (state.open && typeof state.focusedItem != "undefined") {
+                    dispatch({ type: "set_focused_item", item: (state.focusedItem + 1) % items.length });
+                }
+                break;
+            case "ArrowDown":
+                if (state.open && typeof state.focusedItem != "undefined") {
+                    dispatch({ type: "set_focused_item", item: (state.focusedItem - 1 + items.length) % items.length });
+                }
+                break;
+            case "A":
```
slafleche, line 148: I think you should just use the default here and just search for whatever key. Other languages may have accents or other characters that are not in this set.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +64,302 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                            dispatch({ type: "set_button_focus", focus: true });
+                            dispatch({ type: "set_focused_item", item: undefined });
+                            break;
+                        default:
+                            break;
+                    }
+                }
+                break;
+            default:
+                if (state.buttonFocus) {
```
slafleche, line 224: I don't understand why you added this case? As mentioned above, use the default for the search.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +64,302 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+            case "r":
+            case "s":
+            case "t":
+            case "u":
+            case "v":
+            case "w":
+            case "x":
+            case "y":
+            case "z":
+                // Add some handling
```
slafleche, line 200: Looks like you didn't finish this part?

library/src/scripts/flyouts/newPostMenu.story.tsx
```diff
@@ -15,6 +15,7 @@ const items: IAddPost[] = [
         id: "1",
         type: PostTypes.BUTTON,
         action: () => {
+            console.log("New Poll");
```
slafleche, line 18: Get rid of you console logs

library/src/scripts/flyouts/newPostMenuStyles.ts
```diff
@@ -68,6 +100,11 @@ export const newPostMenuClasses = useThemeCache(() => {
     const item = style("item", {
         marginTop: unit(vars.item.position.top),
         marginRight: unit(vars.item.position.right),
+        $nest: {
+            "&:focus": {
+                outline: `px solid ${colorOut(globalVars.mainColors.primaryContrast)}`,
```
slafleche, line 105: What's that "px" doing there?

library/src/scripts/flyouts/NewPostMenu.tsx
charrondev: To clear up a little bit, we have a clear convention defined for how we do prop types, which is generally:

```ts

// We don't always have something else to extend but we sometimes do.
// Less so now with hooks.
interface IProps extends SomeInterface {
   // Optional comment about the prop. Try to add these when you can.
   propName: Type;
}

function MyComponent(props: IProps) {
	const { ... } = props; // Destructuring here.
}


```

library/src/scripts/flyouts/NewPostBackground.tsx
slafleche: Oops, this is the background, this doesn't make sense as a button, nevermind.

tuan-ng: I've moved the key handler to the button menu, so this portal is out of the flow of the page and it seems to me semantically like an aside now.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: Got it! Will do again!

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: Got it! Will change!

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: Ok!

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I got it done!

library/src/scripts/flyouts/newPostMenu.story.tsx
tuan-ng: Already did!

library/src/scripts/flyouts/newPostMenuStyles.ts
tuan-ng: Fixed!

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: > To clear up a little bit, we have a clear convention defined for how we do prop types, which is generally:
> 
> ```ts
> // We don't always have something else to extend but we sometimes do.
> // Less so now with hooks.
> interface IProps extends SomeInterface {
>    // Optional comment about the prop. Try to add these when you can.
>    propName: Type;
> }
> 
> function MyComponent(props: IProps) {
> 	const { ... } = props; // Destructuring here.
> }
> ```

Ok, thanks for the info!

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                if (state.buttonFocus) {
+                    dispatch({ type: "set_button_focus", focus: true });
+                }
+                break;
+        }
+    };
+
+    const onKeyDown = (event: React.KeyboardEvent<any>) => {
+        event.preventDefault();
+        event.stopPropagation();
```
slafleche, line 151: Only call these if you're doing your dispatch. You're blocking all the other shortcuts on the page with this.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+            dispatch({ type: "set_open", open: false });
+            dispatch({ type: "set_focused_item", item: undefined });
         }
     };
 
-    const classes = newPostMenuClasses();
-    const { items } = props;
+    const onBgKeyDown = (event: React.KeyboardEvent<any>) => {
+        event.preventDefault();
+        event.stopPropagation();
```
slafleche, line 130: Only call these if you're doing your dispatch. You're blocking all the other shortcuts on the page with this.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+        switch (event.key) {
+            case "Escape":
+                if (state.open) {
+                    dispatch({ type: "set_open", open: false });
+                    dispatch({ type: "set_button_focus", focus: true });
+                    dispatch({ type: "set_focused_item", item: undefined });
+                }
+                break;
+            default:
+                // We don't want to lose focus in case we already have
```
slafleche, line 141: Don't set a default here, the focus will be where the user is, don't move it without them doing it.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+            case "ArrowUp":
+                if (state.open && typeof state.focusedItem != "undefined") {
+                    dispatch({ type: "set_focused_item", item: (state.focusedItem + 1) % items.length });
+                }
+                break;
+            case "ArrowDown":
+                if (state.open && typeof state.focusedItem != "undefined") {
+                    dispatch({ type: "set_focused_item", item: (state.focusedItem - 1 + items.length) % items.length });
+                }
+                break;
```
slafleche, line 180: You've got the up and down arrows mixed up here :)

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                        dispatch({ type: "set_focused_item", item: next.index });
+                    }
+                }
+                break;
+        }
+    };
+
+    const onMenuButtonKeyDown = (event: React.KeyboardEvent<any>) => {
+        event.preventDefault();
+        event.stopPropagation();
```
slafleche, line 218: Same here, only call inside a case

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+            default:
+                // We don't want to lose focus in case we already have
+                if (state.buttonFocus) {
+                    dispatch({ type: "set_button_focus", focus: true });
+                }
+                break;
+        }
+    };
+
+    const onKeyDown = (event: React.KeyboardEvent<any>) => {
```
slafleche, line 149: Can we call this "onToggleKeyDown" or something similar to differentiate it from the other handlers? Feel free to also break up this file into smaller components if it's getting too big.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                            dispatch({ type: "set_button_focus", focus: true });
+                            dispatch({ type: "set_focused_item", item: undefined });
+                            break;
+                        default:
+                            break;
+                    }
+                }
+                break;
+            default:
+                const itemList = items
```
slafleche, line 203: There's a bug here. If I open the menu with the keyboard, I am able to search for "t" to find the "to Google" link. However, after if i type "n" it should wrap around to the next item. 

If I'm on "New Poll", i should be able to hit "n" and jump to "New Idea". "n" again should jump to "New Discussion" and so on.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                    .map((item, index) => ({ ...item, index }))
+                    .filter(item => item.label.startsWith(event.key));
```
slafleche, line 205: This may be part of your problem, you remember I told you to do a string to lower on your event key and on the label?

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                    dispatch({ type: "set_open", open: true });
+                }
+                break;
+            case "ArrowUp":
+                if (state.buttonFocus) {
+                    dispatch({ type: "set_focused_item", item: items.length - 1 });
+                    dispatch({ type: "set_open", open: true });
+                }
+                break;
+            default:
```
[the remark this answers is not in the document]
tuan-ng: Ok, will change.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,285 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                    role="menu"
+                    aria-labelledby={buttonID}
+                    tabIndex={-1}
+                    aria-activedescendant={state.focusedItem}
                 >
-                    {item => props => <ActionItem key={item.id} item={item} style={props} />}
-                </Trail>
+                    {trail.map(({ opacity, transform, ...rest }, index) => (
+                        <ActionItem
+                            aid={`ID-${index}`} // accessibility id
```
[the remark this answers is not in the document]
tuan-ng: It's not really hard-coded, but yeah I will use the element id to make it even more robust.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I'll think through is comment again.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I'll think through this comment

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I think it's consistent with the way the items are counted numerically (but visually it might be confusing yes).

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I'll think through this comment.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: Ok, will change.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I don't think it's a bug, if you press N it will work. I don't see the spec requiring case-insensitive. I'll talk to you more.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: This is related to the above comment.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I see, will change.

library/src/scripts/flyouts/NewPostMenu.tsx
slafleche: Make it wrap, it feels weird to be "suck" on the bottom.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,288 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                break;
+        }
+    };
+
+    const onMenuItemKeyDown = (event: React.KeyboardEvent<any>) => {
+        const keys = ["Escape", "Home", "End", "ArrowUp", "ArrowDown", "Enter"];
+
+        // Only handling keys in these situations so preventDefault and stopPropagation in these
+        if (
+            keys.includes(event.key) ||
```
slafleche, line 149: This does not work, you also check for the open state. You're blocking the user from tabbing here, that's bad. I told you these are probably not needed and cause problems. If they are needed, you need to put them in your case, in the "if" statement.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,288 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                        dispatch({ type: "set_focused_item", item: next.index });
+                    }
+                }
+                break;
+        }
+    };
+
+    const onMenuButtonKeyDown = (event: React.KeyboardEvent<any>) => {
+        const keys = ["Escape", "ArrowDown", "Enter", " ", "ArrowUp"];
+        if (keys.includes(event.key)) {
```
slafleche, line 222: See other comment about this.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I don't understand. I think it looks fine.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,288 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                if (state.open) {
+                    event.preventDefault();
+                    event.stopPropagation();
+                    dispatch({ type: "set_open", open: false });
+                    dispatch({ type: "set_button_focus", focus: true });
+                    dispatch({ type: "set_focused_item", item: undefined });
+                }
+                break;
+            default:
+                break;
```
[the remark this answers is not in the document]
tuan-ng: I think it's ok there. I always have the break case.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,288 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                            dispatch({ type: "set_focused_item", item: undefined });
+                            break;
+                        case PostTypes.BUTTON:
+                            (item.action as () => void)();
+                            dispatch({ type: "set_open", open: false });
+                            dispatch({ type: "set_button_focus", focus: true });
+                            dispatch({ type: "set_focused_item", item: undefined });
+                            break;
+                        default:
+                            break;
```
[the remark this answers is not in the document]
tuan-ng: Again I think it's ok, I won't delete it.

library/src/scripts/flyouts/NewPostMenu.tsx
```diff
@@ -59,47 +71,288 @@ function ActionItem({ item, style }: { item: IAddPost; style: ITransition }) {
...
+                <animated.ul
+                    onKeyDown={onMenuItemKeyDown}
+                    style={menu}
+                    ref={accessMenuRef}
+                    id={menuID}
+                    role="menu"
+                    aria-labelledby={buttonID}
+                    tabIndex={-1}
+                    // focusedItem is also the index of an item (see aid below)
+                    aria-activedescendant={`${ID}-${state.focusedItem}`}
```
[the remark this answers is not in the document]
tuan-ng: Ok.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: I just deleted those according to you.

library/src/scripts/flyouts/NewPostMenu.tsx
tuan-ng: Again I deleted those according to you.

library/src/scripts/flyouts/NewPostMenu.tsx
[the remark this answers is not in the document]
slafleche: No, delete it, it's useless.

library/src/scripts/flyouts/NewPostMenu.tsx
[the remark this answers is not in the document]
slafleche: It's useless, delete.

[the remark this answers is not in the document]
tuan-ng: I explained above already, I won't delete it.

library/src/scripts/flyouts/NewPostMenu.tsx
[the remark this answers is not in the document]
tuan-ng: I explained above already, I won't delete it.

PR #10570, opened 2020-05-11
Added new components for Events Details to Storybook

slafleche:
This PR adds the components needed to make the new Event Details page. Note that it's not hooked up and the interactions won't work.

- New dataList component. I was going to use a data list here, but seeing as there are some accessibility concerns, I opted for a table.
- The radio as tabs components have been updated to accommodate a radio as buttons variant. It uses the same context and components, but with alternate classes and wraps the old components into the new Button styled ones.

Companion PR: https://github.com/vanilla/knowledge/pull/1820


library/src/scripts/dataLists/dataList.tsx
```diff
@@ -0,0 +1,56 @@
...
+import React from "react";
+
+import { dataListClasses } from "@library/dataLists/dataListStyles";
+import { t } from "@vanilla/i18n/src";
+import classNames from "classnames";
+
+export interface IData {
+    key: React.ReactNode;
+    value: React.ReactNode;
+}
```
charrondev, line 15: Could me make a more descriptive name for this?

Maybe `IDataListNode`?

library/src/scripts/dataLists/dataList.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export interface IDataList {
+    className?: string;
+    data: IData[];
+}
```
charrondev, line 20: Do we need to export this? Doesn't seem to be used anywhere so maybe we can just do the standard `IProps`

library/src/scripts/dataLists/dataList.tsx
```diff
@@ -0,0 +1,56 @@
+/**
```
charrondev, line 1: The capitalization of this file seems incorrect (lower cased). For case sensitive renames in git try this https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git

library/src/scripts/dataLists/dataList.tsx
```diff
@@ -0,0 +1,56 @@
...
+ */
+export function DataList(props: IDataList) {
+    const classes = dataListClasses();
+    if (props.data.length === 0) {
+        return null;
+    }
+    return (
+        <div className={classNames(classes.root, props.className)}>
+            <table className={classes.table}>
+                <caption>{t("Event Details")}</caption>
```
charrondev, line 34: This seems a little bit too coupled.

library/src/scripts/dataLists/dataList.story.tsx
```diff
@@ -0,0 +1,53 @@
...
+        {
+            key: t("When"),
+            value: (
+                <span
+                    dangerouslySetInnerHTML={{
+                        __html: `${renderToString(startDate)}${
+                            dummyEventDetailsData.dateEnd ? ` - ${renderToString(endDate)}` : ""
+                        }`,
+                    }}
+                />
```
charrondev, line 34: This seems particularly odd. I don't think we want to be using this when not necessary.

Also you're using a react-dom server rendering function here. Any reason it can't just be a normal component?

```suggestion
                <span>
                
                </span>
                    dangerouslySetInnerHTML={{
                        __html: `${renderToString(startDate)}${
                            dummyEventDetailsData.dateEnd ? ` - ${renderToString(endDate)}` : ""
                        }`,
                    }}
                />
```

library/src/scripts/dataLists/dataList.story.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default {
+    title: "Data List",
+    parameters: {},
+};
+
+export function Standard(props: { data: [] }) {
+    const startDate = <DateTime {...dummyEventDetailsData.dateStart} type={DateFormats.EXTENDED} />;
+    const endDate = <DateTime {...dummyEventDetailsData.dateEnd} type={DateFormats.EXTENDED} />;
+
+    const dummyData = [
```
charrondev, line 24: Types should be before, rather than cased.

```suggestion
    const dummyData: IData[] = [
```

library/src/scripts/dataLists/dummyEventData.ts
```diff
@@ -0,0 +1,91 @@
+import { EventAttendance } from "@library/events/eventOptions";
```
charrondev, line 1: This file is missing license information, and I think it's in the wrong folder. Shouldn't it be in `/events`?

library/src/scripts/events/EventDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const eventMetaData = [
+        {
+            key: t("When"),
+            value: (
+                <span
+                    dangerouslySetInnerHTML={{
+                        __html: `${renderToString(startDate)}${
+                            dummyEventDetailsData.dateEnd ? ` - ${renderToString(endDate)}` : ""
+                        }`,
+                    }}
```
charrondev, line 48: Same note about the other dangerously set HTML.

The code also seems duplicated. Maybe it should be it's own component?

library/src/scripts/layout/Heading.tsx
```diff
@@ -11,7 +11,7 @@ import { typographyClasses } from "@library/styles/typographyStyles";
 export interface ICommonHeadingProps {
     id?: string;
     depth?: 1 | 2 | 3 | 4 | 5 | 6;
-    renderAsDepth?: 1 | 2 | 3 | 4 | 5 | 6;
+    renderAsDepth?: 1 | 2 | 3 | 4 | 5 | 6 | "custom";
```
charrondev, line 14: Is this component really bringing much to the table in the custom mode?

At that point we could just do

`<h4 className={customClass} />` (or whatever level you want).

library/src/scripts/dataLists/dataList.tsx
slafleche: Good tip, thanks!

library/src/scripts/dataLists/dataList.tsx
slafleche: Woops, yup, fixed.

PR #10580, opened 2020-05-13
Fix permission check on GET categories endpoint

charrondev:
I'm not sure why this was manage before, but we do have category permissions. We have category-specific permissions.

Here's the check on the normal category controller.

https://github.com/vanilla/vanilla/blob/3c15c05b199331e5677c828b6f8d745052b86688/applications/vanilla/controllers/class.categoriescontroller.php#L321

This category model method checks the PermissionCategoryID of the given category, then goes to the session to check the permission on it.

### DIFF
--- a/applications/vanilla/controllers/api/CategoriesApiController.php
+++ b/applications/vanilla/controllers/api/CategoriesApiController.php
@@ -5,6 +5,7 @@
  */
 
 use Garden\Schema\Schema;
+use Garden\Web\Exception\ForbiddenException;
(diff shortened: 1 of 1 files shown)


charrondev:
The improvement here is the frontend can now fetch a single category for non-administrators. This can be done in the old controller, and through the index endpoint.

As far as I know the exception type is the same here though. It was a `PermissionException` before, and it's still a `PermisisonException` except now it's checking for a permission that more than 3-4 people on a site actually have.

PR #10588, opened 2020-05-14
Fix js files not loading at profile/edit endpoint

blutarch:
Closes vanilla/support#1824.

This PR fixes a problem that results in users being able to upload a photo from the `profile/edit` endpoint if that page includes with a `userphotomodule`. This fix loads the necessary js files that allow the button to function.

### TO TEST
1. Load a version of vanilla that includes the userphotomodule at the `profile/edit`.
1. Go to that endpoint and try to upload a photo.
1. Note that the "Upload New Picture" button doesn't work.
1. Checkout this branch and try again, this time noting that it does work.

### DIFF
--- a/applications/dashboard/controllers/class.profilecontroller.php
+++ b/applications/dashboard/controllers/class.profilecontroller.php
@@ -365,9 +365,13 @@ class ProfileController extends Gdn_Controller {
     public function edit($userReference = '', $username = '', $userID = '') {
         $this->permission(['Garden.SignIn.Allow', 'Garden.Profiles.Edit'], true);
 
+        $this->Head->addScript('js/library/jquery.js');
+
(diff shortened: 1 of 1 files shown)


applications/dashboard/controllers/class.profilecontroller.php
```diff
@@ -365,10 +365,14 @@ public function disconnect($UserReference = '', $Username = '', $Provider) {
     public function edit($userReference = '', $username = '', $userID = '') {
         $this->permission(['Garden.SignIn.Allow', 'Garden.Profiles.Edit'], true);
 
+        $this->Head->addScript('js/library/jquery.js');
```
igraziatto, line 368: you don't need to add jQuery here, jQuery is already added by `initialize()`

PR #10589, opened 2020-05-14
Translate Whoops error message

blutarch:
Closes vanilla/support#1896.

This PR puts the "Whoops! There was an error." string through the translate function.

### DIFF
--- a/library/core/functions.error.php
+++ b/library/core/functions.error.php
@@ -270,7 +270,7 @@ function gdn_ExceptionHandler($Exception) {
 
         if ($DeliveryType != DELIVERY_TYPE_ALL) {
             if (!$Debug) {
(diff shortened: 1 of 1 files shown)


library/core/functions.error.php
```diff
@@ -270,7 +270,7 @@ function gdn_ExceptionHandler($Exception) {
 
         if ($DeliveryType != DELIVERY_TYPE_ALL) {
             if (!$Debug) {
-                die('<b class="Bonk">Whoops! There was an error.</b>');
+                die('<b class="Bonk">' . t("Whoops! There was an error.") . '</b>');
```
nervoustwit, line 273: Suggestion
```
die('<b class="Bonk">'.t("Whoops! There was an error.").'</b>');
```

PR #10597, opened 2020-05-15
Fix theme loading being incorrect on mobile

charrondev:
Closes https://github.com/vanilla/support/issues/1818
Closes https://github.com/vanilla/support/issues/1884

## ThemeModel fixes

This was a little bit complicated to write unit tests for because the only other theme implementation we have is over in `internal`.

To compensate I've made a `MockThemeProvider`, and used that along with our `MockAddonProvider` to write some tests.

- [Tests added](https://github.com/vanilla/vanilla/pull/10597/commits/d5ec888abfd779021895f5de4a466d3e29186496). [They fail CI](https://circleci.com/gh/vanilla/vanilla/57415?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) because I haven't made the fix yet.
- [Changes to fix the bug](https://github.com/vanilla/vanilla/pull/10597/commits/894f331c9ac212b70eb9d33fcb13f8d999d9e40d). CI passed after this.

### New strategy

The new strategy goes like this:

- Anything looking for assets will ignore the mobile theme, and just pull from current or desktop.
- Anything looking for an addon will still see the set mobile addon.

**Tradeoff**

There is a trade-off here. New system will not recognize assets set from a mobile theme. This means no pages will not ever use the mobile theme config. Eg. If make `my-custom-mobile-theme` with some theme assets, and set it _only as my mobile theme_, it's assets will no be applied. Instead the assets from `Garden.CurrentTheme` will always always override it.

## Microsoft Edge issues

We had a test to see if the browser auto-escaped stuff in noscript tags, but it looks like edge only auto-escapes stuff in noscript tags on the first load from the server, meaning we can't do a good client side test for them.

I put a workaround in `prepareShadowRoot`. With [the fix](https://github.com/vanilla/vanilla/pull/10597/commits/0d09537614fade66b0f8d2650b9762e3ac363364) in place, I can confirm through testing in browser stack the contents do not HTML escaped anymore.


library/Vanilla/Models/ThemeModel.php
```diff
@@ -337,10 +338,10 @@ public function getMasterThemeKey($themeKey): string {
     /**
      * Get view theme addon
      *
-     * @return Addon
+     * @return AddonInterface
      */
-    public function getCurrentThemeAddon(): Addon {
-        $currentTheme = $this->getCurrentTheme();
+    public function getCurrentThemeAddon(): AddonInterface {
```
tburry, line 343: Is this a method of the interface? If not, you should keep the Addon type hint. If yes, the docblock should still list the correct type.

Ideally, this would be the right type for those that don't care about the interface and want to use the full Addon class. Type widening isn't something that's supported right now. You would instead do it without a type hint, but doc blocks only and wait until PHP 7.4 for the type hint.

library/Vanilla/Models/ThemeModel.php
```diff
@@ -349,9 +350,9 @@ public function getCurrentThemeAddon(): Addon {
      * Get view theme addon
      *
      * @param string $themeKey
-     * @return Addon
+     * @return AddonInterface
      */
-    public function getThemeAddon(string $themeKey = ''): Addon {
+    public function getThemeAddon(string $themeKey = ''): AddonInterface {
```
tburry, line 355: Same comment as above.

library/Vanilla/Models/ThemeModel.php
```diff
@@ -371,9 +372,14 @@ public function verifyThemeIdentifierIsValid($themeKey) {
...
      *
+     * @param bool $prioritizeAddon We have to decide if we want an accurate addon or not.
+     *
+     * If false (default) - The Garden.CurrentTheme will take priority on mobile.
+     * If true - The Garden.MobileTheme will take priority on mobile.
+     *
      * @return array The current theme or the fallback if it fails to load.
      */
-    public function getCurrentTheme(): array {
+    public function getCurrentTheme(bool $prioritizeAddon = false): array {
```
tburry, line 382: You may want to consider a string enum with a constant on this class rather than a bool. When the bool isn't clear to the reader, I prefer an enum.

library/Vanilla/Models/ThemeModel.php
charrondev: Done

library/Vanilla/Models/ThemeModel.php
charrondev: Done. I've made 2 constants.

Definitely cleaner, and gives me a better spot to put longer descriptions.

PR #10617, opened 2020-05-20
Fix Visits recording

igraziatto:
Closes https://github.com/vanilla/support/issues/1535
Relevant: https://github.com/vanilla/support/issues/1535#issuecomment-631628920

**Steps to test:**
1. Login with a user on your local and logoff.
2. On the DB verify that the "DateLastActive", it should be now. Make a note of the current "CountVisits".
3. Change this user "DateLastActive" to any time more than 20 minutes ago from the time you logged off.
4. Open an incognito tab and login again.
5. Verify that the CountVisits has not increased. You can even put a breakpoint in https://github.com/vanilla/vanilla/blob/d363af7b520c95b1a043275332175b1bb5b3ad28/library/core/class.session.php#L270 and see that the `DateLastActive` is set to now, and not the tie you've manually set.
6. Checkout this branch, close the incognito tab.
7. Repeat steps, the `CountVisits` should increase.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -1591,7 +1591,7 @@ class UserModel extends Gdn_Model implements UserProviderInterface {
      *
      * @param int $iD The ID of the user.
      * @param string|false $datasetType Whether to return an array or object.
-     * @param array $options Additional options to affect fetching. Currently unused.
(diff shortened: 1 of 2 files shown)


igraziatto:
Closing this PR because it was not the right approach to this issue.

Issue #10585, opened 2020-05-14
Pockets - By role/permission


slafleche:
- [x] Add a new section for roles/permissions
~- [ ] When it is checked, a section on the bottom with the `<MultiRoleInput />` component appears.~
- [x] Save the data with the rest of the pocket data
- [x] Update pocket code to check for this condition and check user role
- [x] Pocket administrators can always see pockets filtered by permission in test mode.

Working towards: https://github.com/vanilla/vanilla/issues/10435

PR #10618, opened 2020-05-20
Pockets by Role

slafleche:
This PR adds a new section to the pockets edit page to filter by role. Note that if you are in **test mode** and you have the `Plugins.Pockets.Manage` permission, you can see the pocket, even if your role is not explicitly set.

@charrondev I find the loading of the initial values for the input quite slow. I tried to disable it until it finishes loading. I tried multiple different ways to get a callback, but it was either called before the component rendered, or created loop problems with the useEffect functions. I'm open to any suggestions. 

Closes: https://github.com/vanilla/vanilla/issues/10585


slafleche:
I fixed the logic for the testing/admin case.

slafleche:
@charrondev please review ASAP this is blocking the rest of my sprint.

PR #10619, opened 2020-05-20
Updates for reCaptchav3

cchabilall83:
- Adds VanillaWhitelistContentSecurityPolicyProvider
- Add ReCaptcha public key to siteMeta

required for https://github.com/vanilla/knowledge/issues/1826

### DIFF
--- a/library/Vanilla/Web/ContentSecurityPolicy/VanillaWhitelistContentSecurityPolicyProvider.php
+++ b/library/Vanilla/Web/ContentSecurityPolicy/VanillaWhitelistContentSecurityPolicyProvider.php
@@ -1,0 +1,21 @@
+<?php
+
+
+namespace Vanilla\Web\ContentSecurityPolicy;
+
+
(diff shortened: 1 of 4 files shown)


library/Vanilla/Web/ContentSecurityPolicy/VanillaWhitelistContentSecurityPolicyProvider.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Vanilla\Web\ContentSecurityPolicy;
+
+/**
+ * Vanilla whitelist content security policy provider
+ */
+class VanillaWhitelistContentSecurityPolicyProvider implements ContentSecurityPolicyProviderInterface {
+
+    const VANILLA_WHITELIST = [
+        'https://www.google.com/recaptcha/api.js'
```
charrondev, line 15: This is supposed to be just the domain. AFAIK it's not possible to restrict to just a particular script. Doing this allows the whole domain anyways, so it better to be more explicit.

```suggestion
        'https://www.google.com'
```

PR #10620, opened 2020-05-21
Use the container for dba/counts model instantiation

bleistivt:
Currently the `counts` method can't be used if you want to use autowired dependencies in your models contructor, because it is called using `new`.

This checks if a class is available through the container first.


bleistivt:
Still relevant, please consider removing the stale bot.

https://blog.benwinding.com/github-stale-bots/

bleistivt:
Still relevant

PR #10621, opened 2020-05-21
Add a UUID request ID to logs

tburry:
Generate a UUID for each request for the purposes of logging.

_Note: I added some methods to `Gdn_Request` that model the PSR-7 `ServerRequestInterface`_

### DIFF
--- a/tests/Library/Core/RequestTest.php
+++ b/tests/Library/Core/RequestTest.php
@@ -480,7 +480,7 @@ class RequestTest extends SharedBootstrapTestCase {
     public function testAttributeAccessors(): void {
         $r = new Gdn_Request();
 
-        $this->assertNotFalse($r->getAttribute('foo'));
+        $this->assertNull($r->getAttribute('foo'));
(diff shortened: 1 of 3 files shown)


tests/Library/Core/RequestTest.php
```diff
@@ -473,4 +473,17 @@ public function testUrlEquivalence() {
...
         $this->assertSame($req->getUrl(), $req->url('', true));
     }
+
+    /**
+     * Test basic attribute accessors.
+     */
+    public function testAttributeAccessors(): void {
+        $r = new Gdn_Request();
+
+        $this->assertNotFalse($r->getAttribute('foo'));
```
initvector, line 483: Since the default of `getAttribute` is `null`, would this be a better assertion to make? 

```suggestion
        $this->assertNull($r->getAttribute('foo'));
```

tburry: Autocomplete fail right there.

PR #10640, opened 2020-05-25
Add utility method to ensure reCaptcha script is loaded

cchabilall83:
required for https://github.com/vanilla/knowledge/pull/1833


library/src/scripts/utility/appUtils.tsx
```diff
@@ -8,6 +8,7 @@
 import gdn from "@library/gdn";
 import { PromiseOrNormalCallback } from "@vanilla/utils";
 import isUrl from "validator/lib/isURL";
+import { ensureScript } from "@vanilla/dom-utils/src";
```
charrondev, line 11:
```suggestion
import { ensureScript } from "@vanilla/dom-utils";
```

PR #10641, opened 2020-05-25
Remove media embed code from plain text excerpts of Rich text.

nervoustwit:
This PR will close https://github.com/vanilla/support/issues/1907

In all of our formatters, when we convert to plain text we remove image embed code and video embed code and replace it with a placeholder string. With Rich editor we are leaving the URL to the original image which causes issues when you don't have a lot of real estate to display an excerpted text.

This PR fixes that.

### To Test
 - With the Rich editor for formatting posts, create a discussion with an image or video embedded early in the discussion.
 - Turn on the Discussions Excerpt addon.
 - Go to the discussions page and see that in the list of discussions, your discussion has the url of the image in it.
 - Checkout this branch.
 - See that it is fixed.

I have already done some investigation to see that this does not cause unwanted behaviour in other PlainText conversion use cases on the application. More input or investigation would be welcome.

@charrondev if you have any ideas...


library/Vanilla/Formatting/Quill/BlotGroup.php
```diff
@@ -394,6 +394,10 @@ public function getUnsafeText(): string {
                 $text .= $blot->getContent();
             }
 
+            if ($blot instanceof ExternalBlot) {
+                return '(Embedded Media)';
```
charrondev, line 398: This is going to need to be translated. Maybe double check with todd if it's the actual wording we want to use.

nervoustwit:
We need to add 'link', 'file', 'image', 'error', and 'quote' to the basic strings to be translated. All the other embed types are proprietary, like 'vimeo' and 'giphy'.

library/Vanilla/Formatting/Quill/BlotGroup.php
nervoustwit: The Embedded Media is now the "last resort" if there is no embed type specified (ie youtube, image, file, etc.) I didn't bother Todd with it.

PR #10642, opened 2020-05-25
Overhaul theme API

charrondev:
Working towards https://github.com/vanilla/internal/issues/2521

Companion PRs in other repos

https://github.com/vanilla/internal/pull/2576
https://github.com/vanilla/knowledge/pull/1836

## Goals

- Have `/theme` API endpoint request bodies and response bodies be roughly symmetrical in the types of the requests and responses.
- Fix numerous validation issues in the various post/patch places.
- Remove some awkwardness from the internal structure.

## Renaming & Refactoring

- There was a bit of renaming and refactoring. All core classes for theming were centralized in the `Vanilla\Theme` namespace. Some of it was already there, but previously a bunch was in `Vanilla\Models` which is a lot less specific. Additionally pretty much none of this directly touches the DB.
- Anything non-DB related with `ThemeModel` in the name was renamed to `ThemeService`
 
  This one in particular was confusing because we have an actual DB `ThemeModel` class. The core stuff doesn't go to the DB and isn't really a model. It's a collector of providers. Our other collectors of providers have `Service` in the name (`FormatService`, `EmbedService`).

- Theme related `Vanilla\*Asset` ->  `Vanilla\Theme\Asset\*ThemeAsset`
 
  We have webpack assets as well. The naming made it a little confusing as a glance because very few people would use the namespaced name.

- `ThemeProviderInterface` and `FsThemeMissingTrait` have been refactored.

  `ThemeProviderInterface` -> `ThemeProviderInterface` & `ThemeProviderWriteInterface`.

  The service now checks for instances of the write interface.

- Remove direct knowledge of any theme `type`. The service no longer cares about 2 hardcoded theme types. It now calls `ThemeProviderInterface::handlesThemeID()` to know if a provider can handle a particular theme.

## New classes

**`Vanilla\Theme\Theme`**

The contents of a `Theme` grew and grew and grew over time as we added more features. It was getting particular confusing to pass around arrays and have no idea what state the array was in.

The `ThemeService` and related classes now use the `Vanilla\Theme\Theme` class to store theme data now.

**`Vanilla\Theme\ThemeAssetFactory`**

This new class is responsible for validating and creating theme assets.

**`Vanilla\Theme\ThemePreview`**

Take preview generation logic out of the various theme classes and centralize it in this class.

## API changes

_**There are breaking changes in the API. I've discussed them with todd.**_

- `PUT` and `GET` asset endpoints now have 2 variations
  - PUT w/ extensions. Eg. `PUT or GET /themes/:id/assets/assetName.ext`
     - Returns & accepts asset content directly. Type of the asset maps to the `content-type`.
  - PUT with no extensions. Eg. `PUT or GET /themes/:id/assets/assetName`
    - Returns and accepts json. Values are the same as an asset from the `GET/POST/PATCH` /themes endpoint.
- Simplified the preview field of a theme.

I've updated the openapi schemas to accomodate the new changes, and updated our frontend as well where necessary.


applications/dashboard/controllers/api/ThemesApiController.php
```diff
@@ -189,77 +199,70 @@ public function put_current(array $body): array {
     public function put_preview(array $body): array {
```
cchabilall83, line 199: invalid return type, we're getting a THEME back

cchabilall83:
Tested endpoints
Test creating a new theme, updating, applying 
Errors on put_current and put_preview were fixed.

library/Vanilla/Theme/Asset/NeonThemeAsset.php
```diff
@@ -0,0 +1,80 @@
...
+    public function render(string $asType = null): Data {
+        switch ($asType) {
+            case ThemeAssetFactory::ASSET_TYPE_NEON:
+                $data = new Data($this->neonString);
+                $data->setHeader('Content-Type', $this->getContentType());
+                return $data;
+            case ThemeAssetFactory::ASSET_TYPE_JSON:
+            default:
+                $data = new Data($this->jsonString);
+                $data->setHeader('Content-Type', parent::getContentType());
```
alex-mtl, line 76: Is it intentional to return the same mime type as for `ASSET_TYPE_NEON`?

library/Vanilla/Theme/ThemePreview.php
```diff
@@ -0,0 +1,96 @@
...
+    public function addVariablePreview(JsonThemeAsset $jsonAsset): ThemePreview {
+        $this->variablePreview = [];
+        $variables = $jsonAsset->getValue();
+
+        $preset = $variables['global']['options']['preset'] ?? null;
+        $bg = $variables['global']['mainColors']['bg'] ?? $preset === 'dark' ? "#323639" : "#fff";
+        $fg = $variables['global']['mainColors']['fg'] ?? $preset === 'dark' ? '#fff' : '#555a62';
+        $primary = $variables['global']['mainColors']['primary'] ?? null;
+        $this->variablePreview['globalPrimary'] = $primary;
+        $this->variablePreview['globalBg'] = $bg ?? null;
```
alex-mtl, line 69: I can't see how `$bg` could be `null` here. So `?? null` is useless?

library/Vanilla/Theme/ThemePreview.php
```diff
@@ -0,0 +1,96 @@
...
+        $this->variablePreview['globalFg'] = $fg ?? null;
```
alex-mtl, line 70: I can't see how `$fg` could be `null` here. So `?? null` is useless?

library/Vanilla/Theme/Asset/NeonThemeAsset.php
charrondev: It's different. Note `parent::getContentType()`. Json is our parent class.

library/Vanilla/Theme/ThemePreview.php
charrondev: Fixed.

library/Vanilla/Theme/ThemePreview.php
charrondev: Fixed.

PR #10644, opened 2020-05-26
Add ReCaptchaVerification Class and update reCaptchaV3 plugin

cchabilall83:
- Add a ReCaptchaVerification Class
- Updates reCaptchaPlugin to allow adding reCaptchaV3 keys

required for https://github.com/vanilla/knowledge/pull/1833

### DIFF
--- a/library/Vanilla/ReCaptchaVerification.php
+++ b/library/Vanilla/ReCaptchaVerification.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license Proprietary
+ */
+namespace Vanilla;
+
+
+use Garden\Http\HttpClient;
+
+/**
+ * Class reCaptchaVerification
+ */
+class ReCaptchaVerification {
+
+    CONST RECAPTCHA_V3_URL = "https://www.google.com/recaptcha/api/siteverify";
+
+    /** @var HttpClient */
+    private $httpClient;
+
+    /**
+     * reCaptchaVerification constructor.
+     */
+    public function __construct() {
+        $this->httpClient = new HttpClient();
+    }
+
+    /**
+     * Verify is a challenge is valid, using siteVerify endpoint.
+     *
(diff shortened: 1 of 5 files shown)


library/Vanilla/ReCaptchaVerification.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Verify is a challenge is valid, using siteVerify endpoint.
+     *
+     * @param string $privateKey
+     * @param string $responseToken
+     * @return bool
+     */
+    public function siteVerify(string $privateKey = '', string $responseToken = ''):bool {
```
charrondev, line 35:
```suggestion
    public function siteVerifyV3(string $privateKey = '', string $responseToken = ''): bool {
```

I'd recommend putting a v3 in the name, since v2 is a thing. Don't want people getting confused.

charrondev: My thinking is it should also not require the private key as a prop. It should just pull it from the config in the constructor.

library/Vanilla/ReCaptchaVerification.php
```diff
@@ -0,0 +1,45 @@
...
+    const RECAPTCHA_V3_URL = "https://www.google.com/recaptcha/api/siteverify";
+
+    /** @var HttpClient */
+    private $httpClient;
+
+    /**
+     * ReCaptchaVerification constructor.
+     */
+    public function __construct() {
+        $this->httpClient = new HttpClient();
```
charrondev, line 25: We should be dependency injecting this (so any global middlewares get applied.)

library/Vanilla/ReCaptchaVerification.php
cchabilall83: done

library/Vanilla/ReCaptchaVerification.php
cchabilall83: done

PR #10649, opened 2020-05-26
Implement caching for the theme service

charrondev:
Closes https://github.com/vanilla/internal/issues/2581

Depends on https://github.com/vanilla/vanilla/pull/10642 so I'm targeting that for now.

After a recent issue in some high-traffic production sites, I've added theming to our `ThemeService`.


library/Vanilla/Theme/ThemeCache.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public function cacheKey($themeID, array $args = []): string {
+        // Put args in a stable order.
+        ksort($args);
+
+        $cacheKey = self::CACHE_KEY . '_' . $themeID;
+        foreach ($args as $key => $val) {
+            $cacheKey .= '_' . $key . '-' . json_encode($val);
+        }
+        return $cacheKey;
```
alex-mtl, line 98: This code will generate huge `$cacheKey`.

To make unique key we can just generate 
```suggestion
        ksort($args);

        $cacheKey = self::CACHE_KEY . '_' . $themeID.md5(json_encode($args));
        
        return $cacheKey;
```

Issue #10417, opened 2020-04-14
Ability to insert a pocket after the banner


rhysboswarva:
Ability to insert a pocket after the banner

From a selling point being able to put any front end code throughout the page is nice to be able to say.  However with Foundation and the new Theme Editor (which is awesome) you can no longer place Custom pockets.  This is fine, but there is one key position on the page that is missing.  The ability to place code right after the banner.  e.g. Tiles are a popular GUI feature that can't be done, without either Custom Dev, or placing a pocket after the banner.

Examples:

![image](https://user-images.githubusercontent.com/923781/79239496-a0dafe00-7e3e-11ea-8473-7d275a91e869.png)

![image](https://user-images.githubusercontent.com/923781/79239555-b3edce00-7e3e-11ea-948b-4cd997a01675.png)

Thanks 
Rhys

slafleche:
@rhysboswarva am I right to assume the client would be satisfied with a customized home page? What exactly do they want to add?

rhysboswarva:
No, it's the ability to add content below the banner.   Not sure what a "customized home page" is.  

like this:
![image](https://user-images.githubusercontent.com/923781/81183840-a6c78900-8f7d-11ea-8645-803781c5ac86.png)

slafleche:
We do plan to allow the client to customize their home page with widgets. I feel like what they want here could be achieved with custom tile widgets on their home page?

charrondev:
@slafleche That is not in our pipeline until much later in the year, and even then doesn't have the ability to customize through the dashboard.

The scope here, is there's a position beneath the banner where many people want to put some custom HTML widget.

The scope between another pocket location and a overhauled homepage system with some UI is massive.

vrjrobitaille:
100% Example here is King's community, they've placed a "Join now!" CTA below the banner for Guest users. Smartsheet has also placed a Covid banner below their community bannner

<img width="1421" alt="Screen Shot 2020-05-06 at 5 20 23 PM" src="https://user-images.githubusercontent.com/18470831/81229496-e6ae6080-8fbd-11ea-8862-74edacd46710.png">

<img width="1410" alt="Screen Shot 2020-05-06 at 5 21 08 PM" src="https://user-images.githubusercontent.com/18470831/81229543-ff1e7b00-8fbd-11ea-8b5a-8d4bcab92e40.png">

charrondev:
Note, this location would only be available for foundation.

slafleche:
[...]

PR #10653, opened 2020-05-27
Added pocket location after banner

slafleche:
This PR adds a pocket location under the banner, only for data driven themes.

Note I also found a bug with the category filter not being cleared properly on save.

Closes: https://github.com/vanilla/vanilla/issues/10417

### DIFF
--- a/applications/dashboard/models/BannerImageModel.php
+++ b/applications/dashboard/models/BannerImageModel.php
@@ -71,6 +71,16 @@ class BannerImageModel {
         } else {
             $html = "<div data-react='community-content-banner' data-props='$propsJson'><div style=\"minHeight='500px'\"></div></div>";
         }
+
+        if (Gdn::themeFeatures()->useDataDrivenTheme()) {
+            /** @var \Garden\EventManager $eventManager */
+            $eventManager = Gdn::getContainer()->get(\Garden\EventManager::class);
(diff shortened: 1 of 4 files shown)


applications/dashboard/models/BannerImageModel.php
```diff
@@ -71,6 +71,16 @@ public function renderBanner(array $props = []): \Twig\Markup {
...
         }
+
+        if (Gdn::themeFeatures()->useDataDrivenTheme()) {
+            /** @var \Garden\EventManager $eventManager */
+            $eventManager = Gdn::getContainer()->get(\Garden\EventManager::class);
+            $afterBanner = $eventManager->fire('AfterBanner', $this);
+            if (!empty($afterBanner)) {
+                $html .= implode("", $afterBanner);
+            }
+        }
```
charrondev, line 82:
```suggestion
        /** @var \Garden\EventManager $eventManager */
        $eventManager = Gdn::getContainer()->get(\Garden\EventManager::class);
        $afterBanner = $eventManager->fire('AfterBanner', $this);
        if (!empty($afterBanner)) {
            $html .= implode("", $afterBanner);
        }
```

We shouldn't need this check here. Only data driven themes call this function, and if something else did call it we probably want to render anyways.

PR #10658, opened 2020-05-27
Add the ObjectUtils class

tburry:


tests/Library/Vanilla/Utility/ObjectUtilsTest.php
```diff
@@ -0,0 +1,81 @@
...
+            private $a;
+        };
+
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionCode(403);
+        ObjectUtils::hydrate($a, ['a' => 'a']);
+    }
+
+    /**
+     * You can't set a non-public property.
```
initvector, line 57:
```suggestion
     * You can't set a nonexistent property.
```

library/Vanilla/Utility/ObjectUtils.php
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * The array is parsed where keys represent properties to be set by the values. To set a property, the following is done:
+     *
+     * 1. Look for a method in the form: `set$property`.
+     * 2. Look for a method in the form: `with$property`.
+     * 3. Look for an actual property.
+     *
+     * @param object $object The object to hydrate.
+     * @param array $data The data to hydrate with.
+     * @return object Returns the same class or a new one in the case of with methods.
```
initvector, line 25: There doesn't seem to be any "with" support in this particular method. Is this from an earlier iteration?

initvector, line 25: Since we're always modifying the actual `$object` instance, what's the intention of returning a value too? A fluent interface?

library/Vanilla/Utility/ObjectUtils.php
```diff
@@ -0,0 +1,59 @@
...
+            try {
+                if (method_exists($object, $setter = "set$key")) {
+                    call_user_func([$object, $setter], $value);
+                } elseif (property_exists($object, $key)) {
+                    $object->$key = $value;
+                } else {
+                    throw new \InvalidArgumentException("Object doesn't have a $key property.", 400);
+                }
+            } catch (\Error $ex) {
+                // Private access error, chang to an exception.
```
initvector, line 38:
```suggestion
                // Private access error, change to an exception.
```

library/Vanilla/Utility/ObjectUtils.php
tburry: I had put that in, but decided to remove it.

tburry: I was thinking of this:

```php
$o = ObjectUtils::hydrate(new Something(), ['foo' => 'bar', ...]);
```

It's also useful for regular fluent calls.

tests/Library/Vanilla/Utility/ObjectUtilsTest.php
tburry: In my test non-public refers to private or protected.

initvector: For the test above, `testHydratePrivate`, yeah. But this is the doc block for `testHydrateMissing`. It looks like a copy-paste error. We're testing setting a property that doesn't exist.

PR #10666, opened 2020-05-28
Add a URI library to Vanilla

tburry:
This PR is mainly to fix a bug when we are stripping our `?sso=` parameter by bringing in a proper URL library to do this kind of operation rather than using regular expressions.

Closes https://github.com/vanilla/support/issues/968.


library/Vanilla/Utility/UrlUtils.php
```diff
@@ -35,4 +38,46 @@ public static function domainAsAscii(string $url): ?string {
...
+     *
+     * @param UriInterface $uri The base URI.
+     * @param array $replace The querystring replacement.
+     * @return UriInterface
+     */
+    public static function replaceQuery(UriInterface $uri, array $replace): UriInterface {
+        parse_str($uri->getQuery(), $query);
+        $query = array_replace($query, $replace);
+        $result = $uri->withQuery(http_build_query($query));
+        return $result;
```
charrondev, line 53: Do we still need the `http_build_query` and `parse_str` here?

Looks likes we could do this:

```
return UriModifier::mergeQuery($uri, QueryString::build($replace));
```

tburry: I guess so, but I don't know if it's enough reason to change the implementation.

initvector: For what it's worth, I don't think this is possible without requiring the [league/uri-components](https://packagist.org/packages/league/uri-components) package. We're only introducing [league/uri](https://packagist.org/packages/league/uri) here.

PR #10672, opened 2020-05-29
Accessibility improvements - IDs, Landmarks, Roles, Headings

slafleche:
Fixes various accessibility issues:

- Duplicate IDs
- Bad or missing Landmarks
- Bad or redundant roles
- Fixed heading order

(there's also an auto formatted json file that snuck in there)


library/Vanilla/Web/MasterView.twig
```diff
@@ -30,7 +30,7 @@
                 <header id="titleBar"></header>
                 <div id="app" class="page-minHeight">
                     {%- if seoContent|default(false) -%}
-                        <noscript>
+                        <noscript aria-hidden="true">
```
charrondev, line 33: We can't put this here. This exists for users search engines and users without javascript enabled.

It gets entirely replaced if the react app mounts.

slafleche:
@charrondev if you find more pages missing landmarks let me know. However, I don't want this PR to linger for too long. If what I have is ok let's get it merged and I'll make another PR after.

PR #10676, opened 2020-06-01
Fix theme current crash

charrondev:
Fixes https://github.com/vanilla/knowledge/issues/1840

We had an edge case where a theme could be cached, but the value of whether it was current or not was changed by something that didn't trigger a cache clear (like manually editing a theme).

This would cause the manage themes page to get the incorrect "current" theme. In order to mitigate this, our "normalize" method now set the proper current theme for every theme that passes through.


charrondev:
Merging over unrelated test failures (pruneModel and dateRange).

PR #10687, opened 2020-06-02
Fix broken Facebook Social Connect button

blutarch:
Closes vanilla/support#1527.

This PR fixes a problem where clicking on the facebook "connect" button at the `profiles/connections` endpoint causes an error because the redirect url is dynamic, which facebook doesn't allow. This changes the redirect url to `entry/connect/facebook`.

### TO TEST
You have to have the facebook social connect addon enabled and configured (meaning you have to add a domain in your `etc/hosts` file that facebook will accept).

1. Using an account that has an email address linked to a facebook account, go to the `profile/connections` endpoint and click on the "Connect" button associated with facebook.
1. Note that facebook sends back an error message.
1. Check out this branch and try again.
1. Note that you can connect successfully.

### DIFF
--- a/plugins/Facebook/class.facebook.plugin.php
+++ b/plugins/Facebook/class.facebook.plugin.php
@@ -319,7 +319,7 @@ class FacebookPlugin extends SSOAddon {
         $this->EventArguments['User'] = $sender->User;
         $this->fireEvent('AfterConnection');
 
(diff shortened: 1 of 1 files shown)


blutarch:
> It sounds like Facebook Login does allow dynamic redirects, even in strict mode:
> Have you investigated this option?

I had not! But I will now.

PR #10689, opened 2020-06-02
Make the GuestModule more flexible

igraziatto:
Reference: https://github.com/vanillaforums/oracle/issues/10

**Goal:** Make the GuestModule more flexible
**What this PR does:**
- Add CSS classes to elements on the view so those can be targeted more easily.
- Pass Signin and Register URLs as data leveraging logic from the view.

### DIFF
--- a/applications/dashboard/views/modules/guest.php
+++ b/applications/dashboard/views/modules/guest.php
@@ -1,24 +1,27 @@
 <?php if (!defined('APPLICATION')) exit(); ?>
 <div class="Box GuestBox">
-    <h4><?php echo t('Howdy, Stranger!'); ?></h4>
+    <h4 class="GuestBox-title">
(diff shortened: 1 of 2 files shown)


applications/dashboard/views/modules/guest.php
```diff
@@ -1,24 +1,27 @@
...
+    <p class="GuestBox-beforeSignInButton">
+        <?php $this->fireEvent('BeforeSignInButton'); ?>
+    </p>
 
     <?php
-    $signInUrl = signInUrl($this->_Sender->SelfUrl);
-
-    if ($signInUrl) {
-        echo '<div class="P">';
+        if ($this->data('signInUrl')) {
```
tburry, line 15: Can you put this at the same indentation level as before so the diff is cleaner?

PR #10693, opened 2020-06-02
Initial moving of foundation variables into forumStyleVariables

alexbrohman:

* [ ] Title is formatted in present imperative (ex: "Fix X by doing Y").


library/src/scripts/forums/forumStyleVars.ts
```diff
@@ -29,6 +30,29 @@ export const forumVariables = useThemeCache((forcedVars?: IThemeVariables) => {
...
+                    active: "",
+                },
+            },
+            meta: {
+                colors: {
+                    noState: colorOut(globalVars.mainColors.fg),
+                    focus: colorOut(globalVars.links.colors.focus),
+                    keyboardFocus: colorOut(globalVars.links.colors.keyboardFocus),
+                    hover: colorOut(globalVars.links.colors.hover),
+                    active: colorOut(globalVars.links.colors.active),
```
charrondev, line 52: These are not styles so I would recommend against using `colorOut` here. You can just use the color directly.

`colorOut()` is to be used only when creating the final CSS class.

applications/dashboard/src/scripts/compatibilityStyles/index.ts
```diff
@@ -347,7 +349,7 @@ compatibilityStyles = useThemeCache(() => {
     });
 
     cssOut(`.Item.Read`, {
-        background: "none",
+        backgroundColor: formVars.lists.colors.backgroundColorRead,
```
charrondev, line 352:
```suggestion
        backgroundColor: colorOut(formVars.lists.colors.backgroundColorRead),
```

After the other changes this should become colorOut instead.

slafleche:
These changes were moved to another branch. Closing PR. See https://github.com/vanilla/vanilla/tree/feature/forum-variables2

PR #10694, opened 2020-06-03
Accessibility improvements - landmarks, pagination, tabindex, headings, tags, ids

slafleche:
Various accessibility improvements:

- Landmakrs
- Pagination
- tabindex
- heading order
- tags
- ids
- added aria-label to give context to links with only a date


applications/vanilla/views/discussion/tag.php
```diff
@@ -1,6 +1,11 @@
-<?php if (!defined('APPLICATION')) exit(); ?>
-    <h1><?php echo $this->data('Title'); ?></h1>
-<?php
+<?php if (!defined('APPLICATION')) exit();
+$tag = 'h1';
+if ($this->data['mainHeadingLevel']) {
+    $tag = "h2";
+}
```
charrondev, line 5: Shouldn't this be using the `headingTag` function?

PR #10698, opened 2020-06-03
Sort category dropdown list alphabetically

blutarch:
Closes vanilla/support#1941.

Currently, the categories in the dropdown list are sorted according to the "Tree Left" value. In most cases, admins can order categories as they choose. However, when a category is set to display as flat, it isn't possible to change the order of the child categories. In the dashboard, the children categories appear in alphabetical order, but in the category dropdown menu, they aren't sorted alphabetically, which makes it difficult for users to find the category they want. This PR sorts the children categories alphabetically in the dropdown menu by adding the sortFlatCategories() method.

This PR also introduces the method loadAllCategoriesDb(), which is refactored code from loadAllCategories().

### TO TEST
1. Make sure you have one or more categories are set to display "flat" and have multiple child categories.
1. Create a new discussion.
1. Go to select a category.
1. Note that the child categories are not sorted alphabetically.
1. Check out this branch.
1. Refresh the page and note that the child categories are sorted alphabetically.


tburry:
This is not the right approach. Only flat categories have an automatic sort. The other methods sort as they should because a community manager sorts them manually.

applications/vanilla/models/class.categorymodel.php
```diff
@@ -40,6 +40,8 @@ class CategoryModel extends Gdn_Model {
     /** Flag for aggregating discussion counts. */
     const AGGREGATE_DISCUSSION = 'discussion';
 
+    const DISPLAY_FLAT = 'Flat';
```
tburry, line 43: Since this constant was introduced, you should introduce the other constants from the display as and see where they might be referenced within this class.

tests/Models/CategoryModelTest.php
```diff
@@ -0,0 +1,626 @@
...
+        $sort($categories);
+    }
+
+    /**
+     * Test sortFlatCategories() for ordering alphabetically.
+     */
+    public function testSortFlatCategoriesOrdersCorrectly(): void {
+        $categories = [
+                10 =>
+                array (
```
tburry, line 60: Reformat to short array syntax.

PR #10708, opened 2020-06-05
Accessibility Improvements - aria labels, duplicate IDs, alt text, discernible text, landmarks

slafleche:
Various accessibility improvements:

- Fixed aria labels and translations
- Fixed duplicate atwho ID
- Fixed text in buttons to be visible to screen readers
- Fixed broken text for link to category
- Fixed alt text issue
- Fixed landmark issue
- Fixed mismatched aria-described by IDs for flyouts and for the editor in the comments section


tests/InternalClient.php
```diff
@@ -160,6 +165,15 @@ public function getUserID() {
         return $this->userID;
     }
 
+    /**
+     * Get the user ID that will be used to make requests.
+     *
+     * @return int Returns the activityName.
+     */
+    public function getActivityName() {
```
charrondev, line 173: Where is this getting set? In any case it definitely shouldn't be part of our Http client.

tests/APIv2/NotificationsApiTest.php
```diff
@@ -32,6 +32,7 @@ private function addNotification(): int {
             "Emailed" => ActivityModel::SENT_PENDING,
             "NotifyUserID" => $this->api()->getUserID(),
             "Notified" => ActivityModel::SENT_PENDING,
+            "activityName" => $this->api()->getActivityName(),
```
charrondev, line 35: Looks like this is always going to be null.

PR #10723, opened 2020-06-08
Add crawl support

tburry:
The PR does quite a bit so let me try and break it down.

1. I added the `ModelFactory` class to create models from record types.
2. I added a `CrawlableInterface` that models can implement to say they support crawling.
3. I use the factory to create new `/resources` endpoint that can be used to inspect record types in the system.
4. I also needed to implement sorting on many of our basic API index endpoints because crawling needs reliable primary key sorting.

The sorting in particular proved to be fairly tedious to implement because we have a lot of inconsistent legacy code around sorting in the models and `GDN_SQLDriver`. To clean this up I did the following.

1. I added support and tests for our various legacy and new sorting to `GDN_SQLDriver::orderBy()`.
2. I added `GDN_SQLDriver::orderByPrefixed()` for those few models that have a join and have a need to prefix sort fields.
3. I added some test utilities to make asserting sorting easier.
4. I added some specific tests to the various sorting methods in the `DiscussionModel`.
5. I added some more specific tests to to the `DiscussionsApiController`.

It was my goal to get all of my changed code as close to 100% covered as possible.

### DIFF
--- a/library/database/class.sqldriver.php
+++ b/library/database/class.sqldriver.php
@@ -662,7 +662,8 @@ abstract class Gdn_SQLDriver {
             $this->limit($limit, $offset);
         }
 
-        $result = $this->query($this->getSelect());
+        $sql = $this->getSelect();
+        $result = $this->query($sql);
         return $result;
     }
 
--- a/library/database/class.sqldriver.php
+++ b/library/database/class.sqldriver.php
@@ -1428,9 +1428,9 @@ abstract class Gdn_SQLDriver {
         } elseif ($value !== null) {
             $this->_Options[$key] = $value;
             return $this;
-        } elseif (isset($this->_Options[$key]))
+        } elseif (isset($this->_Options[$key])) {
             return $this->_Options[$key];
-        else {
+        } else {
             return null;
         }
     }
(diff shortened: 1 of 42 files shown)


library/database/class.sqldriver.php
```diff
@@ -1427,30 +1428,67 @@ public function options($key, $value = null) {
...
+     * Order by a field or fields, adding a prefix to each fields.
+     *
+     * @param string $px
+     * @param string|array|null $fields
+     * @param string $direction
+     * @return $this
+     * @see Gdn_SQLDriver::orderBy()
+     */
+    public function orderByPrefix(string $px, $fields, $direction = 'asc') {
         if (!$fields) {
```
tburry, line 1478: This little refactor in `GDN_SQLDriver` could have been its own PR, but basically I was finding it to be such a hassle wrangling all of our various sorting methods that I decided to make it easier on myself and add support for them here.

tests/APIv2/AbstractAPIv2Test.php
```diff
@@ -144,7 +144,7 @@ protected function pagingTest($resourceUrl) {
         $pagingTestUrl = $resourceUrl.(strpos($resourceUrl, '?') === false ? '?' : '&').'limit=1';
         $resourcePath = parse_url($resourceUrl, PHP_URL_PATH);
 
-        $result = $this->api()->get($pagingTestUrl);
+        $result = $this->api()->get($pagingTestUrl, ['pinOrder' => 'mixed']);
```
tburry, line 147: This is a bit of a kludge, but we know that the discussions endpoints have an optimization for pinned posts that make asserting definitive orders and paging difficult. This setting turns that off for most regular tests. The `DiscussionsTest` then has specific tests for pinned posts.

tests/APIv2/DiscussionsTest.php
```diff
@@ -289,4 +294,36 @@ public function testEmptyDiscussionTitle() {
...
+        $ids = array_column($rows, 'DiscussionID');
+
+        $fields = ['discussionID', '-discussionID'];
+
+        foreach ($fields as $field) {
+            $rows = $this->api()->get($this->baseUrl, ['discussionID' => $ids, 'pinOrder' => 'first', 'sort' => $field])->getBody();
+            $this->assertNotEmpty($rows);
+            $this->assertSorted($rows, '-pinned', $field);
+        }
+    }
```
tburry, line 328: These two tests round off the test coverage of `DiscussionsApiController::index()`.

tests/BootstrapTrait.php
```diff
@@ -142,4 +142,22 @@ protected static function getTestLogger(): TestLogger {
...
+        /** @var \Gdn_Database $database */
+        $database = static::container()->get(\Gdn_Database::class);
+
+        $database->init([
+            'Host' => $model->getDbHost(),
+            'Dbname' => $model->getDbName(),
+            'User' => $model->getDbUser(),
+            'Password' => $model->getDbPassword(),
+        ]);
+    }
```
tburry, line 162: Allow us to do some tests that touch the database without installing the entire site. This has to be called explicitly because it's hard to automate it without having to fix site installs, but I would like to make it automatic in the future.

tburry:
Closing in favor of #10780.

PR #10731, opened 2020-06-08
add a blockexcpetion method to the dispatcher

dannimajdalani:
related https://github.com/vanilla/internal/pull/2608
This add a dispatcher method addBlockException for https://github.com/vanilla/vanilla/blob/885b7dca3759a02c3489209a565898afa40b141f/library/core/class.dispatcher.php#L34


library/core/class.dispatcher.php
```diff
@@ -707,6 +707,19 @@ public function getBlockExceptions() {
...
+     * Adds a dispatcher block exception.
+     *
+     * @param string $exceptionMatch Exception string to match.
+     * @param int $exceptionType Type of block exception.
+     */
+    public function addBlockException(string $exceptionMatch, int $exceptionType) {
+        $allowedExceptionTypes = [self::BLOCK_NEVER, self::BLOCK_PERMISSION, self::BLOCK_ANY];
+        if (in_array($exceptionType, $allowedExceptionTypes)) {
+            $this->blockExceptions[$exceptionMatch] = $exceptionType;
+        }
```
tburry, line 720: Throw an exception if the type is wrong. Don't silently fail.

PR #10733, opened 2020-06-08
add HtmlSnippet class

dannimajdalani:
related https://github.com/vanilla/internal/pull/2608

Add HtmlSnippet utility class.


tests/Library/Vanilla/Utility/HtmlSnippetTest.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Test Dom saveHtml method.
+     */
+    public function testBasicSaveHtml(): void {
+        $html = <<<HTML
+<b>Hello World!.🤓</b>
+HTML;
+        $this->html->loadHtml($html);
+        $result = $this->html->saveHtml();
+        $this->assertSame($html, substr($result, 22, -6));
```
tburry, line 51: This smells. You don't change the test. You fix the code so the test passes as it should.

library/Vanilla/Utility/HtmlSnippet.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Return the html content.
+     *
+     * @param string $htmlString
+     */
+    public function getHtmlContent(string $htmlString): string {
+        $htmlString = preg_replace('#<div id="'.self::CONTENT_ID.'">#', '', $htmlString);
+        $htmlString = preg_replace('#</div>#', '', $htmlString);
```
tburry, line 83: This is better. I also did a quick Google search and Stack Overflow has a [robust looking solution](https://stackoverflow.com/questions/2087103/how-to-get-innerhtml-of-domnode). Always remember to use Google as part of your toolbox.

If you want to go the regex route, your specific example has some bugs. Here would be a suggestion to play with:

```php
$contentID = preg_quote(self::CONTENT_ID);
$htmString = preg_replace('`^<div id="$contentID">(.*)</div>$`', '${1}', $html);
```

library/Vanilla/Utility/HtmlSnippet.php
```diff
@@ -0,0 +1,86 @@
...
+     * @param null $options
+     * @return string
+     */
+    public function saveXML(DOMNode $node = null, $options = null) {
+        if ($node === null) {
+            $content = $this->getElementById(self::CONTENT_ID);
+            $htmlBodyString = $this->saveXML($content, LIBXML_NOEMPTYTAG);
+            $htmlBodyString = $this->getHtmlContent($htmlBodyString);
+        } else {
+            $htmlBodyString = parent::saveXML($node, $options);
```
tburry, line 51: This seems wrong. Don't you only need to strip the content div if `$node === null`?

tburry:
@charrondev, this was at my direction.

PR #10740, opened 2020-06-09
Fix encoding on file names.

nervoustwit:
This PR will close vanilla/support#1962

In Rich editor we use actual names of files for images embedded on sites which is great for SEO, etc. Unfortunately when image that have "foreign" characters in their names they are saved URL encoded. But when you fetch them the URLs are decoded, so that they are not found.

This PR will double encode the saved URL so that when we fetch the image from storage we have the correct address.

### Testing
 1. Turn on Rich editor.
 2. Re-name an image on your local computer to 'pic-我的照片.jpg'
 3. Upload it in a post.
 4. See that it is not found.
 5. Check out this branch.
 6. Upload the same image.
 7. Make sure English named images still work.
 8. Test on other editors.
 9. Approve PR.

### DIFF
--- a/library/Vanilla/UploadedFile.php
+++ b/library/Vanilla/UploadedFile.php
@@ -196,7 +196,7 @@ class UploadedFile {
         $ext = strtolower(pathinfo($this->getClientFilename(), PATHINFO_EXTENSION));
         $baseName = basename($this->getClientFilename(), ".${ext}");
         $baseName = sprintf($nameFormat, $baseName);
-        $baseName = \Gdn_Format::url($baseName);
+        $baseName = \Gdn_Format::url(urlencode($baseName));
         $uploadPath = FileUtils::generateUniqueUploadPath($ext, true, $baseName, $persistDirectory);
         return $uploadPath;
     }
--- a/tests/Library/Vanilla/UploadedFileTest.php
+++ b/tests/Library/Vanilla/UploadedFileTest.php
@@ -102,7 +102,6 @@ class UploadedFileTest extends SharedBootstrapTestCase {
 
         // Save the upload.
         $file->persistUpload();
-        $f = $file->getPersistedPath();
         $this->assertFileExists(PATH_UPLOADS.'/'.$file->getPersistedPath(), 'Final upload file is persisted');
         $this->assertStringContainsString('my-25e5-259c-2596-25e7-2589-2587.png', $file->getPersistedPath());
     }
--- a/tests/Library/Vanilla/UploadedFileTest.php
+++ b/tests/Library/Vanilla/UploadedFileTest.php
@@ -90,12 +90,39 @@ class UploadedFileTest extends SharedBootstrapTestCase {
     }
 
     /**
+     * Test that we can safely persist files with url_encoded characters in their name.
+     *
+     * @param string $name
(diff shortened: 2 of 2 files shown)


library/Vanilla/UploadedFile.php
```diff
@@ -196,7 +196,7 @@ public function generatePersistedUploadPath(string $persistSubDirectory = '', st
         $ext = strtolower(pathinfo($this->getClientFilename(), PATHINFO_EXTENSION));
         $baseName = basename($this->getClientFilename(), ".${ext}");
         $baseName = sprintf($nameFormat, $baseName);
-        $baseName = \Gdn_Format::url($baseName);
+        $baseName = \Gdn_Format::url(urlencode($baseName));
```
tburry, line 199: I think I see that this is just working by coincidence, but will ultimately create URLs that are garbled in many cases. Consider a file with the name `"cät"`. That should give us a file with the name `"cat"`. But your changes will give us something like `"c-c3-a4t"`. Can we not remove the call to `Gdn_Format::url()` and just go with `rawurlencode()`?

tburry:
@alex-mtl, I had heard you have an approach to this? Can you weigh in here?

alex-mtl:
I've tested this solution. It looks better then mine I was trying to implement outside the function `generatePersistedUploadPath`.

This solution works for my issue and fixes it. Same time all the rest seem still functional and nothing is broken.

I've tested new article to drop a file with "bad" name:
```
cät ? & () really bad | name.png
```

And it still works fine with new implementation. And it fails on demo site (old release with no patch).

New url generated for my localhost:
```
https://dev.vanilla.localhost/uploads/219/ca-cc-88t-3f-26-28-29-really-bad-7c-name.png
# But full html element is good

<img class="embedImage-img" src="https://dev.vanilla.localhost/uploads/219/ca-cc-88t-3f-26-28-29-really-bad-7c-name.png" alt="cät ? &amp; () really bad | name.png">
```


Yes url becomes less readable. But I think that is the best approach we can apply for now without increasing the scope.

tests/Library/Vanilla/UploadedFileTest.php
```diff
@@ -89,6 +89,23 @@ public function testPersistFileWithSpaces(string $name) {
...
+     * @dataProvider provideImagesEncodedChars
+     */
+    public function testPersistFileEncodedChars(string $name) {
+        // Perform some tests related to saving uploads.
+        $file = UploadedFile::fromRemoteResourceUrl($name);
+
+        // Save the upload.
+        $file->persistUpload();
+        $this->assertFileExists(PATH_UPLOADS.'/'.$file->getPersistedPath(), 'Final upload file is persisted');
+        $this->assertStringContainsString('my-25e5-259c-2596-25e7-2589-2587.png', $file->getPersistedPath());
```
initvector, line 106: This test seems weirdly split. It looks like it's setup for a provider, but the assertion is hard coded for a particular expected value. If it can't be more dynamic, it probably shouldn't use a provider. However, if you can figure out how to make it _really_ work with a provider, it'll be a much more valuable test.

tburry:
Okay there are three approvals now. I've said my piece, but I guess we can merge this then.

PR #10748, opened 2020-06-10
Fix urls splitting on braces in BBCode

blutarch:
Closes vanilla/support#1930

While using the Advanced Editor's BBCode option, you cannot post a url that includes braces. This PR allows a url with braces as long as it occurs after the `//` in an address.

#### Before checking out branch
1. Enable Advanced Editor and select the BBCode posting option.
1. Post this url in a discussion: 
`https://www.example.com/this{has-braces}`
1. Verify that the link ends right before the opening brace

#### After checking out branch
1. Repeat steps 1-2 above.
1. Verify that the whole url is a link

### DIFF
--- a/tests/Library/Vanilla/Formatting/GdnFormatLinksTest.php
+++ b/tests/Library/Vanilla/Formatting/GdnFormatLinksTest.php
@@ -139,6 +139,40 @@ HTML;
     }
 
     /**
+     * Test that braces are allowed after the '//', but not before.
+     *
+     * @param string $input String with braces to test.
+     * @param string $expected Expected output.
+     * @dataProvider provideTestBracesData
+     */
+    public function testBraces($input, $expected) {
+        $actual = Gdn_Format::links($input);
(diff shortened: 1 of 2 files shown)


tests/Library/Vanilla/Formatting/GdnFormatLinksTest.php
```diff
@@ -138,6 +138,40 @@ public function provideTestCommasData(): array {
...
+                'h{tt}p://www.foo.bar',
+            ],
+            'bracesBetweenSlashes' => [
+                'http:/{/www.foo.bar',
+                'http:/{/www.foo.bar',
+            ],
+            'bracesAfterSlashes' => [
+                'http://www.{foo}.bar',
+                '<a href="http://www.{foo}.bar" rel="nofollow">http://www.{foo}.bar</a>'
+            ],
```
tburry, line 169: Can you add a test for the URL from the ticket? Maybe change the domain to example.com though.

PR #10759, opened 2020-06-11
Add tests and fixes around count calculation

tburry:


tburry:
Closed in favor of #10795.

PR #10761, opened 2020-06-11
add DomUtilsClass

dannimajdalani:
closes https://github.com/vanilla/internal/issues/2680
related https://github.com/vanilla/internal/pull/2690
@tburry This is the PR.


library/Vanilla/Formatting/Html/DomUtils.php
```diff
@@ -0,0 +1,98 @@
...
+
+use DOMDocument;
+
+/**
+ * Class for stripping images and truncating text from a Dom.
+ */
+final class DomUtils {
+
+    /** @var array */
+    const EMBED_CLASSES = ['js-embed', 'embedResponsive', 'embedExternal', 'embedImage', 'VideoWrap', 'iframe'];
```
tburry, line 19: Our version of PHP allows constant visibility to be declared now.

library/Vanilla/Formatting/Html/DomUtils.php
```diff
@@ -0,0 +1,98 @@
...
+        $xpath = new \DomXPath($dom);
+        foreach ($embedClasses as $key => $value) {
+            $xpathQuery = $xpath->query(".//*[contains(@class, '$embedClasses[$key]')]");
+            $xpathDivQuery = $xpath->query("//div[@data-embedjson]");
+            $dataClassItem = $xpathQuery->item(0);
+            $dataDivItem = $xpathDivQuery->item(0);
+            if ($dataClassItem) {
+                $dataClassItem->parentNode->removeChild($dataClassItem);
+            } elseif ($dataDivItem) {
+                $dataDivItem->parentNode->removeChild($dataDivItem);
```
tburry, line 37: If you install the code coverage chrome extensions, you'll see that this line of code wasn't hit by a test. Can you conceive of a test that hits this?

library/Vanilla/Formatting/Html/DomUtils.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Remove embeds from the dom.
+     *
+     * @param DOMDocument $dom
+     * @param array $embedClasses
+     */
+    public static function stripEmbeds(DOMDocument $dom, array $embedClasses = self::EMBED_CLASSES): void {
+        $xpath = new \DomXPath($dom);
+        foreach ($embedClasses as $key => $value) {
+            $xpathQuery = $xpath->query(".//*[contains(@class, '$embedClasses[$key]')]");
```
tburry, line 30: What about `iframe` and `video` tags?

library/Vanilla/Formatting/Html/DomUtils.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param mixed $element Dom element.
+     * @param int $wordCounter Counter of number of word to trucnate to.
+     * @param int $wordCount Number of words to truncate to.
+     * @return int Return limit used to count remaining tags.
+     */
+    private static function truncateWordsRecursive($element, int $wordCounter, int $wordCount): int {
+        if ($wordCounter > 0) {
+            // Nodetype text
+            if ($element->nodeType == 3) {
```
tburry, line 80: Avoid magic numbers when possible. It looks like there are [constants](https://www.php.net/manual/en/dom.constants.php) for these.

tests/Library/Vanilla/Utility/DomUtilsTest.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+/**
+ * @author Dani M. <dani.m@vanillaforums.com>
+ * @author Isis Graziatto <isis.g@vanillaforums.com>
+ * @copyright 2009-2020 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace VanillaTests\Library\Vanilla\Utility;
```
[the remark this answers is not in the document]
tburry: Resolve conversations when you've read them so that I know that you've dealt with them.

library/Vanilla/Formatting/Html/DomUtils.php
dannimajdalani: Very useful, that was the case of uploading a file in rich-editor which we haven't added a test case for yet. Isis will be adding that.

library/Vanilla/Formatting/Html/DomUtils.php
igraziatto: Dani and I took a look and we couldn't find any cases where `iframe` or `video` tags are actually embedded in the discussions. It's all `span` and `div` tags with specific CSS classes flagging that those elements are embeds.

tburry: Right, but as a library function don't you think those should be stripped? Many of our old embeds to embed iframes I do believe.

dannimajdalani: an iframe tag gets stripped when passed Gdn::formatService()->renderHTML()
As for video, i'll have to add that.

dannimajdalani: I've added a check for video tag and a test for it.

library/Vanilla/Formatting/Html/DomUtils.php
```diff
@@ -28,13 +28,13 @@ public static function stripEmbeds(DOMDocument $dom, array $embedClasses = self:
         $xpath = new \DomXPath($dom);
         foreach ($embedClasses as $key => $value) {
             $xpathQuery = $xpath->query(".//*[contains(@class, '$embedClasses[$key]')]");
-            $xpathDivQuery = $xpath->query("//div[@data-embedjson]");
+            $xpathTagsQuery = $xpath->query("//div[@data-embedjson] | //video");
```
tburry, line 31: Can you add `iframe`?

dannimajdalani: added.

PR #10763, opened 2020-06-11
Release optimizations

tburry:
These are some small optimizations that are hopefully okay to do this close to a release.

- There is an every page component that was querying without an available cached endpoint.
- Models were not shared which would cause duplicates to query for their schema info multiple times.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -11,6 +11,7 @@ use Vanilla\InjectableInterface;
 use Vanilla\Contracts;
 use Vanilla\Models\CurrentUserPreloadProvider;
 use Vanilla\Models\LocalePreloadProvider;
+use Vanilla\Models\Model;
 use Vanilla\Permissions;
 use Vanilla\Site\SingleSiteSectionProvider;
 use Vanilla\Theme\ThemeFeatures;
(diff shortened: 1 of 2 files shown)


bootstrap.php
```diff
@@ -318,6 +318,9 @@
     ->rule('Gdn_Model')
     ->setShared(true)
 
+    ->rule(Model::class)
+    ->setShared(true)
```
charrondev, line 322:
```suggestion
    ->rule(Model::class)
    ->setInherit(true)
    ->setShared(true)
```

Does this actually do anything without a `setInherit`? All of our model instances are subclasses, and we never pull them out of the container as just `Model`.

tburry: I had checked the code and I think inherit is the default:

```
// Don't add the rule if it doesn't say to inherit.
if (!isset($this->rules[$class]) || (isset($this->rules[$class]['inherit']) && !$this->rules[$class]['inherit'])) {
    continue;
}
```

tburry: I'll confirm with a debug session too. One sec...

tburry: Okay. I confirmed that inherit works, but there was a bug with the rule. I didn't import the namespace.

PR #10764, opened 2020-06-12
Add a model with full record caching

charrondev:
Related https://github.com/vanilla/internal/issues/2691

Implement a core model and some utilities that implement full record caching.

To help lower the risk (because this will likely need to be backported for performance considerations), I've added a feature flag to disable it.

`Feature.DisableNewModelCaching.Enabled = true`

## Pieces

- A pipeline processor for invalidating.
- Nice utility methods
  - `getCachedOrHydrate(array $cacheArgs, callable $hydrator)` -> Get a cached value or hydrate it.
  - `invalidateAll()` -> Invalidate the whole cache for the model.
- A extra convienience model base class anything that passes through the pipeline.

## Invalidation

- Items could be manually invalidated with the invalidate method.
- Invalidation happens by rolling over an incrementing ID thats used in every cache key.
- Pipeline invalidation occurs on any insert, delete, or update.

## Other notes

- The `PipelineProcessor` did not run the pipeline for delete operations. Now it does.

## Example usages

Initial PRs
- https://github.com/vanilla/multisite/pull/384
- https://github.com/vanilla/knowledge/pull/1868

### DIFF
--- a/library/Vanilla/Models/FullRecordCacheModel.php
+++ b/library/Vanilla/Models/FullRecordCacheModel.php
@@ -20,11 +20,12 @@ class FullRecordCacheModel extends PipelineModel {
      * Model constructor.
      *
      * @param string $table Database table associated with this resource.
+     * @param \Gdn_Cache $cache The cache instance.
      * @param array|null $defaultCacheOptions Default options to apply for storing cache values.
      */
-    public function __construct(string $table, ?array $defaultCacheOptions = []) {
+    public function __construct(string $table, \Gdn_Cache $cache, ?array $defaultCacheOptions = []) {
         parent::__construct($table);
-        $this->modelCache = ModelCache::fromModel($this, $defaultCacheOptions);
+        $this->modelCache = new ModelCache($this->getTable(), $cache, $defaultCacheOptions);
         $this->addPipelineProcessor($this->modelCache->createInvalidationProcessor());
     }
 
--- a/library/Vanilla/Models/FullRecordCacheModel.php
+++ b/library/Vanilla/Models/FullRecordCacheModel.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * @copyright 2009-2019 Vanilla Forums Inc.
+ * @license Proprietary
+ */
+
+namespace Vanilla\Models;
+
+/**
+ * A model that caches all queries that come through, and invalidates the cache on every change.
+ *
+ * Ideal for DB tables that are infrequently updated and have smaller record sets.
+ */
+class FullRecordCacheModel extends PipelineModel {
+
+    /** @var ModelCache $cache */
+    private $modelCache;
+
+    /**
+     * Model constructor.
+     *
+     * @param string $table Database table associated with this resource.
+     * @param array|null $defaultCacheOptions Default options to apply for storing cache values.
+     */
+    public function __construct(string $table, ?array $defaultCacheOptions = []) {
+        parent::__construct($table);
+        $this->modelCache = ModelCache::fromModel($this, $defaultCacheOptions);
+        $this->addPipelineProcessor($this->modelCache->createInvalidationProcessor());
+    }
+
+    /**
+     * Get resource rows from the cache or a database table.
+     *
+     * @param array $where Conditions for the select query.
+     * @param array $options Options for the select query.
+     *    - orderFields (string, array): Fields to sort the result by.
+     *    - orderDirection (string): Sort direction for the order fields.
+     *    - limit (int): Limit on the total results returned.
+     *    - offset (int): Row offset before capturing the result.
+     * @param array|null $cacheOptions Options for the cache storage.
+     * @return array Rows matching the conditions and within the parameters specified in the options.
+     */
+    public function get(array $where = [], array $options = [], ?array $cacheOptions = []): array {
+        return $this->modelCache->getCachedOrHydrate(
+            [
+                'where' => $where,
+                'options' => $options,
+            ],
+            function () use ($where, $options) {
+                return parent::get($where, $options);
+            }
+        );
+    }
+
+    /**
+     * Get all records.
+     */
+    public function getAll() {
+        return $this->get();
+    }
+
+    /**
+     * @return ModelCache
+     */
+    public function getModelCache(): ModelCache {
+        return $this->modelCache;
+    }
+}
--- a/library/Vanilla/Models/FullRecordCacheModel.php
+++ b/library/Vanilla/Models/FullRecordCacheModel.php
@@ -45,6 +45,7 @@ class FullRecordCacheModel extends PipelineModel {
             [
                 'where' => $where,
                 'options' => $options,
+                'function' => __FUNCTION__, // For uniqueness.
             ],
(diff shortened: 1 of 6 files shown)


library/Vanilla/Models/ModelCache.php
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Create a cache key for some parameters.
+     *
+     * @param array $keyArgs Some arguments to generate the cache key from.
+     *
+     * @return string
+     */
+    public function createCacheKey(array $keyArgs): string {
+        $key = $this->namespace . '-' . $this->getIncrementingKey() . '-' . md5(json_encode($keyArgs));
```
tburry, line 89:
```suggestion
        $key = $this->namespace . '-' . $this->getIncrementingKey() . '-' . sha1(json_encode($keyArgs));
```

library/Vanilla/Models/ModelCache.php
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Create a cache from a model.
+     *
+     * @param Model $model The model to create from.
+     * @param array|null $defaultCacheOptions Default options to apply for storing cache values.
+     *
+     * @return ModelCache
+     */
+    public static function fromModel(Model $model, ?array $defaultCacheOptions = []): ModelCache {
+        $cache = \Gdn::getContainer()->getArgs(ModelCache::class, [$model->getTable(), $defaultCacheOptions]);
```
tburry, line 56: This is not a long term solution. Is it possible to make this class a trait instead in order to not call out to the container?

library/Vanilla/Models/FullRecordCacheModel.php
```diff
@@ -0,0 +1,68 @@
...
+     *    - orderDirection (string): Sort direction for the order fields.
+     *    - limit (int): Limit on the total results returned.
+     *    - offset (int): Row offset before capturing the result.
+     * @param array|null $cacheOptions Options for the cache storage.
+     * @return array Rows matching the conditions and within the parameters specified in the options.
+     */
+    public function get(array $where = [], array $options = [], ?array $cacheOptions = []): array {
+        return $this->modelCache->getCachedOrHydrate(
+            [
+                'where' => $where,
```
tburry, line 46: My concern here is with an example you shared with me. You wanted to be able to add an aggregate method that was different with get. Isn't it possible to have two methods with the same where clause? Maybe you could add `__METHOD__` or something like that to the cache key.

library/Vanilla/Models/ModelCache.php
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Constructor.
+     *
+     * @param string $namespace
+     * @param array|null $defaultCacheOptions Default options to apply for storing cache values.
+     */
+    public function __construct(string $namespace, ?array $defaultCacheOptions = []) {
+        $this->namespace = $namespace;
+        $this->defaultCacheOptions = array_merge(self::GLOBAL_DEFAULT_OPTIONS, $defaultCacheOptions ?? []);
+        $this->isFeatureDisabled = \Gdn::config(self::DISABLE_FEATURE_FLAG);
```
tburry, line 69: I'm okay with a global call since this is just a feature flag. However, please use the `FeatureFlagHelper` methods.

charrondev: Done.

library/Vanilla/Models/FullRecordCacheModel.php
charrondev: Done.

library/Vanilla/Models/ModelCache.php
charrondev: Done.

library/Vanilla/Models/ModelCache.php
charrondev: Changed.

library/Vanilla/Models/FullRecordCacheModel.php
```diff
@@ -0,0 +1,70 @@
...
+    protected $modelCache;
+
+    /**
+     * Model constructor.
+     *
+     * @param string $table Database table associated with this resource.
+     * @param \Gdn_Cache $cache The cache instance.
+     * @param array|null $defaultCacheOptions Default options to apply for storing cache values.
+     */
+    public function __construct(string $table, \Gdn_Cache $cache, ?array $defaultCacheOptions = []) {
```
tburry, line 26: Do you think this might be easier with `array` instead of `?array`?

library/Vanilla/Models/FullRecordCacheModel.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param array $where Conditions for the select query.
+     * @param array $options Options for the select query.
+     *    - orderFields (string, array): Fields to sort the result by.
+     *    - orderDirection (string): Sort direction for the order fields.
+     *    - limit (int): Limit on the total results returned.
+     *    - offset (int): Row offset before capturing the result.
+     * @param array|null $cacheOptions Options for the cache storage.
+     * @return array Rows matching the conditions and within the parameters specified in the options.
+     */
+    public function get(array $where = [], array $options = [], ?array $cacheOptions = []): array {
```
tburry, line 44: This signature doesn't match the base class. Can you make the cache options part of the `$options` array?

library/Vanilla/Models/FullRecordCacheModel.php
```diff
@@ -0,0 +1,70 @@
...
+     * Get all records.
+     */
+    public function getAll() {
+        return $this->get();
+    }
+
+    /**
+     * @return ModelCache
+     */
+    public function getModelCache(): ModelCache {
```
tburry, line 67: Can this be protected?

library/Vanilla/Models/ModelCache.php
```diff
@@ -0,0 +1,170 @@
...
+     * @return ModelCacheInvalidationProcessor
+     */
+    public function createInvalidationProcessor(): ModelCacheInvalidationProcessor {
+        return new ModelCacheInvalidationProcessor($this);
+    }
+
+    /**
+     * @param \Gdn_Cache $cache
+     */
+    public function setCache(\Gdn_Cache $cache): void {
```
tburry, line 122: Call this in the constructor and you get your lines covered ;)

library/Vanilla/Models/FullRecordCacheModel.php
charrondev: Done.

library/Vanilla/Models/FullRecordCacheModel.php
charrondev: Done.

library/Vanilla/Models/FullRecordCacheModel.php
charrondev: I actually just removed it. It was not longer used since I update to put the cache dependency in the constructor.

library/Vanilla/Models/ModelCache.php
charrondev: Done.

PR #10765, opened 2020-06-12
Remove the fuzzy category watch from the discussion model

tburry:
The discussion model has some logic to read a select set of discussions in order to try and mark the category read. I’m not super sure this has ever really worked, but is troubling from a scalability point of view.

This PR puts that logic behind a feature flag so that we can try a deploy with it turned off, but flip it back on if there really is some real functionality required.

### DIFF
--- a/applications/vanilla/models/class.discussionmodel.php
+++ b/applications/vanilla/models/class.discussionmodel.php
@@ -3880,6 +3880,7 @@ class DiscussionModel extends Gdn_Model implements FormatFieldInterface, EventFr
      *
      * @param object $discussion
      * @deprecated
+     * @codeCoverageIgnore
      */
     protected function markCategoryReadFuzzy($discussion): void {
         /**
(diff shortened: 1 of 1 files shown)


charrondev:
A little context.

- This existed for a while, but was a bit buggy.
- https://github.com/vanilla/vanilla/issues/6856 was filed.
- This PR https://github.com/vanilla/vanilla/pull/8574 was introduced to fix the issue.
- It was merged and went out in 2019.006.

tburry:
It's just a lot of querying for every page hit to the discussion endpoint.

bleistivt:
This may apply to very large forums where no user can realistically read every new discussion, but I know that there are plenty of users on small to mid-sized forums who read *everything*.

If the forum software doesn't automatically track the read status on the category level, I don't see any advantage over using tags and may even have to ask myself why I am using categories at all.

initvector:
> If the forum software doesn't automatically track the read status on the category level, I don't see any advantage over using tags and may even have to ask myself why I am using categories at all.

We're not planning to outright removing the feature at this time. This update is only moving it behind a feature flag as an experiment. Sites can still enable the feature from the config.

Issue #10751, opened 2020-06-11
When I create a Event in root category, the application is crashing.


GeetikaMuthyam:
**Steps to reproduce:**

1. Go to root page.
2. Click on the 'NewDiscussions' dropdown and select "New Event".
3. Then fill all the mandatory fields and create an event and save.
4. Then, navigate to root page, i see the error below.

**Screenshot**
![Screen Shot 2020-06-10 at 12.03.55 **PM.png](https://images.zenhubusercontent.com/5d9e3755aba55400015a6b2a/986bc542-02d4-43be-b2a0-6d5c47216798)**

**ScreenRecording**
https://recordit.co/E2kVpuKDX3

PR #10775, opened 2020-06-12
Fix events created in root category failing validation

cchabilall83:
When expanding a the root category we were assigning the name of the category to the value in config for `$Configuration['Garden']['Title'] `.  If that was empty it would return an empty string which would fail validation on the expand param in the \events endpoint.  This fallback to the value set in our config-defaults which is `Vanilla`, if the config value is empty.

closes https://github.com/vanilla/vanilla/issues/10751


tburry:
Closing in favor of #10780.

PR #10780, opened 2020-06-13
Add crawl support (REBASED)

tburry:
The PR does quite a bit so let me try and break it down.

1. I added the `ModelFactory` class to create models from record types.
2. I added a `CrawlableInterface` that models can implement to say they support crawling.
3. I use the factory to create new `/resources` endpoint that can be used to inspect record types in the system.
4. I also needed to implement sorting on many of our basic API index endpoints because crawling needs reliable primary key sorting.

The sorting in particular proved to be fairly tedious to implement because we have a lot of inconsistent legacy code around sorting in the models and `GDN_SQLDriver`. To clean this up I did the following.

1. I added support and tests for our various legacy and new sorting to `GDN_SQLDriver::orderBy()`.
2. I added `GDN_SQLDriver::orderByPrefixed()` for those few models that have a join and have a need to prefix sort fields.
3. I added some test utilities to make asserting sorting easier.
4. I added some specific tests to the various sorting methods in the `DiscussionModel`.
5. I added some more specific tests to to the `DiscussionsApiController`.

It was my goal to get all of my changed code as close to 100% covered as possible.

### DIFF
--- a/applications/dashboard/controllers/api/ResourcesApiController.php
+++ b/applications/dashboard/controllers/api/ResourcesApiController.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * @author Todd Burry <todd@vanillaforums.com>
+ * @copyright 2009-2020 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace Vanilla\Dashboard\Controllers\API;
+
+use Garden\Schema\Schema;
+use Garden\Web\Data;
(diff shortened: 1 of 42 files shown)


tburry:
@charrondev, my reasoning for the expand is the following:

1. It's more of a forwards compatible thing. I would like to add more to that endpoint, but the record type does have some use.
2. The crawl does perform some queries so that it could be a resource hit you don't always want.
3. Easier to make it expand later than to remove it later.

PR #10781, opened 2020-06-14
Add PSR-16 simple cache support

tburry:
This PR adds suport for PSR-16's `CacheInterface` in core. It uses an adapter from `Gdn_Cache`. Although this PR is fully tested against PSR-16, it did require some bug fixes in some core classes which could cause regressions. Here is a list:

- The increment and decrement functions had inconsistent and buggy behavior when there are no items. I added the fix to test the initial values. There are some places that work around this bug and thus depend on it. I changed the code where these exist. I ended up rolling back the changes I made to incrementing EXCEPT on the `Gdn_DirtyCache` class. Instead I made it match the behavior of the `Gdn_Memcached` class. I added a test to ensure the default behavior is respected and added a decent doc block to explain that fact.
- I had to add the `Gdn_Cache::FEATURE_DEFAULT` option in order to properly adapt to simple cache.

I founder a composer library that has a pretty thorough base class for cache tests, so the tests in this PR have pretty decent coverage.

Note that the specific memcached tests don't currently run on CI. You can run them locally, but need to set an environment variable. See the tests/README.md file in this PR.

Closes https://github.com/vanilla/vanilla/issues/9519.


library/Vanilla/Cache/CacheCacheAdapter.php
```diff
@@ -0,0 +1,141 @@
...
+use Gdn_Cache;
+use Psr\SimpleCache\CacheInterface;
+
+/**
+ * An adapter from `Gdn_Cache` to `Psr\SimpleCache\CacheInterface`.
+ *
+ * This class adapts the functionality, but not all of the validation of the `CacheInterface`. If you want to validate
+ * the keys going into and out of the cache then use the `ValidatingCacheCacheAdapter` class.
+ */
+class CacheCacheAdapter implements CacheInterface {
```
tburry, line 19: This was essentially extracted from the same class at the root namespace.

library/core/class.cache.php
```diff
@@ -23,6 +23,11 @@ abstract class Gdn_Cache {
     /** @var string Type of cache this this: one of CACHE_TYPE_MEMORY, CACHE_TYPE_FILE, CACHE_TYPE_NULL. */
     protected $cacheType;
 
+    /**
+     * @var null The current cache config.
+     */
+    private $activeConfig = null;
```
tburry, line 29: This was a static variable of a method below. I extracted it to a property for easier testing.

PR #10787, opened 2020-06-15
Fix location data missing when exporting profiles

blutarch:
Closes vanilla/support#2015.

Currently, if you have a profile extender field named "location," the location data that is stored in the `Gdn_User` table is overwritten by the query for location data in the `Gdn_UserMeta` table. The problem is that the location data is never stored in the `Gdn_UserMeta` table in the first place, so the location field will always be empty. This PR fixes the problem by skipping the `Gdn_UserMeta` query if the field is named "location".

### TO TEST
1. Enable Profile extender and add a "Location" field.
1. Create or edit a user to give their profile a location.
1. Go to the `utility/exportprofiles` endpoint and press enter.
1. Look at the .csv file and note that there is no location data for that user.
1. Check out this branch and repeat step 3.
1. Look at the .csv file and verify that the first location field is populated for that user.

### DIFF
--- a/plugins/ProfileExtender/class.profileextender.plugin.php
+++ b/plugins/ProfileExtender/class.profileextender.plugin.php
@@ -582,14 +582,14 @@ class ProfileExtenderPlugin extends Gdn_Plugin {
                 continue;
             }
             // Add this field to the output
-                $columnNames[] = val('Label', $fieldData, $slug);
+            $columnNames[] = val('Label', $fieldData, $slug);
 
-                // Add this field to the query.
-                $quoted = Gdn::sql()->quote("Profile.$slug");
-                    Gdn::sql()
-                        ->join('UserMeta a' . $i, "u.UserID = a$i.UserID and a$i.Name = $quoted", 'left')
-                        ->select('a' . $i . '.Value', '', $slug);
-                    $i++;
+            // Add this field to the query.
+            $quoted = Gdn::sql()->quote("Profile.$slug");
+            Gdn::sql()
+                ->join('UserMeta a' . $i, "u.UserID = a$i.UserID and a$i.Name = $quoted", 'left')
+                ->select('a' . $i . '.Value', '', $slug);
+            $i++;
         }
 
         // Get our user data.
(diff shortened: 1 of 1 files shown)


plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -576,15 +576,19 @@ public function utilityController_exportProfiles_create($sender) {
 
         $i = 0;
         foreach ($fields as $slug => $fieldData) {
+            // Don't overwrite Location data (it's always stored in the user table, never the userMeta table).
+            if (strtolower($slug) === 'location') {
```
tburry, line 580: I think you would be better to check against the `$columnNames` variable where `"Location"` is added above and then you will prevent bugs for any column that has already been added.

tburry: Make sure to use a similar case-insensitive comparison.

blutarch: It's worth noting that this still may not work for all column names because of some other behaviors --  for instance, creating a profile extender field called "Points" automatically ends up creating a "Points1" field, so you end up getting this: `$Configuration['ProfileExtender']['Fields']['Points1']['Label'] = 'Points';`. So that field will show up twice, as in the screenshot below. There may be other instances of this (although I think it's unlikely these will create problems). In any case, that's why I went with the narrower fix earlier.

image

plugins/ProfileExtender/class.profileextender.plugin.php
```diff
@@ -574,17 +574,22 @@ public function utilityController_exportProfiles_create($sender) {
...
             // Add this field to the output
-            $columnNames[] = val('Label', $fieldData, $slug);
-
-            // Add this field to the query.
-            $quoted = Gdn::sql()->quote("Profile.$slug");
-            Gdn::sql()
-                ->join('UserMeta a'.$i, "u.UserID = a$i.UserID and a$i.Name = $quoted", 'left')
-                ->select('a'.$i.'.Value', '', $slug);
-            $i++;
+                $columnNames[] = val('Label', $fieldData, $slug);
```
tburry, line 585: Your indentation here seems all messed up. Any reason for this?

blutarch: > Your indentation here seems all messed up. Any reason for this?

I'm not sure if this is a sincere question or not. If it is, the answer is that it was simply an oversight. I believe it's fixed now.

PR #10796, opened 2020-06-16
Add convenience method for replacing shared instances in the container

initvector:
This update adds a `replace` method to `ContainerUtils` that performs the following:

1. Replaces a shared instance of a class, if available.
1. Optionally adds an alias from the original class to the replacement class.

The method is used to replace the introduction of the log decorator in the bootstrap.

The `ContainerUtils` class has been cleaned up a little to satisfy my IDE.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -604,9 +604,11 @@ register_shutdown_function(function () use ($dic) {
     $dic->get(\Garden\EventManager::class)->fire('SchedulerDispatch');
 });
 
-// Construct the logger earlier so that its dependencies are satisfied.
-$log = $dic->get(\Vanilla\Logging\LogDecorator::class);
-// Replace the logger interface with the decorator.
-$dic->rule(\Vanilla\Logging\LogDecorator::class)
-    ->addAlias(\Psr\Log\LoggerInterface::class);
+// Add the log decorator.
+ContainerUtils::replace(
+    $dic,
+    \Psr\Log\LoggerInterface::class,
+    \Vanilla\Logging\LogDecorator::class,
+    true
+);
 Logger::setLogger(null);
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -608,7 +608,6 @@ register_shutdown_function(function () use ($dic) {
 ContainerUtils::replace(
     $dic,
     \Psr\Log\LoggerInterface::class,
-    \Vanilla\Logging\LogDecorator::class,
-    true
+    \Vanilla\Logging\LogDecorator::class
 );
 Logger::setLogger(null);
--- a/library/Vanilla/Utility/ContainerUtils.php
+++ b/library/Vanilla/Utility/ContainerUtils.php
@@ -88,8 +88,7 @@ class ContainerUtils {
      */
     public static function replace(Container $container, string $target, string $replacement): void {
         if ($container->hasInstance($target)) {
-            $new = $container->get($replacement);
-            $container->setInstance($target, $new);
+            $container->setInstance($target, null);
         }
 
         $container->rule($replacement)
--- a/library/Vanilla/Utility/ContainerUtils.php
+++ b/library/Vanilla/Utility/ContainerUtils.php
@@ -75,7 +75,7 @@ class ContainerUtils {
 
     /**
      * Replace one type of object in the container with another type of object. Existing shared instances will be
-     * overwritten. Optionally, an alias may be created from the original type to the new type.
+     * overwritten. An alias will be created from the original type to the new type.
      *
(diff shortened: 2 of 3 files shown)


library/Vanilla/Utility/ContainerUtils.php
```diff
@@ -67,9 +67,33 @@ public static function currentTheme(): ReferenceInterface {
...
+                $cacheBuster = $dic->get(DeploymentCacheBuster::class);
                 return $cacheBuster->value();
             }
         );
     }
+
+    /**
+     * Replace container instance of a class, if present, and optionally add it as an alias for its replacement.
+     *
+     * @param Container $container
```
tburry, line 79: I'm trying to get us in the habit of doing a bit more documentation in utility function doc blocks. This is a good candidate because it's not immediately obvious to me what this method does. In particular the `$target` and `$replacement` parameters should get a good sentence to say that they are rules (class name, alias, or custom rule).

tburry: An example in the doc block that explains the problem it solves too.

library/Vanilla/Utility/ContainerUtils.php
```diff
@@ -67,9 +67,39 @@ public static function currentTheme(): ReferenceInterface {
...
+     * @param Container $container Container to configure.
+     * @param string $target Container rule to target for replacement. Shared instances will be overwritten.
+     * @param string $replacement Container rule used to determine what will be replace the target in the container.
+     * @param bool $addAlias Should an alias from the target rule be made to the replacement?
+     */
+    public static function replace(
+        Container $container,
+        string $target,
+        string $replacement,
+        bool $addAlias = false
```
tburry, line 94: I would say this should default to true or not be a parameter at all. Otherwise, if there isn't really a replacement happening and the method has inconsistent behavior depending on whether or not the container has the instance.

IMO you think about the following:

1. Go for it and get rid of `$addAlias` and just make this the behavior.
2. Just remove the instance if it is there and add a rule. Let the container do the creation when it is called. This one may be a bit controversial since we've had problem with the log in particular getting injected and called before it is instantiated. On the other hand, the caller could always do the instantiation themself. If they don't want to instantiate then they are left with having to write the code themselves.

initvector: > Go for it and get rid of $addAlias and just make this the behavior.

I was waffling on this. There might be a scenario where you'd only want to replace the current instance, but we don't have that need now and I can't think of a concrete scenario where it'd be warranted. I'm A-OK with removing the parameter.

> Just remove the instance if it is there and add a rule. Let the container do the creation when it is called. This one may be a bit controversial since we've had problem with the log in particular getting injected and called before it is instantiated. On the other hand, the caller could always do the instantiation themself. If they don't want to instantiate then they are left with having to write the code themselves.

If there's already an instance, I like the idea of having a little more control over when the replacement will be instantiated. But again, our immediate need doesn't require this and I'm okay with revisiting it in the future, as necessary.

library/Vanilla/Utility/ContainerUtils.php
```diff
@@ -67,9 +67,31 @@ public static function currentTheme(): ReferenceInterface {
...
+     * Sometimes an object has limitations or shortcomings that could be resolved by something like a decorator, where
+     * a drop-in replacement wraps existing functionality in enhancements or customizations. This method could configure
+     * a container to use the decorator, replacing stored instances and ensuring new requests to the container would
+     * receive the decorator instead of the original class.
+     *
+     * @param Container $container Container to configure.
+     * @param string $target Container rule to target for replacement. Shared instances will be overwritten.
+     * @param string $replacement Container rule used to determine what will be replace the target in the container.
+     */
+    public static function replace(Container $container, string $target, string $replacement): void {
```
tburry, line 89: I'm going to just leave this as a comment and approve the PR, but...

Do you think `$old` and `$new` are more clear names for this method? I always thought that the needle/haystack were silly paradigms until I realized it was never confusing.

bootstrap.php
```diff
@@ -604,9 +604,11 @@
...
-$log = $dic->get(\Vanilla\Logging\LogDecorator::class);
-// Replace the logger interface with the decorator.
-$dic->rule(\Vanilla\Logging\LogDecorator::class)
-    ->addAlias(\Psr\Log\LoggerInterface::class);
+// Add the log decorator.
+ContainerUtils::replace(
+    $dic,
+    \Psr\Log\LoggerInterface::class,
+    \Vanilla\Logging\LogDecorator::class,
+    true
```
tburry, line 612: Whoops.

PR #10797, opened 2020-06-16
Fix deleting user attachments when deleting user content

dannimajdalani:
closes https://github.com/vanilla/support/issues/2053

### Details

When deleting user with "Delete user content option" user's media does not get deleted.
Screen Shot 2020-06-16 at 6 56 22 PM

**To Test**

- Delete a user
- Check if the user's media files got deleted in GDN_Media


bleistivt:
While I obviously do not know which problem is being fixed here, this change causes the data to be in a worse state than before in my opinion. It leaves orphaned files on the server or CDN which can't easily be tracked through the DB since it doesn't delete the actual file. (Especially once the log gets pruned.)

If this gets merged the way it is, I would never fee comfortable to use the "Delete User Content" option ever again.

dannimajdalani:
We closing this PR. We will create a new issue and address this in the product.

Issue #10799, opened 2020-06-17
Create the domutilities processors/tests


dannimajdalani:
related https://github.com/vanilla/vanilla/pull/10761

PR #10803, opened 2020-06-17
Add DomUtils processors

igraziatto:
Closes https://github.com/vanilla/vanilla/issues/10799

**Notes:**
- I couldn't use `{@inheritdoc}` for some reason, it was failing to declare the `@dataProvider`.
- I had to make a new `provideTrimWordsTests` to fit the `trimWords` default value for `wordCount`. I wasn't sure if I could do something like `TrimWordsProcessor->setWordCount()`.


library/Vanilla/Formatting/Html/Processor/StripEmbedsProcessor.php
```diff
@@ -0,0 +1,36 @@
...
+        $this->applyStripEmbeds();
+        return $this->document;
+    }
+
+    /**
+     * Apply DomUtils::stripEmbeds()
+     */
+    public function applyStripEmbeds() {
+        DomUtils::stripEmbeds($this->document->getDom());
+        $this->document->getDom()->saveHTML();
```
tburry, line 34: This line isn't needed, is it?

tests/Library/Vanilla/Utility/DomUtilsProcessorsTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $actual = $domDocument->getInnerHtml();
+        $this->assertHtmlStringEqualsHtmlString($expected, $actual);
+    }
+
+    /**
+     * Provide tests for `TestTruncateWords()`.
+     *
+     * @return array
+     */
+    public function provideTrimWordsTests(): array {
```
tburry, line 84: This shouldn't be here. It should be in the parent class.

tburry: I read your comment above. If it's necessary to have this in the subclass try a `return parent::provideTrimWordsTests()` instead of copy/pasting. You should never be satisfied with copy/pasting a bunch of code and should always be looking for another way.

PR #10807, opened 2020-06-17
Bypass mods from image upload restrictions

igraziatto:
Closes https://github.com/vanilla/support/issues/1975

**Steps to reproduce:**
1. On Dashboard > Posting, enable "Enable Image Upload Limit" and set the limits to something visually visible, like 20px.
2. Login on your local with a user with `Community.Manage` permission on higher (an admin)
3. On any editor try uploading an image and see it getting resized to the number you've set before.
4. Checkout this branch.
5. Open an incognito tab and login with a user without `Community.Manage` permission (a regular member).
6. Repeat step **3** with both users. The admin should not have their image resized.

**Further testing**
- Repeat step **6** with the Rich Editor (or advanced, if you had tested with Rich before).

**Dependencies:**
https://github.com/vanilla/locales/pull/313

### DIFF
--- a/applications/dashboard/controllers/api/MediaApiController.php
+++ b/applications/dashboard/controllers/api/MediaApiController.php
@@ -48,6 +48,15 @@ class MediaApiController extends AbstractApiController {
     /** @var ConfigurationInterface */
     private $config;
 
+    /** @var bool */
+    private $resizeImages;
+
+    /** @var ?int Max image upload height */
+    private $maxImageHeight;
+
+    /** @var ?int Max image upload width */
+    private $maxImageWidth;
+
     /**
      * DI.
      * @inheritdoc
@@ -57,6 +66,9 @@ class MediaApiController extends AbstractApiController {
         $this->embedService = $embedService;
         $this->imageResizer = $imageResizer;
         $this->config =  $config;
+        $this->resizeImages = $this->config->get('ImageUpload.Limits.Enabled');
+        $this->maxImageHeight = $this->config->get('ImageUpload.Limits.Height');
+        $this->maxImageWidth = $this->config->get('ImageUpload.Limits.Width');
     }
 
     /**
@@ -250,17 +262,25 @@ class MediaApiController extends AbstractApiController {
         ], 'in')->setDescription('Add a media item.');
 
         $body = $in->validate($body);
-        //Bypass ImageUpload.Limits if user has 'Garden.Community.Manage' permission or higher
-        $bypassUploadLimits = $this->getSession()->getPermissions()->hasRanked('Garden.Community.Manage');
+        $fileData = [
+            'foreignType' => 'embed',
+            'foreignID' => $this->getSession()->UserID
+        ];
 
-        $row = $this->mediaModel->saveUploadedFile(
-            $body['file'],
-            [
-                'foreignType' => 'embed',
-                'foreignID' => $this->getSession()->UserID,
-                'bypassUploadLimits' => $bypassUploadLimits
-            ]
-        );
+        if ($this->resizeImages) {
+            //Bypass ImageUpload.Limits if user has 'Garden.Community.Manage' permission or higher
+            $bypassUploadLimits = $this->getSession()->getPermissions()->hasRanked('Garden.Community.Manage');
+
+            if ($bypassUploadLimits) {
+                $fileData['maxImageHeight'] = MediaModel::NO_IMAGE_DIMENSIONS_LIMIT;
+                $fileData['maxImageWidth'] = MediaModel::NO_IMAGE_DIMENSIONS_LIMIT;
+            } else {
+                $fileData['maxImageHeight'] = $this->maxImageHeight;
+                $fileData['maxImageWidth'] = $this->maxImageWidth;
+            }
+        }
+
+        $row = $this->mediaModel->saveUploadedFile($body['file'], $fileData);
         return $row;
     }
 
--- a/applications/dashboard/controllers/api/MediaApiController.php
+++ b/applications/dashboard/controllers/api/MediaApiController.php
@@ -238,7 +238,6 @@ class MediaApiController extends AbstractApiController {
         $this->permission('Garden.Uploads.Add');
 
         //Bypass ImageUpload.Limits.Enabled if user has 'Garden.Community.Manage' permission or higher
-        //$bypassUploadLimits = $this->getSession()->checkRankedPermission('Garden.Community.Manage');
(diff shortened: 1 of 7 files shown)


charrondev:
This doesn't look like the scope that was set out in https://github.com/vanilla/support/issues/1975.

This removes all image upload limits for these users.

> ## Other options
> 
> - Have a different configured limit for administrators.
> - Have a different hard-coded limit for administrators.
> - Have individual dashboard uploads be able to pass through max dimensions to the media endpoint.
> 
> I just don't think in 2020, blindly accepting any image of any size is a reasonable behaviour of a web application.


> I'm switching this to a feature from a bug because the scope has increased.

library/Vanilla/UploadedFile.php
```diff
@@ -295,6 +297,15 @@ public function moveTo($targetPath) {
         $this->moved = true;
     }
 
+    /**
+     * Get whether or not this file should bypass "ImageUpload.Limits.Enabled" config.
+     *
+     * @return bool
+     */
+    public function getBypassUploadLimits() {
```
tburry, line 305: Type hint.

library/Vanilla/UploadedFile.php
```diff
@@ -322,6 +333,16 @@ public function getFile() {
         return $this->file;
     }
 
+    /**
+     * Set whether or not this file should bypass "ImageUpload.Limits.Enabled" config.
+     *
+     * @param mixed $bypassLimits
+     * @return bool
+     */
+    public function setBypassUploadLimits($bypassLimits) {
```
tburry, line 342: Type hints. Also, we usually return `$this` or `void`.

charrondev:
I had discussed this with @initvector  yesterday and we came to the conclusion that rather than bypassing entirely we should still cap to some hardcoded dimensions (say 3k by 3k)?

Anyone uploading larger than that is going to be doing it because they don't know the performance impact it will have on their pages.

applications/dashboard/models/class.mediamodel.php
```diff
@@ -218,6 +218,10 @@ public function saveUploadedFile(UploadedFile $file, array $extraArgs = []): arr
             'Size' => $file->getSize(),
         ]);
 
+        if (in_array('bypassUploadLimits', $extraArgs)) {
```
tburry, line 221: This should be wrong. The `in_array()` functions looks at values not keys. What I would rather see is a PHP best practice where you declare defaults on `$extraArgs` as the first line of the method like this:

```php
$extraArgs += [
    'bypassUploadLimits' => false,
];
```

This shows other developers what can be in `$extraArgs` and also allows you to use the values below without worrying whether or not they exist. Then you call is a simple call to a setter. Also good for code coverage in tests as you don't have another code path with the if statement.

library/Vanilla/UploadedFile.php
```diff
@@ -245,7 +247,7 @@ private function tryApplyImageProcessing(): void {
             "width" => $width ?? 0,
         ];
 
-        if (\Gdn::config("ImageUpload.Limits.Enabled")) {
+        if (\Gdn::config("ImageUpload.Limits.Enabled") && !$this->getBypassUploadLimits()) {
```
tburry, line 250: I would probably be happier to see this class have a basic property called '$maxImageWidth`. Set this in the constructor from the config with a value of `null` meaning there is no max width. That way you can just set this value to whatever. It allows developers to do the following:

1. Set a different width.
2. Set no enforcement.
3. It's clearer that the limit is just on image widths which also allows us to set other limits in the future.
4. It does this all with one property and not a strange combination of config and properties.

applications/dashboard/controllers/api/MediaApiController.php
```diff
@@ -237,6 +237,9 @@ public function patch_attachment(int $id, array $body): array {
     public function post(array $body) {
         $this->permission('Garden.Uploads.Add');
 
+        //Bypass ImageUpload.Limits.Enabled if user has 'Garden.Community.Manage' permission or higher
```
tburry, line 240: Move this call closer to where you are using it. Here is a quote:

> Keep Variables "Live" for as Short a Time as Possible
> A concept that's related to variable span is variable "live time," the total number of statements over which a variable is live. A variable's life begins at the first statement in which it's referenced; its life ends at the last statement in which it's referenced.

*McConnell, Steve. Code Complete (2nd Edition) (Developer Best Practices) (Kindle Locations 6160-6164). Pearson Education. Kindle Edition.*

library/Vanilla/UploadedFile.php
```diff
@@ -295,6 +312,40 @@ public function moveTo($targetPath) {
...
 
+    /**
+     * Get max image upload height
+     * If $this->maxImageHeight is set to NO_IMAGE_DIMENSIONS_LIMIT, fallback to return MAX_IMAGE_HEIGHT
+     *
+     * @return ?int
+     */
+    public function getMaxImageHeight(): ?int {
+        switch($this->maxImageHeight) {
+            case null:
```
tburry, line 323: Have you tested this? I am pretty sure that a switch is the same as `==` rather than `===`.

applications/dashboard/models/class.mediamodel.php
```diff
@@ -204,6 +207,16 @@ private function normalizeAndValidate(array $row): array {
      * @inheritdoc
      */
     public function saveUploadedFile(UploadedFile $file, array $extraArgs = []): array {
+        if (in_array('maxImageHeight', $extraArgs)) {
```
tburry, line 210: This appears to be a bug. `in_array` tests values, not keys. The PHP best practice for options arrays is this:
```php
$extraArgs += [
    'maxImageHeight' => self::NO_IMAGE_DIMENSIONS_LIMIT,
   'maxImageWidth' => self::NO_IMAGE_DIMENSIONS_LIMIT,
];
```

Then you can make your checks to change it below.

library/Vanilla/UploadedFile.php
```diff
@@ -246,11 +256,15 @@ private function tryApplyImageProcessing(): void {
         ];
 
         if (\Gdn::config("ImageUpload.Limits.Enabled")) {
```
tburry, line 258: Can this call to config be removed in a similar way to what you did elsewhere?

igraziatto: I was discussing this with @initvector , his reasoning behind not removing this statement is because we are assuming all usages of UploadedFile are coming from the MediaModel, where that might not always be the case.

tburry: What usages are there currently?

PR #10808, opened 2020-06-18
Keep text field lengths the same on alters

tburry:
This feature will ensure that text fields are not altered when the database alters if the old field is already a text field.

Sometimes a text field changes in size for whatever reason or sometimes a particular site requires a larger field. We want to be able to support those scenarios without being spammed on database alters.

This feature does not come without some risk: we are hiding inconsistency in our system and we are changing some important, low-level code. However, that risk is mitigated with tests and best practices. We should use this feature rarely and carefully.

Closes https://github.com/vanilla/internal/issues/2730

### DIFF
--- a/applications/dashboard/settings/class.hooks.php
+++ b/applications/dashboard/settings/class.hooks.php
@@ -939,4 +939,15 @@ class DashboardHooks extends Gdn_Plugin {
 
         $args['Path'] = $localPath;
     }
+
+    /**
+     * Keep text fields their lengths when altering tables.
+     *
+     * @param \Gdn_DatabaseStructure $structure
+     */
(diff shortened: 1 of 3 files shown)


nervoustwit:
To test I changed my Discussions  Body from TEXT to MEDIUMTEXT

in my Config I set:

```
$Configuration['Feature']['alterTextFieldLengths'] = true;
```

When I do a utilty/structure scan it finds nothing. If I "Run Structure and Scripts" the database does not change. I have put break points in the code and it is executing the `keepTextFieldLengths` function which makes me think there is something else wrong in the `Utility::Structure()`

tburry:
Feature flags are represented with an extra "Enabled" key like this:

```php
$Configuration['Feature']['alterTextFieldLengths']['Enabled'] = true;
```

PR #10814, opened 2020-06-18
Fix RangeExpression parameters not being passed in link headers

initvector:
When an object is used internally in an API endpoint, such as `RangeExpression`, pagination headers are not properly including these values. This is due to `http_build_query` not stingifying objects.

`ApiUtils::pagerUrlFormat` has been updated to cast values to strings before attempting to build the query.


tburry:
@initvector is this ready to go?

initvector:
> is this ready to go?

Yes, indeed.

PR #10816, opened 2020-06-18
cron feature - draft release

egarciajulia:
***  Warning - Please **do not merge** ***

### General

The PR introduces a new feature to the Scheduler, the ability to run jobs on specific date/time or at regular intervals, aka cron-jobs.

This is a draft implementation open to suggestions that I will gladly implement or help to implement.

### Main changes

`JobInterface` was simplified. `setPriority` and `setDelay` were moved to aware-like interfaces. As a consequence, `LocalJobInterface` doesn't need to implement those two methods, therefore, local jobs are also simplified. I refactored the pre-existing jobs implementing `LocalJobInterface` to remove the unneeded/unused methods.

The concept of `JobDescriptorInterface` was introduced. Using `JobDescriptorInterface` to describe a job is more robust, extensible and object-oriented. There are two implementations of the interface `NormalJobDescriptor` and `CronJobDescriptor`. A JobDescriptor doesn't change a job in any form, it just helps the Scheduler to decide if the job needs to be run or not.

The concept of `JobExecutionType` was introduced. The `JobExecutionType` will hint the Scheduler about what kind of job should dispatch. By default, `JobExecutionType::normal()` should be processed in a regular request. 

`DummyScheduler` has a new `addJobDescriptor` method to add jobs to the Scheduler. The old `addJob` method was deprecated but still fully operational, it wraps the parameters into a `NormalJobDescriptor`.

`DummyScheduler` will process job with `NormalJobDescriptor` when the execution type is `JobExecutionType::normal()`. On the other hand, jobs with `CronJobDescriptor` will be executed when the type is `JobExecutionType::cron()`. If the job doesn't match the execution type, the job will not be executed and the status of the job will be set to `JobExecutionStatus::abandoned()`

`DummyScheduler` will receive any kind of job independently of the JobExecutionType. The decision to execute or abandon a job is made at the dispatch event.

`CronJobDescriptor` describes a cron-like schedule when the job should be run. Ex: `*/10 * * * *` to execute a job every 10 minutes. Cron expressions are handled by `dragonmantank/cron-expression` library ([Here](https://github.com/dragonmantank/cron-expression)). The minimum granularity of a cron expression is 1 minute. The library needs to be part of `composer.json`

`DummyScheduler` didn't provide any mechanism for storing the state of a job. It just threw an event after dispatching all the jobs. The approach proved to be limited for handling some edge cases like reaching the maximum execution time. In that case, there was no way to know the job was interrupted or never executed.

The concept of `SchedulerMeta` was introduced to tackle the previous limitation. The scheduler will save some metainformation every time it receives, executes or abandons a job.
In order to avoid introducing a new Model or a Database's scheme change, the DAO Pattern was used. There are two plain-old-php-object (POPO) acting as entities and a DAO implementation based on `Gdn::userMetaModel()`. UserMetaModel was especially useful because it already implements a caching mechanism. The use of a DAO Pattern will allow reimplementing the storage layer to whatever it is needed in the future.
The metainformation will be pruned at regular interval using a cron job (`$this->scheduler->addJobDescriptor(new CronJobDescriptor(TidySchedulerMetaJob::class, "*/10 * * * *"));`). The default TTL for metainformation is 24 hours.

`SchedulerApiController` provides a way to trigger cron jobs and get the metainformation.

`POST /api/v2/scheduler/cron` will do the following:
+ Set the Scheduler on `JobExecutionType::cron()`
+ Fire the event `CRON_TRIGGER_EVENT`
+ Schedule a con job for `TidySchedulerMetaJob` to run every 10 minutes.
Why firing `CRON_TRIGGER_EVENT`? The event gives the opportunity to any listener to add cron jobs into the Scheduler. Hooking into the event is NOT needed, it just gives you an additional opportunity to do so. The use case in mind for this will be the following: In order to schedule a cron-job you need to run an intensive process (querying the database, contacting a remote service, etc), it would make no sense to do the calculation for every request, it will be better to do it just when you are sure a cron execution is running. The event helps you to be sure of that.

`POST /api/v2/scheduler/details` will return you all the available metainformation.

Circling back to "the ability to run jobs on specific date/time or at regular intervals", by now you might be noticing the statement is not completely true. Cron-jobs will only be triggered when `POST /api/v2/scheduler/cron`, so there are some special situations to consider.

The scheduler will run cron-jobs at the defined minimum time span (default of 1 minute). If you hit the endpoint more than once during the time span, the jobs triggered on the second or following will be abandoned. For a cron-job being abandoned is the same as no hitting the endpoint. Why? Because the Scheduler will take the last time it ran cron-jobs to establish is any given cron-job should run or not (based on the descriptor's schedule). As a consequence of this:
+ No cron-job will be missed in case the endpoint is not hit. They will run the next time.
+ No cron-job will run more than once in any given interval between two endpoint calls. EX: "*/10 * * * *", should run every 10 minutes, however, if you are hitting the endpoint once an hour, the job will run one time (not 6)
+ If no information is available to know when the Scheduler ran cron-jobs for the last time, one day in the past will be used as default.

`SchedulerApiController` can be used with the integrated API's permission mechanism (`Garden.Scheduler.Cron` and `'Garden.Scheduler.Details`) or with an AdHoc `Bearer Token`. If the configuration `Garden.Scheduler.Token` is set, then you can use that token to authenticate. The token can not be null. The purpose of the AdHoc Token is to allow you to configure the call to the endpoint for example with an Operating System Cron Task without the risk of leaking credentials. The scope of the Token is quite narrow.

Configurations:
`Garden.Scheduler.CronMinimumTimeSpan`: minimum amount of time between two consecutive cron triggers. Default 60 seconds.
`Garden.Scheduler.JobDetailsPruneAge`: Scheduler metainformation TTL. Default: 24 hours.


egarciajulia:
The PR was superseded

PR #10823, opened 2020-06-19
Fix confirmCommentDelete redirect to point to current discussion

blutarch:
Closes vanilla/support#1298.

Currently, when you delete one or more comments of a discussion via the checkbox (rather than the cogwheel), you are redirected to the main discussions page, which is undesirable if you're not done working on that discussion. This PR sets the redirect to the current discussion.

I have done functional tests within subcommunities and on a hub-node setup.

### TO TEST
1. Go to any discussion with comments and delete a comment via the checkbox.
1. Note that you get redirected to the main discussion page.
1. Repeat step 1 and verify that you get redirected to the current discussion.

### DIFF
--- a/applications/vanilla/controllers/class.moderationcontroller.php
+++ b/applications/vanilla/controllers/class.moderationcontroller.php
@@ -280,7 +280,7 @@ class ModerationController extends VanillaController {
             unset($checkedComments[$discussionID]);
             Gdn::userModel()->saveAttribute($session->UserID, 'CheckedComments', $checkedComments);
             ModerationController::informCheckedComments($this);
-            $this->setRedirectTo('discussion/'.$discussionID);
+            $this->setRedirectTo(discussionUrl($discussion));
         }
 
         $this->render();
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/class.moderationcontroller.php
```diff
@@ -280,7 +280,7 @@ public function confirmCommentDeletes($discussionID = '') {
             unset($checkedComments[$discussionID]);
             Gdn::userModel()->saveAttribute($session->UserID, 'CheckedComments', $checkedComments);
             ModerationController::informCheckedComments($this);
-            $this->setRedirectTo('discussions');
+            $this->setRedirectTo('discussion/'.$discussionID);
```
tburry, line 283: If this is what you want to do then you want to call the `discussionUrl()` function which takes a full discussion. Fortunately, it looks like you have one above in the `$discussion` variable above.

blutarch:
> This looks good on its face but I would check to see if it works with sub communities. Alternatively you can pass true to one of the discussion url parameters to return. The full domain and manually test to see if that works.

The $withDomain parameter defaults to true, so it returns the full domain. I did functional tests with subcommunities, and it works the same.

PR #10826, opened 2020-06-19
Clear user cache on insertInternal

igraziatto:
Closes https://github.com/vanilla/vanilla-patches/issues/702

**Steps to reproduce:**
1. Register with an email and username (basic registration method) and wait for the confirmation email to appear.
1. Take the supplied link in the register confirmation  email (for example: https://open.vanillaforums.com/entry/emailconfirm/72565/ZKID6I2f6FwDeyshAXUCEkJybbkRAzYU) and increment the value "72565" (or whatever number appears in the confirmation link) by 1 (thus: "72566").
1. Open the changed link in a browser.
1. Logout.
1. Register with a different email and different username directly after opening the link. The error message will now pop up and say "value is not a valid object" preventing the user from registering.

### DIFF
--- a/applications/dashboard/models/class.usermodel.php
+++ b/applications/dashboard/models/class.usermodel.php
@@ -1070,6 +1070,9 @@ class UserModel extends Gdn_Model implements UserProviderInterface, EventFromRow
         $userID = $this->SQL->insert($this->Name, $fields);
 
         if ($userID) {
+            //force clear cache for that UserID
+            $this->clearCache($userID, ['user']);
+
             $user = $this->getID($userID);
             $userEvent = $this->eventFromRow(
                 (array)$user,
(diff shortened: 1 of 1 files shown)


applications/dashboard/models/class.usermodel.php
```diff
@@ -1070,6 +1070,9 @@ private function insertInternal($fields, $options = []) {
         $userID = $this->SQL->insert($this->Name, $fields);
 
         if ($userID) {
+            //force clear cache for that UserID
+            $this->clearCache($userID, ['user']);
```
tburry, line 1074: If you want to go the distance on this PR you could do the following:

1. Extract that 'user' string to a constant on the class. Do it for the other versions of that string so you would have three `CACHE_*` constants. Hint: PhpStorm has a refactor tool for this.
2. Go ahead and add the `@param` doc block for that argument.

This isn't a change request though, just a comment.

tburry:
Beauty. Thanks for the extra commit!

PR #10829, opened 2020-06-21
Check for an authenticated postback on /session/stash

tburry:
This method is only called through our js `stash()` function which will post the proper headers and TK.

Closes https://github.com/vanilla/vanilla-patches/issues/687.


applications/dashboard/controllers/class.sessioncontroller.php
```diff
@@ -22,14 +22,16 @@ public function stash() {
         $this->deliveryType(DELIVERY_TYPE_BOOL);
         $this->deliveryMethod(DELIVERY_METHOD_JSON);
 
-        $name = $this->Request->post('Name', '');
-        $value = $this->Request->post('Value', '');
+        if ($this->Request->isAuthenticatedPostBack(true)) {
```
tburry, line 25: Weird diff, but I'm just wrapping the method in an `isAuthenticatedPostBack(true)`.

PR #10844, opened 2020-06-23
Fixed layout calculations, pagination info and search result icons

slafleche:
Fixed several small miscalculations for the panel variables. The widths set in the config should be matching exactly in the DOM now. I made an effort to make the variables easier to understand for the layout. Feedback is welcome.

To fix the responsive bugs with the search page, i had to make a new context for the two panel layout. Not ideal, but not trivial to do differently.

Fixed search result icon placement on smaller devices.

Fixed the counts for the pagination info, they were off by 1.


charrondev:
Moved to https://github.com/vanilla/vanilla-cloud/pull/43

PR #10846, opened 2020-06-23
Improve discussion counts

tburry:
In an effort to reduce the potential for post counts to drift out of sync, we're revisiting the way mass-deletes are performed during moderation actions. Currently, if discussions deleted via moderation tools, they are added to a scheduled job that is executed at the end of the request, which deletes all of the discussions, en masse. Now, we'll rely on an AJAX feedback loop to manually delete each discussion, one at a time. This is done to simplify the count calculation process: counts are updated for one discussion at a time, instead of for all discussions in a batch.

This update will make deleting multiple discussions slower. This is a conscious trade off of speed for accuracy. We can make performance improvements to increase speed, once we have a solid baseline for count calculation behavior.

### Testing
1. Run a dba-counts on your site to ensure everything is accurate before you begin.
1. Enable `Vanilla.AdminCheckboxes.Use` in the site config.
1. Visit the "Recent Discussions" page.
1. Select multiple discussions across multiple pages (ideally, across multiple categories).
1. Select "Delete" from the moderation window in the lower corner of the page.
1. Confirm the delete. You may see the popup change to a "Please wait..." message, if deleting many discussions. The popup will close on its own, once the operations have completed.
1. Verify counts are accurate for the affected categories.

Closes vanilla/support#1976
Closes https://github.com/vanilla/support/issues/1058


dannimajdalani:
I've tested this on 10012, Wasn't able to delete any discussion because of a JS error. 
Screen Shot 2020-06-30 at 12 53 40 PM

dannimajdalani:
Retested after the above commit.

**Count Discussions (before the delete)**

Screen Shot 2020-06-30 at 1 10 33 PM


Screen Shot 2020-06-30 at 1 11 05 PM

**Count Discussions(after the delete)**
Screen Shot 2020-06-30 at 1 14 43 PM

dannimajdalani:
Deleted 100 discussions across multiple pages. Count is correct.

PR #10847, opened 2020-06-24
Fix embedded discussions link translation 

igraziatto:
Closes https://github.com/vanilla/support/issues/2164

**Steps to test:**
1. Make sure you have embedded comments setup on your local (or Wordpress)
2. If a discussion is already created with the embedded comments local, delete it. (this test requires the discussion to be created).
3. Change your local language to spanish.
4. Make a comment in the embedded comments.
5. Go to your local and open the discussion it created, notice the link to the post says "Read the full story here"
6. Delete the discussion.
7. Checkout this branch and repeat from step **4.**. Observe now that the link matches the language set.

### DIFF
--- a/applications/vanilla/controllers/class.postcontroller.php
+++ b/applications/vanilla/controllers/class.postcontroller.php
@@ -525,49 +525,7 @@ class PostController extends VanillaController {
                 $this->Form->addHidden('vanilla_url', $vanilla_url);
                 $this->Form->addHidden('vanilla_category_id', $vanilla_category_id);
 
-                $PageInfo = fetchPageInfo($vanilla_url);
-
-                if (!($Title = $this->Form->getFormValue('Name'))) {
-                    $Title = val('Title', $PageInfo, '');
-                    if ($Title == '') {
-                        $Title = t('Undefined discussion subject.');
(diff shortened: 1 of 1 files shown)


applications/vanilla/controllers/class.postcontroller.php
```diff
@@ -540,7 +540,7 @@ public function comment($DiscussionID = '') {
 
                 $Description = val('Description', $PageInfo, '');
                 $Images = val('Images', $PageInfo, []);
-                $LinkText = t('EmbededDiscussionLinkText', 'Read the full story here');
+                $LinkText = t('EmbededDiscussionLinkText', t('Read the full story here'));
```
tburry, line 543: Can't you replace the call to `fetchPageInfo()` above to `DiscussionModel::fetchPageInfo()`. That is what I was asking for in the original issue.

igraziatto: I've refactored into using `DiscussionModel->fetchPageInfo()`. I took a look at some communities that embed comments (coming from Wordpress or any other platform) and it seems that the title twice (as discussion name and within the content) it's not as much of an issue as I thought, so I left as it is.

Now both `PostController->comment()` and `DiscussionController->refetchPageInfo()` handle data the same way.

PR #10852, opened 2020-06-25
Fix kb dashboard dependency on kb

charrondev:
Fixes issue in the prod build where the dashboard won't load.

I'm not 100% sure why not implicitly specifying the loader for this page caused the issue, but this resolves it.


dannimajdalani:
@initvector I've recheked my setup, without HotReload enabled, no js errors, I was able to access the dashboard.

PR #10857, opened 2020-06-26
Add Psalm static code analysis

tburry:
Add a basic Psalm static code analysis capability and fixes to the project. This analysis starts with a very lax analysis. This is to give us time to manually deal with issues before the build starts breaking.

_Note: You may notice the library/Vanilla/Web/EndpointSchema.php schema class being removed. This was a class that wasn't really every used. Moreover it had a bug in it that would have caused a crash so I'm fairly confident it wasn't used or else we would have noticed the crash._


library/Vanilla/EmbeddedContent/Factories/TwitchEmbedFactory.php
```diff
@@ -118,7 +118,7 @@ private function idFromUrl(string $url): ?string {
             return null;
         }
 
-        if ($host === "clips.twitch.tv" || preg_match("`/(?<channel>[^/]+)/clip/(?<clipID>[^/]+)`", $path, $clipsMatch)) {
+        if ($host === "clips.twitch.tv" && preg_match("`/(?<channel>[^/]+)/clip/(?<clipID>[^/]+)`", $path, $clipsMatch)) {
```
tburry, line 121: This is one of the only material code changes, but I can't see how it wasn't a bug before.

PR #10858, opened 2020-06-26
Verify if user information is defined.

cchabilall83:
This fixes https://github.com/vanilla/support/issues/2166.  Not all the user information is sent back from `/conversations`, so trying access the authors prop info would throw an error.

closes https://github.com/vanilla/support/issues/2166
closes https://github.com/vanilla/support/issues/1758

### DIFF
--- a/library/src/scripts/headers/mebox/pieces/MeBoxDropDownItem.tsx
+++ b/library/src/scripts/headers/mebox/pieces/MeBoxDropDownItem.tsx
@@ -62,12 +62,16 @@ export default class MeBoxDropDownItem extends React.Component<IProps> {
             const authorCount = this.props.authors.length;
             if ("authors" in this.props && authorCount > 0) {
                 authors = this.props.authors!.map((user, index) => {
-                    return (
-                        <React.Fragment key={`meBoxAuthor-${index}`}>
-                            <strong>{user.name}</strong>
-                            {`${index < authorCount - 1 ? `, ` : ""}`}
-                        </React.Fragment>
-                    );
+                    if (user) {
+                        return (
+                            <React.Fragment key={`meBoxAuthor-${index}`}>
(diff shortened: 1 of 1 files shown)


cchabilall83:
> Hi Chris, I have a couple of questions: 1) I've never seen using key check for a component props, should we make a change (maybe just destructing and check if it's not undefined)? 2) Is it correct that the check`"authors" in this.props && authorCount > 0` already makes sure this.props.authors is a proper list, so `user` already always has a value?

1) I agree and changed the current code
2) we can have an array element of undefined authors[0] = undefined authors[1]={authoinfo...} that is why we need to check is the user is defined

library/src/scripts/headers/mebox/pieces/MeBoxDropDownItem.tsx
```diff
@@ -55,19 +55,24 @@ export default class MeBoxDropDownItem extends React.Component<IProps> {
...
-            if ("authors" in this.props && authorCount > 0) {
-                authors = this.props.authors!.map((user, index) => {
-                    return (
-                        <React.Fragment key={`meBoxAuthor-${index}`}>
-                            <strong>{user.name}</strong>
-                            {`${index < authorCount - 1 ? `, ` : ""}`}
-                        </React.Fragment>
-                    );
+            if (authors && authorCount > 0) {
+                authorList = this.props.authors!.map((user, index) => {
```
tuan-ng, line 65: I think due to the check we can write `authors.map((user, index) => ...` here

charrondev:
```suggestion
                authorList = authors.map((user, index) => {
```

Typescript notices that `authors` has been checked already, but not `this.props.authors`.

PR #10862, opened 2020-06-26
Add unread conversations to GET /me

cchabilall83:
Adds the users unread conversations count to the /me endpoint.  Applies it in the front for the message count.

closes https://github.com/vanilla/support/issues/2168


applications/dashboard/controllers/api/UsersApiController.php
```diff
@@ -388,6 +388,10 @@ public function get_me(array $query) {
                 "description" => "Total number of unread notifications for the current user.",
                 "type" => "integer",
             ],
+            "countUnreadConversations" => [
+                "description" => "Total number of unread conversations for the current user.",
```
tburry, line 392: As an FYI: You no longer have to put descriptions on the PHP side of endpoints because the description is in the OpenAPI YAML now.

PR #10866, opened 2020-06-26
Conversations: Add "Cache-Control: no-store" to inbox

bleistivt:
This prevents opened conversations from still being shown as "new" when navigating back to the inbox list using the back button. (this issue mostly exists in mobile chrome which seems to cache more aggressively)

This is similar to how back button navigation is handled in category and discussion lists. See:
https://github.com/vanilla/vanilla/search?q=no-store&unscoped_q=no-store


bleistivt:
Still relevant

bleistivt:
Still relevant

PR #10868, opened 2020-06-27
Apply psalm and fixes to the rest of library

tburry:
This PR adds the rest of the "library" directory to our psalm analysis. There were a lot of purely docblock fixes that were already merged in [a different PR](https://github.com/vanilla/vanilla/pull/10867). The changes in the PR do involve code changes which will require some review.

The one of note is the `HeadModule`. This appears to be a full file change, but all I did was remove the wrapping `class_exists`. This led to a five star lint police alert so I also had to fix the lint errors. Of note, I renamed the private and protected properties to follow our naming conventions. I checked for subclasses and there are none.


applications/dashboard/modules/class.headmodule.php
```diff
@@ -8,636 +8,634 @@
  * @since 2.0
  */
 
-if (!class_exists('HeadModule', false)) {
```
tburry, line 11: I just removed the `class_exists` check from this class.

library/core/class.authenticator.php
```diff
@@ -184,7 +184,7 @@ public function fetchData($dataSource, $directSupplied = []) {
 
                     case self::DATA_COOKIE:
                         $this->_DataHooks[$dataTarget]['value'] = $this->_DataSource->getValueFrom(
-                            Gdn_Authenticator::INPUT_COOKIES,
+                            Gdn_Request::INPUT_COOKIES,
```
tburry, line 187: I am not sure if this is correct, but `Gdn_Authenticator::INPUT_COOKIES` does not exist.

library/core/class.cache.php
```diff
@@ -370,7 +370,6 @@ public function fail($server) {
 
         $fails = &$storeServer['Fails'];
         $fails++;
-        $active = $isActive ? 'active' : 'inactive';
```
tburry, line 373: The `$active` variable was never used.

library/core/class.format.php
```diff
@@ -1339,7 +1339,38 @@ public static function timespan($timespan) {
...
+    protected static $_UrlTranslations = [
```
tburry, line 1342: This change was just to remove duplicate array keys. I had to format the array on multiple lines while I was at it as per our lint rules.

library/core/functions.general.php
```diff
@@ -2014,16 +2014,16 @@ function walkAllRecursive(&$input, $callback) {
             }
             foreach ($input as $key => &$val) {
                 if (is_array($val) || is_object($val)) {
-                    call_user_func_array($walker, [&$val, $callback, $key]);
```
tburry, line 2017: This is not a material change. However, calling a callable this way is more efficient.

library/database/class.dataset.php
```diff
@@ -31,35 +31,38 @@ class Gdn_DataSet implements IteratorAggregate, Countable, JsonSerializable {
...
     /**
-     * @var intDetermines what type of result is returned from the various methods by default.
+     * @var string Determines what type of result is returned from the various methods by default.
      * Either DATASET_TYPE_OBJECT or DATASET_TYPE_ARRAY.
      */
     protected $_DatasetType = DATASET_TYPE_OBJECT;
 
     /** @var bool  */
-    protected $_EOF = false;
+    protected $eof = false;
```
initvector, line 43: Renaming the private properties is A-OK, but renaming the protected property is a little riskier. Are you confident in this change?

initvector: Sounds like we're confident it's very likely anything out there is subclassing `Gdn_DataSet`, so this is not a concern.

Issue #7329, opened 2018-06-07
Conversations:  Comma-separated recipients in URL no longer possible


bleistivt:
URLs like this used to work:
https://open.vanillaforums.com/messages/add/Linc,Todd

This was useful for providing a link where users could write the moderation team, for example

This regression was introduced in defa82c2081ff7c72d29eb12af05ad6b4a1894c6

bleistivt:
This just recently came up in the OS forum, I will provide a PR once I find the time.

PR #10872, opened 2020-06-30
Fix sending messages to users with comma in username

dannimajdalani:
### Details

Sending messages to users with comma in their username is failing because of https://github.com/vanilla/vanilla/blob/ec766e44dd12ee0fcbcbc6ee1d0b7de9978e2690/applications/conversations/controllers/class.messagescontroller.php#L75

### To Test
Send a messages to users with comma in their username.


bleistivt:
Looks like this will close #7329 too.

dannimajdalani:
moved to https://github.com/vanilla/vanilla-cloud/pull/148

PR #10873, opened 2020-06-30
Fix markdown quote code block

igraziatto:
Closes https://github.com/vanilla/support/issues/1171

**Notes:**
- This changes markdown default syntax of code blocks from indented to fenced code.
- I had to override `transform()` which is not ideal, but it was the only way to get all the functions running in the right order.

**Steps to test:**
- Make sure you have "Makrdown" as your posting format
- Make a comment with some block html, for example:
```
<ul>
    <li>something</li>
    <li>something else</li>
    <li>another thing</li>
</ul>
```
- Quote that comment and observe the rendering of the quote is broken.
- Checkout this branch.

**Futher testing:**
- Some examples of content you can test to see if they render correctly:

```
> @Tester_McTesterson said:
> ~~~
> <ul>
>     <li>something</li>
>     <li>something else</li>
>     <li>another thing</li>
> </ul>
> ~~~
> 
>     <ul>
>         <li>something</li>
>         <li>something else</li>
>         <li>another thing</li>
>     </ul>
> 
> 
> ```
> <ul>
>     <li>something</li>
>     <li>something else</li>
>     <li>another thing</li>
> </ul>
> ```
> `<ul><li>something</li><li>something else</li><li>another thing</li></ul>`
> 
> 
>         public function addBreaks() {
>             $this->span_gamut = array_replace($this->span_gamut, [
>                 'doStrikeout' =>  15,
>                 'doSoftBreaks' => 80,
>             ]);
>         }
> 
```
a bunch of code
```

### CODE REVIEW

**Comment by tburry** [comment]
Where are you stuck?

**Comment by igraziatto** [comment]
@tburry marking `$this->no_markup = true;` didn't work because it broke spolier styles, so I'm basically back to square 1.

What happens is that `hashHTMLBlocks` tries to parse HTML, but it only accepts valid HTML and it's a bit buggy in some regex because it interprets the parent opening tag wrong when it is quoted (has a trailing `>`).

File: library/core/class.markdownvanilla.php [hunk 1]
```diff
@@ -224,4 +224,54 @@ protected function makeCodeSpan($code) {
         $code = str_replace(["\r", "\n"], ' ', $code);
         return parent::makeCodeSpan($code);
     }
+
+    /**
+     * Main function. Performs some preprocessing on the input text and pass
+     * it through the document gamut.
+     *
+     * @api
```
tburry, line 232: What is this?

library/core/class.markdownvanilla.php
```diff
@@ -224,4 +224,54 @@ protected function makeCodeSpan($code) {
...
+     *
+     * @param  string $text
+     * @return string
+     */
+    public function transform($text) {
```
tburry, line 237: Overriding the main method make me nervous. Can you document the changes you made better here?

PR #10879, opened 2020-07-01
Clean up the code in the database layer

tburry:
This PR contains mostly doc block fixes. It also has some light refactoring such as:

1. I extracted some constants for LIKE expressions.
2. I added curly braces to some if statements in accordance to our standards.
3. There may be some misc code tweaks too.

### DIFF
--- a/library/core/class.model.php
+++ b/library/core/class.model.php
@@ -510,7 +510,7 @@ class Gdn_Model extends Gdn_Pluggable {
             deprecated('The $limit parameter is deprecated in Gdn_Model->delete(). Use the limit option.');
             $limit = $options;
         } else {
-            $options += ['reset' => true, 'limit' => null];
+            $options += ['rest' => true, 'limit' => null];
             $limit = $options['limit'];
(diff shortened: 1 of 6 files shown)


library/core/class.model.php
```diff
@@ -510,7 +510,7 @@ public function delete($where = [], $options = []) {
             deprecated('The $limit parameter is deprecated in Gdn_Model->delete(). Use the limit option.');
             $limit = $options;
         } else {
-            $options += ['rest' => true, 'limit' => null];
+            $options += ['reset' => true, 'limit' => null];
```
tburry, line 513: This is one of those things that requires special attention. When you combine this bug with [this one](https://github.com/vanilla/vanilla/pull/10879/files#diff-cc39bda16852db3b521fa7c8dd7f6cdeL1396) you'll see the situation is a bit delicate. I think we have a decent number of tests that delete and would fail if we all of a sudden stopped resetting the `Gdn_SQLDriver` though.

PR #10886, opened 2020-07-02
Fix Gdn_SQLDriver::noReset()

tburry:
There seems to be a bug in `Gdn_SQLDriver::noReset()`. This is one of those things that is definitely a fix, but could also lead to a regression. The only place I saw a usage in our entire code base is in `Gdn_Model` There was actually a typo above that was making that code path be hit in tests too which I fixed.

Finally, some tests were throwing exceptions because of an error in `Gdn_SQLDriver::truncate()` where they pass a non-existent option of ‘truncate’ to `Gdn_Database::query()`.

*Note: I made this fix on a different PR and it initially caused a flurry of failures which I addressed. This shows the code is getting hit a reasonable amount in our tests.*

### DIFF
--- a/library/core/class.model.php
+++ b/library/core/class.model.php
@@ -502,16 +502,16 @@ class Gdn_Model extends Gdn_Pluggable {
             $where = [$this->PrimaryKey => $where];
         }
 
-        $resetData = false;
+        $resetData = true;
         if ($options === true || val('reset', $options)) {
             deprecated('Gdn_Model->delete() with reset true');
-            $resetData = true;
         } elseif (is_numeric($options)) {
             deprecated('The $limit parameter is deprecated in Gdn_Model->delete(). Use the limit option.');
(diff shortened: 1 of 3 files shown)


bleistivt:
Is this functionality even needed? I proposed to remove it some time ago, because it hasn't been working forever.

Should I close #10206?

tburry:
I guess the answer to that is beyond the scope of this PR.

bleistivt:
If the need for not cleaning up the SQL objects properties hasn't come up internally, there hasn't been one since 2009:
https://github.com/vanilla/vanilla/blob/474d00d3dadb287ae0b3ca4da2177745c32c4c3e/library/database/class.sqldriver.php#L1207

So if there are fears of regressions caused by this, why not just change the `ModelTest` to test for basic deletion and remove the "reset/noReset" code path?

PR #10905, opened 2020-07-16
Syncing upstream commits

charrondev:
Linter errors are on the plugins that were moved from `vanilla/addons`.


charrondev:
> _No description provided._

PR #10925, opened 2020-08-10
Prevent count() warning in class.profilecontroller.php

bleistivt:
`PHP Warning:  count(): Parameter must be an array or an object that implements Countable`
due to a mismatched closing brace


alex-mtl (approved):

bleistivt:
fixed in master

PR #10997, opened 2021-01-20
Fix structure for MySQL 8

bleistivt:
MySQL > 5.7 seems to report the INT type of standard length differently when using `SHOW COLUMNS`:

`int unsigned` instead of `int(11) unsigned`

This causes the structure to fail as Gdn_MySQLDriver::fetchTableSchema just trims everything after the first parenthesis.

### DIFF
--- a/library/database/class.mysqldriver.php
+++ b/library/database/class.mysqldriver.php
@@ -167,6 +167,9 @@ class Gdn_MySQLDriver extends Gdn_SQLDriver {
                 }
             }
 
+            // Remove "unsigned" suffix on MySQL > 5.7
+            $type = str_replace(' unsigned', '', $type);
+
             $object = new stdClass();
             $object->Name = $field->Field;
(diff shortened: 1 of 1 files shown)


bleistivt:
For reference: 
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-19.html#mysqld-8-0-19-feature

> Display width specification for integer data types was deprecated in MySQL 8.0.17, and now statements that include data type definitions in their output no longer show the display width for integer types [...]

> For DESCRIBE statements and INFORMATION_SCHEMA queries, output is unaffected for objects created in previous MySQL 8.0 versions because information already stored in the data dictionary remains unchanged. This exception does not apply for upgrades from MySQL 5.7 to 8.0, for which all data dictionary information is re-created such that data type definitions do not include display width. (Bug #30556657, Bug #97680)

charrondev:
@bleistivt this should be resolved in the 2022.002 release onwards (and the latest master).

PR #11003, opened 2021-01-27
Fix HTML legacy search results

bleistivt:
HtmlFormat::parseImages outputs the wrong data stucture causing the schema validation on search results to fail.

    Validation of result failed.
    APP Log:  Caught Garden\Schema\ValidationException: images[0] is not a valid object.


bleistivt:
https://github.com/vanilla/vanilla/pull/11017

PR #11004, opened 2021-01-27
Fix ProfileExtender clearing fields when editing profiles

bleistivt:
A wrong method call causes all fields showing up as empty wehen editing profiles causing users to delete all their extended info.


bleistivt:
https://github.com/vanilla/vanilla/commit/7cc3ce7c31e181d934514b78c6f8406f7b5c5e3a

PR #11005, opened 2021-01-27
Fix Gdn_Form notice: add "additionalpermissions" as reserved attribute

bleistivt:
`Gdn_Form` options that are not actually attributes have to be added to `$reservedAttributes`.


bleistivt:
still relevant

bleistivt:
still relevant

PR #11019, opened 2021-02-13
ProfileExtender: Fix DateOfBirth formatting

bleistivt:
ProfileExtender changed the way it saves its configuration array from associative to indexed.
Thus `c('ProfileExtender.Fields.DateOfBirth.OnProfile')` will not work anymore.
This fix puts the condition in the loop below.


bleistivt:
still relevant

bleistivt:
still relevant

PR #11038, opened 2021-03-22
User status clear: Align frontend permission with backend permission

bleistivt:
The frontend permission for clearing a users status does not match the backend permission:
https://github.com/vanilla/vanilla/blob/7cc3ce7c31e181d934514b78c6f8406f7b5c5e3a/applications/dashboard/controllers/class.profilecontroller.php#L207

Since moderators are usually allowed to moderate activities, they should be able to clear a users status too.


bleistivt:
still relevant

bleistivt:
still relevant

Issue #11073, opened 2021-06-23
EXIF-based image rotation broken due to incorrect parameter typing in ImageResizer.php


jamesinc:
Bug is here: https://github.com/vanilla/vanilla/blob/8b460a8d64df77f5bb1ce7a1b4e2b09efb3a38f6/library/Vanilla/ImageResizer.php#L347

Expected function signature: `exif(string $source, int $imageType)`
Actual function signature: `exif(string $source, string $imageType)`

This bug breaks EXIF image rotation (and probably other EXIF functions too I didn't dig into it any further)

bleistivt:
still a problem

PR #11171, opened 2023-01-21
Fix EXIF-based image rotation in ImageResizer.php

bleistivt:
PR for #11073
The PHP IMAGETYPE_* constants are ints.


bleistivt:
fixed in master

PR #11215, opened 2025-09-24
VNLA-9967: Merge 010 into master snyk dependency tracking

Crousem:
Snyk doesn't have a way to keep changing branches it targets with its scans. So we will need periodically merge release branches into master (or whatever) to keep scanning in sync with the latest version.


maneeshchiba-hl (approved):
When this is merged, we should be careful to not have `release/2025.010` automatically delete

